PR #108, opened 2014-09-02
Scala Grammar

lrlucena:
Scala Grammar based on the "Scala Language Specification 2.11"
http://www.scala-lang.org/files/archive/spec/2.11/


igstan:
I don't think it works. For the following Scala code snippet, I get this error:

> line 1:0 mismatched input 'object Main' expecting {<EOF>, 'private', 'override', 'package', 'abstract', 'sealed', '@', 'protected', 'final', 'lazy', 'trait', 'class', 'case', 'implicit', 'import', Semi}

```scala
object Main
  def main(args: Array[String]): Unit = {
    println("hello")
  }
}
```

PR #138, opened 2015-02-04
ECMAScript : Fixed missing lookahead

danilaml:
Before 
```javascript
function f() {return 100}
```
or similar would be parsed as FunctionExpression while it should be parsed as FunctionDeclaration. This change should fix it.

### DIFF
--- a/ecmascript/ECMAScript.g4
+++ b/ecmascript/ECMAScript.g4
@@ -270,7 +270,7 @@ emptyStatement
 /// ExpressionStatement :
 ///     [lookahead ∉ {{, function}] Expression ;
 expressionStatement
- : expressionSequence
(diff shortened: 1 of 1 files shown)


ecmascript/ECMAScript.g4
```diff
@@ -270,7 +270,7 @@ emptyStatement
 /// ExpressionStatement :
 ///     [lookahead ∉ {{, function}] Expression ;
 expressionStatement
- : expressionSequence
+ : {_input.LA(1) != OpenBrace}? {_input.LA(1) != Function}? expressionSequence
```
parrt: This would be faster if you combined the predicates with `&&`.

danilaml: Done

PR #153, opened 2015-04-29
initial contribution of grammar for SUO-KIF language.

AppSorcery:
@see http://www.adampease.org/OP/
@see
http://sigmakee.cvs.sourceforge.net/viewvc/sigmakee/sigma/suo-kif.pdf


parrt:
thanks!

PR #154, opened 2015-04-30
initial CLIF grammar

AppSorcery:
@see ISO/IEC STANDARD 24707 First edition
http://metadata-standards.org/24707/index.html


parrt:
thanks!

Issue #158, opened 2015-05-13
vhdl grammar allows invalid identifiers


fer-rum:
Identifiers in VHDL are restricted with respect to underscores as follows:

````
Underscores are significant characters in an identifier 
and basic identifiers may contain underscores, 
but it is not allowed to place an underscore as a first or last character of an identifier. 
Moreover, two underscores side by side are not allowed as well. 
Underscores are significant characters in an identifier.
````

[source](http://vhdl.renerta.com/mobile/source/vhd00037.htm)

The current grammar does not reflect this behaviour. It allows two underscores in a row or as the last character.

PR #159, opened 2015-05-13
Addressed issue #158 (vhdl grammar allows invalid identifiers)

fer-rum:
Requiring an underscore to be immediately followed by a letter or digit prevents them from occurring multiple times in a row as well as terminating an identifier with an underscore.

Related issue: #158


parrt:
Fixes #158

PR #161, opened 2015-05-25
Cobol 85

uwol:
Pull request for Cobol 85 grammar. Tested against NIST test suite.


parrt:
Wow. that is a big grammar! Thanks!

PR #162, opened 2015-05-25
Allow E in exponent

z0rgy:
Extend the exponent definition to include either an uppercase E or a lowercase e

### DIFF
--- a/vhdl/vhdl.g4
+++ b/vhdl/vhdl.g4
@@ -1661,7 +1661,7 @@ BACKSLASH     : '\\'  ;
   
 
 EXPONENT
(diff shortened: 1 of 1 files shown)


z0rgy:
The spec under the Decimal literals section states the following: "The letter E of the exponent, if any, can be written either in lowercase or in uppercase, with
the same meaning"

The spec I used was this one: http://edg.uchicago.edu/~tang/1076.1-1999.pdf

parrt:
should i reopen? why did you close?

z0rgy:
I don't know how I did that :( yes please reopen

PR #163, opened 2015-05-25
Exponent of a base literal was invalid

z0rgy:
Modified support for exponents in base literals to adhere to the standard of using an 'E' or 'e' as opposed to an integer. The standard also states:
"An exponent for a based integer literal shall not have a minus sign." hence the addition of the lexer rules.

Also fixed a spelling typo of BINANRY_BASED_INTEGER to BINARY_BASED_INTEGER

### DIFF
--- a/vhdl/vhdl.g4
+++ b/vhdl/vhdl.g4
@@ -1532,21 +1532,21 @@ waveform_element
 
 //------------------------------------------Lexer-----------------------------------------
 BASE_LITERAL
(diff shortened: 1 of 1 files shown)


z0rgy:
Just thought of a better solution rather than limiting the lexer to only allow +, give that responsibility to the visitor/listener by adding in a comment describing the limitations of the input as the lexer cannot give a useful error message as the listener/visitor can.

PR #167, opened 2015-06-01
Rename scala.g4 to Scala.g4

superraz375:
Fixes the grammar name mismatch


z0rgy:
Did you also encounter the issue of when having a lowercase grammar file and grammar name, then the java does not compile?

parrt:
they both have to have the same case. on unix this will usually fail. yep.

MennaEwas:
You just need to make sure that the line "grammar [FileName];"
has the same name as your File.g name.

PR #169, opened 2015-06-21
Update Antlrv4 grammar & add StringTemplate grammar

grosenberg:
Update Antlrv4 grammar & add StringTemplate grammar


parrt:
Hi Gerald, looks interesting. sorry for delay. racing to get 4.5.1 out.

GRosenberg:
No problem.  Was not expecting an automatic accept - the restructuring is worth looking at to see if it is appropriate for the repository version.  I would think that it should be idiomatic (whatever that means) as well as a showcase of the available features.

teverett:
@grosenberg would it be possible for you to add a pom.xml to the STG grammar, add some test cases and reference the STG pom from the root pom?  This would allow the test suite to verify your grammar after every change. It would also allow the test suite to verify that no breaking changes to ANTLR4 are made.

parrt:
A very interesting idea. @teverett  is the mvn expert here so I'll leave it up to him :)

parrt:
@grosenberg  can you remind me from where you derived grammars? I think maybe I will grab the grammar I have for the intellij plug-in and insert that.  I need to get a good antlr v4 version of the StringTemplate grammar two so maybe I can use yours there.

teverett:
If you'll do the pull @parrt, I'l do the pom.xml now and submit a pull for that.

parrt:
ok, maybe I'll use this as a guide and create a new PR based upon Gerald's work here. let me play around a little bit.

parrt:
actually,Pulling into a local branch now to review as it looks pretty good.

PR #174, opened 2015-07-04
replace literal with token

kasonyang:
replace literal with token
for example: 
`annotation* 'package' qualifiedName ';' `
====> `annotation* PACKAGE qualifiedName SEMI`

### DIFF
--- a/java/Java.g4
+++ b/java/Java.g4
@@ -39,18 +39,17 @@
  *  $ grun Java compilationUnit *.java
  */
 grammar Java;
-
 // starting point for parsing a java file
 compilationUnit
     :   packageDeclaration? importDeclaration* typeDeclaration* EOF
     ;
 
 packageDeclaration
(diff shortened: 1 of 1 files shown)


parrt:
Hi. What is the motivation to replace? thanks.

kasonyang:
Yeah,the grammer has many lexer rules(Tokens),but it seems that they have not been used in the parser rules. I don't understand it.
In my oppion,it is helpful to use lexer rules in parser rules when we write a visitor.
For example:
A parser rule:
```
    expression:
    expression DOT Identifier
    |   expression DOT THIS
```
We can tell the parser rule by this java code:
```java
    if(ctx.THIS() != null){
        //matched the second rule
    }else{
        //matched the first rule
    }
```
What do you think?

parrt:
I like to see the explicit '.' not DOT in the grammar but you can still use DOT in your code since DOT : '.' ; exists, right? You've not added any lexical rules.

Issue #179, opened 2015-07-10
ECMAScript.g4 - problems with automatic semicolon insertion


cshapeshifter:
I'm not sure what exactly is going on, but I notice that even in very simple cases, automatic semicolon insertion is not working properly using the [ECMAScript grammar](https://github.com/antlr/grammars-v4/blob/master/ecmascript/ECMAScript.g4), and that the resulting trees are missing branches, essentially mangling follow-up lines into branch with the missing semicolon. I've created a test case gist [here](https://gist.github.com/cshapeshifter/5ff47f7652918e5a5717#file-broken-js)

The resulting output is included, but as long as java/javac is on the PATH, you can run the test case yourself by simply doing
```
git clone https://gist.github.com/5ff47f7652918e5a5717.git
cd 5ff47f7652918e5a5717
chmod +x run-test.bash
./run-test.bash
```
It will download antlr 4.5, generate the parser and use it to parse two files. One with semicolons, and one where some are missing. These are the two examples:

```
function Test() {
  o = {};
  o.n = "alice";
  o.n = null;
  o.n = "bob";
}
```
```
function Test() {
  o = {};
  o.n = "alice"
  o.n = null
  o.n = "bob"
}
```
In the second example, a lot of errors are printed and the resulting tree is essentially mangling the "null" and "bob" lines into the "alice" branch.
```
line 4:2 no viable alternative at input 'function Test() {\n  o = {};\n  o.n = "alice"\n  o'
line 1:16 extraneous input '{' expecting {'[', '(', ';', ',', '=', '?', '.', '++', '--', '+', '-', '*', '/', '%', '>>', '<<', '>>>', '<', '>', '<=', '>=', '==', '!=', '===', '!==', '&', '^', '|', '&&', '||', '*=', '/=', '%=', '+=', '-=', '<<=', '>>=', '>>>=', '&=', '^=', '|=', 'instanceof', 'in'}
line 4:2 extraneous input 'o' expecting {'[', '(', ';', ',', '=', '?', '.', '++', '--', '+', '-', '*', '/', '%', '>>', '<<', '>>>', '<', '>', '<=', '>=', '==', '!=', '===', '!==', '&', '^', '|', '&&', '||', '*=', '/=', '%=', '+=', '-=', '<<=', '>>=', '>>>=', '&=', '^=', '|=', 'instanceof', 'in'}
line 5:2 no viable alternative at input '.n = null\n  o'
[...]

PR #187, opened 2015-07-13
Bug fix in ECMAScript (related to issue #179)

adrianherrera:
An expression statement can also be terminated by a new line character (not
just a semicolon) as per ECMAScript's automatic semicolon insertion rules
(Section 7.9 of Standard ECMA-262).

This does not completely solve the problems raised in issue #179, but goes some way towards solving it.


adrianherrera:
Needs more work.

Issue #214, opened 2015-09-01
Ecmascript grammar does not parse getters/setters


renatahodovan:
Given a the following javascript source:

``` javascript
a = {
    get b() {}
}
```

When trying to parse this with https://github.com/antlr/grammars-v4/blob/master/ecmascript/ECMAScript.g4 it results in a no viable alternative error.

If BailErrorStrategy is set then it shows that the fail comes from the ParserATNSimulator:

``` javascript
line 2:8 no viable alternative at input 'get b'
Exception in thread "main" org.antlr.v4.runtime.misc.ParseCancellationException
    at org.antlr.v4.runtime.BailErrorStrategy.recover(BailErrorStrategy.java:75)
    at ECMAScriptParser.propertyAssignment(ECMAScriptParser.java:2781)
    at ECMAScriptParser.propertyNameAndValueList(ECMAScriptParser.java:2622)
    at ECMAScriptParser.objectLiteral(ECMAScriptParser.java:2565)
    at ECMAScriptParser.singleExpression(ECMAScriptParser.java:3927)
    at ECMAScriptParser.expressionSequence(ECMAScriptParser.java:3075)
    at ECMAScriptParser.singleExpression(ECMAScriptParser.java:4211)
    at ECMAScriptParser.expressionSequence(ECMAScriptParser.java:3075)
    at ECMAScriptParser.expressionStatement(ECMAScriptParser.java:949)
    at ECMAScriptParser.statement(ECMAScriptParser.java:480)
    at ECMAScriptParser.sourceElement(ECMAScriptParser.java:366)
    at ECMAScriptParser.sourceElements(ECMAScriptParser.java:311)
    at ECMAScriptParser.program(ECMAScriptParser.java:255)
    at ECMAScriptProcessor.processJSFile(ECMAScriptProcessor.java:67)
    at ECMAScriptProcessor.walk(ECMAScriptProcessor.java:48)
    at ECMAScriptProcessor.main(ECMAScriptProcessor.java:33)
Caused by: org.antlr.v4.runtime.NoViableAltException
    at org.antlr.v4.runtime.atn.ParserATNSimulator.noViableAlt(ParserATNSimulator.java:1886)
    at org.antlr.v4.runtime.atn.ParserATNSimulator.execATN(ParserATNSimulator.java:486)
    at org.antlr.v4.runtime.atn.ParserATNSimulator.adaptivePredict(ParserATNSimulator.java:412)
[...]

PR #219, opened 2015-09-09
Fixing getter/setter definitions in ECMAScript grammars

renatahodovan:
Earlier, the getter and setter rules only consumed the first
'get'/'set' part of an expression but not the following
property name.
The patch fixes this in the various ECMAScript grammars.

### DIFF
--- a/ecmascript/ECMAScript.CSharpTarget.g4
+++ b/ecmascript/ECMAScript.CSharpTarget.g4
@@ -732,11 +732,11 @@ futureReservedWord
  ;
 
 getter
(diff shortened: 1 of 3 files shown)


renatahodovan:
Proposed solution for #214
cc @bkiers @adrianherrera @vitorelli

bkiers:
Sorry, missed this conversation. Looks good to me @renatahodovan. 

@parrt, could you merge this?

KvanTTT:
@teverett I guess we have to close it.

PR #224, opened 2015-10-15
Completely new PHP grammar (5.6).

KvanTTT:
Removed incorrect test file (var1.php), because of this syntax incorrect:
```
static $a3 = 12.0 e 13.0;		
-static $a4 = 12.0 e 2;
```

Unfortunaly, this grammar contains actions (for C# runtime), because it impossilble to implement full PHP grammar without them. I did not completely finished PHP grammar, but pushed this pull request for watching purpose. Further commits will be pushed to grammars repo. The advantage of final grammar is it is able to parse php code parts inside html, script blocks. Also it is able to parse [alternative syntax](http://php.net/manual/en/control-structures.alternative-syntax.php) structs, [heredoc](http://php.net/manual/en/language.types.string.php#language.types.string.syntax.heredoc) and other complex things. However, this grammar is pretty neat (compare to Phalanger) and have a satisfactory performance.

I tested this grammar on [phpBB](https://github.com/phpbb/phpbb), [Zend Framework](https://github.com/zendframework/zf2), [Yii Framework](https://github.com/yiisoft/yii) except some files with bad syntax. Also I tested the most part of [WebGoatPHP](https://github.com/shivamdixit/WebGoatPHP). I'll add more test files with complex syntax later.


parrt:
Hi. "I did not completely finished PHP grammar, but pushed this pull request for watching purpose."  Does that mean I should wait to merge?

KvanTTT:
Now this grammar tested on many php files. Only that grammar syntax may be improved and cleared. 

Does Maven can support C# runtime for tests? If it does, then maybe you should wait me when I add test samples. Otherwise this pull request can be merged now and I'll add tests data later if it is necessary.

parrt:
mvn should be able to run whatever. :)  Adding examples would be great like the other grammars.

KvanTTT:
Ok, I'll add test samples before merge :)

parrt:
Hi. Trying to catch up. Any examples for php?

KvanTTT:
Hi, parrt. Ok, I upload examples when I will have free time.

PR #225, opened 2015-10-26
update to /cpp/

teverett:
After the latest changes to /cpp/ the unit tests didn't pass.  A quick change to the pom.xm was required.


teverett:
Added travis support so that builds are done automatically when changes are committed.

PR #237, opened 2015-10-31
 added support for preprocessor line directives with multiple lines 

timeraider4u:
added support for preprocessor line directives with multiple lines, see https://gcc.gnu.org/onlinedocs/gcc-2.95.3/cpp_1.html#SEC2

added a test case for preprocessor directives

### DIFF
--- a/c/C.g4
+++ b/c/C.g4
@@ -868,7 +868,14 @@ SChar
     ;
 
 LineDirective
-    :   '#' Whitespace? DecimalConstant Whitespace? StringLiteral ~[\r\n]*
+    :   '#' 
+        ('define' | 'include' | 'ifdef' | 'ifndef' |
+        'undef' | 'if' | 'else' | 'endif' | 
+        'eif')
+        Whitespace 
+        (~[\r\n] 
+          | '\\' [\r\n] 
+         )*
         -> skip
     ;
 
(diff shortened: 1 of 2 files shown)


parrt:
The preprocessor commands are so complex that typically the preprocessor is run before sending the input to the actual compiler. On the other hand, doesn't seem to hurt anything to include those. Users of this grammar should note that the preprocessor allows non-C code. Imagine:

```C
#define FOO int i;
FOO
```

`FOO` is not valid C code.

Hmm... I'm wondering if this really belongs here or in a separate grammar call CPreproc.g4?

timeraider4u:
I totally agree with you that it would be better to run this ANTLR grammar on an intermediate file instead of parsing C files which include (probably nasty) preprocessor directives. On the other hand, a user might want to deal with them independently after parsing the C file, e.g. show me all expressions where an identifier, which has been defined in a preprocessor statement with the name "foo", is used.

parrt:
Hmm...in general preproc and C parsing have to be done separately and then info shared between them. How about creating separate cpreproc grammar?

timeraider4u:
Yes, I think you are right. Okay, I will give it a try, but I can not guarantee for anything ;-).

parrt:
ok, great! thanks for all your help!

PR #267, opened 2015-11-25
PLSQL grammar

KvanTTT:
Translated from https://github.com/porcelli/plsql-parser for ANTLR3.
Actions removed (grammar is runtime-independent).
Also added tests data except **function06.sql**, **string01.sql**, **xmltable01.sql**.
TODO: code fixing with TODO headers.


parrt:
thanks!

PR #276, opened 2015-12-28
version 1.1: DFHRESP and DFHVALUE; CALL BY VALUE; REMARKS paragraph; …

uwol:
…fixes


zenware:
Is there any way to define the preprocessor as a lexical grammar, similar to how the CSharp grammar file imports a CSharpPreprocessor lexical grammar file? I'm currently investigating this independently as well.

PR #285, opened 2016-01-16
`iostream.h` is deprecated.

nabeelomer:
iostream.h is deprecated by those compilers that provide it, iostream is part of the C++ standard. There is no mention of iostream.h at all in the current C++14 standard.


parrt:
hi. but what about earlier compilers?  Also there's no point in changing an example at a location that is stripped out by the preprocessor, right?

nabeelomer:
Sorry I'm late. I was pointing out something that the standard says is deprecated.

PR #304, opened 2016-02-17
IFILE handling changes. Required separation of lexer and parser grammars

NormanDunbar:
Hi Ter,

another change to the tnsnames grammars I'm afraid. It turned out that Oracle allowed IFILE filenames to be unquoted, double or single quoted. Because of this, I had to use modes in the lexer and thus had to separate out the lexer and parser grammars. 

I've also updated a few of the readme files as URLs etc had changed.

Cheers,
Norm.


parrt:
thanks!

NormanDunbar:
Interesting indeed. That's one he'll of a log to trawl through too! From what I can see it's failing to parse the test harness that I use. That hasn't changed from previous releases other than it moved from the examplesdoc to the examples folder, I assume the old examples folder got the new examplesdoc folder when the new xml file appeared? Something to do with testing?

Perhaps the xml file is telling the test harness that my test harness is in the examplesdoc folder rather than examples?

I tested the whole lot in a separate sandbox, as per the readme.md instructions today, and all was well.

I have no idea what else it could be, but I'll check again tomorrow (UK) time.


Cheers,
Norm.

NormanDunbar:
Spooky. My post from last night has now appeared beneath the two from this morning! Ignore the one immediately above this one.

Cheers, 
Norm.

PR #331, opened 2016-03-22
https://developer.apple.com/library/ios/documentation/Swift/Conceptua…

natefrechette:
…l/Swift_Programming_Language/RevisionHistory.html#//apple_ref/doc/uid/TP40014097-CH40-ID459

- added #selector syntax
- fixed inconsistency with swift grammar & what actually works in Xcode (empty closure expression)
- added a new explicit member expression (_:) to postfix expressions (added in swift 2.2)
- Fixed inconsistencies with swift grammar and swift documentation changes for identifiers


parrt:
thanks Nate!

PR #342, opened 2016-04-14
Clean-up the ANTLR4 lexer grammar.

renatahodovan:
* The LexUnicode lexer grammar was imported by LexBasic but
  was not used anywhere (removed).
* The LexBasic lexer grammar also defined many unused rules
  (maybe some kind of legacy code) (removed too).
* Although there is a fragment defined with the alternation of
  horizontal and vertical whitespaces (Ws), it was not used
  but instead the alternation was inserted multiple times (replaced).


parrt:
thanks!

PR #345, opened 2016-04-17
Various changes for new antlr version

teverett:


KvanTTT:
Maybe it's better to add underscore char instead of last char duplication?

teverett:
Good idea.  Fixed.

PR #347, opened 2016-04-18
Lexer rules must not refer to parser rules.

renatahodovan:
Lexer rules can refer only to other tokens or fragments.
However the lexerAtom rule in the antlr4 parser grammar
enables parser rule references too.
The patch removes the wrong alternative.

### DIFF
--- a/antlr4/ANTLRv4Parser.g4
+++ b/antlr4/ANTLRv4Parser.g4
@@ -345,7 +345,6 @@ ebnfSuffix
 lexerAtom
 	:	characterRange
 	|	terminal
-	|	RULE_REF
 	|	notSet
 	|	LEXER_CHAR_SET
 	|	DOT elementOptions?


whitten:
I'm sure that this has been discussed, but what about a language which depends on the interpretation of a an ID into a SUBROUTINE NAME or a VARIABLE NAME depending on if it is referenced in an IF or a GOSUB ?

parrt:
That is a symbol table issue. This is about antlr ref'ing a parser rule from a lexer rule, which is not semantically valid for antlr grammars. :)

renatahodovan:
I'm not sure I get your point. This patch has fixed an inconsistency between how ANTLRv4 works and how this grammar is written.

Current ANTLRv4 does not allow rule references in lexer rules. See this short example:

```
grammar Test;

start
    : Token
    ;

parser_rule
    : Token2
    ;

Token
    : 'a' parser_rule
    ;

Token2
    : 'b'
    ;
```

If you try to pass it to ANTLR, you'll get the following error:

```
error(160): Test.g4:12:10: reference to parser rule parser_rule in lexer rule Token
```

However, if you write your own tool to interpret antlrv4 grammars, you will most probably use the lexer/parser here. Which would incorrectly process the above input without signalling any failure. This patch just fixed that inconsistency.

Issue #355, opened 2016-04-21
PL/SQL exception declaration not parsed


stephendrew:
It falls into the variable declaration instead as EXCEPTION is treated as regular ID rather than reserved keyword.

PR #356, opened 2016-04-21
Do not treat EXCEPTION as regular ID

stephendrew:
As per issue #355

### DIFF
--- a/plsql/plsql.g4
+++ b/plsql/plsql.g4
@@ -2203,7 +2203,7 @@ regular_id
     | ERRORS
     | ESCAPE
     | EVALNAME
-    | EXCEPTION
+    //| EXCEPTION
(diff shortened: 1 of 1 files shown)


KvanTTT:
Seems to me that build has been broken due to this PR: https://travis-ci.org/antlr/grammars-v4/jobs/124887003

stephendrew:
Indeed.  There seems to be some general confusion in the grammar between where certain identifiers can be used.

EXCEPTION cannot be used a standard identifier (rule: regular_id) but SQL column_name also maps to regular_id, and EXCEPTION is both a valid table name and column name.

Maybe this change should be backed out for the time being - I have been working on this grammar quite a bit as it is quite incomplete, and I hope to have a few more additions which I will run through the tests thoroughly next time!

PR #361, opened 2016-04-26
Fix typo in the Python version of LexerAdaptor.

renatahodovan:
Python uses the len() function to detect the size of a list.
However, in the LexerAdaptor of the antlr4 grammar size()
is used (mistakenly kept from the JAVA version).
This is fixed now.


teverett:
@renatahodovan could you update to the tip of grammars-v4 and resubmit your pull please?

renatahodovan:
Updated.

teverett:
ok the tests pass now.  @parrt  can we merge this one?

PR #362, opened 2016-04-30
fixed plsql build

teverett:
build fixed, by removing a test case which now fails.  I'm not sure we should have committed a change that broke a test case.


parrt:
Sorry, I assume it was failing due to another error, which I intended to go back and fix ;)

PR #373, opened 2016-05-13
Add support for negative integer and floating point literals in Java.g4.

npgall:
This PR will hopefully add support for negative integer and floating point literals to the Java.g4 grammar.

I am not an expert on Antlr4, but in my project when I imported this grammar I was not able to get it to parse negative/signed integer or floating point literals "-1" or "-1.5".

Please double-check that this makes sense before merging!

Many thanks!
Niall

### DIFF
--- a/java/Java.g4
+++ b/java/Java.g4
@@ -671,7 +671,7 @@ IntegerLiteral
 
 fragment
 DecimalIntegerLiteral
-    :   DecimalNumeral IntegerTypeSuffix?
+    :   Sign? DecimalNumeral IntegerTypeSuffix?
     ;
 
 fragment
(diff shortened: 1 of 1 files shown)


parrt:
Hi. looks to me like the spec doesnt' have sign in front of the literals. that's part of expression. See [spec](https://docs.oracle.com/javase/specs/jls/se8/jls8.pdf) page 25ish.

npgall:
I see! I guess I will need to work around this in my use case then. So I guess you can disregard this PR. Many thanks for checking, and thanks for this grammar!

parrt:
no problem

npgall:
I just realised you are the author of the amazing Antlr itself too Terrence! I'm using it to parse two dialects in my project and it's working great. Impressive stuff thanks so much for releasing something like this!

parrt:
hooray! Glad you enjoy.

PR #381, opened 2016-05-17
Rfc822

teverett:


parrt:
date time would be great to have a separate grammar. could we break apart? dates are really useful.

teverett:
Current that is what this is; DateTime. If I add more of the RFC, I am happy to break it apart at that time.

parrt:
Maybe we could call it datetime.g4 then? ;)

parrt:
nobody will find anything as Rfc822

teverett:
done

parrt:
awesome. what is khubla.com by the way?

Issue #368, opened 2016-05-05
What should be a block comment: /**/ in antlr4.g is not recognized as such 


ibilenjkij:
The character sequence: /**/ will be recognized as a start of a doc comment and will consume everything up until it finds another */ sequence.

Issue #272, opened 2015-12-18
Antlr4 Grammar does not work on the CPP Grammar


h4ck3rm1k3:
Attempting to parse the CPP14  grammar 

copy the cpp/CPP14.g4 into antlr4/examples/CPP14.g4
changed the pom to be verbose and print tree.
run mvn test
the output is here: 
antlr4/examples/CPP14.g4.err
https://github.com/h4ck3rm1k3/grammars-v4/commit/4cd2d16966fd422c0134fb9ebf9c20a92976df2e#diff-bfb944532751425acb6fcf0c7eeeb498

teverett:
You are correct that the Antlr4 grammar doesn't part CPP14.g4.  Here is a more correct output from the failed test:

```
oscar:antlr4 tom$ mvn clean package
[INFO] Scanning for projects...
[WARNING] 
[WARNING] Some problems were encountered while building the effective model for com.antlr.grammarsv4:antlr4:jar:1.0-SNAPSHOT
[WARNING] 'build.plugins.plugin.version' for org.antlr:antlr4test-maven-plugin is missing. @ com.antlr.grammarsv4:antlr4:[unknown-version], /Users/tom/projects/grammars-v4/antlr4/pom.xml, line 41, column 12
[WARNING] 
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
[WARNING] 
[WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
[WARNING] 
[INFO]                                                                         
[INFO] ------------------------------------------------------------------------
[INFO] Building ANTLR4 grammar 1.0-SNAPSHOT
[INFO] ------------------------------------------------------------------------
[INFO] 
[INFO] --- maven-clean-plugin:2.5:clean (default-clean) @ antlr4 ---
[INFO] Deleting /Users/tom/projects/grammars-v4/antlr4/target
[INFO] 
[INFO] --- antlr4-maven-plugin:4.5.3:antlr4 (default) @ antlr4 ---
[INFO] ANTLR 4: Processing source directory /Users/tom/projects/grammars-v4/antlr4
[INFO] Processing grammar: ANTLRv4Lexer.g4
[INFO] Processing grammar: ANTLRv4Parser.g4
[INFO] 
[INFO] --- maven-resources-plugin:2.6:resources (default-resources) @ antlr4 ---
[INFO] Using 'UTF-8' encoding to copy filtered resources.
[...]

PR #393, opened 2016-05-26
fix for #368 and #272

teverett:


ibilenjkij:
With regards to #368, can you add the following test grammar:
```antlr
/**/
grammar three;
/**/
```

PR #395, opened 2016-05-29
update

paul-weiss:


teverett:
@paul-weiss ok, this is closed now.  Is it still broken?  If so, Paul can you explain what's broken?

PR #414, opened 2016-06-24
Added CREATE/DROP TYPE and enhanced OPENROWSET

vijaygill:
@teverett 
Hi Tom,

I had to rename and existing rule "view_name" to "simple_name" because same rule could be used for CREATE/DROP TYPE also.

Please share your views.

Thanks.


KvanTTT:
Also please add yourself to contributors list at the top of grammar.

vijaygill:
@KvanTTT 
Hi Tom,

I will do it in next commit. Don't want to have one commit just for that :)

Thanks.

teverett:
@vijaygill I am not a tsql expert, so my views are that the grammar should work properly and more test cases are better :)

vijaygill:
@teverett 

Hi Tom,

I agree with you. In the limited time I got so far, I added these few things. I will add more examples as soon as I get a bit of time.

If you think this PR does not meet the desired level of quality, I will try to get more examples added.

Thanks

PR #473, opened 2016-08-25
VBA - fix operator precedences

foolmoron:
Many operators had incorrect precedences. Fixed to match the order described here: https://msdn.microsoft.com/en-us/library/fw84t893.aspx


teverett:
Could we get some examples added which validate the fix?

foolmoron:
Added some!

PR #480, opened 2016-09-13
Add lexer support for multiline macros in C++

adamtornhill:
The existing C++ grammar skips all pre-processor directives.  However, the grammar fails to handle multiline macros. These macros have to be skipped too, otherwise the parser will fail. This commit adds the necessary support.

### DIFF
--- a/cpp/CPP14.g4
+++ b/cpp/CPP14.g4
@@ -1317,6 +1317,12 @@ noexceptspecification
 ;
 
 /*Preprocessing directives*/
+
+MultiLineMacro
+:
+    '#' (.*? '\\' [\r\n]+)+ (.*? ~[\r\n]*) -> skip
+;
+
 Directive
 :
 	'#' ~[\r\n]* -> skip
--- a/cpp/CPP14.g4
+++ b/cpp/CPP14.g4
@@ -1320,7 +1320,7 @@ noexceptspecification
 
 MultiLineMacro
 :
-    '#' (.*? '\\' [\r\n]+)+ (.*? ~[\r\n]*) -> skip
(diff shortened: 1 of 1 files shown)


adamtornhill:
Sure, I have several examples. Where do you want me to add the tests, because I couldn't find any test folder for the C++ grammar?

Here's a small example that fails without the patch and pass with it:

```
#define Verify(cond, msg)                                                                                              \
    do                                                                                                                 \
    {                                                                                                                  \
        if (!(cond))                                                                                                   \
        {                                                                                                              \
            verRaiseVerifyExceptionIfNeeded(INDEBUG(msg) DEBUGARG(__FILE__) DEBUGARG(__LINE__));                       \
        }                                                                                                              \
    } while (0)
    
void f() {
  int b = 0;
}
```
In the example above, the unpatched grammar would skip the first line (`#define Verify...`) and then fail on the `do` keyword on line two. The reason is that the existing lexer rule for pre-processor directives only handle single line directives.

cpp/CPP14.g4
```diff
@@ -1317,6 +1317,12 @@ noexceptspecification
 ;
 
 /*Preprocessing directives*/
+
+MultiLineMacro
+:
+    '#' (.*? '\\' [\r\n]+)+ (.*? ~[\r\n]*) -> skip
```
KvanTTT: Can it be simplified to `'#' (.*? '\\' [\r\n]+)+ ~[\r\n]+ -> skip`?

adamtornhill:
@KvanTTT Yes, indeed. Thanks for the suggestion. I've pushed an updated version with the simplification.

Issue #487, opened 2016-09-27
Missing whitespaces in VisualBasic6.g4


uwol:
Hi, with 64da7695421acc25143e5d25732d0779ae0c10d0 the whitespaces token (WS) was removed from VisualBasic6.g4. I get the point, that parsing whitespace induces suboptimal parsing performance. However, in Visual Basic 6.0 whitespace is relevant for semantics.

With WS in the grammar _SomeProcedure .Name_ is parsed as ICS_B_ProcedureCall and _SomeObject.Name_ as ICS_B_MemberProcedureCall. Without WS both calls are parsed as ICS_B_MemberProcedureCall.

```
WITH Customer
    SomeProcedure .Name
END WITH

WITH Customer
    SomeObject.Name
END WITH
```

Distinguishing these both cases is relevant for semantic analysis of the AST, i. e. whether to look up procedures or variables. Insofar: Could you please review the commit?

teverett:
I will certainly review the commit.  I suspect that the problem is not with WS but with handling <token>.<token> as a single token rather than two tokens separated by a ".", however I will review.

PR #494, opened 2016-10-04
pr #487

teverett:
- reverted grammar
- added .errors files for form1.vb and form2.vb

Now ready for Ulrich to add .tree files

### DIFF
--- a/vb6/VisualBasic6.g4
+++ b/vb6/VisualBasic6.g4
@@ -5,6 +5,7 @@
 * This software may be modified and distributed under the terms
 * of the BSD 3-clause license. See the LICENSE file for details.
 */
+
 /*
 * Visual Basic 6.0 Grammar for ANTLR4
 *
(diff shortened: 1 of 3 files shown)


vb6/VisualBasic6.g4
```diff
@@ -35,93 +36,96 @@
...
 *
 * v1.0 Initial revision
 */
 
 grammar VisualBasic6;
 
+options
+{
+   language = Java;
+}
```
KvanTTT: What necessity of language declaration here? This grammar does not use actions.

vb6/VisualBasic6.g4
```diff
@@ -35,93 +36,96 @@
...
-   : moduleBodyElement (moduleBodyElement)*
-   ;
-
-moduleBodyElement
-   : moduleBlock
+
+startRule : module EOF;
+
+module : 
+   WS? NEWLINE*
```
uwol: In Visual Basic 6.0 whitespace is relevant for semantics (cf. #487 ). 

```
WITH Customer
    SomeProcedure .Name
END WITH

WITH Customer
    SomeObject.Name
END WITH
```

With WS in the grammar SomeProcedure .Name is parsed as ICS_B_ProcedureCall and SomeObject.Name as ICS_B_MemberProcedureCall. Without WS both calls are parsed as ICS_B_MemberProcedureCall. Distinguishing these both cases is relevant for semantic analysis of the AST, i. e. whether to look up procedures or variables.

vb6/VisualBasic6.g4
uwol: Correct, I suppose, the language declaration can be removed.

PR #512, opened 2016-11-12
 CPP14.g4 bugfix - relational ops in template arguments

shravanrn:
Consider the signature
void TestFunc(vector< ClassA > args, vector < ClassB > args2)

Right now this is parsed as 
- a declaration of a variable args2 
- the type of args2 is vector <A, B> 
    where A = (ClassA > args) 
               B = (vector < B) 
   i.e. a a template instantiation that has two relational expressions that are the arguments

This stems from the priority order of the template argument list which is 
templateargument : constantexpression | typeid | idexpression;

Changing the order to : typeid | constantexpression| idexpression; fixes this.

Tested that this is the same behavior as observed in Microsoft c++ compiler, gcc, clang and icc.

### DIFF
--- a/cpp/CPP14.g4
+++ b/cpp/CPP14.g4
@@ -1239,8 +1239,8 @@ templateargumentlist
 
 templateargument
 :
-	constantexpression
-	| typeid
+	typeid
+	| constantexpression
 	| idexpression
 ;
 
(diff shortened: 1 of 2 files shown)


teverett:
Are you able to add some test cases that fail without this patch, and pass with it?

shravanrn:
Sorry, was a bit busy. Have added the test now.

shravanrn:
Oops, accidentally added a different fix to this pull request as well... just as well though. I was going to submit that after this... can just submit it together instead

PR #536, opened 2016-12-28
Add SwiftSupport.swift

therealbnut:
I've attempted a direct translation of [swift/src/main/java/SwiftSupport.java](https://github.com/therealbnut/grammars-v4/blob/therealbnut-swift-swift-support/swift/src/main/java/SwiftSupport.java) into Swift.

I haven't put much effort into making it more idiomatic Swift as I don't know antlr or its codebase well enough. It would be nice to remove the `-1` sentinel values, perhaps use `nil` instead. Also `EOF` is used inconsistently. I'm not happy with the error handling in this, someone that know antlr better may need to verify how to handle failing `try` calls rather than just `try!`.

I think ideally the `{SwiftSupport. ...}?` parts of the grammar would be split out into separate `.g4` files, so they can be conditionally included based on the target language. This isn't necessary yet though (See [ANTLRv4Lexer.PythonTarget.g4](https://github.com/antlr/grammars-v4/blob/master/antlr4/ANTLRv4Lexer.PythonTarget.g4) for an example).

FYI @parrt

### DIFF
--- a/swift/SwiftSupport.swift
+++ b/swift/SwiftSupport.swift
@@ -1,0 +1,238 @@
+import Foundation
+import Antlr4
+
+fileprivate let TokenEOF = -1
+
+public final class SwiftSupport {
+    /* TODO
+    There is one caveat to the rules above. If the ! or ? predefined operator
+     has no whitespace on the left, it is treated as a postfix operator,
+     regardless of whether it has whitespace on the right. To use the ? as
+     the optional-chaining operator, it must not have whitespace on the left.
+      To use it in the ternary conditional (? :) operator, it must have
+      whitespace around both sides.
+    */
+
+    /*
+    operator-head : /  =  -  +  !  *  %  <  >  &  |  ^  ~  ?
+      | [\u00A1-\u00A7]
+      | [\u00A9\u00AB]
+      | [\u00AC\u00AE]
+      | [\u00B0-\u00B1\u00B6\u00BB\u00BF\u00D7\u00F7]
+      | [\u2016-\u2017\u2020-\u2027]
+      | [\u2030-\u203E]
+      | [\u2041-\u2053]
+      | [\u2055-\u205E]
+      | [\u2190-\u23FF]
+      | [\u2500-\u2775]
+      | [\u2794-\u2BFF]
+      | [\u2E00-\u2E7F]
+      | [\u3001-\u3003]
+      | [\u3008-\u3030]
+      
+     */
+    private static let operatorHead: Set<Int> = {
+        var out = Set<Int>()
+        out.insert(SwiftParser.Tokens.BANG.rawValue)
+        out.insert(SwiftParser.Tokens.LT.rawValue)
+        out.insert(SwiftParser.Tokens.GT.rawValue)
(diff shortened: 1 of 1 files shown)


parrt:
Yeah, we're not really set up for actions in grammars, at least not non-java languages. @teverett any thoughts?

KvanTTT:
I think we should accept grammars with actions for custom runtimes. At present time we have some grammars for not only java runtimes: PHP (Java, C# Sharwell, Python), C# (Java, C# Sharwell) etc., but they are not covered by tests unfortunately. So, we have to improve our testing plugin.

parrt:
Yeah, we've have to port lots of the code from antlr4 tool that launches various runtimes to test things. Not sure I'm in favor of making this repo anything more than a grammar repo...

therealbnut:
@parrt I agree, making this repo grammar only is ideal. Although I think something like [this](https://github.com/therealbnut/grammars-v4/blob/therealbnut-swift-swift-support/swift/Swift.g4#L959) means that SwiftSupport is a necessary part of the grammar (for now):
```antlr
assignment_operator : {SwiftSupport.isBinaryOp(_input)}? '=' ;
```

>   If an operator has whitespace around both sides or around neither side,
     it is treated as a binary operator. As an example, the + operator in a+b
     and a + b is treated as a binary operator.

It's providing information to the parser that's hard/impossible to do with pure `g4`.

I can't think of a better place to put those dependencies, we already have `SwiftSupport.java` in this repo, and only this grammar depends on it.

The alternative is something like [Python3.g4](https://github.com/antlr/grammars-v4/blob/master/python3/Python3.g4#L38) where a big chunk of java code is embedded in the grammar file. Which would be okay IMO if that java code was instead in a file `Python3.JavaTarget.g4`.

therealbnut:
We could possibly simplify it like this (partial implementation):

**Swift.TargetJava.g4**
```antlr
LOOKBEHIND_WS  : { _input.LT(-1).equals(WS) }? ;
LOOKAHEAD_WS   : { _input.LT(1).equals(WS) }? ;
LOOKBEHIND_DOT : { _input.LT(1).equals(DOT) }? ;
```

**Swift.g4**
 
```diff
-assignment_operator : {SwiftSupport.isBinaryOp(_input)}? '=' ;
+assignment_operator :  LOOKBEHIND_WS '='  LOOKAHEAD_WS
+                    | ~LOOKBEHIND_WS '=' ~LOOKAHEAD_WS ;
```

etc.

parrt:
Well, given that it's one predicate, I think we should avoid duplicating the grammar and/or support code for 8 ANTLR targets. One grammar, with sem preds in Java could be considered the "standard" I guess. @teverett what do you think?

teverett:
The core of the problem here is that Antlr is fundamentally a Java tool, that generates Java.  The test suite used by antlr4-grammars is also Java.  The purpose of the antlr4-grammars tree is to provide a library of grammars that are tested and known to work with the latest version of Antlr.  So, as long as there is still SwiftSupport.java, the tests can be run (with the current Java test suite) and the tests pass, I have no objection to SwiftSupport.swift.

therealbnut:
Sorry, I missed the merge - thank you :)

PR #554, opened 2017-02-06
Fix issue in grant statement

ohyesboy:
After fix, it should be able to parse:
GRANT EXECUTE ON [dbo].[closeLead] to Public,XYZPublic

Ref: https://msdn.microsoft.com/en-us/library/ms187965.aspx

### DIFF
--- a/tsql/tsql.g4
+++ b/tsql/tsql.g4
@@ -530,7 +530,7 @@ security_statement
     // https://msdn.microsoft.com/en-us/library/ms188354.aspx
     : execute_clause ';'?
     // https://msdn.microsoft.com/en-us/library/ms187965.aspx
-    | GRANT (ALL PRIVILEGES? | grant_permission ('(' column_name_list ')')?) (ON on_id=table_name)? TO (to_principal=id) (',' to_principal+=id)* (WITH GRANT OPTION)? (AS as_principal=id)? ';'?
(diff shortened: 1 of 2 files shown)


tsql/tsql.g4
```diff
@@ -522,7 +522,7 @@ security_statement
     // https://msdn.microsoft.com/en-us/library/ms188354.aspx
     : execute_clause ';'?
     // https://msdn.microsoft.com/en-us/library/ms187965.aspx
-    | GRANT (ALL PRIVILEGES? | grant_permission ('(' column_name_list ')')?) (ON on_id=table_name)? TO (to_principal=id) (WITH GRANT OPTION)? (AS as_principal=id)? ';'?
+    | GRANT (ALL PRIVILEGES? | grant_permission ('(' column_name_list ')')?) (ON on_id=table_name)? TO (to_principal=id) (',' to_principal=id)* (WITH GRANT OPTION)? (AS as_principal=id)? ';'?
```
KvanTTT: You should use `+=` operator for repeated syntax with star or plus: `to_principal+=id`.

teverett:
@ohyesboy any chance we can fix this?

zcao:
@teverett Do I need to apply += to all `to_pricipal=id` or just from the 2nd on?

teverett:
@zcao I'm not an expert in this grammar.  Perhaps @KvanTTT or @ohyesboy can help.

KvanTTT:
@zcao you should use `+=` operator for repetition elements: star or plus. In this grammar `+=` should be used for all `to_principal` elements.

ohyesboy:
Done, applied += to both places. BTW, @zcao is myself too.

PR #555, opened 2017-02-06
Add support for empty statement that only has ';'

ohyesboy:


tsql/tsql.g4
```diff
@@ -113,6 +115,11 @@ cfl_statement
         state=constant_LOCAL_ID (',' constant_LOCAL_ID)* ')' ';'?                   #raiseerror_statement
     ;
 
+empty_statement
```
KvanTTT: Please use common formatting style as in entire tsql file.

zcao: Done

PR #559, opened 2017-02-07
TSQL-DBCC clause

ohyesboy:
Reference https://msdn.microsoft.com/en-us/library/ms175008.aspx


teverett:
please fix the merge conflict

zcao:
@teverett  Resolved

teverett:
@KvanTTT is this good to merge?

tsql/tsql.g4
```diff
@@ -586,6 +588,14 @@ use_statement
     : USE database=id ';'?
     ;
 
+dbcc_clause
+    : DBCC name=simple_id ('(' expression_list ')')? (WITH dbcc_options)? ';'?
```
KvanTTT: Do we need for `name` label here? `simple_id` is single in this rule.

KvanTTT:
@ohyesboy can you add yourself to contributors list in the header of file?
@teverett I added one comment but overall LGTM :)

tsql/tsql.g4
ohyesboy: This is a simplified version, DBCC clause is quite versatile in MSDN https://msdn.microsoft.com/en-us/library/ms176064.aspx, there could be more simple_id added to the clause in future.

KvanTTT: Ok, got it.

PR #561, opened 2017-02-13
Graphstream DGS grammar

timhemel:
A grammar for the graphstream DGS format:
http://graphstream-project.org/doc/Advanced-Concepts/The-DGS-File-Format/


teverett:
Could you add this into the root pom.xml also please?

timhemel:
Sorry, I was not aware of needing to do that, although the thought had crossed my mind. Entry is added to the root pom.xml.

teverett:
thanks for the new grammar!

PR #565, opened 2017-02-24
Restrict LineTerminators in ECMAScript grammars.

renatahodovan:
According to the ECMAScript 5.1 Standard (and the comments in the existing
ANTLR grammars) LineTerminator is not allowed between a 'return', 'throw',
'break' or 'continue' keyword and the following Identifier or Expression
or between a LeftHandSideExpression and a postfix operator. However,
while the latter case is handled correctly, this restriction is missing
from the prior one. The patch fixes this.


teverett:
Thanks

eugenega:
This commit broke the JS Grammar, see issue https://github.com/antlr/grammars-v4/issues/594

PR #566, opened 2017-02-26
Char escaping, collision and other issues

KvanTTT:
I tested all grammars with upcoming ANTLR 4.7 version.


teverett:
Amazing work!

PR #575, opened 2017-03-15
Extended Unicode grapheme parser grammar

bhamiltoncx:
This PR depends on https://github.com/teverett/antlr4test-maven-plugin/pull/4  as well as ANTLR 4.7.

In this PR, I introduce an extended Unicode grapheme parser based on the following standards:

http://unicode.org/reports/tr29/#Grapheme_Cluster_Boundaries
http://unicode.org/reports/tr51/#def_emoji_sequence

I included test inputs and confirmed everything passed with:

`mvn -Dantlr.version=4.7 -Dantlr4test-maven-plugin.version=1.8 test`


bhamiltoncx:
OK, ANTLR 4.7 is out and we should be able to update grammars-v4!

teverett:
Brilliant.  I am flying tomorrow, so it will have to be this weekend.

PR #596, opened 2017-04-05
Fixing exception in the ECAMScript JavaScript grammar

eugenega:
Fix for:
Parser generated from ECMAScript grammar throws exception.
see issue https://github.com/antlr/grammars-v4/issues/594

### DIFF
--- a/ecmascript/ECMAScript.JavaScriptTarget.g4
+++ b/ecmascript/ECMAScript.JavaScriptTarget.g4
@@ -269,21 +269,21 @@ iterationStatement
 ///     continue ;
 ///     continue [no LineTerminator here] Identifier ;
 continueStatement
- : Continue ({!this.here(LineTerminator)}? Identifier)? eos
+ : Continue ({!this.here(ECMAScriptParser.LineTerminator)}? Identifier)? eos
  ;
 
 /// BreakStatement :
 ///     break ;
 ///     break [no LineTerminator here] Identifier ;
 breakStatement
- : Break ({!this.here(LineTerminator)}? Identifier)? eos
+ : Break ({!this.here(ECMAScriptParser.LineTerminator)}? Identifier)? eos
  ;
 
 /// ReturnStatement :
 ///     return ;
 ///     return [no LineTerminator here] Expression ;
 returnStatement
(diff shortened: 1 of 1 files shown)


teverett:
Could we please get a test case submitted for this?

eugenega:
Can you give me an example of the test?
The test will require this. Is it even possible?
1 getting ECAMScript.JavaScript.g4 file, renaming it to ECAMScript.g4 (since it contains ECAMScript grammar)
2. generating parser and lexer 
java -jar antlr-4.7-complete.jar -Dlanguage=JavaScript ECMAScript.g4
3. Run this code:
var antlr4 = require('antlr4/index');
var ECMAScriptLexer = require('./ECMAScriptLexer').ECMAScriptLexer;
var ECMAScriptParser = require('./ECMAScriptParser').ECMAScriptParser;

var inputStream = new antlr4.InputStream("a=x(10)");
var lexer = new ECMAScriptLexer(inputStream);
var tokens = new antlr4.CommonTokenStream(lexer);
var parser = new ECMAScriptParser(tokens);
parser.buildParseTrees = true;
var tree = parser.program();
4. Check that no exceptions.

teverett:
ok, on step 1, I understand.  The current testing framework doesn't account for the fact that there may be multiple grammar files for a language, depending on the output target.  That is something that the author may need to address.

With respect to the rest, the purpose of the /examples directory is to allow the testing framework to test against a number of grammars so that as the grammar evolves and antlr4 evolves we can ensure that all grammars in the tree work properly.

You've made a significant fix, so I'd like to see a test added that fails if yourself or someone else breaks your fix.

eugenega:
I do not mind writing test if that is possible. However, according to your note above it is not possible.
Grammar is broken at the moment and needs to be fixed. Fix is simple/safe/ready. 
Please advise what to do.

teverett:
@eugenega.   Is it possible for you to commit a JS file to /examples that will only parse properly if your change is committed?

eugenega:
It is not possible to generate JS Parser and Lexer to be tested w/o renaming file first. JS grammar name and file name differ and antlr throws exception.

PR #653, opened 2017-05-03
Update css3.g4

msbruno:
Adding suport to @import url(STRING)
Ex.:

@import url("Style.css");
@import "Style.css";

.c {
    color: black;
}


teverett:
@msbruno could you please add a test case into the /examples dir that uses this new functionality?

msbruno:
Sure, it is done :)

Issue #655, opened 2017-05-04
Issue with parsing declaration in C


prateekgulati:
When I try to parse code like 
`int x;` or `long int x;` then in declaration, all the tokens get parsed as declarationSpecifiers.
Can I try to fix this issue?

teverett:
@prateekgulati, anyone is welcome to fix any issue and to submit pull requests.

KvanTTT:
Try to use the following declaration rule
`declarationSpecifiers initDeclaratorList | declarationSpecifiers`

instead of the old one:
`declarationSpecifiers initDeclaratorList?`

I'll upload an updated version of Objective-C grammar soon with such improvement. It works fine.

prateekgulati:
This works fine. Thanks

KvanTTT:
Could you also suggest a pull request for this?

PR #677, opened 2017-05-09
Fixed declaration rule

prateekgulati:
resolve issue #655

### DIFF
--- a/c/C.g4
+++ b/c/C.g4
@@ -175,7 +175,8 @@ constantExpression
     ;
 
 declaration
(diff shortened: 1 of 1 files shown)


teverett:
@prateekgulati  could you please add a test case to your commit for this fix?

prateekgulati:
Some programs in example directory already use the updated functionality in the commit.

KvanTTT:
Perhaps we should use a **.tree** file with proper parse tree here.

PR #694, opened 2017-05-12
antlr4: do not refer parser rules from lexer.

renatahodovan:
Until now, the labeledLexerElement rule of the antlr4 parser grammar
enabled having a block element on its right hand side, which meant
enabling parser rule references in lexer rules. The patch fixes this.


teverett:
@renatahodovan  thank-you.

PR #695, opened 2017-05-12
Add `ENCRYPTION` and `SCHEMABINDING` options for function options.

Muhamedali:
According with https://docs.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql function options has `ENCRYPTION` and `SCHEMABINDING` values

### DIFF
--- a/tsql/tsql.g4
+++ b/tsql/tsql.g4
@@ -370,7 +370,9 @@ procedure_option
     ;
 
 function_option
-    : RETURNS NULL ON NULL INPUT
+    : ENCRYPTION
+    | SCHEMABINDING
+    | RETURNS NULL ON NULL INPUT
     | CALLED ON NULL INPUT
     | execute_clause
     ;


teverett:
@Muhamedali thank you for your contribution.  Could you please add a test case that uses ENCRYPTION and SCHEMABINDINGS?

Muhamedali:
So, my test case hasn't applied sense, because same operation as in function possible do another way.

```
CREATE TABLE dataTable (
  intValue int,
  dateValue date
);

CREATE OR ALTER FUNCTION [dbo].[get_some_int] (@date varchar(10))   
RETURNS int  
WITH SCHEMABINDING, RETURNS NULL ON NULL INPUT, EXECUTE AS 'dbo' 
AS   
BEGIN  
    declare @res int; 
    SELECT @res = intValue FROM [dbo].[dataTable] WHERE dateValue > @date; 
    return @res; 
END;

```
But it demonstrates main purpose of  SCHEMABINDING option - reference to a table inside function in full-qualified style - `[dbo].[dataTable]`

PR #698, opened 2017-05-14
Normalized line breaks (\n)

KvanTTT:
We should use identical line breaks for grammar and pom files.


teverett:
@KvanTTT oh that is awesome.  learn something new every day!

PR #706, opened 2017-05-16
ecmascript: objectLiteral must not contain a single comma only

renatahodovan:
The objectLiteral rule must match on zero or more elements,
separated by commas and if it contained at least one element,
then it may end with an optional trailing comma. However,
the current grammar enables a single comma, without any other
elements. The patch fixes this in every ecmascript targets.


teverett:
@renatahodovan  thanks!

PR #707, opened 2017-05-17
COOL: Declaring fragment lexical rules

linonetwo:
Make case-insensitive rules clear.

### DIFF
--- a/cool/COOL.g4
+++ b/cool/COOL.g4
@@ -78,28 +78,28 @@ COMMENT: OPEN_COMMENT (COMMENT|.)*? CLOSE_COMMENT -> channel(HIDDEN);
 ONE_LINE_COMMENT: '--' .*? '\n' -> channel(HIDDEN);
 
 // key words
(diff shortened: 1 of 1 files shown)


linonetwo:
This CI takes too long a time to build...

KvanTTT:
@linonetwo you can comment out all languages except COOL and then test it locally with maven plugin.
@teverett can we leave only one oraclejdk7 and remove other oraclejdk8, openjdk7 JDKs from .travis configuration? What aspects do they check?

linonetwo:
Hold this for a while, I'm building a code transformer, and find some bug in this grammar.  :P

teverett:
@linonetwo- holding

@KvanTTT hm.  Well the advantage is that there have been instances in the past where a grammar worked on on JDK and not another.  I'm not sure what the justification to remove checking across JDK's would be.  It is a very long build, but actually the 3x JDK builds run in parallel.

linonetwo:
OK, I've finished it!

teverett:
@linonetwo  thanks and congrats

Issue #714, opened 2017-05-19
PLSQL separate package spec and body


stephendrew:
Create separate rules for `create_package` and `create_package_body`.

Issue #717, opened 2017-05-19
PLSQL return statements minor fix


stephendrew:
...returns an `expression` not a `condition`

Issue #716, opened 2017-05-19
PLSQL allow correct function/proc specs & bodies in declare sections


stephendrew:
Previously, `declare_spec` only allowed:

- `create_procedure_body`
- `create_function_body`

Fixed to:

- `procedure_spec`
- `procedure_body`
- `function_spec`
- `function_body`

Issue #715, opened 2017-05-19
PLSQL multipart identifiers not allowed in certain places


stephendrew:
The following rules were allowing identifiers with periods (`id_expression*`) in them:

- `procedure_spec`
- `function_spec`
- `procedure_body`
- `function_body`
- `variable_declaration`
- `subtype_declaration`
- `exception_declaration`
- `type_declaration`

These have been replaced with `identifier`.

PR #718, opened 2017-05-19
Declare spec fixes

stephendrew:
Fixes issues

#714 
#715 
#716 
#717

### DIFF
--- a/plsql/plsql.g4
+++ b/plsql/plsql.g4
@@ -575,12 +575,12 @@ variable_declaration
     ;
 
 subtype_declaration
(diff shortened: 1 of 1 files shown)


plsql/plsql.g4
```diff
@@ -41,6 +41,7 @@ unit_statement
     | create_function_body
     | create_procedure_body
     | create_package
+	| create_package_body
```
KvanTTT: @stephendrew could you please replace tabs on spaces in this grammar and check it in future (stick a grammar indentation)?

stephendrew: Sure, sorry.

PR #725, opened 2017-05-25
Rewrite CSS3 grammar according to W3C spec

huan086:
Works for the style sheets I use on my website, unlike the original which does not handle CSS3.


teverett:
@huan086  thanks

PR #761, opened 2017-06-17
Update Java identifier fragments per spec

BurtHarris:
This is intended to make the Java and Java8 grammars both more compliant with the official specs, and cross-target.   It appears they were mistakenly extended to include non-ASCII characters.   See http://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-Identifier for official definitions.

This should address issue antlr/antlr4#1885, but is untested at this point.   I'm not a Java dev, and running `mvn test` is about all I know.   It fails.

### DIFF
--- a/java/Java.g4
+++ b/java/Java.g4
@@ -975,26 +975,22 @@ Identifier
     :   JavaLetter JavaLetterOrDigit*
     ;
 
+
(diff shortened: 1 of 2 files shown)


BurtHarris:
I'll re-submit

PR #762, opened 2017-06-17
Java language identifier fragments per spec

BurtHarris:
This is intended to make the Java and Java8 grammars both more compliant with the official specs, and cross-target (by removing unnecessary predicates.)  It appears these were mistakenly extended to include non-ASCII characters.    Intended to fix antlr/antlr4#1885

Based on http://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-Identifier as "official" definitions.   I am **not** a Java developer, so I can only go by what I read.

Note: There are tests failing, but I *think* not due to this change.    Master seems to be failing as well.   First failure message on my machine.  If it breaks on travis-ci, I may need a Java guy to help me understand why.

```
[INFO] Unicode TR29 Grapheme Cluster Boundary Parsing ..... FAILURE [  0.251 s]
````

### DIFF
--- a/java/Java.g4
+++ b/java/Java.g4
@@ -975,22 +975,22 @@ Identifier
     :   JavaLetter JavaLetterOrDigit*
     ;
 
+
+// The "Java letters" include uppercase and lowercase ASCII Latin letters A-Z (\u0041-\u005a), 
+// and a-z (\u0061-\u007a), and, for historical reasons, the ASCII underscore (_, or \u005f) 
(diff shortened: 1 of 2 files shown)


jundl77:
This does stop the generation of invalid code, at least for me.

cowang:
See my [comments ](https://github.com/antlr/antlr4/issues/1885)for more than you ever wanted to know.

BurtHarris:
I'm going to let the Java experts hash this out.

Issue #528, opened 2016-12-14
Java8 grammar unbearably slow


ghost:
I work on a source level static analysis tool and tried to use the Java8 grammar. It's impossible to use, because parsing a moderate Java project takes almost *a minute* ! I've tried my parser on multiple Java projects, and it's the same. I understand the Java7 grammar is faster, but that doesn't help me since we must target Java 8 (Java 7 is EOL'ed anyway)

What can be done? As it is, the grammar is complete, yet unusable.

pureconfig:
I should add that Eclipse (JDT) parses the same sample project in less than two seconds.

KvanTTT:
I'am working on Java8 optimized grammar but not so fast as I would like to.

pureconfig:
@KvanTTT Not working as fast as you would like to, or is the grammar not as fast as you would like to? :)

KvanTTT:
Oops, ambiguity :) the statement related to me.

maybeec:
Any progress here? Maybe an intermediate version of the optimized grammar, I can try? :)

KvanTTT:
No :(

2017-03-01 0:06 GMT+03:00 Malte Brunnlieb <notifications@github.com>:

jundl77:
Having the same issue, but with Javascript target

KvanTTT:
Relative simple advice: rewrite all expression rules to left-recursive. I mean [conditionalExpression](https://github.com/antlr/grammars-v4/blob/master/java8/Java8.g4#L1225) and descendants. This should decrease rules count and improve performance.

jundl77:
Is there no Java8 grammar out there that has this already done? Seems like a very tedious thing to do.

KvanTTT:
Sure. Java7 for example: https://github.com/antlr/grammars-v4/blob/master/java/Java.g4#L497

maybeec:
Is there any way of checking some complexity metrics for a grammar? I found the antlr4 IntelliJ plugin already quite useful, but it provides just a profiler for specific inputs. I would like to have something like given a grammar, the ANTLR4 parser would have X DFA states, Y NFA states, etc. Or more easy as you indicated, the grammar summarizes Z productions. Is there any tool providing such analysis results?

KvanTTT:
Try a new optimized [java8-pt](https://github.com/antlr/grammars-v4/tree/master/java8-pt) grammar which based on Java7.

KvanTTT:
[...]

PR #770, opened 2017-06-27
Small improvements of java8 grammar

maybeec:


java8/Java8.g4
```diff
@@ -348,7 +348,8 @@ fieldModifier
 	;
 
 variableDeclaratorList
-	:	variableDeclarator (',' variableDeclarator)*
+	:	variableDeclarator
+	|	variableDeclaratorList ',' variableDeclarator
```
KvanTTT: What is the point of this change? I like the previous format.

maybeec: I thought it comes with better parse performance when turning it into left recursion. Just referring to one of your posts: https://github.com/antlr/grammars-v4/issues/528#issuecomment-304441606

KvanTTT: In this case, your rule has a worse performance. It does not reduce the number of rules in comparison of replacing a sequence of rules with a single left-recursive rule.

maybeec: Ok. So one essential metric for ANTLR performance is rule count?

java8/Java8.g4
```diff
@@ -705,7 +705,8 @@ arrayInitializer
 	;
 
 variableInitializerList
-	:	variableInitializer (',' variableInitializer)*
+	:	variableInitializer 
+	|	variableInitializerList ',' variableInitializer
```
maybeec: so this would be the same right?

KvanTTT: Correct.

java8/Java8.g4
KvanTTT: I adhere the following priority:
1. Rule count.
2. If possible replace left-recursive rule with non-left-recursive or no-recursion rule (your case).

maybeec: @KvanTTT How do fit no. of ATN states into the picture? I implemented a grammar transformation and had to make the decision whether I introduce new ATN states or new rules by alternatives.
As DFAs and NFAs are build up on demand during parsing, I was not aware of any other metric to take into account besides the ATN, which somehow indicates the language complexity.
So in my case also introducing new ATN states would have made more sense, right?

teverett:
@KvanTTT Am i ok to pull this?

KvanTTT:
Single-line change, but ok :)

java8/Java8.g4
KvanTTT: I don't understand clearly, but I think the number of ATN states depend on rule number. You can do your own research and compare performance results.

PR #778, opened 2017-07-09
Support for line comment only.

smoelker:
There was not support for lines that only consist of a line comment with no trailing space.

### DIFF
--- a/mysql/MySQLLexer.g4
+++ b/mysql/MySQLLexer.g4
@@ -32,7 +32,7 @@ channels { MYSQLCOMMENT, ERRORCHANNEL }
 SPACE:                               [ \t\r\n]+    -> channel(HIDDEN);
 SPEC_MYSQL_COMMENT:                  '/*!' .+? '*/' -> channel(MYSQLCOMMENT);
 COMMENT_INPUT:                       '/*' .*? '*/' -> channel(HIDDEN);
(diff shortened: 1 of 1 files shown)


teverett:
@smoelker you could add a test case into /examples that uses this functionality?

mysql/MySQLLexer.g4
```diff
@@ -32,7 +32,7 @@ channels { MYSQLCOMMENT, ERRORCHANNEL }
 SPACE:                               [ \t\r\n]+    -> channel(HIDDEN);
 SPEC_MYSQL_COMMENT:                  '/*!' .+? '*/' -> channel(MYSQLCOMMENT);
 COMMENT_INPUT:                       '/*' .*? '*/' -> channel(HIDDEN);
-LINE_COMMENT:                        ('-- ' | '#') ~[\r\n]* ('\r'? '\n' | EOF) -> channel(HIDDEN);
+LINE_COMMENT:                        (('-- ' | '#') ~[\r\n]* ('\r'? '\n' | EOF) | '--' ('\r'? '\n' | EOF)) -> channel(HIDDEN);
```
KvanTTT: What the sense of this change? Line comments should be recognized too due to the star `*` sign.

smoelker: The issue with the original version is the required trailing space. A comment existing only out of 2 dashes and a line break was not recognized and threw an error.

PR #780, opened 2017-07-09
emailaddress: fix the lwsp-char rule.

renatahodovan:
According to the RFC 822 standard, a linear whitespace character
should either be a space or a horizontal tab. However, the current
grammar tries to match on the TAB lexer rule (which is not even
defined). The patch fixes this.


teverett:
@renatahodovan  thanks

PR #781, opened 2017-07-09
muMATH: add missing token definitions for AND, NOT and MOD.

renatahodovan:


teverett:
@renatahodovan could we get some examples that use these new tokens pls?

renatahodovan:
@teverett I'm not an expert of muMATH. I was just testing my ANTLR-grammar-to-test-case generator while recognised that these tokens were lost during the ANTLR3 to ANTLR4 conversion.

PR #789, opened 2017-07-13
Added ++ and -- to regex delimiter check

bkiers:
From an expression like `foo++/bar`, the `/` was interpreted as the regex delimiter instead of the division operator. Fixed it by adding the tokens `++` and `--` to the `isRegexPossible()` method/function.


teverett:
@bkiers thanks!

PR #792, opened 2017-07-14
COMMENT ON COLUMN command implemented.

aniskop:
Nothing more to say, actually :) Everything's in the title.


teverett:
@aniskop thanks for this and for the tests.

aniskop:
@teverett, more to come soon on other DDL commands :)

PR #814, opened 2017-07-24
Removed code from Z grammar which is GPL

dhait:
The Unicode grammar, specifically the classify.g4 file, is licensed under GPL 3.0.  The z grammar is licensed under BSD 3-clause.  In order to prevent the GPL restrictions from polluting the z grammar, this pull requests contains modifications the the ZLexer.g4 file to remove any code pulled from the unicode grammar, and replace it with native Antlr4 lexer constructs (specifically, the "Unicode properties" included in Antlr 4.7.


teverett:
@dhait  thanks

PR #815, opened 2017-07-24
Prevent variables and numbers with spaces

ghareeb-falazi:
Because white spaces are suppressed, having the following rules:

number
   : MINUS? DIGIT + (POINT DIGIT +)?
   ;
DIGIT
   : ('0' .. '9')
   ;
MINUS
   : '-'
   ;
POINT
   : '.'
   ;
WS
   : [ \r\n\t] + -> channel (HIDDEN)
   ;
 will lead to the acceptance of numbers with spaces such 1 2 3


teverett:
@ghareeb-falazi nice fix!

teverett:
@ghareeb-falazi ok this example worked fine before the PR, so I feel like a fix is needed here pls.

ghareeb-falazi:
So, a couple of things I would like to say: First, the proposed edits did indeed have troubles, and I found a way to fix them BUT I noticed that the test that is failing (as well as other tests) has a number in the scientific format with spaces within (e.g., -12.3 e -13.2 ). As far as I can tell, this is not allowed in programming languages and I am not sure it should be allowed within this grammar.
Finally, I noticed that the grammar does not support chaining unary minus or plus such as --++1.

teverett:
@ghareeb-falazi ok well that makes sense.  If feels like you're saying "the test case is wrong", in which case, can we just fix the test case?

ghareeb-falazi:
I have changed the grammar and the test cases, but the .tree files that I am changing to match the new grammar always get a new-line at the end when they are committed! this is causing the check to fail and is driving me crazy. Do you know how I can get rid of this new line?

ghareeb-falazi:
I have finally managed to fix it

PR #824, opened 2017-07-29
newline isn't a valid char,  ~['\\] -> ~['\\\r\n]

fonlang:
char invalidChar = '
';


teverett:
@KvanTTT ok to merge this?

KvanTTT:
It would be great to fix it also in other Java grammars (java8, java8-pt)

PR #826, opened 2017-07-29
added 'var' as local type (token and keyword)

ChristianWulf:


teverett:
@ChristianWulf could you add an example to the /examples dir for this?

teverett:
@ChristianWulf thanks!

PR #829, opened 2017-07-30
Newline cannot be in char or string literals in Java

fonlang:
The follow code will be compiled fail.

char newline_char = '
';

String newline_string = "abc
";


fonlang:
Same as : https://github.com/antlr/grammars-v4/pull/824

PR #837, opened 2017-08-04
allow optional schema name in package name rule

serii833:


teverett:
@serii833 could you commit an example file which uses this please?

PR #841, opened 2017-08-08
Added some more create statement rules, cleaned up some of my previous rules

madams51703:
Added some more create statement rules, cleaned up some of my previous rules


madams51703:
Looking into failed check

madams51703:
Removed rule reference that I was having issues with.  Forgot to remove the calling reference.

PR #842, opened 2017-08-08
Added dice notation grammar

Bernardo-MG:
Added dice notation grammar. Dice notation is used in tabletop games and roleplaying games, and as I couldn't find any parser for it on Java I prepared this grammar.


teverett:
@Bernardo-MG thanks!

PR #845, opened 2017-08-09
to_char, to_date, least, greatest

serii833:
* to_char with nls_params parameter
* added least, greatest, to_date functions


teverett:
@serii833 it looks like there are some test failures

PR #853, opened 2017-08-21
Added backup commands

madams51703:


madams51703:
recently modified delete_statement_from  rule ordering to pick a ddl_object before a table_alias.

PR #856, opened 2017-08-23
Added Keywords in pl/sql for future buildout of rules and a variable in tsql

madams51703:
Mainly added keywords for pl/sql so we can move this grammar forward faster.  In Oracle, the V$RESERVED_WORDS view lists keywords, so I pulled that information and added it to the pl/sql Lexer for future growth.  I commented out a number of the new lexer rules so that the existing tests would work.  We can go back in and retro the grammar to utilize the new keywords.


KvanTTT:
As I understand not all T-SQL keywords are PL/SQL keywords. What the sense of this big addition?

madams51703:
The big addition of keywords are not T-SQL kewords.  The large set of keywords I added are PL/SQL keywords.  In Oracle, there is a view called V$RESERVED_WORDS.  I have access to an Oracle database that I extracted the Keywords from, to eliminate the need to produce them by hand.  Since a large part of the tsql grammar has been added, I thought I would start giving us a boost on the Oracle(Pl/SQL) side.

I just happened to have a small change to the T-SQL side in the pull request as well that came along for the ride.

I hope my explanation makes sense.

Once these PL/SQL keywords are added, I can go back in and start adding them in in the appropriate locations in the grammar.

It might take me awhile as I am going to be starting a new job.

madams51703:
Here is a short example of how the keywords will immediately allow for the expansion of the grammar
(I know the parameters need added, just wanted to supply a quick example of the expansion to the existing rules that the new set of keywords will provide)

string_function
    : SUBSTR '(' expression COMMA expression (COMMA expression)? ')'
    | TO_CHAR '(' (table_element|standard_function|expression) (COMMA quoted_string)? (COMMA quoted_string)? ')'
    | DECODE '(' expression (COMMA expression)*  ')'
    | CHR '(' concatenation USING NCHAR_CS ')'
    | NVL '(' expression COMMA expression ')'
    | TRIM '(' ((LEADING | TRAILING | BOTH)? quoted_string? FROM)? concatenation ')'
    | TO_DATE '(' expression (COMMA quoted_string)? ')'
    | CONCAT
    | INITCAP
    | LOWER
    | LPAD
    | LTRIM
    | NCHR
    | NLS_INITCAP
    | NLS_LOWER
    | NLS_UPPER
    | NLSSORT
    | REGEXP_REPLACE
    | REGEXP_SUBSTR
    | REPLACE
    | RPAD
    | RTRIM
    | SOUNDEX
    | TRANSLATE
    | UPPER
    ;

KvanTTT:
Ok then.

KvanTTT:
Just noticed that these keywords are not used in parse rules.

teverett:
ok.. care to submit a PR to resolve it, however is most appropriate?

madams51703:
I realize that there are not yet parse rules utilizing the token rules, but
there will be.  I have started working on the string functions.

My reasoning for providing the tokens up front is that I know we will need
these tokens.  I am attempting to make it quicker to generate the parse
rules by generating the tokens up front.

teverett:
I'm fine with that as long as the existing tests pass, and you continue to add tests as you go.  Let's make sure that people can download a working grammar.

madams51703:
Will do.  I spent several hours churning through the tests when I added
them to make sure I did not break any tests.

I will have to look into commenting out the entries I added to further
prevent the grammar from breaking if somebody uses a token I defined. But
have it available for quick usage.

I appreciate the two of you pointing that kind of stuff out!

teverett:
The reason grammars-v4 and the test framework was created was to ensure that people could download working, up-to-date grammars that are tested with the latest version of Antlr. As long as we can stick to that, I'm fine.

madams51703:
Ok, the best thing for me to do is comment out the tokens I added so they
are there when I need them, but do not interfere with existing behaviour.
I only concentrated on passing the tests, not the bigger picture.  And I
will throw together a script to automate uncommenting them when I need
them.

I think that is a sound approach that will still save me effort without
taking the risk of breaking existing behaviour.  It will be easy enough to
go back and comment them out using a script to do so.

PR #885, opened 2017-09-18
Disallow inner enums/annotations in Java8PT grammar

PureCS:
Reference: [http://docs.oracle.com/javase/specs/jls/se7/html/jls-8.html#jls-8.1.3](http://docs.oracle.com/javase/specs/jls/se7/html/jls-8.html#jls-8.1.3)


PureCS:
Whoops this was a mistake.

PR #886, opened 2017-09-18
Disallow local enums/annotations in Java8PT grammar

PureCS:
Reference: [http://docs.oracle.com/javase/specs/jls/se7/html/jls-8.html#jls-8.1.3](http://docs.oracle.com/javase/specs/jls/se7/html/jls-8.html#jls-8.1.3)

### DIFF
--- a/java8-pt/JavaParser.g4
+++ b/java8-pt/JavaParser.g4
@@ -49,6 +49,12 @@ typeDeclaration
     | ';'
     ;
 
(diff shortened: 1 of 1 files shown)


KvanTTT:
Could you also add a code fragment to AllInOne8.java?

PureCS:
I'll work on your suggestions, much appreciated.

PR #906, opened 2017-10-17
LexerAdaptor for csharp target

aboubkr90:


KvanTTT:
What the meaning of this file in grammars repository?

aboubkr90:
This is the base for the lexer (LexerAdaptor) in c#

PR #909, opened 2017-10-21
A grammar for Java 9

chungkwong:
A grammar for Java 9 is generated by updating the Java 8 grammar to caught up with the latest Java Language Specification, including support for module declaration and modifed try-with-resources-statement.


teverett:
@chungkwong could you add some examples that validate what you changed, such as try-with-resources?

chungkwong:
Two tiny examples are added

teverett:
Thanks!

PR #912, opened 2017-10-24
Multiple minor fixes to the pom.xml files

bramp:
Specifically ensures the correct grammar is referenced, all the required included files are listed, and the filenames have the correct case.


teverett:
@bramp thanks!

PR #915, opened 2017-10-27
Replace old Java grammar with the new one (Java8) and other fixes

KvanTTT:
* Java 8 grammar fixed and simplified.
* AllInOne8.java test files splitted on AllInOne7.java and AllInOne8.java.
* Added benchmark and correctness tests for different projects (see Benchmarks.md).


parrt:
Looking great! Why have java and javascript mixed up in same PR though?

KvanTTT:
There are small fixes in README and license notes not related to grammar itself :)

parrt:
Ok, @teverett What do you think? the grammar changes don't seem that major and if it parses 8, i'm cool with it.  We should however keep the pure java8 grammar that is close to the spec.

teverett:
I'm good!

PR #917, opened 2017-10-29
Issue916

hvogeler:
Added the optional parentheses to the create function body.
I wasn't sure what to do with automatic crlf conversions git did.
So I first committed them and then reverted that commit.


teverett:
@hvogeler could you add one or more tests please which validate this change?

tsql/TSqlParser.g4
```diff
@@ -1894,7 +1894,7 @@ func_body_returns_select
   :RETURNS TABLE
   (WITH function_option (',' function_option)*)?
   AS?
-  RETURN select_statement
+  RETURN '('? select_statement ')'?
```
KvanTTT: Incorrect. The proper code:
```
RETURN ('(' select_statement ')' | select_statement)
```

hvogeler: Thank you Ivan. How about two other places in the grammar file that have '('? ? Would your suggestion apply to them as well to enforce balanced parentheses?

`receive_statement
    : '('? RECEIVE (ALL | DISTINCT | top_clause | '*')
      (LOCAL_ID '=' expression ','?)* FROM full_table_name
      (INTO table_variable=id (WHERE where=search_condition))? ')'?
    ;`

`create_or_alter_procedure
    : ((CREATE (OR ALTER)?) | ALTER) proc=(PROC | PROCEDURE) func_proc_name (';' DECIMAL)?
      ('('? procedure_param (',' procedure_param)* ')'?)?
      (WITH procedure_option (',' procedure_option)*)?
      (FOR REPLICATION)? AS sql_clauses
    ;`

tsql/TSqlLexer.g4
```diff
@@ -807,7 +807,7 @@ LINE_COMMENT:       '--' ~[\r\n]* -> channel(HIDDEN);
 DOUBLE_QUOTE_ID:    '"' ~'"'+ '"';
 SINGLE_QUOTE:       '\'';
 SQUARE_BRACKET_ID:  '[' ~']'+ ']';
-LOCAL_ID:           '@' ([A-Z_$@#0-9] | FullWidthLetter)+;
+LOCAL_ID:           '@' ([A-Za-z_$@#0-9] | FullWidthLetter)+;
```
KvanTTT: `a-z`

Remove it. We use UPPER case insensitive input stream, see [caseInsensitiveType](https://github.com/antlr/antlr4test-maven-plugin#caseinsensitivetype)

tsql/TSqlParser.g4
KvanTTT: They also should be fixed.

tsql/TSqlLexer.g4
hvogeler: As as user of the grammar I am not really interested in the antlr4test-maven-plugin plugin. How do I use the grammar to parse lower case parameter names without modifying the Lexer rule?

KvanTTT: You can use case-insensitive input stream in any runtime, see [Runtimes](https://github.com/KvanTTT/DAGE/tree/master/AntlrGrammarEditor/Runtimes).

As a T-SQL grammar creator, I'm not interested in dirty and slow grammar due to the fragment rules.

teverett:
@hvogeler @KvanTTT are we good to pull this?

hvogeler:
Ivan asked to remove commits 724c17c and 7801ae4. I hope I did it right by reverting those commits. From my side only commit 31bece5 should be pulled.

KvanTTT:
Is it possible to fix line breaks?

hvogeler:
It is not only the line breaks. An even bigger problem are superfluous whitespaces in the files before the line end. Git is very picky about those whitespace characters. They should be removed from the files. I wonder what kind of editor produced these files.

KvanTTT:
Ok then.

PR #918, opened 2017-10-30
UPnP search syntax grammar

atamariya:
Antlr grammar for UPnP search sytax


teverett:
Thanks for the contribution.  Please add this into the root pom.xml so that when the unit tests are run, the tests for this grammar are run.

atamariya:
Updated the POM.

PR #926, opened 2017-11-03
PL/SQL CREATE TABLE update with test example

luyseyal:
PL/SQL: Support NOT NULL, ENABLE/DISABLE inline column constraints in CREATE TABLE along with quoted identifiers for TABLESPACE.

### DIFF
--- a/plsql/PlSqlParser.g4
+++ b/plsql/PlSqlParser.g4
@@ -526,7 +526,7 @@ drop_index
 
 create_table
     : CREATE (GLOBAL TEMPORARY)? TABLE tableview_name 
-        LEFT_PAREN column_name datatype (NOT NULL)? (ENABLE | DISABLE)? (',' column_name datatype (NOT NULL)? (ENABLE | DISABLE)?)*
+        LEFT_PAREN datatype_null_enable? (',' datatype_null_enable?)*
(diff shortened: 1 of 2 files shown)


luyseyal:
@KvanTTT  FYI

plsql/PlSqlParser.g4
```diff
@@ -526,7 +526,14 @@ drop_index
 
 create_table
     : CREATE (GLOBAL TEMPORARY)? TABLE tableview_name 
-        LEFT_PAREN column_name datatype (',' column_name datatype)*
+        LEFT_PAREN column_name datatype (NOT NULL)? (ENABLE | DISABLE)? (',' column_name datatype (NOT NULL)? (ENABLE | DISABLE)?)*
```
KvanTTT: Maybe extract duplicated syntax to separated rule? Something like this:

```ANTLR
column_name datatype_null_enable
   :  column_name datatype (NOT NULL)? (ENABLE | DISABLE)?
   ;
```

It improves readability, parse tree bypass, and clarity.

KvanTTT:
LGTM.

luyseyal:
Thank you for the suggestion, @KvanTTT. I am obviously a newbie but expect to have more to contribute soon!

PR #930, opened 2017-11-05
Normalize line endings using correct .gitattributes settings

### LINKED ISSUE #923: git clone leaves modified files (line endings)
Hi,

I did a fresh clone on Mac, Linux and Windows.
On every platform `git status` lists between one and two dozens of modified files right after, without doing anything.

Should I try to fix this (as described in https://help.github.com/articles/dealing-with-line-endings/#platform-all) and file a pull request, or do you already have a plan how to handle this?

Thanks,
Oti.

ohumbel:
This fixes #923

I am now able to clone on Mac, Windows and Linux without getting modified files.
The list of all the extensions is IMHO necessary to prevent those modified files from popping up again.

I strongly followed the rules and commands on https://help.github.com/articles/dealing-with-line-endings/#platform-all


KvanTTT:
I don't like this PR:  there are too many changes without obvious advantages.
It's better to use unified line breaks for all files, at least *.g4 files.
What line break is used by default?

ohumbel:
The PR would make sure that for the defined text files (including *.g4) the line breaks are unified in the repository, but by checking out or cloning are converted to platform specific ones.

As Travis told us, this does not work for `/vb6`. I am confident that I can solve this as well.

I do not understand what you mean by "It's better to use unified line breaks for all files", because that is exactly what I am trying to achieve.

So I plan to work further on my feature branch, pushes will trigger another travis build.

.gitattributes
```diff
@@ -1,3 +1,93 @@
...
+*.csv text
+*.dgs text
+*.dot text
+*.erl text
+*.errors text
+*.faa text
+*.ffn text
+*.fna text
+*.g text
+*.g4 text
```
KvanTTT: Should be changed to eol=lf

ohumbel: Since my commits did not touch a single `.g4` file, I do not see the need to enforce `lf` endings. 
The `*.g4 text` line (together with the `* text=auto` line) make sure that the line endings in the git repo are unified to `lf`. On the working tree they are converted to platform specific line endigs (and vice versa).

ohumbel:
Please let me explain why it took so many pushes and failed Travis builds to get this right.

I started on Mac, made a list of all file extensions in the repo and checked all these if they "look" like (`ASCII`) text. With `rm .git/index` and `git reset` I found local line endings not conforming to `.gitattributes` and committed them (as described here: https://help.github.com/articles/dealing-with-line-endings/). Doing manual checks I noticed that not all occurrences had been discovered, and therefore began to list every file extension in `.gitattributes` separately. Some more files were normalized. The first push broke the build, and therefore I switched to Ubuntu, made a fresh clone of my feature branch and started again. As soon as I had `mvn test` passing, I pushed again. But it turned out that there were some conversions happening only on checkout, so I took a fresh clone again and the tests failed locally again. This way I found the last offending line endings.

I want to emphasize that I did not touch any file by hand, except `.gitattributes`.

ohumbel:
I have 2 platforms running now - trying to get `mvn test` passing on Windows , for desert.

ohumbel:
There is only one last thing I struggle with: the graphemes (`ascii.txt` and etc.) on Windows.

ohumbel:
So, after a bit more work than expected initially, I can now confirm that `git clone` produces no modified files any more, and that `mvn test` passes on three platforms. I verified both of these multiple times on:
 - Linux (Ubuntu 16.04 LTS), git version 2.7.4
 - Windows 10, git version 2.14.0.windows.1
 - macOS High Sierra 10.13.1, git version 2.13.6 (Apple Git-96)

ohumbel:
From my point of view, this pull request is ready to be merged.

Advantages would be:
 - Neither existing nor future contributors are irritated by modified files after `git clone` or `git pull`. And nobody has to invent workarounds using local `.gitattributes`.
 - `git clone` just works on all platforms
 - `mvn test` just works on all platforms
 - `* text=auto` makes sure future file extensions are treated reasonably
 - If a grammar has special requirements concerning line endings (like the one in `/vb6`), those can be specified in `.gitattributes`

Hope to have helped a bit - best wishes
Oti.

teverett:
Any objections to me pulling this?

ohumbel:
No objections from my side - thanks!

teverett:
@KvanTTT could we have your feedback pls?

KvanTTT:
Ok, let it be.

ohumbel:
Cool, thanks very much.

I am confident that the situation has now greatly improved for all contributors. If you hear any complaints, please let me know! Chances are high that I already experienced a similar problem.

PR #933, opened 2017-11-08
fix queue_setting

PrVrSs:
Fix ALTER QUEUE according https://docs.microsoft.com/en-us/sql/t-sql/statements/alter-queue-transact-sql. I write simple tool(fuzzer) for check grammar. Fuzzer builds query's according grammar file(*.g4) and checks for correctness.  And next query: ALTER QUEUE WORK WITH STATUS = ON , RETENTION = ON , ACTIVATION( , STATUS = ON DROP ); exception  Incorrect syntax in Microsoft SQL Server 2017
https://github.com/PrVrSs/AntlrGrammarFuzzer - fuzzer

### DIFF
--- a/tsql/TSqlParser.g4
+++ b/tsql/TSqlParser.g4
@@ -1654,13 +1654,22 @@ queue_settings
        (RETENTION EQUAL (ON | OFF) COMMA?)?
        (ACTIVATION
          LR_BRACKET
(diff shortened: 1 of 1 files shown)


KvanTTT:
Nice tool!
LGTM.

PR #954, opened 2017-11-20
TSQL. Fix for_clause

PrVrSs:
TSQL. Fix for_clause according https://docs.microsoft.com/en-us/sql/t-sql/queries/select-for-clause-transact-sql


KvanTTT:
LGTM 👍

PR #961, opened 2017-11-22
Enable parser reuse and facilitate grammar extension

arcanefoam:
These changes address two issues: 

1. Hard-coded grammar name and location:
 * Line 43 made a hard-coded import of the Python3Parser class (required for TOKEN references). As a result users have little flexibility of where to place the generated grammar, e.g. in my case the grammar should go in a sub-module.
* This hard coded import also made it impossible to extend the grammar (e.g. write a piggy back DSL on-top of python), as the generated lexer of the new language will still try to import the Python3Parser. 
 -> The dynamic import allows arbitrary locations and also grammar extensions

2. Class variables for INDENT/DEDENT management:
* As a result of the supporting variables being class variables, it was hard to reuse a single parser (or use several parsers) as they shared state, causing strange behaviours (e.g. erroneous INDENT/DEDENT level)
-> The use of Properties  the attributes are now at the instance level and the state is no longer shared between instances (or instantiations). Further, the overriden reset method returns this attributes to their initial values.


teverett:
@arcanefoam thanks for the commit.  I've got to presume that you have some test cases of your own that you used when making these changes.  Any chance you could commit them to /examples?

arcanefoam:
I do... but atm I can not share my grammar extensions... I will try to make a smaller example and add it. Hopefully some time next week.

arcanefoam:
@teverett Does your CI (or whichever you use) provides a python interpreter? I am thinking we could actually run some tests on the generated python code (I assuming your infrastructure is currently only testing parsers/lexers generated in Java).

PR #966, opened 2017-11-24
support timestamp auto init and update

scadgek:
Support special syntax for specifying TIMESTAMP and DATETIME columns auto-update. Documentation: https://dev.mysql.com/doc/refman/5.7/en/timestamp-initialization.html


KvanTTT:
LGTM.

PR #969, opened 2017-11-24
add dimensions support for YEAR data type

scadgek:
`YEAR` data type can also be specified as `YEAR(4)` (and `YEAR(2)` in older MySQL versions. See docs: https://dev.mysql.com/doc/refman/5.7/en/year.html


teverett:
@scadgek can we get examples submitted for these too please?

scadgek:
@teverett sure, I've added it to the test

PR #984, opened 2017-11-30
Support for SQL (implicit) cursor attribute

bedorlan:
https://docs.oracle.com/cd/B28359_01/appdev.111/b28370/sql_cursor.htm


plsql/PlSqlParser.g4
```diff
@@ -1845,7 +1845,7 @@ other_function
     | COALESCE '(' table_element (',' (numeric | quoted_string))? ')'
     | COLLECT '(' (DISTINCT | UNIQUE)? concatenation collect_order_by_part? ')'
     | within_or_over_clause_keyword function_argument within_or_over_part+
-    | cursor_name ( PERCENT_ISOPEN | PERCENT_FOUND | PERCENT_NOTFOUND | PERCENT_ROWCOUNT )
+    | ( cursor_name | SQL ) ( PERCENT_ISOPEN | PERCENT_FOUND | PERCENT_NOTFOUND | PERCENT_ROWCOUNT )
```
KvanTTT: I think `SQL` should be added to the `regular_id` rule.

bedorlan: hmm .. i'm not sure. i tried to name a variable SQL and i got this error:
```
declare
    sql number := 2;
begin
    dbms_output.put_line(sql);
end;
/

ORA-06550: line 2, column 5:
PLS-00103: Encountered the symbol "SQL" when expecting one of the following:

   begin function pragma procedure subtype type <an identifier>
   <a double-quoted delimited-identifier> current cursor delete
   exists prior  ...
```

Issue #986, opened 2017-12-01
generate asn.g4 fail while language is Go


runner-mei:
````
D:\developing\src\mibexplorer>java -jar D:\developing\java\antlr4-4.7\antlr-4.7-complete.jar -Dlanguage=Go ASN.g4
warning(154): ASN.g4:124:0: rule sequenceType contains an optional block with at least one alternative that can match an empty string
warning(154): ASN.g4:377:0: rule setType contains an optional block with at least one alternative that can match an empty string
error(134): ASN.g4:356:0: symbol type conflicts with generated code in target language or runtime
error(134): ASN.g4:144:35: symbol type conflicts with generated code in target language or runtime
error(134): ASN.g4:194:11: symbol type conflicts with generated code in target language or runtime
error(134): ASN.g4:260:47: symbol type conflicts with generated code in target language or runtime
error(134): ASN.g4:268:52: symbol type conflicts with generated code in target language or runtime
error(134): ASN.g4:348:6: symbol type conflicts with generated code in target language or runtime
error(134): ASN.g4:352:6: symbol type conflicts with generated code in target language or runtime
error(134): ASN.g4:421:22: symbol type conflicts with generated code in target language or runtime
error(134): ASN.g4:423:23: symbol type conflicts with generated code in target language or runtime
error(134): ASN.g4:484:25: symbol type conflicts with generated code in target language or runtime
error(134): ASN.g4:510:18: symbol type conflicts with generated code in target language or runtime
error(134): ASN.g4:516:7: symbol type conflicts with generated code in target language or runtime
error(134): ASN.g4:158:96: symbol type conflicts with generated code in target language or runtime
error(134): ASN.g4:241:5: symbol type conflicts with generated code in target language or runtime
error(134): ASN.g4:258:58: symbol type conflicts with generated code in target language or runtime
error(134): ASN.g4:380:70: symbol type conflicts with generated code in target language or runtime
error(134): ASN.g4:166:2: symbol type conflicts with generated code in target language or runtime
````

KvanTTT:
Try to rename `type` identifier to `type_`.

runner-mei:
it is ok.

KvanTTT:
It would be great if you upload this change here :)

PR #987, opened 2017-12-01
fix generate fail while target language is golang

runner-mei:
fix generate fail while target language is golang
https://github.com/antlr/grammars-v4/issues/986


teverett:
@runner-mei thanks for the PR.  I'm going to pull it, and politely ask that you contribute an example or two.  The more examples we have on the tree, the less likely it is that a future change will break this grammar.

Issue #989, opened 2017-12-03
Java method single receiver parameter


MartijnDwars:
According to the JLS, the list of [formal parameters](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.4.1) can consist of a single `ReceiverParameter`. However, according to the Java 8 grammar in this repository, when a [FormalParameterList](https://github.com/antlr/grammars-v4/blob/master/java8/Java8.g4#L458) contains only one parameter, it must be a `lastFormalParameter` which cannot be `receiverParameter`. Hence, the following program is syntactically correct (according to JLS 1.8) but cannot be parsed (according to the Java 8 grammar in this repository):

```java
class Foo {
  void bar(boolean this);
}
```

PR #990, opened 2017-12-03
Java method single receiver parameter

MartijnDwars:
Fixes issue #989 in the [java8](https://github.com/antlr/grammars-v4/blob/master/java8/) grammar (not the [java](https://github.com/antlr/grammars-v4/blob/master/java/) grammar) and adds a test to prevent regressions.


teverett:
@MartijnDwars thanks

PR #998, opened 2017-12-05
Fixes #997

### LINKED ISSUE #997: Java shift expressions
[Java shift operator](https://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.19) does not allow whitespace between the less-than and greater-than signs, i.e. the following should not be a legal Java program:

```
class X {
        public X() {
                int x = 1;
                System.out.println(x > > 1);
        }
}
```

The [Java8 grammar](https://github.com/antlr/grammars-v4/blob/master/java8/Java8.g4#L1273-L1275) and [Java grammar](https://github.com/antlr/grammars-v4/blob/master/java/JavaParser.g4#L480), however, allow whitespace between the less-than and greater-than signs.

MartijnDwars:
Fixes #997. I couldn't add a test, because that would require "negative testing" (i.e. test that some string _cannot_ be parsed).

### DIFF
--- a/java/JavaLexer.g4
+++ b/java/JavaLexer.g4
@@ -142,6 +142,9 @@ BITAND:             '&';
 BITOR:              '|';
 CARET:              '^';
 MOD:                '%';
+LSHIFT:             '<<';
+RSHIFT:             '>>';
+URSHIFT:            '>>>';
 
 ADD_ASSIGN:         '+=';
 SUB_ASSIGN:         '-=';
--- a/java/JavaParser.g4
+++ b/java/JavaParser.g4
@@ -477,7 +477,7 @@ expression
     | prefix=('~'|'!') expression
     | expression bop=('*'|'/'|'%') expression
     | expression bop=('+'|'-') expression
-    | expression ('<' '<' | '>' '>' '>' | '>' '>') expression
+    | expression ('<<' | '>>>' | '>>') expression
     | expression bop=('<=' | '>=' | '>' | '<') expression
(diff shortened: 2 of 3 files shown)


cowang:
There is a situation where multiple nested annotations can create <<<. Is there a test for that?

KvanTTT:
As I guess this is not correct. Will nested generics work? See https://stackoverflow.com/q/22806202/1046374

MartijnDwars:
Hmz, you're both right that these changes would break parsing of such programs. I'm used to scannerless parsing where issues like these do not exist. I'm not sure how to solve this problem in ANTLR yet.

KvanTTT:
Maybe with using of semantic [predicates](https://github.com/antlr/antlr4/blob/master/doc/predicates.md) and actions.

MartijnDwars:
Indeed, that seems like the best way to solve this in ANTLR. I found a [blog post](http://madbean.com/2003/mb2003-12/) that describes this exact problem and suggests a solution using semantic predicates and actions. Having to rely on semantic predicates and actions seems like a serious shortcoming of ANTLR though.

I'll try to push a fix using semantic predicates and actions later this week.

KvanTTT:
There is an issue related to Unified Actions: https://github.com/antlr/antlr4/issues/1045 But it won't be implemented.

> I'll try to push a fix using semantic predicates and actions later this week.

I suggest not to pollute a grammar with runtime dependent actions for such small task.

So, I'm closing PR.

PR #1006, opened 2017-12-09
Fix NEWLINE embedded code in python3-py grammar

gVirtu:
* `self.skip()` was incorrectly called as `skip()` causing runtime errors
 * `indents.peek()` had been translated as `self.indents[0]`, however the stack top is always the last element if we use `append` as our push operation.

Cheers!


teverett:
@gVirtu thanks

PR #1013, opened 2017-12-12
Partitioning should now be properly working for create table.

madams51703:


KvanTTT:
What the meaning of binary `plsql/examples/.partitioned_tables.sql.swp` file? It's even not displayed.

madams51703:
Yeah, it is a temp file that the vi editor creates.  I had that example
file open when I commited and realized it afterwards, So I went and removed
it then pushed. Wasn't sure of the right way to fix that situation.  So did
it that way.

On Dec 13, 2017 12:23, "Ivan Kochurkin" <notifications@github.com> wrote:

What the meaning of binary plsql/examples/.partitioned_tables.sql.swp file?
It's even not displayed.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/antlr/grammars-v4/pull/1013#issuecomment-351461507>,
or mute
the thread

KvanTTT:
git allows removing files. Try to use [`git rm`](https://stackoverflow.com/q/2047465/1046374) command.

madams51703:
Ok, I tried that and comitted/pushed. How that resolved it.

PR #1014, opened 2017-12-13
Added options to create index statement

madams51703:


plsql/PlSqlParser.g4
```diff
@@ -521,9 +521,145 @@ sequence_start_clause
...
+index_partitioning_clause
+    : PARTITION partition_name? VALUES LESS THAN '(' 
+                                        (','? (CHAR_STRING
+                                              | string_function
+                                              | numeric
+                                              | MAXVALUE
+                                              )
+                                        )+ 
+                                                 ')'
+                                        segment_attributes_clause?
```
KvanTTT: Could such big indent be reduced?

madams51703:
Yes, I already had created a literal rule in my next release and started to
use it.  I can go and modify these large indents to use the literal rule.
I have the next release ready to queue up pending reviewing formatting.
Can I fix this with the pending release since I have already started down
that road?

KvanTTT:
Ok then!

madams51703:
Thank you!

PR #1017, opened 2017-12-16
Alignment with current ProLeap 2.3.0 parser grammars

uwol:
We were provided with a large batch of COBOL files from banking and thus could identify and fix several remaining bugs. The COBOL parser now is able to parse all of these real-world COBOL files.

Additional unit tests have been added in the [COBOL project repo](https://github.com/uwol/cobol85parser). Test suites both for COBOL and Visual Basic 6.0 pass.

COBOL grammar
-----------------

- added a lot of preprocessor compiler options
- added exception clauses for ACCEPT statement
- made DIVIDE statement consistent 
- fixed identifier in MOVE CORR statement
- fixed identifier in STRING statement
- added <> operator
- switched from BSD 3-clause license to MIT license for simplification reasons

Visual Basic 6.0 grammar
-------------------------

- switched from BSD 3-clause license to MIT license for simplification reasons


teverett:
@uwol thanks very much for this.

PR #1018, opened 2017-12-16
Python target parser fixes and added test bench

arcanefoam:
Fixes some issues with the Python target parser, mainly with respect to indentation. Adds a test class for testing the generated Python parser and lexer that uses the Python's project test_grammar.py file to test the parser.

### DIFF
--- a/python3-py/Python3.g4
+++ b/python3-py/Python3.g4
@@ -124,8 +124,8 @@ def createDedent(self):
     dedent.line = self.lastToken.line
     return dedent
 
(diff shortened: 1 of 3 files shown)


arcanefoam:
If I am pointed in the right direction I might be able to add execution of the test class as part of the integration. 
Probably the test_grammars.py file should also be added to the java (and others) Python target examples for testing.

teverett:
@arcanefoam thanks!

PR #1023, opened 2017-12-19
plsql: add row limiting clause support to SELECT [12c]

luyseyal:
Support row limiting clauses like:
```
SELECT *
FROM green_table
OFFSET 1 ROW
FETCH FIRST 1 ROWS ONLY
```

FYI, @MilesPerrySEU

### DIFF
--- a/plsql/PlSqlParser.g4
+++ b/plsql/PlSqlParser.g4
@@ -5260,7 +5260,7 @@ non_reserved_keywords
     | PENDING
     | PERCENTILE_CONT
     | PERCENTILE_DISC
-    | PERCENT
+    | PERCENT_KEYWORD
     | PERCENT_RANKM
     | PERCENT_RANK
     | PERFORMANCE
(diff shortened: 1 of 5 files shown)


luyseyal:
@KvanTTT Do I need to add the uncommented keywords into the regular_id definition?

KvanTTT:
@madams51703 please take a look at this PR.

madams51703:
@KvanTTT 
I verified the
 SELECT *
FROM green_table
OFFSET 1 ROW
FETCH FIRST 1 ROWS ONLY
query works at livesql.oracle.com ( did that since I can't find the syntax diagram in Oracle's documentation.)

so I am good with it.  I just can't find the syntax diagram.

@luyseyal Can you add the url where the syntax diagram is at just so we know where the syntax diagram is defined?  This new addition just taught me something cool about the language.

luyseyal:
@madams51703 
https://docs.oracle.com/database/121/SQLRF/statements_10002.htm#BABBADDD

luyseyal:
I tried to handle some of the logical options (like, you can't use it with a `FOR UPDATE`) but the code was super ugly and redundant so I gave up.

KvanTTT:
Could you resolve conflicts, please?

luyseyal:
@KvanTTT Done

PR #1024, opened 2017-12-20
Added alter_database

madams51703:


luyseyal:
Do the uncommented keywords need to be added to the ever-expanding `regular_id` list in the parser?

madams51703:
Stephan,  yes the keywords need added to the regular_id list.  Almost all oracle keywords can be used as identifiers in PL/SQl.  I run the parser against real world scripts as I am adding rules.  As I run into usage of the keywords as identifiers, I add them to the regular_id list. So theoretically almost every token should be in the regular_id list, except a few that can not be used as identifiers, and that list is very small.  I only add them as I encounter them in the real world scripts.  That begs the question of if I should just go ahead and add all the keywords except those that can not be used that way into the regular_id list or just continue to trickle them into the regular_id list as I encounter their usage.  I also use livesql.oracle.com to do some validation on additions that I question.  
This link defines reserved words for PL/SQL
https://docs.oracle.com/cd/B19306_01/server.102/b14237/dynviews_2048.htm#REFRN30204

SQL> select 'RESERVED', count(*) from V$RESERVED_WORDS where reserved = 'Y'
  2  union
  3  select 'RES_TYPE', count(*) from V$RESERVED_WORDS where res_type = 'Y'
  4  union
  5  select 'RES_ATTR', count(*) from V$RESERVED_WORDS where res_attr = 'Y'
  6  union
  7  select 'RES_SEMI', count(*) from V$RESERVED_WORDS where res_semi = 'Y'
  8  /

'RESERVE   COUNT(*)
-------- ----------
RESERVED        100
RES_ATTR          2
RES_SEMI         29
RES_TYPE          0

SQL> select count(*) from v$reserved_words;

  COUNT(*)
----------
      1854

So I really ought to have 1723 keyword entries in regular_id. (As of  11GR2 ) Adding them would make testing and releases go  somewhat faster.

Thoughts?  Should I add those 1723 entries in the next release or continue to dribble them in as I find them by encountering their absence in regular_id.

Sorry for the long post, wanted to make sure I was as clear as possible.

luyseyal:
1. I'm not an Antlr expert by any stretch of the imagination, but surely, SURELY there is a better way to do this than having the whole list in two places (Lexer AND Parser). :smile: 
2. I think dribbling is fine in the short run unless there is a good answer for [1]
3. Please include 12c keywords if you decide to go ahead with a full keyword patch. I am starting to add in 12c stuff we need. We use this for sanity checking code before it gets dumped in the database (12c)

madams51703:
I currently don't have Direct access to a 12c instance, but might have
saved off the 12c keywords. If I don't have 12c, I do have 11gr2 list

Ivan, do you have an opinion on dribble in vs put in all the unreserved
keywords into the regular_id  rule?

luyseyal:
I can send you a list direct from 12cR1 if you need

madams51703:
Sure, I'll take a list of the 12c keywords.  Can you dump v$reserved words
into a text file?

KvanTTT:
> Ivan, do you have an opinion on dribble in vs put in all the unreserved
keywords into the regular_id  rule?

I think this is an acceptable solution. Moreover, ANTLR doesn't support universal semantic predicates and grammar will become runtime dependent if we want to detect some keywords on a parsing stage. See the JavaScript parser for example: https://github.com/antlr/grammars-v4/blob/master/javascript/JavaScriptParser.g4#L413

luyseyal:
[reserved_12c.zip](https://github.com/antlr/grammars-v4/files/1576796/reserved_12c.zip)

@madams51703, Find attached a tab-separated dump of V$RESERVED_WORDS from 12cR1

madams51703:
Thanks.  Can I endeavor to implement this on my next release?

madams51703:
Stephen,

           How about sending me the same for 11G.  I might have grabbed
the12c list when I grabbed the list, I still had access to 12C when I
grabbed that list of keywords, So we may already be using the 12c KEYWORD
list.  I probably future proofed us when I generated the list.

Mark

luyseyal:
@madams51703 I don't actually have access to an 11g box at the moment

madams51703:
Stephen,

       Ok,  I have access to 11g, I was just wanting another set of
eyes/hands so I didn't mess it up.  I have filled out a new rule
oracle_reserved_keywords rule and will be releasing it momentarily my test
with the examples just finished.  I basically uncommented all the lexer
rules for the keywords so they would get matched in the keyword rule in the
parser.Maybe there is a better way that we will figure out in the future,
but at least the unreserved keywords will be in one rule that regular_id
will call.

      I need to digest Ivan' s property information before I do anything
with it so I can try to use it correctly. Can you mark 12c specific stuff
that you add with a comment so I know where it is 12c specific.  I will
endeavor to do the same.

Mark

madams51703:
Ok, be aware of large number of changes to lexer to make tokens active(uncommented lines) and new rule in parser for them to land in.  No more worrying about tables name SOURCE or something like that anymore or some other table with a non reserved keyword.

luyseyal:
If this is merged first, I can rebase my PR on it

madams51703:
Thanks!

madams51703:
Do you have a page/section you can reference in Terrance's book on the
runtime dependencies.  If so I can read it there and ponder it.

KvanTTT:
Hi, guys. Can I merge this Pull Request eventually?

luyseyal:
@KvanTTT  I'm OK for this to be merged. I'll rebase my PR on this. @madams51703 are you good-to-go?

madams51703:
I am good to go.

PR #1025, opened 2017-12-23
Hypertalk

defano:
Corrected issue with `scriptlet` rule in which portions of ambiguous input could be ignored. Added support for `number`, `selectedChunk`, `selectedLine`, `selectedField` and `selectedText` properties.


teverett:
@defano thanks!

Issue #580, opened 2017-03-22
Issue with parsing shift operation in Objective-C grammar


prateekgulati:
Right Shift operation is parsed as two RT siblings and might create ambiguity with greater than operator.
Same for left shift. 
Eg: for 0x01 >> 1;
![img](https://cloud.githubusercontent.com/assets/9799617/24200422/0661a28e-0f33-11e7-94b7-044e0caa57e8.PNG)

These two RT should be considered a single child

KvanTTT:
It does not create an ambiguity with less than operator `<`. On the other hand, a single `>>` operator can be incorrectly recognized with several generic specifiers.

prateekgulati:
Just like Not equal `!=` is considered a single child and not two children `!` and `=`
And so are `<=`, `>=` `==` 
Similarly, right/left shift `>>` or `<<` is a single operator. So it should be considered as a single child and not two children.

KvanTTT:
This is not working for token-based parser such as ANTLR. Please, read "Avoiding the Maximal Munch Ambiguity" paragraph in "The definitive antlr 4 reference" book. Also see the same [expression rule](https://github.com/antlr/grammars-v4/blob/master/java/Java.g4#L497) in Java grammar.

KvanTTT:
@teverett this issue should be closed.

prateekgulati:
If you want, I can try to fix this issue.

KvanTTT:
This is not a bug and I already wrote why.

pejobo:
For future readers, it's the ambiguity between a generic type definition and e.g. the lshift:
`Map<<String>,<String>>` vs. `0x1F << 3`
This has to be resolved in the parser, where this could be done easily based on context information.

PR #1033, opened 2018-01-01
added tokens for bit shift operators <<, >>, >>>

pejobo:


KvanTTT:
I'm closing this due to the following items:

* Tests failed.
* Java grammar uses only `<` and `>`, see discussion: https://github.com/antlr/grammars-v4/issues/580

pejobo:
For future readers, it's the ambiguity between a generic type definition and e.g. the lshift:
`Map<<String>,<String>>` vs. `0x1F << 3`
This has to be resolved in the parser, where this could be done easily based on context information.

PR #1044, opened 2018-01-17
fixes: ISSUE#1043 PLSQL create table AS should be followed by select statement instead of subquery

WolfeeTJ:
create table should be followed by select_statement instead of subquery
added drop table PURGE clause

### DIFF
--- a/plsql/PlSqlParser.g4
+++ b/plsql/PlSqlParser.g4
@@ -2078,7 +2078,7 @@ upgrade_table_clause
     ;
     
 drop_table
-    : DROP TABLE tableview_name (PURGE)? SEMICOLON
+    : DROP TABLE tableview_name PURGE? SEMICOLON
     ;
 
 comment_on_column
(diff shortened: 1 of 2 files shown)


plsql/PlSqlParser.g4
```diff
@@ -2078,7 +2078,7 @@ upgrade_table_clause
     ;
     
 drop_table
-    : DROP TABLE tableview_name SEMICOLON
+    : DROP TABLE tableview_name (PURGE)? SEMICOLON
```
KvanTTT: Please remove redundant parenthesis.

WolfeeTJ:
@KvanTTT thank you for your review. I've removed the parenthesis and added an example. I'm new to contributing to grammars-v4 and not sure how unit test here works. So please let me know if the updates are good. Thank you.

KvanTTT:
Just add a code fragment which failed on the previous version of grammar to [examples](https://github.com/antlr/grammars-v4/tree/master/plsql/examples). Also, check other samples.

PR #1049, opened 2018-01-30
Add Solidity grammar

federicobond:
See https://github.com/solidityj/solidity-antlr4/issues/14


teverett:
@federicobond thanks!

PR #1081, opened 2018-03-15
[BugFix]: [C] Function call with single argument should be recognized as "statement" but actually as "declaration".

smwikipedia:
This is to fix Issue https://github.com/antlr/grammars-v4/issues/1082.
With the old C.g4, the function call with single argument will be mistakenly recognized as `declaration`. The correct recognization result should be `statement`.
E.g.
For this input:

```
int f(int arg1, char arg2)
{
	a1(arg1);
	a2(arg1, arg2);
	a3();
}
```

**The buggy version:**
bug_funccallwithsinglearg

**The fixed version:**
bugfix_funccallwithsinglearg

**（The commits include the fix and 3 example files.）**

### DIFF
--- a/c/C.g4
+++ b/c/C.g4
@@ -478,8 +478,8 @@ blockItemList
     ;
 
 blockItem
-    :   declaration
-    |   statement
+    :   statement
+    |   declaration
     ;
 
 expressionStatement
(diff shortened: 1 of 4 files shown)


teverett:
@smwikipedia  thanks!

PR #1113, opened 2018-04-11
Fix Java compile

### LINKED ISSUE #1112: PLSQL Java parser not compiling
Looks like 2 recently Base Class do not compile.

changed it to:
PlSqlBaseParser.java

```java
import org.antlr.v4.runtime.*;

public abstract class PlSqlBaseParser extends Parser
{
    private boolean _isVersion12 = true;

    public PlSqlBaseParser(TokenStream input) {
        super(input);
    }

    public boolean isVersion12() {
        return _isVersion12;
    }

    public void setVersion12(boolean value) {
        _isVersion12 = value;
    }
}
```

PlSqlBaseLexer .java

```java
import org.antlr.v4.runtime.*;

public abstract class PlSqlBaseLexer extends Lexer
{
    public PlSqlBaseLexer(CharStream input) {
        super(input);
    }

    protected boolean IsNewlineAtPos(int pos) {
         int la = _input.LA(pos);
         return la == -1 || la == '\n';
    }
}
```

Now compile! Checking if working...

sinopsysHK:
Fix typo or copy/paste issue to compile, fixed #1112

### DIFF
--- a/plsql/Java/PlSqlBaseLexer.java
+++ b/plsql/Java/PlSqlBaseLexer.java
@@ -2,9 +2,9 @@ import org.antlr.v4.runtime.*;
 
 public abstract class PlSqlBaseLexer extends Lexer
 {
-    public PlSqlBaseLexer(ICharStream input)
-        : base(input)
-    {
+    public PlSqlBaseLexer(CharStream input) 
+	{
+		super(input);
     }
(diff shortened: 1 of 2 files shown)


KvanTTT:
Thanks!

PR #1124, opened 2018-04-26
Add single dot to function definition and calls

TDiazT:
An expression such as "function (.) x" or "myFunction(.)"

Also the ';' or NL are not always necessary in an expression (it was throwing warnings on files without a terminating NL or ';')


KvanTTT:
Ok for me, but I don't "maintain" this grammar :)

PR #1132, opened 2018-05-09
Use lazy rather than greedy regex for raw and interpreted strings

### LINKED ISSUE #1131: Golang grammar string patterns should be lazy not greedy
The Golang.p4 grammar has two bugs that can cause raw and interpreted strings to encompass to much of the input file. Specifically they are currently specified as:
```
//raw_string_lit         = "`" { unicode_char | newline } "`" .
fragment RAW_STRING_LIT
    : '`' ( UNICODE_CHAR | NEWLINE )* '`'
    ;

//interpreted_string_lit = `"` { unicode_value | byte_value } `"` .
fragment INTERPRETED_STRING_LIT
    : '"' ( UNICODE_VALUE | BYTE_VALUE )* '"'
    ;
```
but should be lazy rather than greedy:
```
//raw_string_lit         = "`" { unicode_char | newline } "`" .
fragment RAW_STRING_LIT
    : '`' ( UNICODE_CHAR | NEWLINE )*? '`'
    ;

//interpreted_string_lit = `"` { unicode_value | byte_value } `"` .
fragment INTERPRETED_STRING_LIT
    : '"' ( UNICODE_VALUE | BYTE_VALUE )*? '"'
    ;
```
I have experienced the raw string running away due to greediness. I assume the same is possible for interpreted string but have not explicitly verified that scenario.

baron1405:
This fixes #1131 by making the raw and interpreted string regex lazy.


KvanTTT:
Could you add unit-test that covers these cases?

baron1405:
@KvanTTT glad you asked for a test case. I was able to distill down the Go file in which I hit the problem to a very short and simple test case. Turns out that only the raw string should use a lazy regex. The `strings.go` file will fail to parse if the raw string regex is greedy or the interpreted string is lazy. It successfully parses if the raw string is made lazy and the interpreted string remains greedy.

PR #1134, opened 2018-05-10
csharp: add 'out type' in argument

gosp:
Support for `out var` and `out type`,

For example:

```csharp
if (_loggerAction.TryGetValue(level, out var loggingAction))
{
    loggingAction(message);
}
if (!Enum.TryParse(siteType, out SiteType site))
    throw new ArgumentException($"Unsupported site type: {siteType}.");
```

### DIFF
--- a/csharp/CSharpParser.g4
+++ b/csharp/CSharpParser.g4
@@ -77,7 +77,7 @@ argument_list
 	;
 
 argument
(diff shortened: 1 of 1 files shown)


KvanTTT:
Could you also add your sample to `AllInOne.cs` file?

PR #1136, opened 2018-05-11
xpath: fix qName/functionName rules

rovarga:
Parser grammar is incorrect in its definition of QName, which is
taken to be constrained the way FunctionName is -- to exclude
NodeType. This means that NameTest and VariableReference do not
accept legal values.

Fix this by having nCName rule contain NodeType, automatically
fixing all users of qName rule and breaking functionName.
functionName is reformulated without using qName and spelling out
the two alternatives of PrefixedName and UnprefixedName -- excluding
NodeType from UnprefixedName case.

Signed-off-by: Robert Varga <nite@hq.sk>

### DIFF
--- a/xpath/examples/example-function.txt
+++ b/xpath/examples/example-function.txt
@@ -1,0 +1,1 @@
+foo:comment()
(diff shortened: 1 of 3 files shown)


teverett:
@rovarga thanks

PR #1140, opened 2018-05-13
java9: fix identifier parsing

JanUlrich:
grammar treated words like 'with', 'open', 'to' as keywords.
But they are none. They can be used as normal Identifiers.


teverett:
@JanUlrich thanks

Issue #729, opened 2017-05-27
Javascript target + Java8.g4 -> out of memory


jundl77:
When generating the parser with the [Java8.g4](https://github.com/antlr/grammars-v4/blob/master/java8/Java8.g4) grammar, and Javascript as the target language, the parser does not terminate until it runs out of heap-memory. 

More specifically, the following code proposed in the [docs](https://github.com/antlr/antlr4/blob/master/doc/javascript-target.md) does not work with the Java 8 grammar:

```js
const chars = new antlr4.InputStream(input)
const lexer = new Java8Lexer.Java8Lexer(chars)
const tokens = new antlr4.CommonTokenStream(lexer)
const parser = new Java8Parser.Java8Parser(tokens)
parser.buildParseTrees = true
const tree = parser.compilationUnit()
```

It is the  ```const tree = parser.compilationUnit()``` that does not terminate until it runs out of memory.

Using other grammars like the [Java.g4](https://github.com/antlr/grammars-v4/blob/master/java/Java.g4) grammar works like a charm, it just seems to be the Java8.g4 that causes the out-of-memory exception.

I saw a similar problem on [SO](https://stackoverflow.com/questions/32910013/antlr-v4-java8-grammar-outofmemoryexception) with a proposed solution in Java. I tried porting the solution to Javascript with limited success. 

Perhaps it has something to do with [this issue](https://github.com/antlr/antlr4/issues/1885)?

KvanTTT:
Try a new [java8-pt](https://github.com/antlr/grammars-v4/tree/master/java8-pt) grammar.

KvanTTT:
I don't think this is a grammar issue.

PR #1144, opened 2018-05-16
Edit the documentation comment show correct target

sujeet:
The documentation said this file is for Python 2, most likely a copy-paste error.

### DIFF
--- a/python2-js/Python2.g4
+++ b/python2-js/Python2.g4
@@ -7,7 +7,7 @@
  * Added lexer rules, and code to handle INDENT's, DEDENT's,
  * line continuations, etc.
  *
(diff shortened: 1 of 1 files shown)


KvanTTT:
Incorrect, it's JavaScript runtime grammar.

sujeet:
Hi @KvanTTT, https://github.com/antlr/antlr4/blob/master/doc/targets.md seems to suggest that "target" means the language in which the parser runs, and not the language which the parser parses...

Here are other examples: https://github.com/antlr/grammars-v4/commit/663fc536f4530a687785c2521df2ef5d6a578ceb "Python3 alternative targets" says the commit message and the commit is about generating the parsers for various runtimes.

Look at the wording for https://github.com/antlr/grammars-v4/issues/729 too

PR #1148, opened 2018-05-30
Fixed issue with multiline raw string literals

ritchiecarroll:


teverett:
@ritchiecarroll thanks

PR #1151, opened 2018-06-04
 fix a bug for golang grammar

lixi520:
golang grammar would fail when parse code

`
fields := map[string]interface{}{   
"extra":{"test_update_fields":"update_fields_by_test"},  
 }  
`


teverett:
@lixi520 thanks

PR #1155, opened 2018-06-09
dot: skip all lines beginning with a '#' character

hao-lee:
The following rule doesn't work if there is no `\n` at the end of the line:
```
'#' .*? '\n'
```

I change it to:
```
'#' ~[\r\n]*
```

Signed-off-by: Hao Lee <haolee.swjtu@gmail.com>


teverett:
@hao-lee thanks

PR #1156, opened 2018-06-11
Multi-class selectors and misc fixes

wambat:


teverett:
@wambat could you add some examples pls?

wambat:
It wasn't parsing `.a.b .c.d` ,`not(...)`, `.a.b > .c.d ~ .e.f`, kinds of selectors `import (css,multiple) '...' ` and `... !important` properties. I had to change the element meaning from `class` to `class group` in the lexer. I don't think it's ass CSS-compartible, as it should be, but neither is the `origin`'s, and this version digests better.

teverett:
Makes sense.  Could you add some example files onto the tree that parse with this fix?

PR #1169, opened 2018-07-02
golang: Fixed issue with parsing interface and structure promotions

ritchiecarroll:
Both methodSpec and fieldDecl were improperly ignoring new lines and
picking up declaration on the next line as the result type. Correction
required addition of a new semantic predicates and restructuring of
methodSpec since result is optional.


teverett:
@ritchiecarroll thanks

PR #1170, opened 2018-07-02
golang: Fixed issue with parsing multi-value type declarations

ritchiecarroll:
The typeDecl should accept both ';' and new line as valid terminators, i.e., `eos` , like all other declarations that accept multiple values within a parenthetical set.


teverett:
@ritchiecarroll thanks

PR #1177, opened 2018-07-07
Update Solidity grammar with latest constructs

federicobond:


teverett:
@federicobond could you please add some more examples so that we can be sure that when new Antlr versions come out, that the grammar still works properly?

federicobond:
Sure, I'll add more examples!

PR #1193, opened 2018-07-26
ADD: PROV-N grammar

TomasKulhanek:
new grammar for W3C PROV-O standard notation as defined at https://www.w3.org/TR/prov-n/ with 2 basic examples


teverett:
@TomasKulhanek thanks!

Could you add this grammar into the root pom pls?

https://github.com/antlr/grammars-v4/blob/master/pom.xml

TomasKulhanek:
@teverett ok. Added as a new module next to existing one.

teverett:
@TomasKulhanek thanks for your contribution!

PR #1197, opened 2018-08-02
mysql lexer, enable lower case letters in variables

PHaroZ:
Hi, 

According to https://dev.mysql.com/doc/refman/5.7/en/user-variables.html
variable names could contains lower case letters, yet the lexer does not permit that (only ```A-Z0-9._$```). 

Here a small pull request to fix that.

### DIFF
--- a/mysql/MySqlLexer.g4
+++ b/mysql/MySqlLexer.g4
@@ -1128,14 +1128,14 @@ STRING_USER_NAME:                    (
                                      );
 LOCAL_ID:                            '@'
                                 (
-                                  [A-Z0-9._$]+ 
+                                  [a-zA-Z0-9._$]+ 
                                   | SQUOTA_STRING
                                   | DQUOTA_STRING
                                   | BQUOTA_STRING
                                 );
 GLOBAL_ID:                           '@' '@' 
                                 (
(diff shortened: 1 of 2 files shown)


teverett:
@PHaroZ thanks

KvanTTT:
Why did you merge this request? mysql lexer uses [CaseChangingCharStream](https://github.com/antlr/antlr4/blob/master/doc/case-insensitive-lexing.md), so there is no need to use lower-case chars.

Please let me merge *sql grammar pull requests (at least MySQL, T-SQL, PL/SQL).

PHaroZ:
Hi,

After other issues about case sensitivity (with sql keyword like ```SELECT```) and some hours spend for searching about that, I find the CaseChangingCharStream doc. I've also read the whole discussion at antlr/antlr4#2046 .
I'm ok with the solution but I think it would be nice if these informations were in the mysql/README.md : this would avoid to people like me to lose some time and, first of all, this would prevent to antlr's maintenairs to merge PR that are not required.

Anyway, thanks for your job.

KvanTTT:
Ok, you can do it if you want. Also, it makes sense to add such info to the wiki.

PHaroZ:
PR is #1200

PR #1200, opened 2018-08-03
enrich README with information about case sensitivity

PHaroZ:
refs #1197


KvanTTT:
Could you please add such info to other case-insensitive SQL grammars (PL/SQL, T-SQL)?

PHaroZ:
Sure. But I lost my source fork/branch so I created a new PR #1215 which could replace this one.

PR #1203, opened 2018-08-08
fix erase and string declaration

devjunhong:
Hello, I have 2 things to correct the grammar. 

1. [Support multiple args for Erase](https://docs.microsoft.com/en-us/dotnet/visual-basic/language-reference/statements/erase-statement) 

ex) In vba, this is possible but the previous one is limited. So, I do change **eraseStmt** in vba.g4
```vba
Sub main()
    Erase array1, array2, array3
End Sub
```

2. [Support fixed length string declaration](https://www.example-code.com/vb/fixedLength.asp)

ex) In vba, there's special declaration for string type with fixed length. Please check out **baseType**.
```vba
Sub main()
    Dim str As String * 3
End Sub 
```

Thank you! Regards, 
Junhong

### DIFF
--- a/vba/examples/example1.bas
+++ b/vba/examples/example1.bas
@@ -1,4 +1,4 @@
 Public Sub Module()
     Dim sd As Boolean
-   
+    Erase array, list
 End Sub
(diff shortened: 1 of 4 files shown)


teverett:
@devjunhong thanks

PR #1204, opened 2018-08-08
golang: Fixed issue parsing structures with functionType parameters

ritchiecarroll:
The `signature` of `functionType` was improperly picking up its optional result from the next line and consequently failing with "no viable alternative at input ...". In this case the parser should ignore anything after the line feed. Correction simply invoked the existing predicate to verify no line terminator exists after function parameters and providing an alternative that has no result. Example `struct_with_func_type.go` provided.


teverett:
@ritchiecarroll thanks

Issue #1201, opened 2018-08-07
Python Grammar for python target needs an update for new syntax


virresh:
The current python grammar for python target cannot handle syntax like:
```python
variable = [{**{'name': bear}, **content} for bear, content in bears.items()]
```

Also it has `async` and `await` nodes missing

KvanTTT:
What version of Python? 2 or 3?

virresh:
[Python3 grammar for Python3 target](https://github.com/antlr/grammars-v4/tree/master/python3-py)

I replaced the rules from java grammar and they seem to [work fine](https://travis-ci.org/virresh/grammars-v4/builds/412075702?utm_source=github_status&utm_medium=notification). Don't know if that's the appropriate way to deal though.

PR #1205, opened 2018-08-08
Python3.g4: Add support for newer python features

virresh:
Synced with the Java version
The Python version can too have ASYNC and AWAIT nodes

My attempt at it
Issue at https://github.com/antlr/grammars-v4/issues/1201

### DIFF
--- a/python3-py/Python3.g4
+++ b/python3-py/Python3.g4
@@ -36,14 +36,14 @@ grammar Python3;
 
 tokens { INDENT, DEDENT }
 
-//Note that indentation of code inside
+//Note the indentation of code inside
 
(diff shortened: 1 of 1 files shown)


teverett:
@virresh thanks for your commit.  Could you add some examples that leverage the new features pls.

virresh:
Sure, could you point me to some resource so that I can understand the testing mechanisms in use ?

Not sure where should the tests should go in

teverett:
@virresh it's pretty simple: put some python3 files that use these features into the /examples directory.

https://github.com/antlr/grammars-v4/tree/master/python3-py/examples

PR #1209, opened 2018-08-19
 Fix int options parsing in ANTLRv4Lexer.g4

AmatanHead:


teverett:
@AmatanHead thanks for the PR.  Could you add some examples for these fixes?

AmatanHead:
@teverett as I understand, there are no integer options in Antlr4 at the moment.

PR #1211, opened 2018-08-22
CHARSET option for MySQL database

jpechane:


KvanTTT:
Could you please also extend examples with the new construction?

jpechane:
@KvanTTT An example statement was added

KvanTTT:
Thanks!

PR #1239, opened 2018-09-26
system verilog grammar

rnpittman:


rnpittman:
Added my name to the contributor lists in the antlr4 repo.  That pr is still pending.

PR #1256, opened 2018-10-07
Fix drive_strength and Time_Identifier

Thomasb81:
Fix drive_strength 
Fix Time_Identifier to have optional mnpf. Issue forbid time to be second.


verilog/Verilog2001.g4
```diff
@@ -164,6 +164,7 @@ module_or_generate_item
    | attribute_instance* parameter_override
    | attribute_instance* continuous_assign
    | attribute_instance* gate_instantiation
+//   | attribute_instance* udp_instantiation
```
Thomasb81: udp_instantiation is comment. But this a diff with the ebnf of the Ieee document

PR #1261, opened 2018-10-09
[JavaScript] Improve classElement parsing.

renatahodovan:
Make sure to parse static method definitions, even if strict
mode is not applied in the global scope (the content of
JavaScript classes are inside an implicit strict scope).
In this case, the 'static' word will be recognised as an
Identifier, hence it also has to be considered in classElement
parsing.


renatahodovan:
CC @KvanTTT 
(Note: the C Sharp change is heuristic.)

PR #1265, opened 2018-10-12
Support nested statement expression for ObjectiveCParser

KaltZK:
like, 
self.attr = ({
  1 + 1
});


teverett:
@KaltZK could you add an example for this change?

PR #1267, opened 2018-10-14
clu

teverett:


clu/clu.g4
```diff
@@ -0,0 +1,304 @@
...
+iterator
+   : idn '=' 'iter' parms? args yeilds? signals? where? routine_body 'end' idn
+   ;
+
+cluster
+   : idn '=' 'cluster' parms? 'is' idn (',' idn)* where? cluster_body 'end' idn
+   ;
+
+parms
+   : (param (',' param)*)?
```
KvanTTT: Why do you use excess parentheses? They are excess. Here and below.

PR #1287, opened 2018-11-02
register is a CPP keyword

teverett:


purpl3F0x:
oh, I see thank u

teverett:
Any time

PR #1293, opened 2018-11-08
Cto

mario-s:
Grammar for [Hyperledger Composer Modeling Langauge](https://hyperledger.github.io/composer/latest/reference/cto_language.html)


teverett:
@mario-s  thanks

PR #1296, opened 2018-11-17
Forces brackets to match.

mzattera:
Accordingly to [Wikipedia](https://en.wikipedia.org/wiki/Brainfuck):

_[ and ] match as parentheses usually do: each [ matches exactly one ] and vice versa, the [ comes first, and there can be no unmatched [ or ] between the two._

Current implementation does not enforce the above.


teverett:
@mzattera thanks

PR #1299, opened 2018-11-19
MySQL allows quotes around engine names

jpechane:


mysql/MySqlLexer.g4
```diff
@@ -685,17 +685,17 @@ UTF8MB4:                             'UTF8MB4';
...
-CSV:                                 'CSV';
-FEDERATED:                           'FEDERATED';
-INNODB:                              'INNODB';
-MEMORY:                              'MEMORY';
-MRG_MYISAM:                          'MRG_MYISAM';
-MYISAM:                              'MYISAM';
-NDB:                                 'NDB';
-NDBCLUSTER:                          'NDBCLUSTER';
-PERFOMANCE_SCHEMA:                   'PERFOMANCE_SCHEMA';
+ARCHIVE:                             (SINGLE_QUOTE_SYMB | DOUBLE_QUOTE_SYMB | REVERSE_QUOTE_SYMB)? 'ARCHIVE' (SINGLE_QUOTE_SYMB | DOUBLE_QUOTE_SYMB | REVERSE_QUOTE_SYMB)?;
```
gunnarmorling: I don't think the engine list should be hard-coded, as it's a pluggable mechanism, so using a 3rd party engine such as RocksDB would fail I suppose.

Also the proposed form allows for assymetric quotes.

teverett:
@KvanTTT ?

teverett:
@jpechane thanks

PR #1301, opened 2018-11-29
remove unused token

mario-s:
There was an unused token in the cto lexer grammar. I removed it and did some grouping of the tokens for better reading.


teverett:
@mario-s thanks

PR #1309, opened 2018-12-14
Fixed java embedded code

viversba:
Name of the input class in CSharp is `InputStream` rather than `_input`. And method name changes from `La` to `LA`. This way, grammar can be generated in C# without errors.


teverett:
@viversba thanks

KvanTTT:
There are incorrect changes, it's not Java, but optimized C# runtime (https://github.com/tunnelvisionlabs/antlr4cs). Please, create another directory `CSharp` and put your changes to it.

PR #1283, opened 2018-10-30
Added JavaScript support for javascript grammar

cetfor:
This PR adds JavaScript support for the `javascript` grammar.

There is one catch, which I would like to discuss. Adding support for JavaScript requires prefixing `this.` to function calls within the grammar (`.g4`) files.  

I am wondering if this will completely break the existing `Java` and `CSharpSharwell` implementations or not.

I will elaborate on this more in another comment why `this.` needs to be prefixed to calls in the grammar files, but in short, it has to do with the way JavaScript targets are generated by ANTLR4. 

If the proposed changes to the grammar files will break existing implementations, I would like to discuss ways to make this work. 

**How to test**
**Step 1:** Save this code to `test.js`
```

const antlr4 = require("antlr4/index");
const fs = require("fs");

const { JavaScriptLexer } = require("./JavaScriptLexer");
const { JavaScriptParser } = require("./JavaScriptParser");

function readJavaScriptFile(file) {
    const filePath = `${__dirname}/${file}`;
    if (!fs.existsSync(filePath)) {
        console.log(`ERROR: Unable to locate "${filePath}"`);
        return null;
    }
    const script = fs.readFileSync(filePath, "utf8");
    return script;
}

let tests = [];
if (process.argv.length <= 2) {
    console.log(`Usage: node ${process.argv[1]} <file1, file2, ...>`);
    process.exit(-1);
} else {
    tests = [...process.argv].splice(2);
}

tests.forEach(s => {
    console.log(`Analyzing file '${s}'`);
    const script = readJavaScriptFile(s);
    if (script) {
        const inputStream = new antlr4.InputStream(script);
        const lexer = new JavaScriptLexer(inputStream);
        lexer.strictMode = false;
        const tokens = new antlr4.CommonTokenStream(lexer);
        const parser = new JavaScriptParser(tokens);
        parser.buildParseTrees = true;
        const tree = parser.program();

        // Print the LISP-formatted tree
        console.log(tree.toStringTree(parser.ruleNames));
    }
});
```

**Step 2:** Run the following commands (have node.js installed)
```
cd grammars-v4/javascript
cp JavaScript/*.js .
antlr4 -Dlanguage=JavaScript JavaScriptLexer.g4
antlr4 -Dlanguage=JavaScript JavaScriptParser.g4
npm init
npm install antlr4
node test.js examples/ArrowFunctions.js
```

PR #1314, opened 2018-12-19
[javascript] support for Go

shixzie:
Port of the grammar to Go. More detailed information in the `javascript/Go/README.md` file.


shixzie:
The problem is that in  #1283 the `this.` prefix was added to in-grammar actions. In Go the keyword `this` does not exist to refer to the struct (class in C#, Java) containing the parsing/lexing state. In Go I could make that _state_ global but would have to remove the `this.` prefix added in the previous mentioned pr, thus breaking the JavaScript target.

KvanTTT:
So, we need some additional preprocessing to make universal grammars with semantic predicates...

KvanTTT:
> I could remove the grammar files inside the Go folder and simply add a note in the README inside that folder specifying the steps to fix this problem. That way we just stick with the original grammar files.

Yes, I think this is the best solution at least for now (while CI for not-java runtimes does not work).

KvanTTT:
Thanks!

PR #1316, opened 2018-12-19
[javascript/Go] Added missing step to grammar tweaks

shixzie:
Without the added step, the generated parser file will have conflicting function names.


teverett:
@shixzie thanks

PR #1321, opened 2018-12-25
Align Kotlin grammar closer to the official one

vlsi:
Of course, official grammar is not very formal, however it is so much better to compare grammars when the same rules have the same names.

On top of that, the PR improves parsing performance significantly (6.6sec -> 0.5sec) .

Note: Test.kt still cannot be parsed in SLL mode.

Profiling "before":

Profiling "after":


KvanTTT:
@shadrina please take a look.

KvanTTT:
@vlsi maybe it also makes sense to transform expression rules to recursive ones? It should bring performance improvements.

vlsi:
@KvanTTT , It does not look like expression grammar causes slowness here. Why do you think recursive expression rules would make difference here?

The current major time consumers (according to ANLTR profiling) are as follows:

1) type 109ms, 0 ambiguous, 2436 max lookahead symbols examined.

```
type
    : typeModifierList?
    ( functionType
    | parenthesizedType
    | nullableType
    | typeReference) // <-- this alternative is slow
    ;
```

I've tried to redo `type` parsing as follows, however it is still slow: 

```
type
    : typeModifiers type # typeWithModifier
    | type NL* DOT NL* functionTypeParameters NL* ARROW (NL* type) # functionWithReceiverType
    | functionTypeParameters NL* ARROW (NL* type) # functionType
    | type NL* QUEST+ #nullableType
    | parenthesizedType #parenthesizedTyp
    | userType # userTyp
    ;
```

2) delegationSpecifier 54ms, 6 ambiguous, 

```
delegationSpecifier
    : constructorInvocation
    | userType
    | explicitDelegation
    ;
```

The problematic case is https://github.com/antlr/grammars-v4/blob/6939d301e8d0d26db497d676aaa8407214506cf0/kotlin/examples/Test.kt#L828-L831 where `b {` is ambiguous between "function literal" and class body.

3) postfixUnaryExpression 33ms, 21 ambiguous

```
postfixUnaryExpression
    : (atomicExpression | callableReference) postfixUnaryOperation*
    ;
```

The problematic input is https://github.com/antlr/grammars-v4/blob/6939d301e8d0d26db497d676aaa8407214506cf0/kotlin/examples/Test.kt#L424-L440

Note that hanging function literals are parsed as `postfixUnaryOperation` repetition (~ annotatedLambda)

vlsi:
On the other hand, it might make sense to make performance measurements for real-life like code (e.g. pick a random Kotlin project).

teverett:
Given that the previous tests pass and new tests have been added, it seems to me that this version is better than the previous version, and therefore should be merged.  after that, if people want to  discuss more, they can.

KvanTTT:
> It does not look like expression grammar causes slowness here. Why do you think recursive expression rules would make difference here?

I've made such a conversion for [PHP](https://github.com/antlr/grammars-v4/tree/master/php) grammar and got a performance improvement. But you are also right: we have to test this improvement on the real code.

vlsi:
@belyaev-mikhail , what's your suggestion then?

>The grammar in kotlin docs is incorrect and cannot be used as a source of truth for anything.

Of course there might be issues, however I do think it is still better to stick to the same naming otherwise it becomes hard to cross-check.

Was `assignableExpression` in ANTLR grammar a part of "improving the correctness of the grammar"?

In the mean time, would you please mind to review and merge https://github.com/JetBrains/kotlin/pull/2043?

KvanTTT:
Hi!

Our priority is practical usage and performance, not the formal correctness. But we can just add a new grammar `kotlin_formal` similar to [java8](https://github.com/antlr/grammars-v4/tree/master/java8) one ([java](https://github.com/antlr/grammars-v4/tree/master/java) grammar supports Java 8 but much faster).

belyaev-mikhail:
> Was assignableExpression in ANTLR grammar a part of "improving the correctness of the grammar"?

Yes. What's the problem with it?

> In the mean time, would you please mind to review and merge JetBrains/kotlin#2043?

I'm not a part of the Kotlin team, so no, I can't help you with that.

> Our priority is practical usage and performance, not the formal correctness.

Practical usage includes applicability to real-life code, doesn't it?

belyaev-mikhail:
> what's your suggestion then?

We'll make an alternative PR with our current version of the grammar and let the mantainers of this repository decide it. We are not currently actively working on the project so unless somebody else does the work of trying to merge and (more importantly) test it, I have no better strategy for resolving this.

Alternatively, we may add a new grammar, as suggested above.

belyaev-mikhail:
Please try using your grammar to parse files in this project:

https://github.com/belyaev-mikhail/kt-grammar-tester

It gave me a number of issues from the get-go. Note that this is all correct, compilable Kotlin.

belyaev-mikhail:
The other PR is here: https://github.com/antlr/grammars-v4/pull/1324

teverett:
@belyaev-mikhail if you will make a new pull request under "kotlin-formal", that would be fantastic.  Thanks for contributing to grammars-v4.

teverett:
@belyaev-mikhail thank-you very much

teverett:
@belyaev-mikhail @vlsi I'm ok to pull this PR?

belyaev-mikhail:
It is an improvement over the original, so it's ok with me.
Lots of work still needs to be done though.

PR #1323, opened 2018-12-26
Add CMake grammar

zbq:
CMake grammar reference:
https://cmake.org/cmake/help/v3.12/manual/cmake-language.7.html


teverett:
@zbq  thanks

PR #1324, opened 2018-12-26
Current spec-oriented version of the Kotlin grammar

belyaev-mikhail:
Copied from https://github.com/shadrina/kt-grammar-tester

This is the current version of the grammar we've been working on as a part of Kotlin specification initiative.

Pros:
- It has been tested on a number of real-life projects (including itself, the Kotlin compiler sans the tests and some other tens-of-thousand-line projects) and the infrastructure from https://github.com/shadrina/kt-grammar-tester easily allows adding new ones;
- It is kinda-synced with the actual parser from the Kotlin compiler using tree comparison and a big-ish dataset, including the results from the [Kotlin fuzzer](https://github.com/ItsLastDay/KotlinFuzzer)  project;
- It is most likely to be the base for the next grammar in both the Kotlin reference and Kotlin specification, modulo ANTLR4 specifics.

Cons:
- Performance was not our goal, working grammar was. It might be even slower than the original;
- It is still not Kotlin 1.3 (but so is the original grammar).


belyaev-mikhail:
To explain the changes in tests:

- The lines removed from Tests.kt are (as far as I know) not valid Kotlin code;
- This grammar (unlike the previous one) does distinguish between declaration and type modifiers and some of the usages of modifiers are now illegal;
- The `fuzzer` folder is a randomly selected subset from the files produced by the [Kotlin fuzzer](https://github.com/ItsLastDay/KotlinFuzzer) project.

vlsi:
@belyaev-mikhail , the grammar fails for the following inputs:

kotlinFile and script rules:
```
import kotlin

;
;
```

Is it something you want to explicitly forbid?

belyaev-mikhail:
@vlsi This seems stupid (and probably is), but the actual kotlin compiler forbids this, too. We're just replicating them.

belyaev-mikhail:
Just put in another import or package directive after the second semicolon, you'll get the picture right in your IDE.

belyaev-mikhail:
Same goes for strange semicolon rules between getters/setters and in some other places where the grammar is wrong, the parser is wrong or both are wrong. 
We have a special tag for these in YT: https://youtrack.jetbrains.com/issues?q=tag:%20parser-spec-needed

vlsi:
That was executed from within kt-grammar-tester, so I assume Kotlin runtime was 1.2.51

belyaev-mikhail:
Ok, different case. Seems like a bug then.

kotlin-formal/KotlinParser.g4
```diff
@@ -0,0 +1,892 @@
...
+    ;
+
+propertyDeclaration
+    : modifiers? ('val' | 'var')
+    (NL* typeParameters)?
+    (NL* receiverType NL* '.')?
+    (NL* (multiVariableDeclaration | variableDeclaration))
+    (NL* typeConstraints)?
+    (NL* ('=' NL* expression | propertyDelegate))?
+    (NL+ ';')? NL* (getter? (NL* semi? setter)? | setter? (NL* semi? getter)?)
```
vlsi: This does look complicated, however what do you think of 

```
(semi getter (NL* semi? setter)? | semi setter (NL* semi? getter)?)?
```
?

I guess it is pretty much the same grammar as you write, however it does not cause ambiguity for the case when no getter/setter is present.
If both getter and setter are absent, then "closing `(NL+ ';')? NL*` causes ambiguity with `semis` in-between `statements`.

kotlin-formal/KotlinParser.g4
```diff
@@ -0,0 +1,892 @@
...
+
+unaryPrefix
+    : annotation
+    | label
+    | prefixUnaryOperator NL*
+    ;
+
+postfixUnaryExpression
+    : primaryExpression
+    | primaryExpression postfixUnarySuffix+
```
vlsi: Is your intention here to parse "without postfixUnarySuffix first", and only then try parse with `postfixUnarySuffix`?
Unfortunately, current wording causes ANTLR to parse primaryExpression twice.

If that is the case, then single `primaryExpression postfixUnarySuffix+?` probably makes more sense.

If the alternation was carefully picked, then a comment would be really helpful for the readers.

belyaev-mikhail: IIRC we had some specific reason to do this. 
@shadrina, any ideas?

kotlin-formal/KotlinParser.g4
belyaev-mikhail: This semicolon mess is trying to replicate the compiler behaviour, which is quite illogical here. The difference between `';'` and `semi` here is important, as far as I remember.
Your variant seems legit, but I need to do some additional testing to be sure.

KvanTTT:
Can we use common samples set for kotlin performance and formal grammars?

kotlin-formal/KotlinParser.g4
KvanTTT: > If that is the case, then single primaryExpression postfixUnarySuffix+? probably makes more sense.

Maybe better `primaryExpression postfixUnarySuffix*`?

kotlin-formal/KotlinParser.g4
KvanTTT: > The difference between ';' and semi here is important, as far as I remember.

What's the difference? They are completely equal tokens, excluding the way of designation.

belyaev-mikhail: `semi` is not a token, it's a full blown rule.

belyaev-mikhail:
I'm pretty sure that the current "performance" grammar is not able to parse our full test set. But anyway, we used this projects:
https://github.com/shadrina/kt-grammar-tester
https://github.com/jetbrains/kotlin minus the tests
https://bitbucket.org/vorpal-research/kotoed

and ~86КiB of randomized samples generated by kotlin fuzzer, which I can easily provide if you want.

The current grammar (and the recently merged fork) cannot parse neither the first project nor the generated set.

kotlin-formal/KotlinParser.g4
belyaev-mikhail: There is an interesting point about these identical rules. We have an additional thing to consider every time we change something in the grammar — that is, the shape of the tree produced as the result.

You see, the only source of truth we have right now is the kotlin compiler, and the only way to assert that the result of parsing is correct is comparing the shapes of the syntax trees between ANTLR and the compiler. Even small and absolutely correct changes may lead to big changes in the shape of those trees, meaning that even if we don't break anything, we cannot be sure.

teverett:
@belyaev-mikhail is this ready to pull?

belyaev-mikhail:
@teverett Yes. Will follow up with another PR if and when the speed fixes proposed here are applied.

kotlin-formal/KotlinParser.g4
vlsi: @KvanTTT , right you are. It might be either `primaryExpression postfixUnarySuffix*?` or `primaryExpression (postfixUnarySuffix+)??`
The latter should be equivalent to the original, however it depends on whatever semantics is preferred.

shadrina: There are some cases where an expression can be parsed both like `primaryExpression` and `primaryExpression` with some suffix. It was an attempt to give a higher priority to the first option. Maybe this can be done in another way, but I remember that `primaryExpression postfixUnarySuffix*` gives unwanted result.

kotlin-formal/KotlinParser.g4
KvanTTT: Sorry, it's my fault.

kotlin-formal/KotlinParser.g4
KvanTTT: Yes, it can be. We faced with [the similar problem](https://github.com/antlr/grammars-v4/blob/master/objc/ObjectiveCParser.g4#L386) in Objective-C grammar. I.e.

```antlr
varDeclaration
    : (declarationSpecifiers initDeclaratorList | declarationSpecifiers) ';'
    ;
```

is not equvalent to

```antlr
varDeclaration
    : declarationSpecifiers initDeclaratorList? ';'
    ;
```

vlsi: @shadrina ,  please note that `(postfixUnarySuffix+)??` is (it should be) different from `postfixUnarySuffix*`.
`postfixUnarySuffix*` is a greedy rule while `(postfixUnarySuffix+)??` is non-greedy rule (i.e. it should try parse without `postfixUnarySuffix+` first, then it would retry with `postfixUnarySuffix+` (which is greedy, so it would try to match as many `postfixUnarySuffix` as it is possible).

Alternative option is `postfixUnarySuffix*?` which is non-greedy, so it would try to parse with NO `postfixUnarySuffix` first, then it would try to parse with one `postfixUnarySuffix` then it would try to parse with two `postfixUnarySuffix`, and so on.

Intuitively, I guess the most reasonable approach should be `postfixUnarySuffix*?`

shadrina: @vlsi, yeap, in that case, `postfixUnarySuffix*?` looks the most appropriate.

PR #1327, opened 2018-12-26
fixes for #568

teverett:


cpp/CPP14.g4
```diff
@@ -21,2345 +22,2013 @@
...
-	| '\\U' Hexquad Hexquad
-;
-
-Identifier
-:
-/*
+   : Assign val = Octalliteral{if($val.text.compareTo("0")!=0) throw new InputMismatchException(this);};
+   
+   /*Derived classes*/
+   baseclause
```
KvanTTT: Why do you use whitespaces before rule declarations? It does not make sense.

teverett: Antlr4Formatter does it.   I have no objection at all to you making a change to Antlr4Formatter.

PR #1336, opened 2018-12-30
Cto: enhancements

mario-s:
- namespace is mandatory
- space after REF and VAR
- LetterAndDigits at least once
- link to specific version of documentation


teverett:
@mario-s thanks

PR #1362, opened 2019-02-07
Add initial support of Rego

anadon:
The Rego Language, created by the folks at Open Policy Agent, is a high level description language to express security policies and rules.

This adds initial, but still very alpha quality, support of Rego.

More may be found at https://www.openpolicyagent.org/docs/language-reference.html .

@tsandall @timothyhinrichs @ashutosh-narkar Have some code.  I'm sorry it isn't production quality yet, but it should still save time and open up some good language clarification questions.  I'd like to see a lot more pushed from in language validation calls to grammar and lexer level rules than what is happening in the current .PEG version.


teverett:
@anadon thanks

PR #1371, opened 2019-03-04
Updated dice notation grammar

Bernardo-MG:
Fixed some errors with the grammar and added multiplication and division, along parenthesis.


teverett:
@Bernardo-MG thanks!

PR #1372, opened 2019-03-06
Adding support for keywords in CPP14 grammar

hristo-vrigazov:
Adds `and`, `not` and `or` in the CPP14 Grammar. Example programs are included.


teverett:
@hristo-vrigazov thanks!

PR #1379, opened 2019-03-20
Asn

gsongsong:
```
modules: moduleDefinition+;
```

One file may have multiple module definitions

```
 | BOOLEAN_LITERAL
 | NULL_LITERAL
```

`BOOLEAN` and `NULL` are also builtin types

```
 |  WITH_LITERAL COMPONENTS_LITERAL L_BRACE componentPresenceLists R_BRACE
;

componentPresenceLists:
   componentPresenceList? (EXTENSTIONENDMARKER (COMMA componentPresenceList)?)?
  |  ELLIPSIS (COMMA componentPresenceList)?
;

componentPresenceList: (componentPresence) (COMMA componentPresence)*
;

componentPresence: IDENTIFIER (ABSENT_LITERAL | PRESENT_LITERAL)
```

Constraint expression `Blahblah(WITH COMPONENTS{Something PRESENT, ..., Something ABSENT})`

```
	|   BSTRING
```

Default value can have `BIT STRING`

PR #1380, opened 2019-03-22
Fix ASN and add working example

gsongsong:
Fix of https://github.com/antlr/grammars-v4/pull/1379

And working example added


teverett:
@gsongsong thanks!

Issue #1295, opened 2018-11-15
TSQL: Create function/procedure name cannot have server name and database name


dxingda:
From MS document, [create function](https://docs.microsoft.com/en-us/sql/t-sql/statements/create-function-transact-sql?view=sql-server-2017) and [create procedure](https://docs.microsoft.com/en-us/sql/t-sql/statements/create-procedure-transact-sql?view=sql-server-2017), name cannot include server/database name.

```sql
func_proc_name
    : (database=id '.' (schema=id)? '.' | (schema=id) '.')? procedure=id
    | server=id '.' database=id '.' (schema=id)? '.' procedure=id
    ;
```

should be 

```sql
func_proc_name
    : ((schema=id) '.')? procedure=id
    ;
```

siggemannen:
This is fixed =)

PR #1382, opened 2019-03-26
Update TSqlParser.g4

brikken:
Resolving issue #1295


KvanTTT:
Some T-Sql tests are failed. Please take a look and fix.

brikken:
Looks like it finally passed, after a couple of typos :)
And I renamed as you suggested

Issue #1110, opened 2018-04-06
tsql:  table name not supported on create_or_alter_ddl_trigger


garymazz:
Table name "`TRIGGER1`" in `CREATE TRIGGER` statement causes parse error:
````sql
CREATE TRIGGER **TRIGGER1**
ON T1
FOR UPDATE

garymazz:
This looks like an issue surrounding selection between `create_or_alter_dml_trigger` and
`create_or_alter_ddl_trigger` rules. Further investigation is needed.

As a quick fix:
Add `table_name` rule to `ON` dependency to rule `create_or_alter_ddl_trigger` in file TSqlParser.g4:
Original:
````antlr
create_or_alter_ddl_trigger
    : ((CREATE (OR ALTER)?) | ALTER) TRIGGER simple_name
      ON (ALL SERVER | DATABASE)
      (WITH dml_trigger_option (',' dml_trigger_option)* )?
      (FOR | AFTER) ddl_trigger_operation (',' dml_trigger_operation)*
      AS sql_clauses
    ;
````
New:
````antlr
create_or_alter_ddl_trigger
    : ((CREATE (OR ALTER)?) | ALTER) TRIGGER simple_name
      ON (ALL SERVER | DATABASE | table_name)
      (WITH dml_trigger_option (',' dml_trigger_option)* )?
      (FOR | AFTER) ddl_trigger_operation (',' dml_trigger_operation)*
      AS sql_clauses
    ;

brikken:
Your table name is not a valid table identifier, ref. the [documentation](https://docs.microsoft.com/en-us/sql/relational-databases/databases/database-identifiers?view=sql-server-2017#rules-for-regular-identifiers). It should be enclosed in square brackets.

PR #1383, opened 2019-03-26
Schema not allowed in server and database triggers

brikken:
Inspired by #1110 I noticed that schema is not allowed in server and database triggers (see [documentation](https://docs.microsoft.com/en-us/sql/t-sql/statements/create-trigger-transact-sql?view=sql-server-2017#remarks-for-ddl-triggers))


brikken:
@KvanTTT Any comments? It should be ready to merge.

KvanTTT:
Yes, sure. Thanks for notification.

PR #1384, opened 2019-04-02
Add parser rule for alter table enum/set collection options.

Naros:
During an alter table modify column statement, we found we needed a clear way to identify where the collection options for an enum/set started/stopped.  It made sense to add an additional parser rule that would identify those points in the statement syntax allowing the implementation to better facilitate handling the collection options combined with the optional `charsetName` and `collationName` fragments.

### DIFF
--- a/mysql/MySqlParser.g4
+++ b/mysql/MySqlParser.g4
@@ -2003,11 +2003,7 @@ dataType
     ;
 
 collectionOptions
-    : '(' collectionOption (',' collectionOption)* ')'
-    ;
-
-collectionOption
-    : STRING_LITERAL
+    : '(' STRING_LITERAL (',' STRING_LITERAL)* ')'
     ;
 
 convertedDataType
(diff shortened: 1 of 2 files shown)


mysql/MySqlParser.g4
```diff
@@ -1994,14 +1994,22 @@ dataType
...
+      collectionOptions BINARY?
       (CHARACTER SET charsetName)? (COLLATE collationName)?         #collectionDataType
     | typeName=(
         GEOMETRYCOLLECTION | LINESTRING | MULTILINESTRING
         | MULTIPOINT | MULTIPOLYGON | POINT | POLYGON
       )                                                             #spatialDataType
     ;
 
+collectionOptions
+    : '(' collectionOption (',' collectionOption)* ')'
```
KvanTTT: Maybe it makes sense to simplify and replace `collectionOption` rule to `STRING_LITERAL`?

Naros: It does not, I've pushed that change to this PR.

PR #1386, opened 2019-04-11
Add textual WebAssembly grammar.

renatahodovan:


wat/WatLexer.g4
```diff
@@ -0,0 +1,280 @@
...
+CALL_INDIRECT: 'call_indirect' ;
+
+LOCAL_GET: 'local.get' ;
+LOCAL_SET: 'local.set' ;
+LOCAL_TEE: 'local.tee' ;
+GLOBAL_GET: 'global.get' ;
+GLOBAL_SET: 'global.set' ;
+
+LOAD : NXX '.load' (MEM_SIZE '_' SIGN)? ;
+STORE : NXX '.store' (MEM_SIZE)? ;
```
KvanTTT: Redundant parentheses.

wat/WatLexer.g4
```diff
@@ -0,0 +1,280 @@
...
+
+VAR : Name ;
+
+SPACE
+  : [ \t\r\n] -> skip
+  ;
+
+COMMENT
+  : ( '(;' .*? ';)'
+  | ';;' .*? '\n')-> skip
```
KvanTTT: Maybe it makes sense to put at least comments to the `HIDDEN` channel? But I would like to avoid the `skip` command at all.

wat/WatParser.g4
```diff
@@ -0,0 +1,398 @@
...
+value_type
+  : VALUE_TYPE
+  ;
+
+elem_type
+  : FUNCREF
+  ;
+
+global_type
+  : value_type | LPAR MUT value_type RPAR
```
KvanTTT: Please always write the first-level alternative on a new line:

```antlr
global_type
  : value_type
  | LPAR MUT value_type RPAR
  ;
```

wat/WatParser.g4
```diff
@@ -0,0 +1,398 @@
...
+  ;
+
+def_type
+  : LPAR FUNC func_type RPAR
+  ;
+
+func_type
+  : (LPAR (RESULT value_type* | PARAM value_type* | PARAM bind_var value_type) RPAR)*
```
KvanTTT: It's not a very good style (Don't you have warnings during parser generation?). I suggest replacing star `*` with plus `+` and adding questions `?` to parent rules. But the best way consists of removing `*` in this rule at all and adding `*` to parent rules. This produces more easy to handle parse trees.

wat/WatParser.g4
```diff
@@ -0,0 +1,398 @@
...
+  | BINARY
+  | CONVERT
+  ;
+
+call_instr
+  : CALL_INDIRECT type_use? call_instr_params
+  ;
+
+call_instr_params
+  : (LPAR PARAM value_type* RPAR)* (LPAR RESULT value_type* RPAR)*
```
KvanTTT: Possible zero alternative. Try to refactor.

wat/WatLexer.g4
```diff
@@ -0,0 +1,280 @@
...
+GLOBAL_SET: 'global.set' ;
+
+LOAD : NXX '.load' (MEM_SIZE '_' SIGN)? ;
+STORE : NXX '.store' (MEM_SIZE)? ;
+
+OFFSET_EQ_NAT : 'offset=' Nat ;
+ALIGN_EQ_NAT : 'align=' Nat ;
+
+UNARY
+  : IXX '.clz'
```
KvanTTT: Maybe it makes sense to introduce a token type for every instruction? It removes the need to check string values during parse tree handling (because tokens will be available).

The final result looks like this:

```antlr
unary
    : IxxClz
    | ...
    ;

IxxClz : IXX '.clz';
```

teverett:
@renatahodovan Thank-you for this contribution!

Issue #1108, opened 2018-04-03
line 2:8 mismatched input 'Hello' expecting {TOKEN_REF, RULE_REF}


psllc:
In a simple (c#/visual studio) project with the three antlr4 grammar files (LexBasic.g4, ANTLRv4Lexer.g4, ANTLRv4Parser.g4) and LexerAdapter.cs ...

attempting to parse:

```antlr
// define a grammar called Hello
grammar Hello;
r   : 'hello' ID;
ID  : [a-z]+ ;
WS  : [ \t\r\n]+ -> skip ;
```

fails with: 

```
line 2:8 mismatched input 'Hello' expecting {TOKEN_REF, RULE_REF}
line 3:0 mismatched input 'r' expecting {<EOF>, TOKEN_REF, RULE_REF, DOC_COMMENT, 'options', 'tokens', 'channels', 'import', 'fragment', 'protected', 'public', 'private', 'mode', AT}
```

parsing other more complex grammars (that build and operate as expected) also fails in the same manner... 

should note that I'm using the following packages:

```
id="Antlr4" version="4.6.5-beta001" targetFramework="net462" 
id="Antlr4.CodeGenerator" version="4.6.5-beta001" targetFramework="net462" 
id="Antlr4.Runtime" version="4.6.5-beta001" targetFramework="net462"
```

Here is the test :

```csharp
using System;
using Antlr4.Runtime;
using Antlr4.Runtime.Tree;
using Microsoft.VisualStudio.TestTools.UnitTesting;

namespace Tool.Antlr.Test
{
    [TestClass]
    public class TestANTLRv4Grammar
    {
        [TestMethod]
        public void TestMethod1()
        {
            var source = @"// define a grammar called Hello
grammar Hello;
r   : 'hello' ID;
ID  : [a-z]+ ;
WS  : [ \t\r\n]+ -> skip ;

";
            new ParseTreeWalker()
                .Walk(Activator.CreateInstance<ANTLRv4ParserBaseListener>(),
                    ((ANTLRv4Parser) Activator
                        .CreateInstance(typeof(ANTLRv4Parser),
                            new CommonTokenStream((ANTLRv4Lexer) Activator
                                .CreateInstance(typeof(ANTLRv4Lexer),
                                    new AntlrInputStream(source))))).grammarSpec());
        }
    }
}
```
[...]

PR #1387, opened 2019-04-11
Made a working LexerAdaptor

Marti2203:
Currently, Lexer Adaptor is the only obstacle that makes the antlr4 grammar unusable for a .Net project. The Grammar works for Java and now it works for .Net

### DIFF
--- a/antlr4/LexerAdaptor.cs
+++ b/antlr4/LexerAdaptor.cs
@@ -1,18 +1,30 @@
-﻿namespace GrammarGrammar
+﻿using System;
+using System.Reflection;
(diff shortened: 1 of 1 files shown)


Marti2203:
The open issue is #1108

Marti2203:
There is an issue. I forgot that the grammar has the method names with lowercase.

teverett:
it seems that you found an issue?   Should I wait to pull this?

Marti2203:
You can pull. I fixed it with the second commit.

teverett:
@Marti2203 thanks

PR #1388, opened 2019-04-11
Add missing function IIF (Instant IF)

tuxmania87:
Missing Function Instant If is basically the ternary operator of TSQL.

Imagine expression in c#
`a < b ? c  : d`

translates into TSQL
`IIF(a<b, c,d)`


KvanTTT:
Could you please add a test sample for such instruction?

PR #1391, opened 2019-04-14
Prevent String name conflict with Go target in SystemVerilog grammar

zyedidia:
The SystemVerilog grammar uses the name "String" which is reserved in Go. This causes the Go target to not compile when using the SysVerilogHDL grammar. This pull request renames "String" to "SVString" which removes the conflict.

### DIFF
--- a/system_verilog/SysVerilogHDL.g4
+++ b/system_verilog/SysVerilogHDL.g4
@@ -208,7 +208,7 @@ Small : 'small' ;
 Specify : 'specify' ;
 Specparam : 'specparam' ;
 Static : 'static' ;
-String : 'string' ;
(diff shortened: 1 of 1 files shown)


system_verilog/SysVerilogHDL.g4
```diff
@@ -208,7 +208,7 @@ Small : 'small' ;
 Specify : 'specify' ;
 Specparam : 'specparam' ;
 Static : 'static' ;
-String : 'string' ;
```
KvanTTT: Maybe rename it to `String_`?

teverett:
@zyedidia thanks

PR #1411, opened 2019-05-21
Add missing label to function_call IIF rule

Naros:
https://github.com/antlr/grammars-v4/pull/1388 which was recently merged introduced a compilation issue where not all `function_call` rules were properly labeled.  This PR addresses that by adding the missing label.


KvanTTT:
Thanks a lot for fixing!

PR #1413, opened 2019-05-23
Rework some grammars

Marti2203:
I am currently trying to look at different ways to optimize an ANTLR grammar and I saw that some of the example grammars are not very optimal. I would like to discuss various aspects of the optimisation:
* does the grammar become too obfuscated
* is this optimal and are there standards which it does not comply with

### DIFF
--- a/abnf/Abnf.g4
+++ b/abnf/Abnf.g4
@@ -52,7 +52,7 @@ rulelist
    ;
 
 rule_
-   : ID ( '=' | '=/' ) elements
+   : ID '=' '/'? elements
    ;
 
 elements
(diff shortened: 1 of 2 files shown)


teverett:
@Marti2203 

Thanks for the fixes.  Please rectify the build issues before the PR is puled.

Marti2203:
Woops, I have not tested it. Maybe a pre(commit or push) hook will stop this? This will be another thing to add. :smile:

Marti2203:
For now I will not try to rework the AGC grammar, because the examples were not parsed properly.

PR #1415, opened 2019-05-23
Added spaces to %type, %rowtype, etc.

glgraca:
Declarations such as `abc%type` can be written with spaces: `abc%  type`,

So the rule must be changed to `PERCENT_TYPE: '%' SPACE* 'TYPE';`


KvanTTT:
Could you add some samples that cover the changed lexer rules?

Issue #1419, opened 2019-05-25
Algol60 mutually Left Recursive rules


Marti2203:
The current algol60 grammar cannot be run through Antlr as it contains mutually left recursive rules. #1417 is another issue and I will make a pull request to fix both of them.

Marti2203:
Just saw the readme... Closing this issue.

Issue #1417, opened 2019-05-24
Algol60 Brackets issue


Marti2203:
The Algol60 grammar has brackets in some expressions, but they are just groupings, not literals. I will fix them in #1413.

Marti2203:
And it will be reworked as per the goal of #1413

PR #1420, opened 2019-05-25
Fixed algol60 grammar and reworked Arithmetic grammar.

Marti2203:


Marti2203:
This PR is a fix of the algol60 grammar as it had some issues (#1417 and #1419)

Marti2203:
Reworked Arithmetic grammar so it will use the full power of left recursion.

Marti2203:
I cannot understand this error. The difference is a paragraph?

teverett:
The error:

<pre>
Parsing :/home/travis/build/antlr/grammars-v4/arithmetic/examples/precedence3.txt
java.lang.Exception: Parse tree does not match 'precedence3.txt.tree'. Differences: Diff(EQUAL,"(equation (expression (atom (variable a))) (relop =) (expression (expression (atom (scientific 234))) ^ (expression ( (expression (expression (atom (scientific 4.23))) / (expression (atom (scientific 345)))) ))))"), Diff(DELETE,"¶"), 

</pre>

Is from the test framework, not from Antlr.   There is a file on the source tree here:

https://github.com/antlr/grammars-v4/blob/master/arithmetic/examples/precedence3.txt.tree

That is the "correct" output from parsing precedence3.txt.  You've changed the grammar and the results of parsing precedence3.txt.tree are now different.  Just paste the right results in precedence3.txt.tree and you're good to go.

Marti2203:
I have done that. That is why the first two tests work.

KvanTTT:
> The grammar is reworked a lot to be a proper Antlr grammar without any syntactic sugar (custom actions/ labels/ identifiers).

What's wrong with labels and identifiers? They are natural to Antlr as well as other syntactic constructions. Customs actions are a bit awkward because they are not universal.

Marti2203:
I have no issues, I like them(labels and identifiers) and use them in my grammars, but I try to make the code in these grammars without them.

Marti2203:
In the commit, I have changed the trees which are used for the tests. I understand that the error is from the test framework, but I do not understand what I should remove to have no differences.

Marti2203:
Found the issue. A phantom new line was the issue. Now I am reworking the other trees and I think that everything is okay. Sorry if I am spamming you.

teverett:
@Marti2203  thanks

PR #1426, opened 2019-05-26
Unicode letter (category L) should include full CJK range

studentmain:
Affected: Javascript, ECMAScript, Kirikiri-TJS, Golang, Kotlin-formal

Seems caused by a malfunctioned generation script, not sure.

BTW, I noticed that different grammar use different Unicode letter range, should we update them to newest standard?


teverett:
@studentmain Thanks for the contribution.  This affects numerous grammars so I'll let it sit a day or two before I pull it, in order to give interested people a chance to comment.

teverett:
@studentmain  thanks

Issue #231, opened 2015-10-30
"The name '_input' does not exist in the current context"


tianhuat:
I got the error when using ECMAScript.CSharpTarget.g4, with C# target (antlr-csharp-runtime-4.5.1.zip) download here:
http://www.antlr.org/download.html

"The name '_input' does not exist in the current context"

parrt:
likely it has embedded java code or something.

jpwiddy:
I get this same exact error. 

I was using the ECMAScript.g4

ajdnaranjo:
I got the same error when I've "CSharpSharwell" grammar  is generated. I'm using antlr-4.7-complete.jar to generate C# classes and nuget package Antlr4.Runtime.Standard.4.7.1

![img](https://user-images.githubusercontent.com/17502787/32806352-39298ed4-c95a-11e7-8e92-4c1b8c557d42.PNG)

Can anyone helpme whit this issue

viversba:
Were you able to solve it @ajdnaranjo ?
I'm having the exact same problem

Marti2203:
Hello, I think I may have resolved this issue.

eeskildsen:
I get the same error in [`PlSqlLexerBase`](https://github.com/antlr/grammars-v4/blob/805b32407c92cc496f55bfdcfd9ef8c897757214/sql/plsql/CSharp/PlSqlLexerBase.cs#L14).

The problem seems to be that [`_input` is `private` in the C# `Lexer` base class](https://github.com/antlr/antlr4/blob/4779b8f6e45fb88261efb984cf9dbd2243e0cb2b/runtime/CSharp/runtime/CSharp/Antlr4.Runtime/Lexer.cs#L35). Looks like it needs to be `protected`.

Marti2203:
I can work on it if needed @eeskildsen .

PR #1427, opened 2019-05-27
Reworked CLU! And a fix for #231

Marti2203:
@teverett  Asked if  I could try to rework the grammar for CLU and I think I have made it working. I have not tested it for there are no examples in the folder, but the grammar is almost the same as the original one (save for the _list rules which reduce the clutter).

### DIFF
--- a/clu/clu.g4
+++ b/clu/clu.g4
@@ -1,360 +1,222 @@
 /*
-BSD License
-
-Copyright (c) 2018, Tom Everett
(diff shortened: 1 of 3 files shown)


Marti2203:
I saw issue #231 and I made a little fix for it. I hope someone tests it.

teverett:
@Marti2203  thanks

Marti2203:
No problem!

PR #1433, opened 2019-06-05
Kirikiri-TJS parser optimization

studentmain:
- Correct expression order.
- OctetLiteral now is a single token.
- Speed is 1-3 times faster than old one.
- Refactor and rename.
- Switch to Kirikiri official test script


teverett:
@studentmain  thanks!

PR #1444, opened 2019-06-18
Add XSD regular expression grammar

rovarga:
This adds grammars that handle XML Schema 1.0 regular expressions
as defined in http://www.w3.org/TR/2004/REC-xmlschema-2-20041028/.

Since there are some amguities in the spec as to how dashes are
handled in character groups, we borrow parts of XML Schema 1.1
clarifications to provide unsurprising results.


teverett:
@rovarga what a great contribution!  Thanks!

PR #1445, opened 2019-06-18
Fix xpath unicode TODO

rovarga:
antlr-4.7 can handle unicode larger than U+FFFF, use the appropriate
syntax to express them -- thus addressing a leftover TODO.


teverett:
@rovarga any chance you could add an example to /examples for this one?

teverett:
@rovarga thanks

PR #1457, opened 2019-07-03
Update PlSqlBaseLexer.py

em23:
LA(x) returns an int for python (on other targets returns string/char). ASCII int for newline '\n' is 10. Another alternative could be using chr(la) == '\n' or la == ord('\n')


teverett:
@em23 thanks

PR #1460, opened 2019-07-11
Scientific notation exponent must be an integer

SSteve:
- Modified SCIENTIFIC_NUMBER so that the exponent is an integer.
- Edited comment that referred to a non-existent rule.
- Corrected examples that had invalid exponents.


teverett:
@SSteve  thanks

PR #1463, opened 2019-07-15
Support for third party engines

jpechane:


mysql/MySqlParser.g4
```diff
@@ -1889,9 +1889,13 @@ engineName
     : ARCHIVE | BLACKHOLE | CSV | FEDERATED | INNODB | MEMORY
     | MRG_MYISAM | MYISAM | NDB | NDBCLUSTER | PERFORMANCE_SCHEMA
     | TOKUDB
+    | customEngineName
```
KvanTTT: Why don't use just an `ID` token without adding new `customEngineName` rule? Also, you can use an element label in this case: `customEngineName=ID`.

jpechane: @KvanTTT I did it intentionally, so the generated code will contain reference to not generic id. The label is probably a better solution so I'll try it out and modify the PR if it works as expected

jpechane: @KvanTTT Updated

KvanTTT:
Thanks

PR #1471, opened 2019-07-24
Add Lars Pensjö C grammar which being used in games.

hanjianqiao:
Adding grammar for LPMud game language.


teverett:
@hanjianqiao thanks

Issue #1484, opened 2019-08-12
Grammar for goland generates 4 Java functions


fred1268:
When generating the golang grammar in Go (i.e. to parse Go file in Go), the following functions are generated in Java:

lineTerminatorAhead(...), noTerminatorBetween(...) & noTerminatorAfterParams(...) in golang_parser.go file

nextToken(...) in golang_lexer.go

Expected behavior: these 4 functions should be implemented in golang, not in Java.

PR #1497, opened 2019-08-23
Golang

fred1268:
Split Golang.g4 into GolangLexer.g4 and GolangParser.g4
Delete Golang.g4 since it is no more necessary
Create the BaseLexer and BaseParser in both Java and Go

This should solve issue #1484 that I created few days ago.


KvanTTT:
We are also working on it: https://github.com/PositiveTechnologies/grammars-v4/tree/GoLangRefactoring/golang It would be great to merge our changes later.

KvanTTT:
@fred1268 what about CI fixes? Unfortunately, we can not merge this PR without them.

fred1268:
Sorry @KvanTTT . Will try to make it happen shortly. Was pretty busy on other topics.

KvanTTT:
Ok, take your time.

fred1268:
Hello @KvanTTT . Sorry for the delay, I did not even realize that what you asked for was so quick otherwise I would have taken the time to do it earlier.
I hope this will go in the right direction: now at least we do have lexer and parser separated, and all functions are written in Go in addition to Java, on the same model than JavaScript.
Best,
Fred.

KvanTTT:
Thanks a lot!

PR #1500, opened 2019-08-28
Make java8 universal. Resolve #1499

### LINKED ISSUE #1499: Merge java8 java8-js and java8-ts grammar?
As they're same grammar with different target.

Java8 g4 itself use `java.lang.Character.isJavaIdentifierStart`and `java.lang.Character.isJavaIdentifierPart` which is not portable. According to [doc](https://docs.oracle.com/javase/8/docs/api/java/lang/Character.html#isJavaIdentifierPart-int-), seems we can replace them with lexer rules to make it portable.

studentmain:
1. Split Java8.g4 to lexer and parser
2. No longer use JRE to check identifier character
3. That makes other language runtime unnessary


KvanTTT:
As I understand, now it's Java-target only?

studentmain:
No, it's cross platform now.

KvanTTT:
What is the meaning of `java8/CodepointRangeScanner.java`?

studentmain:
Used to generate these:

https://github.com/antlr/grammars-v4/blob/919d1d15baf7328b9262a48911c87377a77f4024/java8/Java8Lexer.g4#L469-L1104

KvanTTT:
Ok then. Thanks for your contribution!

PR #1507, opened 2019-09-08
Changed EOL token in muddb and tinymud grammars to support non-unix line endings

nils91:
...also to make the tests work in non-unix environments.


teverett:
@nils91  Thanks!

PR #1527, opened 2019-10-13
added type system as per June 2018 GraphQL specs

innomon:
added type system as per June 2018 GraphQL specs
https://graphql.github.io/graphql-spec/June2018/


teverett:
@innomon  thanks

PR #1530, opened 2019-10-21
Support for IN/VISIBLE index option

jpechane:


mysql/Positive-Technologies/MySqlParser.g4
```diff
@@ -299,6 +299,7 @@ indexOption
     | indexType
     | WITH PARSER uid
     | COMMENT STRING_LITERAL
+    | (INVISIBLE | VISIBLE)
```
KvanTTT: Simplify:

```antlr
    | INVISIBLE
    | VISIBLE
```

jpechane: @KvanTTT Done

Issue #1532, opened 2019-10-21
[Tsql] An issue in select_list_elem: asterisk


wcx11:
Dear author: I found out the following grammar in TsqlParser.g4 could be an issue.
(about line 3008 to 3011) :
`asterisk
    : '*'
    | table_name '.' asterisk
    ;`
it's equivalent to `asterisk: (table_name '.')* '*'` which means the selected_list_elem could be something like `table1.table1.table1.*`. 
Actually only things like `table1.*` or `*` are valid.
Could you please double check it? 
Thanks for providing such helpful tools~

KvanTTT:
Would you like to suggest a pull request?

wcx11:
Pull request created at : https://github.com/antlr/grammars-v4/pull/1533

PR #1533, opened 2019-10-22
fix Tsql asterisk

wcx11:
change asterisk from `asterisk : '*' | table_name '.' asterisk` to `asterisk : (table_name '.')? '*'` 
Before this change, `selected_list_elem` could be something like `table1.table1.table1.*`. Actually only things like `table1.*` or `*` are valid.
Please refer to issue: https://github.com/antlr/grammars-v4/issues/1532


KvanTTT:
Thanks for the first contribution!

PR #1534, opened 2019-10-26
Added hex integer and block comments.

stefan301:
Added hexadecimal integer constants for enum values and block comments.


teverett:
@stefan301 thanks

PR #1537, opened 2019-10-27
sickbay

teverett:


sickbay/sickbay.g4
```diff
@@ -0,0 +1,95 @@
...
+sickbay
+   : line*
+   ;
+
+line
+   : intExpr stmt (':' stmt)* NL
+   ;
+
+stmt
+   : ('REM' ARBTEXT)
```
KvanTTT: What is the meaning of parentheses for each alternative? They are redundant.

teverett: I find it easier to read

PR #1542, opened 2019-10-29
Tsql KILL Statement

### LINKED ISSUE #1529: TSQL KILL statement not properly supported
It seems that kill statement isn't supported.

following sql file:

```sql
SELECT 5
KILL 1619
```

renders following error:

```
line 2:0 extraneous input 'KILL' expecting {<EOF>, 'ALTER', 'BACKUP', 'BEGIN', '.... '$', '(', ';', '+', '-', '~'}
```

Silvenga:
Resolves #1529


Silvenga:
Not 100% confident in my knowledge of Antlr or TSQL. I would appreciate any criticisms or corrections 😸.

tsql/TSqlParser.g4
```diff
@@ -2489,6 +2490,24 @@ backup_service_master_key
...
+    ;
+
+// https://docs.microsoft.com/en-us/sql/t-sql/language-elements/kill-query-notification-subscription-transact-sql
+kill_query_notification
+    : QUERY NOTIFICATION SUBSCRIPTION (ALL | subscription_id=DECIMAL)
+    ;
+
+// https://docs.microsoft.com/en-us/sql/t-sql/language-elements/kill-stats-job-transact-sql
+kill_stats_job
+    : STATS JOB job_id=DECIMAL
```
KvanTTT: Do you consider the element label `job_id` is necessary here? `DECIMAL` is easily reachable in Visitor. The same not for the rules above.

Silvenga: Not functionally no, I added it to make the generated grammar more descriptive/discoverable without needing to reference the source grammar file.

```csharp
var context = new TSqlParser.Kill_stats_jobContext();
var jobId1 = context.DECIMAL();
var jobId2 = context.job_id;
```

Is it a good practice to not generate an extra field if it's not functionally required? _I'm still reading the book..._

KvanTTT: Actually not so important, but I prefer as such a minimal grammar as possible.

Silvenga: I'll keep that in mind, thanks.

PR #1544, opened 2019-10-30
Fixed the grammar for  Compatibility Acceptance Tests

innomon:
Added Bool, String Block, bug fixes also validates syntax against GraphQL Compatibility Acceptance Tests 
now parses: https://raw.githubusercontent.com/graphql-cats/graphql-cats/master/scenarios/validation/validation.schema.graphql

### DIFF
--- a/graphql/GraphQL.g4
+++ b/graphql/GraphQL.g4
@@ -27,10 +27,14 @@
  AB:10-sep19: replaced type with type_ to resolve conflict for golang generator
  
  AB: 13-oct-19: added type system as per June 2018 specs
+ AB: 26-oct-19: added ID type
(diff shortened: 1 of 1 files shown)


graphql/GraphQL.g4
```diff
@@ -232,3 +248,8 @@ fragment EXP: [Ee] [+\-]? INT;
 // \- since - means "range" inside [...]
 
 WS: [ \t\n\r]+ -> skip;
+
+LineComment
+    :   '#' ~[\r\n]*
+        -> skip
```
KvanTTT: It is better to use `channel(HIDDEN)`  to prevent loss of tokens, even hidden.

teverett:
@innomon  thanks

PR #1548, opened 2019-11-01
Fixed javascript assignment expressions order and a 'sed' script

yanmingsohu:
The README of the 'sed' script needs to be updated, but I can't speak English well.


javascript/Cpp/pitch.sed
```diff
@@ -0,0 +1,41 @@
+#
+# https://github.com/antlr/grammars-v4/blob/master/javascript/Cpp/README.md
+#
+# sed -f pitch.sed JavaScriptParser.g4 > JavaScriptParser-CppRuntime.g4
+# sed -f pitch.sed JavaScriptLexer.g4  > JavaScriptLexer-CppRuntime.g4
+#
+# Replease 'this.' TO 'this->';
```
KvanTTT: Typo: `Replease` -> `Replace`.

javascript/Cpp/pitch.sed
```diff
@@ -0,0 +1,41 @@
...
+
+/}/ {
+  z
+  h
+  d
+}
+
+/.*/ {
+  H
+}
```
KvanTTT: Is it a universal script for targeting .g4 files to C++ runtime? Can we place this file to the root directory and add info to README how to use it? As I understand the similar info is described in https://github.com/antlr/grammars-v4/tree/master/javascript/Cpp

@mike-lischke what do you think about it?

studentmain: > Is it a universal script for targeting .g4 files to C++ runtime?

Seems not yet, as there's hard coded #include, but maybe there's a chance to make it universal?

mike-lischke: Looks like this is some local stuff for the Javascript grammar with a C++ target. Nothing that should go into a global scope.

KvanTTT: Ok, I'm leaving it as is until better times.

teverett:
@yanmingsohu  thanks

PR #1553, opened 2019-11-04
Js: esnext

studentmain:
Update js grammar to recent standard


studentmain:
As CI not works in JS grammar, a manually test may need.

Known problem, help wanted:
- Can't parse `for of`. Works with IDEA plugin, fail in grun (p("of") fail)
  - Tried `Of: "of";`, but it still recognized as Identifier.
- Nested template string won't work: \`outer${\`inner\`}\`

javascript/JavaScriptLexer.g4
```diff
@@ -33,6 +34,7 @@ channels { ERROR }
 
 options { superClass=JavaScriptBaseLexer; }
 
+HashBangLine:                   '#!' ~[\r\n\u2028\u2029]*; // only allowed at start
```
KvanTTT: Have you seen how it is handled in PHP grammar? https://github.com/antlr/grammars-v4/blob/master/php/PhpLexer.g4#L46

javascript/README.md
```diff
@@ -90,6 +90,19 @@ Grammar supports the following list of ECMAScript 6 features taken from
 
 See [examples](examples) directory with test data files.
 
+### ES6 to ES2020
```
KvanTTT: Please always use two line breaks after headers.

javascript/README.md
```diff
@@ -105,6 +118,17 @@ Also this grammar supports outdated syntax such as
   * Cleared & optimized
   * Universal code actions & predicates
   * Support of some outdated syntax (Html comment, CData)
+* Student Main (2019):
+  * Update to ES2020
+
+## Running fuzz test
```
KvanTTT: Insert line break.

javascript/JavaScriptParser.g4
```diff
@@ -184,7 +188,7 @@ tryStatement
     ;
 
 catchProduction
-    : Catch '(' Identifier ')' block
+    : Catch ('(' (Identifier|arrayLiteral|objectLiteral)? ')')? block
```
KvanTTT: Please, always use space before and after `|` character.

javascript/README.md
```diff
@@ -90,6 +90,21 @@ Grammar supports the following list of ECMAScript 6 features taken from
 
 See [examples](examples) directory with test data files.
 
+
```
KvanTTT: But I meant to insert a line break after the header but not before :) Also, see [MD022](https://github.com/DavidAnson/markdownlint/blob/master/doc/Rules.md#md022).

studentmain: My mistake. Fixed.

KvanTTT:
> Known problem, help wanted:

Maybe extract them to a new issue? We can fix them later.

studentmain:
> Maybe extract them to a new issue?

Sounds good. I'll list not implemented feature in a new issue here later.

But about `for of`, I'm not sure where the problem is. Post it in another issue?

studentmain:
Finished, ready to merge.

KvanTTT:
> But about for of, I'm not sure where the problem is. Post it in another issue?

Yes, I think so.

PR #1557, opened 2019-11-10
golang: Updated CSharp GoParserBase to properly compile with run-time

ritchiecarroll:
Updated CSharp `GoParserBase.cs` to compile with Antlr4 CSharp run-time. For example, `_input` in the `Antlr4.Runtime.Parser` for CSharp is private and exposed via `TokenStream` property.

### DIFF
--- a/golang/CSharp/GoParserBase.cs
+++ b/golang/CSharp/GoParserBase.cs
@@ -13,10 +13,12 @@ namespace GoParseTree
         {
         }
 
+    #if ANTLR_STANDARD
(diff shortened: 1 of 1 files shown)


ritchiecarroll:
OK - I'll check this out, I presume this is the optimized repo you were referring to:
https://github.com/tunnelvisionlabs/antlr4cs

ritchiecarroll:
@KvanTTT - OK, I've updated the base parser class to support both "standard" and "optimized" run-times. When the `ANTLR_STANDARD` preprocessor directive is defined, the standard run-time can be used.

KvanTTT:
OK, thanks.

PR #1561, opened 2019-11-14
fix balancedtoken rule.

delguoqing:
should match any token other than bracket, brace, parenthesis.


teverett:
@delguoqing thanks

PR #1566, opened 2019-11-22
url: ipv6 support and few correction

studentmain:
In old rule, 127.0.0.1 is `STRING` token... so I removed `hostnumber` rule


teverett:
@studentmain  thanks

PR #1568, opened 2019-11-24
antlr4: support action blocks inside options spec (fixes #1567)

### LINKED ISSUE #1567: ANTLRv4Lexer does not support action blocks inside options spec
Within an `options` construct, action blocks are not recognized as valid option values:

image

I suppose once an `LBRACE` has been detected, the following `{` should be a `BEGIN_ACTION` and start the `Action` mode.

I'll starting working on a PR to fix this.

bjansen:
Here's my attempt at fixing #1567.

I suppose it could also be fixed by introducing a new mode like this:
```antlr
mode Options;
...
OPT_LBRACE
   : LBrace -> type (LBRACE) , pushMode (OptionsBlock)
   ;
...

mode OptionsBlock;
...
OPT_BEGIN_ACTION
   : LBrace -> type (BEGIN_ACTION) , pushMode (Action)
   ;
...
```
But the grammar feels less readable with an extra mode.

I wonder if you guys can find a better alternative?

Also, I have only adapted `LexerAdaptor.java` at the moment. I'll wait for your feedback before trying to modify other target languages.

### DIFF
--- a/antlr4/ANTLRv4Lexer.g4
+++ b/antlr4/ANTLRv4Lexer.g4
@@ -396,7 +396,8 @@ OPT_LINE_COMMENT
    ;
 
 OPT_LBRACE
-   : LBrace -> type (LBRACE)
+   : LBrace
+   { handleOptionsLBrace(); }
    ;
 
 OPT_RBRACE
--- a/antlr4/examples/Issue1567.g4
+++ b/antlr4/examples/Issue1567.g4
@@ -1,0 +1,11 @@
+grammar Sample;
+
+options {
+ foo = {};
+}
+
+@members {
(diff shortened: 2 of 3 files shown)


bjansen:
I think I broke the parsing of `@members{}` when then contain braces (like method bodies). Please wait until I fix it before merging this PR :)

bjansen:
OK I fixed the problem and updated the test case to check that everything is parsed without errors.

teverett:
Are we ok to merge this?

bjansen:
I'm ok

>

teverett:
@bjansen thanks

PR #1569, opened 2019-11-25
Subprogram body inside package declaration

kicsihobi:
Don't know about the LRM, but having small subprograms with body inside the declarative part of a package is generally used in my experience.


teverett:
@kicsihobi  thanks

kicsihobi:
I thank you the great work you guys are doing!

Issue #1571, opened 2019-11-30
Protobuf3.g4 : not properly recognized if type and field options used with reserved words


bvrygud:
```
syntax = "proto3";
package statustest;

import "envoyproxy/protoc-gen-validate/validate/validate.proto";
import "google/rpc/status.proto";

package helloworld;

// The greeter service definition.
service Greeter {
  // Sends a greeting
  rpc SayHello (HelloRequest) returns (HelloReply) {
    option (google.api.http) = {
        post: "/service/hello"
        body: "*"
    };
  }
}

// The request message containing the user's name.
message HelloRequest {
  string name = 1  [(validate.rules).string.pattern = "^\\w+( +\\w+)*$"]; // Required. Allows multiple words with spaces in between, as it can contain both first and last name;
}

// The response message containing the greetings
message HelloReply {
  string message = 1;
  google.rpc.Status status = 2;
}
```

The above proto fails at following places
1. It is not able to capture google.rpc.Status (rpc is reserved keyword)
2.   string name = 1  [(validate.rules).string.pattern = "^\\w+( +\\w+)*$"];   -- Fails as it contains string (resreved keyword) in between option

loonydev:
Hi, I already added to my protobuf optionBody parsing, it's not a problem. But I'm not sure if it's a part of "vanilla" protobuf syntax.
Regarding rpc in type of variable, i'll fix it soon

bvrygud:
Thanks @loonydev . 
Regarding the issue with field options containing reserved word, even that is not a vanilla syntax as you mentioned.

PR #1572, opened 2019-12-02
Add optionBody and reserved word in type for Protobuf3

loonydev:
Close https://github.com/antlr/grammars-v4/issues/1571


teverett:
@loonydev thanks

PR #1573, opened 2019-12-02
Add Python3_without_actions

RobEin:
Let's add this Python solution to this grammar collection.


KvanTTT:
What advantages of grammar without actions if target-specific code (Java) is used anyway? Moreover, this code is very big.

RobEin:
Hi Ivan,

First of all, this was my first pull request int the GitHub and I would like to apologize if I did something wrong.

"What advantages of grammar without actions if target-specific code (Java) is used anyway?"
There is no (Java) code in the grammar.
Sorry, but I do not understand it.
What do you think exactly?

"Moreover, this code is very big."
What is it big compared to?
My code is about 230 lines, the original (Bart Kiers) code is about 130 lines.
The 100 line difference contains a couple of new features:
- more detailed and accurate token metadata (see grun)
- detection of inconsistent dedent (half dedent)
- mixture of space and tab indentation warning

Best regards,
Robert

KvanTTT:
> There is no (Java) code in the grammar. Sorry, but I do not understand it. What do you think exactly?

Yes, it is. But there is `LexerWithIndentDedentInjector` that is required for lexing. And it is written in Java. If you want to use another target (C#, JavaScript, Python and so on), you have to rewrite it to the respective language despite the fact that grammar doesn't contain actions.

RobEin:
Unlike most languages, I do not think that a parser for an indentation language (e.g. Python) can be implemented purely from a grammar without code.
I haven't seen it in ANTLR4 yet.

teverett:
@RobEin Thanks

RobEin:
I am glad that I could contribute.

Issue #1570, opened 2019-11-30
skipping NEWLINE-only lines any single_input in Python grammar


rockobonaparte:
Note that this is quite possibly a screwup on my end and I have a lot of layers going on here that are making it hard to isolate back with just the original grammar, but I have advanced enough to try to file something, risk sounding stupid, and hopefully learn something at the worst.

I have been trying to use the C# Python3 grammar and I think I'm hitting a hiccup here:

https://github.com/antlr/grammars-v4/blob/8803cbe8aca1b3b6ebf73e13f5c41665f393b381/python3-cs/Python3.g4#L361

It's been parsing great from the file_input rule, but I'm starting to use the single_input rule in a REPL. That REPL is having a problem completing itself if I do a compound statement like:
```python
if True:
  a = 1

```
I'm using a parse error listener for now to try to tell line-by-line that I need to continue and use the secondary prompt. Per the Python REPL, I should terminate that inner block with a newline on its own line. However, I keep seeing an error that I'm missing a NEWLINE at <EOF>. If I look at the tokens, I have a growing pile of newlines as I keep hitting enter to try to send it through. The lexer is trying to make them into newlines.

Anyways, that skip on that line seems to trigger on these empty newlines. I am wondering if that could be a problem with the REPL.

I see that skip triggering for any file where I have newlines even if I have some white space before it. So let's say * is white space in this bit:
```python
if True:
**if True:
****a = 1
**

```

I'm generally curious what the NEWLINE logic is trying to do since it's kind of opaque to me, and I have some sense I might have tripped on a situation here. I don't know if the testing coverage includes REPL scenarios. I am kind of assuming not because it would imply a separate test bench that spoon feeds input into the parser.

rockobonaparte:
Additional information: My setup does start working if I get rid of all the logic for evaluating when to skip around that block of code in the NEWLINE code. It looks like that is stripping out some NEWLINE tokens from the parser that satisfy the single_input rule.

So at this point I'm trying to figure out why that skip logic was added originally.

rockobonaparte:
I was going to email @bkiers about this, but I saw in his profile to @ him first. This logic comes from the original that has the assisting code written in Java.

bkiers:
> what the NEWLINE logic is trying to do

It determines if there is a need to insert `NEWLINE`, `INDENT` and `DEDENT` tokens to the lexer's token stream. For example, the input:

```py
numbers = [
  42
]
```
[...]

PR #1576, opened 2019-12-04
Python NEWLINEs at EOF are no longer stripped

rockobonaparte:
This should address this issue:
https://github.com/antlr/grammars-v4/issues/1570

The lexer rule for NEWLINEs strips spurious newline characters inside of
various declarations like lists and dictionary declarations--anything being
that triggered Opened++ or Opened--. The parser rules for these declarations
don't accommodate NEWLINEs. Generally, this works when parsing using
file_input as the initial rule.

However, the single_input rule requires a NEWLINE at the end of its input.
This is the parser rule used by the Python REPL. The lexer code would strip
out these NEWLINEs. A solution that seems to work is to particularly avoid
stripping a NEWLINE right before EOF.

### DIFF
--- a/python3/Python3.g4
+++ b/python3/Python3.g4
@@ -334,8 +334,12 @@ NEWLINE
    {
      String newLine = getText().replaceAll("[^\r\n\f]+", "");
      String spaces = getText().replaceAll("[\r\n\f]+", "");
+	 
+     // Strip newlines inside open clauses except if we are near EOF. We keep NEWLINEs near EOF to
+     // satisfy the final newline needed by the single_put rule used by the REPL.
      int next = _input.LA(1);
-     if (opened > 0 || next == '\r' || next == '\n' || next == '\f' || next == '#') {
+     int nextnext = _input.LA(2);
+     if (opened > 0 || (nextnext != -1 && (next == '\r' || next == '\n' || next == '\f' || next == '#'))) {
        // If we're inside a list or on a blank line, ignore all indents, 
        // dedents and line breaks.
        skip();
--- a/python3alt/AltPython3.g4
+++ b/python3alt/AltPython3.g4
@@ -866,9 +866,12 @@ NEWLINE
    {
      String newLine = getText().replaceAll("[^\r\n\f]+", "");
      String spaces = getText().replaceAll("[\r\n\f]+", "");
-     int next = _input.LA(1);
 
-     if (opened > 0 || next == '\r' || next == '\n' || next == '\f' || next == '#') {
+     // Strip newlines inside open clauses except if we are near EOF. We keep NEWLINEs near EOF to
+     // satisfy the final newline needed by the single_put rule used by the REPL.
+     int next = _input.LA(1);
+     int nextnext = _input.LA(2);
+     if (opened > 0 || (nextnext != -1 && (next == '\r' || next == '\n' || next == '\f' || next == '#'))) {
        // If we're inside a list or on a blank line, ignore all indents, 
        // dedents and line breaks.
        skip();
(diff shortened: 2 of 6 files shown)


rockobonaparte:
I think the following contributors would want to see this--assuming I didn't blow something in CI:
@bkiers
@teverett
@KvanTTT 

I originally made the changes from the C# grammar and then tried to infer the others. I've never played with TypeScript, and ironically the Python syntax caught me off-guard, but I *think* I did it right.

python3/Python3.g4
```diff
@@ -334,8 +334,12 @@ NEWLINE
    {
      String newLine = getText().replaceAll("[^\r\n\f]+", "");
      String spaces = getText().replaceAll("[\r\n\f]+", "");
+	 
+     // Strip newlines inside open clauses except if we are near EOF. We keep NEWLINEs near EOF to
+     // satisfy the final newline needed by the single_put rule used by the REPL.
      int next = _input.LA(1);
-     if (opened > 0 || next == '\r' || next == '\n' || next == '\f' || next == '#') {
+     int nextnext = _input.LA(2);
+     if (opened > 0 || (nextnext != -1 && (next == '\r' || next == '\n' || next == '\f' || next == '#')))
```
KvanTTT: Missing curly at the end? `{`

python3alt/AltPython3.g4
```diff
@@ -866,9 +866,12 @@ NEWLINE
...
      String newLine = getText().replaceAll("[^\r\n\f]+", "");
      String spaces = getText().replaceAll("[\r\n\f]+", "");
-     int next = _input.LA(1);
 
-     if (opened > 0 || next == '\r' || next == '\n' || next == '\f' || next == '#') {
+     // Strip newlines inside open clauses except if we are near EOF. We keep NEWLINEs near EOF to
+     // satisfy the final newline needed by the single_put rule used by the REPL.
+     int next = _input.LA(1);
+     int nextnext = _input.LA(2);
+     if (opened > 0 || (nextnext != -1 && (next == '\r' || next == '\n' || next == '\f' || next == '#')))
```
KvanTTT: Missing curly at the end? `{`

rockobonaparte:
Yeah it's the curly bracket in both cases. The C# coding style puts them on the next line but I didn't carry that over. I add them at the end of the line like I think the other languages prefer and am waiting on CI.

python3/Python3.g4
rockobonaparte: Fixed, although it took me two shots to post it for some reason.

rockobonaparte:
Is there anything else I need to do? I admit I don't do a lot of editing in GitHub code reviews so I'm hoping I pulled the right levers after making some fixes to keep everybody notified.

KvanTTT:
What about universal python grammar? Does it require this fix?

rockobonaparte:
I never tested it because I hadn't instrumented my code to use that grammar, but I did look at it while auditing which grammars had similar blocks that needed to be fixed. I don't see anywhere in the universal grammar--with an emphasis on scrutinizing the lexer rules--where it's trying to strip newlines. If we want to be thorough then here's where I'd be worried:

1. How list/dict/tuple definitions manage newlines in between elements. It may be that these just get passed along to the visitor and the developer manage, which is ... a thing you can do. If that's not acceptable, then I would guess the whole grammar would have to consider adding newline stripping, and then the problem solved in this PR would crop up.
2. How well it handles `single_input`. This PR discussion gave me a strong impression that most people aren't crazy like me and just use `file_input`. Since I'm basically making my own interpreter, I needed to worry about both `file_input` and `single_input` as entry rules. If `file_input` doesn't work due to lack of testing, then that's a completely different thing.

KvanTTT:
Ok then, thanks for the contribution.

PR #1579, opened 2019-12-10
Remove empty set symbol from tinybasic

matozoid:
Looks like the author misunderstood the meaning of it :-)


teverett:
@matozoid thanks

PR #1584, opened 2019-12-12
TinyBasic: remove stray tokens

matozoid:
I'm pretty sure + and - aren't "relative operators". Doesn't make sense to have them there. They are not found in the probable grammar source either (https://en.wikipedia.org/wiki/Tiny_BASIC).


teverett:
@matozoid thanks

PR #1586, opened 2019-12-17
[ecmascript] Fix expressionStatement condition

renatahodovan:
Fix the condition of `expressionStatement` in ecmascript grammar which prevents parsing nested functions. The lookahead which looks for Function token should guide the decision between `functionDeclaration` and `statement`.

### DIFF
--- a/ecmascript/CSharp/ECMAScript.g4
+++ b/ecmascript/CSharp/ECMAScript.g4
@@ -178,7 +178,7 @@ sourceElements
 ///     Statement
 ///     FunctionDeclaration
 sourceElement
- : statement
+ : {this.InputStream.LA(1) != Function}? statement
  | functionDeclaration
  ;
 
@@ -202,7 +202,7 @@ statement
  : block
  | variableStatement
  | emptyStatement
- | expressionStatement
+ | {this.InputStream.LA(1) != OpenBrace}? expressionStatement
  | ifStatement
  | iterationStatement
  | continueStatement
@@ -263,7 +263,7 @@ emptyStatement
 /// ExpressionStatement :
 ///     [lookahead ∉ {{, function}] Expression ;
 expressionStatement
- : {(this.InputStream.LA(1) != OpenBrace) && (this.InputStream.LA(1) != Function)}? expressionSequence eos
+ : expressionSequence eos
  ;
 
 /// IfStatement :
--- a/ecmascript/CSharpSharwell/ECMAScript.g4
+++ b/ecmascript/CSharpSharwell/ECMAScript.g4
@@ -178,7 +178,7 @@ sourceElements
 ///     Statement
 ///     FunctionDeclaration
 sourceElement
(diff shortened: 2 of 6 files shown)


renatahodovan:
An example snippet that could not be parsed with the old version:

```js
if (true) {
  function f() {}
}
```

KvanTTT:
Does this fix applicable to JavaScript grammar?

studentmain:
> Is this fix applicable to JavaScript grammar?

No, JavaScript grammar already works with this

renatahodovan:
@KvanTTT The structure of JavaScript grammar looks different and it should work already.

studentmain:
Wait. Old version works for me.

图片

renatahodovan:
@studentmain Is your IDE able to work with inline predicates?

Trying to parse the example code with the old version gives me these errors:
```
line 2:2 no viable alternative at input 'function'
line 3:0 extraneous input '}' expecting <EOF>
```

Tester code in Python3:

```py
import sys

from antlr4 import *

from ECMAScriptLexer import ECMAScriptLexer
from ECMAScriptParser import ECMAScriptParser

tokens = CommonTokenStream(ECMAScriptLexer(FileStream(sys.argv[1])))
parser = ECMAScriptParser(tokens)
parser.program()
```

studentmain:
Problem seems not in grammar, again:
```
C:\src\grammars-v4\ecmascript>java -cp .;C:\Data\antlr\antlr-4.7.2-complete.jar org.antlr.v4.gui.TestRig ECMAScript program
if(true){
        function f(){}
}
^Z
line 2:8 no viable alternative at input 'function'
line 3:0 extraneous input '}' expecting <EOF>
```


If you're interested, here's my environment

OS: Windows 10.0.18362.535 (Locale: zh-CN)
ANTLR version: 4.7.2
Java version: 
```
java version "1.8.0_231"
Java(TM) SE Runtime Environment (build 1.8.0_231-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.231-b11, mixed mode)
```

Intellij IDEA version:
图片

ANTLR plugin version: 1.12

renatahodovan:
@studentmain According to your last comment, it seems to me, that we got exactly the same result with the old grammar (parse fail). I've also tried the example test with TestRig (with Java target) and I still believe that the proposed fix makes possible to parse the aforementioned test.

studentmain:
> we got exactly the same result

Yes, seems it works only in Intellij IDEA plugin, strange.

renatahodovan:
@studentmain My guess is that the plugin only "works" because it ignores inline code predicates and it was "lucky" with the precedence.

KvanTTT:
What's the final conclusion on this? :) Should we accept this PR?

renatahodovan:
I think that we agreed on that the the PR is needed.

studentmain:
Yes, we can merge it.

PR #1593, opened 2019-12-27
Fixing error in lexer NEWLINE action

braderman:
The code referenced identifier "nextnexteof", which doesn't exist.
Changed to nextnext_eof, which is likely what was intended.


teverett:
@braderman thanks

Issue #1165, opened 2018-06-24
Antlr4 grammar file can't parse "@parser::member"


peterremote1980:
Hi
   I am using https://github.com/antlr/grammars-v4/blob/master/antlr4/ANTLRv4Parser.g4 to parse the below g4 file and got an error

```
line 7:6 no viable alternative at input '['
Antlr4Parser::syntaxError
```

```antlr
grammar Assembler;

@parser::members{
	public int xxx;
}

WS		:	[ \t]+ -> skip;
NL		:	'\r'? '\n' -> skip;
LINE_COMMENT	:	';' ~[\r\n]*;

AAA		:	'aaa';
AAD		:	'aad';
AAM		:	'aam';
AAS		:	'aas';
ADC		:	'adc';
ADCX	:	'adcx';

IMM32	:	'0x'? [0-9a-fA-F][0-9a-fA-F][0-9a-fA-F][0-9a-fA-F][0-9a-fA-F][0-9a-fA-F][0-9a-fA-F][0-9a-fA-F];
IMM16	:	'0x'? [0-9a-fA-F][0-9a-fA-F][0-9a-fA-F][0-9a-fA-F];
IMM8	:	'0x'? [0-9a-fA-F][0-9a-fA-F]?;

//NUMBER	:	'0x'? [0-9]+;
MEM8	:	'[' IMM8 ']';
STRING	:	'"' ~["]* '"';

RAX		:	'rax';
EAX		:	'eax';
AX		:	'ax';
AH		:	'ah';
AL		:	'al';
RBX		:	'rbx';
EBX		:	'ebx';
BX		:	'bx';
BH		:	'bh';
BL		:	'bl';
RCX		:	'rcx';
ECX		:	'ecx';
CX		:	'cx';
CH		:	'ch';
CL		:	'cl';
RDX		:	'rdx';
EDX		:	'edx';
DX		:	'dx';
DH		:	'dh';
DL		:	'dl';
RBP		:	'rbp';
EBP		:	'ebp';
BP		:	'bp';
RSP		:	'rsp';
ESP		:	'esp';
SP		:	'sp';
RSI		:	'rsi';
ESI		:	'esi';
SI		:	'si';
RDI		:	'rdi';
EDI		:	'edi';
DI		:	'di';
R8		:	'r8';
R8D		:	'r8d';
R8W		:	'r8w';
R8L		:	'r8l';
R9		:	'r9';
R9D		:	'r9d';
R9W		:	'r9w';
[...]

PR #1595, opened 2019-12-28
Fixed the ANTLRv4 lexer to exit correctly from prequel constructs

bjansen:
This is a fix for #1165 (similar to [what was done in the IntelliJ plugin](https://github.com/antlr/intellij-plugin-v4/commit/ab388c97e972757b874d38748732cd2eecf05791)).


teverett:
@bjansen thanks

PR #1598, opened 2019-12-31
[TSQL] Allow for STRING_AGG function within group, and fix non keyword in Select

Szhlopp:
Allows for **STRING_AGG function within group**, and fix non keyword issue with simple_id.

I just completed using TSQL-ANTLR to parse names from a production database. These two issues were all I ran across. 

Sample for `STRING_AGG`:

```tsql
SELECT
	ETLTYPE,
	STRING_AGG(ETLNAME, ',') WITHIN GROUP (ORDER BY TABLENAME) AS NAMES
FROM CONFIG.ETL
GROUP BY
	ETLTYPE
```

Sample for `SPLIT` `simple_id`:

```tsql
SELECT SPLIT.A.VALUE('.', 'INT') AS SPLV 
FROM 
	(SELECT 
		CAST('<SPL>' + REPLACE('1,2,3', ',', '</SPL><SPL>') + '</SPL>' AS XML) AS DTA
	) AS A 
  CROSS APPLY 
	DTA.NODES ('/SPL') AS SPLIT(A)
```


KvanTTT:
Good! Could you also add your SQL samples to [`examples`](https://github.com/antlr/grammars-v4/tree/master/tsql/examples) directory?

tsql/TSqlParser.g4
```diff
@@ -3204,6 +3204,7 @@ function_call
     | aggregate_windowed_function                       #AGGREGATE_WINDOWED_FUNC
     | analytic_windowed_function                        #ANALYTIC_WINDOWED_FUNC
     | scalar_function_name '(' expression_list? ')'     #SCALAR_FUNCTION
+    | STRING_AGG '(' aggcol=expression ',' (expression | STRING) ')' (WITHIN GROUP '(' order_by_clause ')')?  #STRINGAGG
```
KvanTTT: Useless alternative: `expression | STRING`. Because of `expression` already includes `STRING` by such way: `expression -> primitive_expression -> constant -> STRING.

tsql/TSqlParser.g4
```diff
@@ -3992,6 +3993,7 @@ simple_id
     | SNAPSHOT
     | SOURCE
     | SPATIAL_WINDOW_MAX_CELLS
+    | SPLIT
```
KvanTTT: I think introduced `STRING_AGG` also should be added here as an alternative to `simple_id`.

Szhlopp: Didn't think this would work, but just tried it and it does. Will add

tsql/TSqlParser.g4
Szhlopp: Oops, let me remove and retest.

KvanTTT:
Thanks and happy new year! 👨‍💻🌲🎉

PR #1605, opened 2020-01-13
Add TRY_CONVERT function call

csrowell:
Add support for the `TRY_CONVERT` function.


KvanTTT:
Could you also please add a sample that covers such a case?

csrowell:
Yeah, I'd be happy to. When looking around at the examples I see there already is [an example of TRY_CONVERT](https://github.com/antlr/grammars-v4/blob/master/tsql/examples/dml_select.sql#L624).  I am using a copy of the grammar from a few years ago and had to add this line in the version in my project to get that function to be recognized, but perhaps it is being handled in some way I'm not recognizing in this newer version?

KvanTTT:
I think it was being handled by another alternative, likely `SCALAR_FUNCTION`.

PR #1612, opened 2020-01-19
Refactor antlr, asm, java, javascript, kotlin, logo, sql, python, swift, asn, rfc into modules

teverett:


pom.xml
```diff
@@ -228,9 +214,7 @@
 				<module>stellaris</module>
 				<module>stringtemplate</module>
 				<module>suokif</module>
-				<module>swift2</module>
-				<module>swift3</module>
-				<module>swift-fin</module>
```
KvanTTT: This is incorrect, swift-fin is a completely different language, compare https://en.wikipedia.org/wiki/Society_for_Worldwide_Interbank_Financial_Telecommunication and https://en.wikipedia.org/wiki/Swift_(programming_language). So, `siwft-fin` should be returned back.

python/pom.xml
```diff
@@ -0,0 +1,24 @@
...
+		<relativePath>../pom.xml</relativePath>
+    </parent>
+
+	<modules>
+		<module>python2</module>
+		<module>python3</module>
+		<module>python3-js</module>
+		<module>python3-py</module>
+		<module>python3-ts</module>
+		<module>python3_without_actions</module>
```
KvanTTT: Maybe rename to `python3-without-actions` to unify with other directories?

swift/pom.xml
```diff
@@ -0,0 +1,20 @@
...
+		<groupId>org.antlr.grammars</groupId>
+		<artifactId>grammarsv4</artifactId>
+		<version>1.0-SNAPSHOT</version>
+		<relativePath>../pom.xml</relativePath>
+    </parent>
+
+	<modules>
+		<module>swift2</module>
+		<module>swift3</module>
+		<module>swift-fin</module>
```
KvanTTT: Remove `swift-fin` it from here.

KvanTTT:
Also, I suggest considering the following merges:

```
ecmascript, javascript -> javascript
asn, asn_3gpp -> asn
kotlin, kotlin-formal -> kotlin
rfc822-datetime, rfc822-emailaddress -> rfc822
```

pom.xml
teverett: oops

teverett:
ok all submitted

teverett:
should verilog and system-verilog be a module?

KvanTTT:
> should verilog and system-verilog be a module?

Looks like this, they are similar: https://en.wikipedia.org/wiki/Verilog https://en.wikipedia.org/wiki/SystemVerilog SystemVerilog is the extension of Verilog.

KvanTTT:
Seems like everything is fine. Let's merge it!

PR #1619, opened 2020-01-22
Several TSQL fixes

tkvw:
`TRY_CONVERT` was introduced without Lexer `TOKEN`, but was already covered with `CONVERT` token in the lexer. And see commit messages for further info.

### DIFF
--- a/sql/tsql/TSqlParser.g4
+++ b/sql/tsql/TSqlParser.g4
@@ -2982,8 +2982,7 @@ for_clause
     ;
 
 xml_common_directives
-    : ',' (BINARY_BASE64 | TYPE)
-    | (COMMA ROOT ('(' STRING ')')?)
+      : ',' (BINARY_BASE64 | TYPE | ROOT ('(' STRING ')')?)
     ;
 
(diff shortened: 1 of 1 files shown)


sql/tsql/TSqlParser.g4
```diff
@@ -2982,7 +2982,8 @@ for_clause
     ;
 
 xml_common_directives
-    : ',' (BINARY_BASE64 | TYPE | ROOT)
+    : ',' (BINARY_BASE64 | TYPE)
+    | (COMMA ROOT ('(' STRING ')')?)
```
KvanTTT: Excess parentheses:

```sql
    | COMMA ROOT ('(' STRING ')')?
```

KvanTTT: More clear version:

```g4
xml_common_directives
    : ',' (BINARY_BASE64 | TYPE | ROOT ('(' STRING ')')?)
    ...
```

tkvw: Changed

Issue #1620, opened 2020-01-23
Scala Grammar - Issue with Concatenation Expression


igvls:
Hi @loonydev,

Grammar still fails to parse concatenation expressions. See the sample code below:

```scala
package xxx.yyy.zzz.aaa.bbbb.commons.transformation
import java.math.BigDecimal
import xxx.yyy.zzz.aaa.bbbb.commons.constants.Fields
import xxx.yyy.zzz.aaa.bbbb.commons.utils.Functions._
import xxx.yyy.zzz.aaa.bbbb.configuration.SparktacusTransformation
import xxx.yyy.zzz.aaa.bbbb.engine.ProcessStatus
import xxx.yyy.zzz.aaa.bbbb.transformation.TransformationTrait
import org.apache.spark.sql.{DataFrame, Row, functions}
import org.apache.spark.sql.functions._
import org.apache.spark.sql.types.{Decimal, DecimalType, StringType, StructField, StructType}
class CapComer(val transformation: SparktacusTransformation) extends TransformationTrait{
  def runTransformation: DataFrame = {
    val previousDF = ProcessStatus.components(transformation.previous.get(0))
    val outPutDataframe = capComer(previousDF)
    outPutDataframe
  }
  def capComer(previousDF: DataFrame): DataFrame = {
    val capComer = previousDF.
      withColumn(Fields.cap_comer, when(col(Fields.Trel).isNull, col(Fields.capacc)).
        otherwise(when(col(Fields.Trel).isNotNull && col(Fields.capacc) === "", lit("ECE")).
          otherwise(when(col(Fields.Trel).isNotNull && col(Fields.capacc) === "ECE", lit("ECE")).
            otherwise(lit("ECE,ACR")))))
    val rcct = capComer.select(
      Fields.Tper,
      Fields.Cper,
      Fields.cap_comer
    ).map(x => ((x.getString(0), x.getDecimal(1)), x.getString(2)))
      .reduceByKey((a, b) => a + b)
      .mapPartitions(row => row.map {
        case ((tipo_per: String, cod_pers: BigDecimal), cap: String)
      =>
          if(cap.contains("ECE") && cap.contains("ACR")){
            Row.fromSeq(Seq(tipo_per, cod_pers, "ECE,ACR"))
          } else if (cap.contains("ECE") && !cap.contains("ACR")) {
            Row.fromSeq(Seq(tipo_per, cod_pers, "ECE"))
[...]

PR #1622, opened 2020-01-27
Fix case parsing error Scala

loonydev:
Close https://github.com/antlr/grammars-v4/issues/1620

### DIFF
--- a/scala/Scala.g4
+++ b/scala/Scala.g4
@@ -50,14 +50,10 @@ ids
    : Id (',' Id)*
    ;
 
-path
-   : Id ('.' Id)*
-   | (Id '.')? 'this'
-   ;
-
 stableId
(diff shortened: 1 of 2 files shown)


scala/Scala.g4
```diff
@@ -50,14 +50,11 @@ ids
...
-   | (Id '.')? 'this'
-   ;
-
 stableId
-   : (path '.')? Id
-   | Id '.' 'super' classQualifier? '.' Id
+   : Id
+   | stableId '.' Id
+   | (Id '.')? 'this'
+   | (Id '.')? 'super' classQualifier? '.' Id
```
KvanTTT: Maybe `(Id '.')? ('this' | 'super' classQualifier? '.' Id)`

Issue #1623, opened 2020-01-27
Scala Grammar - Issue with direct value assignment


igvls:
Hi @loonydev,

Antlr fails to parse the following code. Kindly verify,

```scala
package uuu.vvv.www.xxx.yyy.zzz

object Fields {
    val a : String = "Apple"
    val b : String = "Ball"
}

Thanks!
```

loonydev:
Hi, 
Yep, checking, good findings.

igvls:
Hi @loonydev ,

I got a RecursionError. 
1. If usage is `hiveContext.sql("DROP DATABASE IF EXISTS demo_test_app CASCADE")` -**PARSE SUCCESS**
2. If usage is `hiveContext.sql("""DROP DATABASE IF EXISTS demo_test_app CASCADE""")` -**PARSE FAILED**

Antlr fails to parse the following code,
```scala
package uuu.vvv.www.xxx.yyy.zzzz
import java.math.BigDecimal
import org.apache.spark.sql.hive.HiveContext

object CreateTable extends CreateTableInterface {
    def execute (hiveContext: HiveContext) = {
        hiveContext.sql("""DROP DATABASE IF EXISTS demo_test_app CASCADE""")
    }
}
```

Kindly help. Thanks!

loonydev:
Thanks, fixed in PR

igvls:
Hi @loonydev & @studentmain,

I found that 
- use of tab inside docstring """**_< tab >_**""" - fails to parse.
```scala
object Test 
{
    // Main Method 
    def main(args: Array[String]) 
    {
        println("""Hello	World!""") 
    }
}
```
- use of comment _**/* < code > */**_ - fail to parse
```scala
/*object Test 
{
    // Main Method 
    def main(args: Array[String]) 
    {
        println("""Hello	World!""") 
    }
}*/
```
`line 8:3 mismatched input '<EOF>' expecting {'implicit', 'lazy', 'case', '@', 'override', 'abstract', 'final', 'sealed', 'private', 'protected', 'import', 'class', 'object', 'trait', 'package'}
compilationUnit`

loonydev:
Second case is invalid. You tried to parse empty file. Minimum it's have package construction. Parsing of `\* *\` in code works well.

First case fix soon.

igvls:
Understood!
Thanks @loonydev

PR #1624, opened 2020-01-27
Fix sep problem in Scala

loonydev:
Semi to ignore. Removed all places where Semi used, cause it makes no sense more.

Also, fixed problem with `postfixExpr`. Scala returns the last action in function. In some cases, the last action concatenates with up line.

I think NL can be ignored too. Added example. Close https://github.com/antlr/grammars-v4/issues/1623

### DIFF
--- a/scala/Scala.g4
+++ b/scala/Scala.g4
@@ -71,7 +71,7 @@ functionArgTypes
    ;
 
 existentialClause
-   : 'forSome' '{' existentialDcl (Sep? existentialDcl)* '}'
+   : 'forSome' '{' existentialDcl+ '}'
    ;
 
 existentialDcl
@@ -108,7 +108,7 @@ types
    ;
 
 refinement
-   : '{' refineStat (Sep? refineStat)* '}'
+   : NL? '{' refineStat+ '}'
    ;
 
 refineStat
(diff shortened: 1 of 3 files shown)


scala/Scala.g4
```diff
@@ -592,6 +595,15 @@ Delim
    : '`' | '\'' | '"' | '.' | ';' | ','
    ;
 
+Semi
```
studentmain: Maybe name it as `Sep`? `Semi` stand for semicolon.

loonydev: Yep, logical, but in [spec](https://scala-lang.org/files/archive/spec/2.13/13-syntax-summary.html) used this definition for `Semi`
```
semi  ::=  ‘;’ |  nl {nl}
```

If you want, I can change.

studentmain: > Yep, logical, but in [spec](https://scala-lang.org/files/archive/spec/2.13/13-syntax-summary.html) used this definition for `Semi`
> 
> ```
> semi  ::=  ‘;’ |  nl {nl}
> ```
> 
> If you want, I can change.

Let's follow the spec.

loonydev:
Upd: Also fix problem with multiline string.

PR #1628, opened 2020-01-31
RCS grammar fixes

isotes:
- BRANCH was erroneously the literal 'BRANCH' instead of a revision number
- STRING parsing was completely broken as relevant code was commented out
- REVISION was broken as it required additional pairs of version numbers

Fixes are based on the upstream RCS grammar
https://www.gnu.org/software/rcs/manual/html_node/comma_002dv-grammar.html

### DIFF
--- a/rcs/RCS.g4
+++ b/rcs/RCS.g4
@@ -81,7 +81,7 @@ head
    ;
 
 branch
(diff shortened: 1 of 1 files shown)


teverett:
@isotes thanks

Issue #1626, opened 2020-01-30
NUMBER conflicts DIGIT in smiles grammar


luvf:
smiles grammar have a conflict with DIGIT and NUMBER = DIGIT+ 
when parsing
C12C1C2 (toy example)
12 is read as a NUMBER, and parsing ends when reading 12
C12 have to be parsed as 
branched_atom(C, ringbond( void ,1),ringbond( void, 2))

fix proposal : inline NUMBER by replacing it by DIGIT+ in isotope rule l.75, class_ rule l.290, and remove it l.352-353.

as i am not an expert in grammar rules there is probably a cleaner solution.

teverett:
could you submit a PR?

teverett:
Is this better?

https://github.com/teverett/grammars-v4/tree/smiles

Here's the output for your example

Parsing :/Users/tom/projects/grammars-v4/smiles/examples/C12C1C2.txt
(smiles (chain (chain (chain (branched_atom (atom (aliphatic_organic C)) (ringbond (number 12)))) (branched_atom (atom (aliphatic_organic C)) (ringbond (number 1)))) (branched_atom (atom (aliphatic_organic C)) (ringbond (number 2)))) (terminator \n))

luvf:
PR send

no this wouldn't work because in C12C1C2, 12 have to be read as 1 and 2 an not as  "12" : twelve. It should parse this : 
![parseTree](https://user-images.githubusercontent.com/20399455/73652689-bdf99080-4687-11ea-8748-5e8803aa725e.png)

and not this :
![parseTreeWrong](https://user-images.githubusercontent.com/20399455/73652612-96a2c380-4687-11ea-9f83-9a9019bdddb1.png)


reminder : in this case numbers describes bonds between pairs of atoms with the same number. you can use

PR #1632, opened 2020-02-03
conflict between NUMBER and DIGIT in ringbonds corrected by removing …

luvf:
…NUMBER (and inlining it)
correcting:
https://github.com/antlr/grammars-v4/issues/1626


teverett:
@luvf thanks

Issue #1591, opened 2019-12-25
Async is not a reserved keyword


phamvinh1712:
According to ESMA
https://tc39.es/ecma262/#sec-future-reserved-words
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Keywords
async is not a reserved or future reserved keyword. 
Why is it implemented as a reserved keyword in the Javascript grammar?
At the moment I cannot parse code that has async as a variable name

studentmain:
Yes, it's an known issue

studentmain:
#1555

phamvinh1712:
Hi @studentmain , do you have any tip to quickly handle this? Or should I just remove everything relating to async as a temporary fix?
Thank you

studentmain:
[js-esnext.zip](https://github.com/antlr/grammars-v4/files/4001126/js-esnext.zip)

Here's an alternate version maybe fixed all problems in that issue. If this one don't work, use your temporary fix.

studentmain:
@phamvinh1712 Does it work?

phamvinh1712:
@studentmain sorry I forgot to comment back. I only take the part you handle the keyword in the alternate version and it works great.
Thank you very much

studentmain:
In case someone interested, that version is from https://github.com/studentmain/grammars-v4/tree/js-esnext . It changed too much, I'm not sure open PR (and merge it) is a good idea.

KvanTTT:
> It changed too much, I'm not sure open PR (and merge it) is a good idea.

But why? It is backward compatible? If so, I don't see a reason not to accept such changes.

studentmain:
> It is backward compatible?

I think no, I modified it to adopt ESTree node naming. (https://github.com/studentmain/grammars-v4/commit/b6a112dbca1db83c1d103a2ddaacd83d218fe68c)

PR #1633, opened 2020-02-04
Fix #1138: Parse lets in non strict mode

### LINKED ISSUE #1138: [JavaScript] Let declarations are not available outside of strict mode
The lexer does not recognize `Let` tokens in non-strict mode. The purpose seems to be to recognize let as Identifier instead, so that they can be used in rules such as `var let = 1;`, which is only valid in non-strict mode.

But this has the unintended side effect of not allowing let declarations at all in non-strict mode! This simplest fix seems to be to make the lexer emit `StrictLet` and `NonStrictLet` tokens, and turn all appropriate `Identifier` references in the parser into references to a new `identifier` parser rule, which accepts `Identifier` and `NonStrictLet`. Then, the `keyword` and `varModifier` rules can accept both `NonStrictLet` and `StrictLet`.

jsyeo:
I tested on a few snippets of Javascript code that includes `let foo = 123;` and `var let = 123;` and `"use strict"; var let = 123;` and it all seems to work as expected. Let me know if more work is required. 😄

### DIFF
--- a/javascript/javascript/JavaScriptLexer.g4
+++ b/javascript/javascript/JavaScriptLexer.g4
@@ -170,7 +170,8 @@ Await:                          'await';
 /// when parsing strict mode
 
 Implements:                     'implements' {this.IsStrictMode()}?;
-Let:                            'let' {this.IsStrictMode()}?;
+StrictLet:                      'let' {this.IsStrictMode()}?;
+NonStrictLet:                   'let' {!this.IsStrictMode()}?;
 Private:                        'private' {this.IsStrictMode()}?;
 Public:                         'public' {this.IsStrictMode()}?;
 Interface:                      'interface' {this.IsStrictMode()}?;
--- a/javascript/javascript/JavaScriptParser.g4
+++ b/javascript/javascript/JavaScriptParser.g4
@@ -155,8 +155,7 @@ iterationStatement
 
 varModifier  // let, const - ECMAScript 6
     : Var
-    | StrictLet
-    | NonStrictLet
+    | let
     | Const
     ;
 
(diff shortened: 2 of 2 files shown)


javascript/javascript/JavaScriptLexer.g4
```diff
@@ -170,7 +170,8 @@ Await:                          'await';
 /// when parsing strict mode
 
 Implements:                     'implements' {this.IsStrictMode()}?;
-Let:                            'let' {this.IsStrictMode()}?;
+StrictLet:                      'let' {this.IsStrictMode()}?;
+NonStrictLet:                   'let' {!this.IsStrictMode()}?;
```
loonydev: Hi, maybe it's better to use `Let` rule which will be like `Let : StrictLet | NonStrictLet`, WDYT?

jsyeo: Oh right, makes sense!

loonydev: Nice, LGTM

studentmain: #1591 has simillar problem

KvanTTT: @loonydev I don't think it's possible. Firstly, it can be simplified by jus `Let: "let'`. Secondly, unified `Let` token substitutes `StrictLet` and `NonStrictLet`, and it becomes impossible to distinguish between `StringLet` and `NonStrictLet` in parser (`NonStrictLet` used in `identifier` rule).

loonydev: In case where we have both of them, it's better to have one Let. But you right about simplification.

KvanTTT:
@jsyeo what about samples? It shouldn't be complicated.

teverett:
@jsyeo there seem to be some merge conflicts

KvanTTT:
I'm closing it in favor of #1669

Issue #1636, opened 2020-02-06
JavaScript throws NoViableAltExcpetion parsing `for in` and `for of` statements


spencerxiao:
The JavaScript parser outputs error when parsing the following code snippet which is the ES6 syntax. 

```javascript
var fibonacci = [1, 3]
for (var n of fibonacci) {
    console.log(n)
}
```
Use the following Junit test which fails for the JavaScript parser because the `ctx.exception` is not null, although the parser can recover from the error, it is nice not to throw the exception at all. 

```java
import org.antlr.v4.runtime.CharStreams;
import org.antlr.v4.runtime.CommonTokenStream;
import org.antlr.v4.runtime.ParserRuleContext;
import org.junit.Test;

import static org.junit.Assert.assertNull;

public class Main {
    @Test
    public void testForInAndForOfStatements(String[] args) {
        String programStr = "var fibonacci = [1, 3]\n"+
                "for (var n of fibonacci) {\n" +
                "    if (n > 1000)\n" +
                "        break\n" +
                "    console.log(n)\n" +
                "}";
        JavaScriptLexer lexer = new JavaScriptLexer(CharStreams.fromString(programStr));
        lexer.setUseStrictDefault(true);
        CommonTokenStream tokenStream = new CommonTokenStream(lexer);
        JavaScriptParser parser = new JavaScriptParser(tokenStream);
        ParserRuleContext program = parser.program();

        new JavaScriptParserBaseVisitor<Void>() {
            @Override
            public Void visitVariableDeclaration(JavaScriptParser.VariableDeclarationContext ctx) {
                super.visitVariableDeclaration(ctx);
                assertNull(ctx.exception);
                return null;
            }
        }.visit(program);
    }
}
```
This is because the original definition of for statements use `variableStatement` rule which requires a `eos` rule.  It can be fixed with the following patch by using `variableDeclarationList` rule and modify the definition of `variableDeclarationList` and `variableStatement` such that the variable definition in for statements does not look for an `eos`.
```diff
--- a/javascript/javascript/JavaScriptParser.g4
+++ b/javascript/javascript/JavaScriptParser.g4
@@ -120,11 +120,11 @@ declaration
     ;
 
 variableStatement
-    : varModifier variableDeclarationList eos
+    : variableDeclarationList eos
     ;
 
 variableDeclarationList
-    : variableDeclaration (',' variableDeclaration)*
+    : varModifier variableDeclaration (',' variableDeclaration)*
     ;
 
[...]

PR #1637, opened 2020-02-06
Recognise for in and for of iteration statements without NoViableAltException

spencerxiao:
This PR tries to fix the issue #1636, let me know your thoughts

### DIFF
--- a/javascript/javascript/JavaScriptParser.g4
+++ b/javascript/javascript/JavaScriptParser.g4
@@ -120,11 +120,11 @@ declaration
     ;
 
 variableStatement
-    : varModifier variableDeclarationList eos
+    : variableDeclarationList eos
     ;
 
 variableDeclarationList
-    : variableDeclaration (',' variableDeclaration)*
+    : varModifier variableDeclaration (',' variableDeclaration)*
     ;
 
 variableDeclaration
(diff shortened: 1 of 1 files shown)


KvanTTT:
Could you please add a code sample that was parsed incorrectly before your changes and is parsed correctly after?

spencerxiao:
The parser although can recognise the following code snippet but it also outputs error message saying that  `line 2:11 no viable alternative at input 'of'. `
```javascript
var fibonacci = [1, 3];
for (var n of fibonacci) {
  console.log(n);
}
```

You can verify with the following unit test. Note that the `ctx.exception` is not null when it parses the for statement and thus the parser outputs the above error message. With this patch, the test case passes. 

```java
    @Test
    public void testForInAndForOfStatements() {
        String programStr = "var fibonacci = [1, 3];\n"+
                "for (var n of fibonacci) {\n" +
                "    console.log(n);\n" +
                "}";
        JavaScriptLexer lexer = new JavaScriptLexer(CharStreams.fromString(programStr));
        lexer.setUseStrictDefault(true);
        CommonTokenStream tokenStream = new CommonTokenStream(lexer);
        JavaScriptParser parser = new JavaScriptParser(tokenStream);
        ParserRuleContext program = parser.program();

        new JavaScriptParserBaseVisitor<Void>() {
            @Override
            public Void visitVariableDeclaration(JavaScriptParser.VariableDeclarationContext ctx) {
                super.visitVariableDeclaration(ctx);
                assertNull(ctx.exception);
                return null;
            }
        }.visit(program);
    }
```

KvanTTT:
Do we have the same sample in Iteratos.js file? https://github.com/antlr/grammars-v4/blob/master/javascript/javascript/examples/Iterators.js#L21-L25

spencerxiao:
Yes, we already have the example, actually the above code snippet is an excerpt from that file.

KvanTTT:
Ok then. Thanks.

PR #1650, opened 2020-02-25
tsql grammer - IDs and Null Columns

xDaeDaLuSx015:
Moved some of the keywords that can also be used as IDs and modified the `column_elem` rule to include the `NULL` keyword as `select null as column_name` is valid.


KvanTTT:
Any chance to add a code sample to [examples](https://github.com/antlr/grammars-v4/tree/master/sql/tsql/examples) directory?

xDaeDaLuSx015:
I don't mind adding one but I'm not sure what i need to add? Would it just
be the sql for the column_elem I've modified, like the example I gave in
the description or should it be sql that covers the keyword changes to be
ids as well?

KvanTTT:
> Would it just be the sql for the column_elem I've modified, like the example I gave in the description

I think it's okay and enough.

xDaeDaLuSx015:
I've added a column_elem.sql file that i think covers the rule i've changed with examples.

KvanTTT:
Thanks.

PR #1653, opened 2020-03-06
JSON.g4 parser generation leads to symbol conflict (array)

prasad83:
antlr4 -Dlanguage=PHP JSON.g4

error(134): JSON.g4:20:0: symbol array conflicts with generated code in target language or runtime

Replacing (array) to (arr) in JSON.g4 worked


teverett:
@prasad83 thanks

PR #1669, opened 2020-03-25
JavaScript: fix let and async tokens processing

KvanTTT:

PR #1672, opened 2020-03-26
Adding Support for React JS Components HTML Tags

### LINKED ISSUE #1671: Support for React JS Components HTML Tags
Hi Guys, any plans to supporting HTML Tags for JavaScript Grammar.
A common convention in ReactJs

See the following Snippet
https://github.com/project-bobon/bobonroastprofile/blob/master/src/components/utils/Button.jsx

asadcr:
Closes #1671

Repo Example for testing: https://github.com/project-bobon/bobonroastprofile/


asadcr:
@KvanTTT Some of the PR changes are pulled from #1669. Will merge into once #1669 is merged

studentmain:
Please create a new folder (`jsx` or `javascript/jsx`) for it, just like what TypeScript did. It's not original ECMAScript.

If you like, you can also add TypeScript based TSX grammar.

asadcr:
@studentmain Good Point. I would like to propose to merged this into JavaScript grammar, because jsx is an extension of JS. Creating 2 grammars will cause confusion since these will be 90% duplicated and future maintenance needs to be done in both grammars.

KvanTTT:
Is jsx an experimental feature? How much it is used?

studentmain:
@KvanTTT It's stable, but not official. It's an extension developed by Facebook for React.js framework. Now it's supported by many front-end framework.

@asadcr Problem is there's many extension for JS, how if someone want develop an extension which isn't compatible with JSX but compatible with original JS.

asadcr:
@studentmain Makes Sense. i will move it to `javascript/jsx` shortly

asadcr:
@studentmain @KvanTTT Could you guys review the PR so we can merge it ?

javascript/jsx/JavaScriptParser.g4
```diff
@@ -0,0 +1,557 @@
...
+    | htmlElements                                                          # htmlElementExpression
+    | '(' expressionSequence ')'                                            # ParenthesizedExpression
+    ;
+
+htmlElements
+    : htmlElement+
+    ;
+
+htmlElement
+    : '<' htmlTagName htmlAttribute* '>' htmlContent '<''/' htmlTagName '>'
```
studentmain: How if 
```jsx
var a = <h1>title</p>
```

javascript/jsx/JavaScriptLexer.g4
```diff
@@ -0,0 +1,815 @@
...
+CDataComment:                   '<![CDATA[' .*? ']]>' -> channel(HIDDEN);
+UnexpectedCharacter:            . -> channel(ERROR);
+CDATA:                          '<![CDATA[' .*? ']]>' -> channel(HIDDEN);
+
+//
+// html tag declarations
+//
+mode TAG;
+
+TAG_OPEN
```
studentmain: Following other part's naming convention is recommended.

javascript/jsx/JavaScriptParser.g4
asadcr: Good Point. Not Sure how to handle that. I certainly need help with that.

javascript/jsx/JavaScriptLexer.g4
asadcr: fixed.

asadcr:
Added code samples

javascript/jsx/JavaScriptParser.g4
studentmain: I think it's somehow tolerable. XML grammar seems has same problem. It can check by other part of compiler.

My solution is push tag name into stack at start(use action), and check it at the end of tag. Not sure it's a good idea, maybe it will cause some performance impact.

asadcr: https://github.com/antlr/grammars-v4/pull/1672/commits/2d7e637155e29df3dd1fe77784de781e9933ce77 Fixes this. the action methods should be added for other platforms.

asadcr: I think performance overhead is negligible.

teverett:
@asadcr  thanks!

PR #1674, opened 2020-03-26
Accept insert with default values

daikoz:
Fix to use:

```sql
INSERT INTO tablename() VALUES(), ();
```

where tablename have primary key with AUTO_INCREMENT and all columns are a default value


KvanTTT:
Could you also add a code sample?

daikoz:
Hi,

Follow the sample SQL code, it is a valid SQL request.

```sql
INSERT INTO tablename() VALUES(), ();
```

KvanTTT:
I mean to add this to examples directory.

Issue #1246, opened 2018-09-28
javascript grammer dont work with python


0x410c:
self explanatory question.

suggest what i can do to fix it?

zhunki:
I have the same problem.
generated python3 lexer doesn't work.

Anoyomouse:
Just found this

Unfortunately it's not a simple task of a 1-1 conversion for the base classes, however it's almsot that simple

YOu also need to change the references in the grammars to point to self instead of this

I've pushed a working example of this here: https://github.com/Anoyomouse/Antlr4-Grammar-JavaScript-Py

Because of the nature of the changes i don't think it's a simple case of making a PR and pushing here

KvanTTT:
@Anoyomouse could you please suggest a pull request with Python files JavaScriptBaseLexer.py, JavaScriptBaseParer.py? I think it's enough.

Anoyomouse:
@KvanTTT I've updated my parser (same repo) and unfortunately it's still not a trivial matter, the translation from Java/C# is not directly compatible with python because of how the rules in the lexer work (the lexer uses this. instead of self.), that being said, it's actually simple enough to transfrom the g4 before running Antlr, and that gives you a working parser (actually tested properly this time)

However, i've done a PR (#1675), but i'm almost totally sure it'll get killed off because of incompatibilities with all the other languages

PR #1675, opened 2020-03-27
Added python scripts for JS

Anoyomouse:
- Base classes for Parser / Lexer
- Transformer script for grammar

As discussed in #1246, this is a PR for the python base class Lexer and Parser, but also needs a transformation script to make the java/c# in the Lexer and Parser compatible with python.

As noted in the Antlr4 python docs, you cannot make the code compatible with python without having to change every other language to suit (self insead of this), so i thought a transformation was better in this case.

However, i see the problem in this PR, and am totally expecting it to be closed.
I have a repo of transformed grammars here: https://github.com/Anoyomouse/Antlr4-Grammar-JavaScript-Py

### DIFF
--- a/javascript/javascript/Python/JavaScriptBaseLexer.py
+++ b/javascript/javascript/Python/JavaScriptBaseLexer.py
@@ -1,0 +1,101 @@
+
+from antlr4 import *
+
+relativeImport = False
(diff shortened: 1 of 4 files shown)


teverett:
@Anoyomouse  thanks

PR #1678, opened 2020-04-01
Fixing Typescript Grammar Bugs From Real CodeBase

### LINKED ISSUE #1683: Typescript parser fails for function with rest argument.
The current version of the [Typescript antlr4 parser](https://github.com/antlr/grammars-v4/tree/master/typescript) fails for the following input:

```typescript
function Greet(greeting: string, ...names: string[]) {
    return greeting + " " + names.join(", ") + "!";
}
```
(from [here](https://github.com/antlr/grammars-v4/blob/master/typescript/examples/Function.ts#L34)).

Error message is:
```
line 2:0 rule expressionStatement failed predicate: {this.notOpenBraceAndNotFunction()}?

line 2:41 no viable alternative at input '(greeting: string, ...names:'
line 2:33 extraneous input '...' expecting {'[', '{', Identifier}
line 2:50 mismatched input ')' expecting {<EOF>, RegularExpressionLiteral, '[', '(', '{', ';', ',', '=', '++', '--', '+', '-', '~', '!', '<', '&', '|', 'null', BooleanLiteral, DecimalLiteral, HexIntegerLiteral, OctalIntegerLiteral, OctalIntegerLiteral2, BinaryIntegerLiteral, 'break', 'do', 'instanceof', 'typeof', 'case', 'else', 'new', 'var', 'catch', 'finally', 'return', 'void', 'continue', 'for', 'switch', 'while', 'debugger', 'function', 'this', 'with', 'default', 'if', 'throw', 'delete', 'in', 'try', 'from', 'readonly', 'async', 'class', 'enum', 'extends', 'super', 'const', 'export', 'import', 'implements', 'let', 'private', 'public', 'interface', 'package', 'protected', 'static', 'yield', 'type', 'namespace', 'abstract', Identifier, StringLiteral, TemplateStringLiteral}
```

It can be parsed however by removing either `...` before `names`, or, `:string[]` behind `names`. Sadly speaking, these constructs are quite often used out there.

asadcr:
* Adding C# Target
* Fixes #1683


KvanTTT:
@gabriel-fallen please also take a look.

max-leuthaeuser:
Could you also port this to [Java](https://github.com/antlr/grammars-v4/tree/master/typescript/Java)?

asadcr:
@max-leuthaeuser I couldn't get the `port this to Java?`. All you need to do is run `java -jar antlr-4.8.jar -Dlanguage=Java -no-listener TypeScriptParser.g4 TypeScriptLexer.g4` and grammar for Java platform is ready.

max-leuthaeuser:
I was referring to the TypescriptParserBase and TypescriptLexerBase classes. So they stay the same?

max-leuthaeuser:
Ok, I tried it with your PR branch. Sadly, this specific function with a rest parameter still fails.

asadcr:
> Ok, I tried it with your PR branch. Sadly, this specific function with a rest parameter still fails.

Please provide the test code

asadcr:
@max-leuthaeuser  https://github.com/antlr/grammars-v4/pull/1678/commits/eeb67ef6096996a4b297f55bf44ef6d09561b3bd fixes your #1683 issue

PR #1657, opened 2020-03-12
Add Haskell grammar

sloboegen98:
Fixes #210 issue
This grammar covers all standard Haskell 2010. Now I am implementing the support of ghc syntastic extensions. For example, MultiWayIf, FlexibleInstances and MagicHash are already supported.

PR #1681, opened 2020-04-02
Fixing issue about haskell grammar

sloboegen98:
issues from this pull request: https://github.com/antlr/grammars-v4/pull/1657


haskell/Java/HaskellBaseLexer.java
```diff
@@ -0,0 +1,367 @@
...
+    @Override
+    public Token nextToken() {
+        if (!tokenQueue.isEmpty()) {
+            return tokenQueue.poll();
+        }
+
+        Token next = super.nextToken();
+
+        if (startIndent == -1
+            && next.getType() != HaskellLexer.NEWLINE
```
KvanTTT, line 186: Maybe it makes sense to calculate this only once and use it in other places? `type = next.getType()`

sloboegen98: Fixed

KvanTTT:
Thanks!

PR #1688, opened 2020-04-12
Fix issue in GraphQL grammar limiting the enum value definition only to only one

cybergrouch:
The current grammar will only allow one enum value definition. Thus, it's failing when parsing enums with multiple enum value definitions like:

```
enum TransportType {
   BICYCLE
   WALK
   CAR
}
```

This is the link to the GraphQL spec particular to `EnumValuesDefinition`: https://spec.graphql.org/draft/#sec-Enums


teverett:
@cybergrouch  thanks

PR #1696, opened 2020-04-19
IDL.g4: Fix omission of OCTAL_LITERAL in rule "literal"; minor cosmetics

okellogg:
* According to the [IDL spec](https://www.omg.org/spec/IDL/4.2/), `OCTAL_LITERAL` should be included in rule `literal`.
* The rule for comma is named COMA. I rename it to `COMMA`.
* I insert line breaks between alternatives in the these rules:
  FLOATING_PT_LITERAL
  BOOLEAN_LITERAL
  fragment ESCAPE_SEQUENCE
  fragment OCTAL_ESCAPE
  fragment LETTER
  fragment ID_DIGIT


teverett:
@okellogg thanks

PR #1712, opened 2020-05-08
fixes Check-sat-assuming and refine the commands

i-walker:
Running 
```smt
(check-sat-assuming (AB))
```
results in an error, though the smt spec allows such definitions also with not. 
Please refer to page [99](http://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2017-07-18.pdf).

I also refined the command definitions for a nicer traversal over the AST, as the former approach would need the information of the parent tree `command` to define a valid ADT for the subtree, for example for to complete `get-info` the parent tree needs to pass the `info-flag` . 

I am also happy to contibute a github actions approach to run the tests instead of using travis.


teverett:
@i-walker  thanks

PR #1720, opened 2020-05-11
fix code smells

XenoAmess:
also, suggest replace all \t to spaces.
or replace all four-spaces-block to tabs(not suggested).
A mixture is trouble causer.


teverett:
@XenoAmess thanks

PR #1729, opened 2020-05-25
Haskell Grammar: add extensions and fix some bugs

sloboegen98:
Grammar was very changed, because when I was implementing a standard grammar, I relied on 'Haskell Language Report 2010', and [GHC official repo](https://gitlab.haskell.org/ghc/ghc/-/blob/master/compiler/GHC/Parser.y) was taken as a basis for extensions. But this grammars are very different about standard Haskell grammar implementation.

This version pass 98/122 tests from [GHC parser tests](https://gitlab.haskell.org/ghc/ghc/-/tree/master/testsuite%2Ftests%2Fparser%2Fshould_compile). 

Errors are similar: the lexer stops to process the token when it is not needed.
For example, I have lexer rule:
```
 AOpenBracket: '(|' 
```
Haskell allows to create your own operators:
```haskell
 (|:) :: Int -> Int -> Int 
```

So parser gets token stream like `AOpenBracket Colon CloseRoundBracket ...` - not what parser needs.

Could you tell, please, what to do in this situation?

### DIFF
--- a/haskell/HaskellLexer.g4
+++ b/haskell/HaskellLexer.g4
@@ -42,6 +42,155 @@ WS : [\u0020\u00a0\u1680\u2000\u200a\u202f\u205f\u3000]+ {
     this.processWSToken();
 } ;
 
+AS            : 'as'               ;
+CASE          : 'case'             ;
+CLASS         : 'class'            ;
+DATA          : 'data'             ;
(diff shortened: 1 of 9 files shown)


teverett:
@sloboegen98  thanks

PR #1748, opened 2020-06-11
IDL.g4 replace literal '*' with lexer rule

mspiegel:
The lexer ruler STAR exists in the grammar and was previous unused.


KvanTTT:
Actually it is used because the generator is able to match string literals with token names in simple cases. But your change is good because the grammar is written in a style where token names are used instead of string literals.

PR #1755, opened 2020-06-20
Fix bug with empty comment

d-markey:
Fix: Empty comments "/**/" parse as a delimited doc comments


teverett:
@d-markey thanks.  If you could add some examples that test this change it would be great.

PR #1764, opened 2020-06-24
New Python3 Grammar for TS

sonny3690:
I rewrote the Python3 TS Grammar to reflect the changes made to the Python3 Java grammar file. 

I was frustrated by how general the rules were and how they couldn't properly lex through a lot of subtle python. In addition to this, the major differences between the TS generated parse tree structure and the one `grun` shows, which you visualize through the Java Parser, only added to the frustration. 

This change should all fix it and work with all the tests.

**Note**: the Python3-JS grammar doesn't work out of box. Probably should open an issue for this.

### DIFF
--- a/python/python3-ts/Python3.g4
+++ b/python/python3-ts/Python3.g4
@@ -1,39 +1,35 @@
 /*
  * The MIT License (MIT)
- *
+ * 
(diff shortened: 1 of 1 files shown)


teverett:
@sonny3690 Thanks!

Issue #1381, opened 2019-03-25
CPP14 grammar:parse error when writing a string in multiple lines


jumormt:
Sometimes because the length of a string is too long, we would like to write the code in multiple lines for tidy.
e.g.

```cpp
string str1 = "abcabc"
             "abcabc";
```

Marti2203:
Hi,
I see this can be closed @KvanTTT .

Issue #1577, opened 2019-12-04
Problems with CPP14.g4


dmedme:
1. The Lexer does not recognise multi-line strings. I have a fix for this if anyone is interested.

2. A C++ source file with just the following contents
/*
 * A comment
 */
static const char * whatever = "A multi-line string\n\
A bit more of it\n";

gives "no viable alternative at input 'staticconstchar'"

Grun displays a tree with ALL of the input (including that beyond where it has given up) classified as I would expect, so it isn't clear to me why it has choked.

KvanTTT:
> 1. The Lexer does not recognize multi-line strings. I have a fix for this if anyone is interested.

Yes, we are interested in it very much :) You are welcome to make a pull request.

> gives "no viable alternative at input 'staticconstchar'"

Maybe support of multiline strings with backslashes are not very good and correct?

dmedme:
Yes, it was the backslash handling. I will endeavour to submit a 'pull' request.

dmedme:
Fallen at the first hurdle. I am not permitted to create the branch Issue-#1577-CPP14.g4 as a precursor to creating the pull request.

KvanTTT:
You can do whatever you want, but at first, you have to create a [fork](https://help.github.com/en/github/getting-started-with-github/fork-a-repo) of this repo.

PR #1788, opened 2020-07-14
Fixed-> CPP14 grammar:parse error when writing a string in multiple l…

rafeeque1:
…ines #1381

Fixed ,Now Its Parsed .


cpp/CPP14.g4
```diff
@@ -32,7 +32,7 @@ translationunit
 
 
 primaryexpression
-   : literal
+   : literal (literal)*
```
KvanTTT, line 35: Please remove redundant parentheses and use `+` operator:

```suggestion
   : literal+
```

cpp/examples/test.txt
```diff
@@ -0,0 +1,18 @@
+#include <iostream>
```
KvanTTT, line 1: Use `.cpp` extension for example files:

```
cpp/examples/test.cpp
```

cpp/CPP14.g4
rafeeque1: Thank you KvanTTT

rafeeque1: Problems with CPP14.g4 #1577  duplicate issues.

KvanTTT: What do you mean? I guess `literal (literal)*` and `literal+` are equivalent forms.

teverett:
@rafeeque1 thanks

PR #1792, opened 2020-07-16
lua: local variable attribute

studentmain:
new syntax added in lua 5.4

`local a <const> = 1`

### DIFF
--- a/lua/Lua.g4
+++ b/lua/Lua.g4
@@ -43,7 +43,7 @@ This grammar file derived from:
 
 Tested by Kazunori Sakamoto with Test suite for Lua 5.2 (http://www.lua.org/tests/5.2/)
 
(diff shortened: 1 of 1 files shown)


teverett:
@studentmain  could you add an example please?

lua/Lua.g4
```diff
@@ -71,7 +71,15 @@ stat
...
-    | 'local' namelist ('=' explist)?
+    | 'local' attnamelist ('=' explist)?
+    ;
+
+attnamelist
+    : NAME attrib (',' NAME attrib)*
+    ;
+
+attrib
+    : ('<' NAME '>')?
```
KvanTTT, line 82: Maybe it's better to extract optional operator to uppermost rules?

studentmain: this is following the official manual.

PR #1794, opened 2020-07-17
Add crossdb data types

jpechane:
Support for alias datatypes as defined in https://dev.mysql.com/doc/refman/8.0/en/other-vendor-data-types.html


KvanTTT:
Thanks again :)

sql/mysql/Positive-Technologies/MySqlLexer.g4
```diff
@@ -245,6 +253,7 @@ VARBINARY:                           'VARBINARY';
 TINYBLOB:                            'TINYBLOB';
 BLOB:                                'BLOB';
 MEDIUMBLOB:                          'MEDIUMBLOB';
+LONG:                                'LONG';
```
rk3rn3r, line 256: @KvanTTT this LONG is used for LONG VARCHAR and LONG VARBINARY. maybe it's better to give the token a better fitting name and split between the 2 usages?

KvanTTT: I don't exactly understand. What are you suggesting? To introduce tokens `LONG VARCHAR` and `LONG VARBINARY` instead of single `LONG`?

rk3rn3r: Exactly. From a database view VARCHAR is linked to encoding and VARBINARY is arbitrary bytes. I was suggesting that the ANTLR grammar should maybe separate this as it might have semantical differences, though technically (like transport level) might be the same.

jpechane: I am not sure about making it a combined a combine token but I agree with the `VARBINARY` problem. I'll prepare a separate PR to handle it differently.

jpechane: @rk3rn3r  Please see followup - https://github.com/antlr/grammars-v4/pull/1796

PR #1797, opened 2020-07-20
Fixed detection of Decimal.

dk2k:
'0' is legal, but '00', '000' and so on - aren't.
Still '00' and others will be picked up by the rule Octal:
Octal:              '0' [0-7]+;


teverett:
@dk2k  thanks!

Issue #586, opened 2017-03-31
Issue with parsing exponents in Php grammar


prateekgulati:
E(exponential) symbol is not parsed by the antlr using Php grammar
eg `$float = 7E-10;` gives an error node in multiplicationExpression

KvanTTT:
It's case insensitive issue. Please read [doc](https://github.com/antlr/antlr4/blob/master/doc/case-insensitive-lexing.md) and use `CaseChangingCharStream` for this grammar.

PR #1798, opened 2020-07-20
'E' letter can be both case in exponent

dk2k:
Please see the demo sample https://sandbox.onlinephpfunctions.com/code/6e977d34696f9635398e6b3c35912955c60e4fa7


KvanTTT:
We are using [CaseInsensitiveInputStream](https://gist.github.com/sharwell/9424666) (lower case) for PHP grammar, also see [README](https://github.com/antlr/grammars-v4/tree/master/php).

Thus such a change is not needed.

KvanTTT:
Also, it's duplicate. See https://github.com/antlr/grammars-v4/issues/586

Issue #1804, opened 2020-07-29
XPath31 grammar missing cr/lf as whitespace contrary to spec.


kaby76:
The XPath31 grammar contains an error in the definition of [whitespace](https://github.com/antlr/grammars-v4/blob/17d3db3fd6a8fc319a12176e0bb735b066ec0616/xpath/xpath31/XPath31.g4#L389). [It should allow CR/LF characters](https://www.w3.org/TR/REC-xml/#NT-S), but the grammar doesn't because it's used in examples/expressions.txt as a delimiter between expressions, so each expression is not in a separate file. Instead, the grammar should follow the spec and another character used as expression delimiter (';' is unused in the spec, so that would be an OK choice). In fact, it's very important to allow CR/LF's in an expression in order to improve readability.

PR #1805, opened 2020-07-29
Fix for XPath31 grammar missing CR/LF as whitespace.

kaby76:
This is a fix for [XPath31 grammar missing CR/LF as whitespace contrary to spec. #1804](https://github.com/antlr/grammars-v4/issues/1804). It turns out that it's very handy to use CR/LF's in large expressions. An example, one expression that I'm working on is this one for finding lexer string literals that participate in rules of the form "FOOBAR : 'foobar' ;":

    //ruleSpec
        /lexerRuleSpec
            /lexerRuleBlock
                /lexerAltList[not(@ChildCount > 1)]
                    /lexerAlt
                        /lexerElements[not(@ChildCount > 1)]
                            /lexerElement[not(@ChildCount > 1)]
                                /lexerAtom
                                    /terminal[not(@ChildCount > 1)]
                                        /STRING_LITERAL

It's used to replace string literals with the corresponding lexer symbol in parse rules, since a partitioned lexer/parser grammar cannot define string literals in parser rules. Admittedly, even this is rather unruly to read, but it's a little better than trying to read the expression without the returns and indents. Anyway, what was there before was not per spec. Instead, I used a semi-colon to delimit XPath expressions for testing.

"mvn clean test" works on Ubuntu and Windows.

--Ken

### DIFF
--- a/xpath/xpath31/XPath31.g4
+++ b/xpath/xpath31/XPath31.g4
@@ -385,7 +385,7 @@ fragment FragChar : '\u0009' | '\u000a' | '\u000d'
   | '\u{10000}'..'\u{10ffff}'
  ;
 
-// Per spec https://www.w3.org/TR/REC-xml/#NT-S
+// https://github.com/antlr/grammars-v4/blob/17d3db3fd6a8fc319a12176e0bb735b066ec0616/xpath/xpath31/XPath31.g4#L389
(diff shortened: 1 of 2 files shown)


teverett:
@kaby76  thanks

Issue #250, opened 2015-11-05
SQLite does not recognize INNER JOIN, NATURAL JOIN etc.


egeogretmen:
When an sql statement is given such as

``` sql
SELECT * FROM t1 INNER JOIN t2 ON t1._id = t2.t1_id
```

"INNER" is perceived as table_alias
The problem is only solved if a table_alias is already given such as

``` sql
SELECT * FROM t1 as t INNER JOIN t2 ON t1._id = t2.t1_id
```

inglesp:
I *think* I was able to fix this by changing

```
any_name
 : IDENTIFIER 
 | keyword
 | STRING_LITERAL
 | '(' any_name ')'
 ;
```

to

```
any_name
 : IDENTIFIER 
 | STRING_LITERAL
 | '(' any_name ')'
 ;
```

I don't know whether this breaks anything else though!

inglesp:
This issue is also reported upstream at https://github.com/bkiers/sqlite-parser/issues/3.

teverett:
Have you considered submitting a pull request?

inglesp:
Yes, I'll follow this up upstream.

Marti2203:
@inglesp was this fixed?

inglesp:
@Marti2203 Doesn't look like it, sorry.  TBH I can't remember what I did four weeks ago, let alone four years...

Marti2203:
No problem, I am gonna play around with it today

Marti2203:
The grammar now catches it properly.

Marti2203:
@teverett  can you close this issue please?

PR #1808, opened 2020-08-04
Reworking SQLite grammar

Marti2203:
Issue #250 mentioned the grammar has an issue and I started to poke it around. I saw that many things were not very structurally okay - there were a little inconsistencies of the names, a strange prefix, some repetitions. Another issue was that the grammar was big so I split it into a lexer and parser. There are still things to be done, but I think that this is a bit better than before.


Marti2203:
@KvanTTT I am sorry to bother you, but I am still new with the tests in the system. Can you tell me where the test is located so I can see what it expects?

KvanTTT:
You should add sqlite to the [global pom.xml](https://github.com/antlr/grammars-v4/blob/master/pom.xml) and change [local pom.xml](https://github.com/antlr/grammars-v4/blob/master/sql/sqlite/pom.xml) with considering of your improvements. After that, you will be able to run maven tests by the following command: `mvn test`.

Marti2203:
@KvanTTT this is strange, I ran the file in `grun` (the ANTLR testrig) and it parsed without issues, but when I tried to do it in `mvn test`, I get a `no viable alternative`.

KvanTTT:
Maybe @teverett can give a piece of advice because he knows maven better :)

As I understand there are some errors during parsing with SQLite grammars, see https://github.com/antlr/grammars-v4/pull/1808/checks?check_run_id=948130080#step:4:9861

Marti2203:
The parser says there are no alternatives while grun shows a perfectly built tree that is strange and I am not experienced with maven a lot such that I can get info.

teverett:
I'm just a wee bit busy with work.  Can i take a look on the weekend?

Marti2203:
Yeah, not a problem. I will make some more pull requests until then

teverett:
@Marti2203  whats the root rule name in your grammar?

Marti2203:
The original author named it parse

KvanTTT:
Yes, and it's the first.

Marti2203:
Okay, it breaks on my machine now. I can fix it and test it!

Marti2203:
@KvanTTT I think it is fixed. It has passed all checks and conforms to the diagrams the SQLite website

PR #1810, opened 2020-08-08
Antlr2 grammar with two complex examples.

kaby76:
This is a conversion of the Antlr v2.7.7 grammar to Antlr4, hand tweaked in several places. I decided to make this grammar because I am using it for my Antlrvsix tool to convert any old Antlr2 grammars to Antlr4. See the [Import.md](https://github.com/kaby76/AntlrVSIX/blob/master/doc/Import.md) page. While there are only two grammars for tests, both are fairly narly. One is the original Antlr2 grammar. The other is a grammar for XQuery that was [converted to v4 just last year or so](https://stackoverflow.com/questions/51423673/migrating-antlr-v2-grammar-to-antlr-v4). "mvn clean test" passes on both Linux and Windows, and there should be no conflicts.


teverett:
@kaby76  thanks!

PR #1813, opened 2020-08-12
Hive join condition clause parsing error, mvn clean test, warnings, etc

oldsong:
Errors fixed:

  - grammar error for` joinSourcePart` which will cause join condition clause parsing failure
  - pom.xml errors causes `mvn clean test` failure
  - test SQL should only contain a single statement and with upper case keywords
  - parser rule references fragment lexer element
 
Warnings:
  - remove undefined keyword names
  
Others:
  - remove v3 semantic predicates
  - rename a rule name 'type' to 'type_db_col' to avoid keyword conflict with Python and Go


teverett:
@oldsong  Thanks!

PR #1815, opened 2020-08-13
Simplify Hive SQL expression grammar using direct left recursion

oldsong:
15 intermediate expressions for precedence are eliminated. The
simplification will dramatically reduce the size of the generated
parse tree.

There will be some topology changes in the generated parse
tree, for example: for expression 1 + 2 + 3, before we got:
     (+ 1 2 3)
now we get:
     (+ (+ 1 2) 3)

Add hive module to sql/pom.xml for mvn test

Add a test SQL for expression grammar testing


teverett:
@oldsong  thanks

PR #1822, opened 2020-08-19
Updated grammar for Antlr2, plus some old grammars I found on the Way…

kaby76:
…back Machine in antlr.org. The older grammars exposed some issues that needed a little tidying up, in octal numbers in string literals. "mvn clean test" passes on Ubuntu and Windows.


teverett:
@kaby76  Thanks!

Issue #491, opened 2016-09-30
EDN Format Grammar.


eprozium:
Please add an EDN Format https://github.com/edn-format/edn grammar too (it's a subset of Clojure).

Thanks.

Marti2203:
Hi. I am gonna make a grammar. Could you please make some examples? I hope it is not too late... @eprozium

aadrian:
@Marti2203 a few EDN file examples:

 - https://github.com/bpsm/edn-java/blob/master/src/test/resources/us/bpsm/edn/serializability.edn
 - all files in https://github.com/bpsm/edn-java/tree/master/src/test/resources/us/bpsm/edn/performance
 - https://github.com/clojure/java.jdbc/blob/master/deps.edn
 - https://github.com/seancorfield/dot-clojure/blob/develop/deps.edn

Marti2203:
Thank you! I am currently working on making it, I just need to add tests so that whenever somebody changes something, it can easily be tested.

PR #1831, opened 2020-08-27
Added grammar for the EDN format

Marti2203:
As per #491, I have created a grammar for EDN. It does not have full UTF-8 character support. And I would like it to be checked by people who use the format.


edn/edn.g4
```diff
@@ -0,0 +1,124 @@
...
+
+Symbol: '/' | Name ('/' Name)?;
+
+Name: (( [-+.] (Alpha | Extra | Special)) | (Alpha | Extra)) (
+		Alpha
+		| Numeric
+		| Extra
+		| Special
+	)*;
+//TODO ADD more UTF-8 Compilant Characters
```
studentmain, line 118: Try unicode category, e.g. `\p{L}` for any letter

Marti2203: Ah, forgot about that feature. Thank you! Will make a commit.

teverett:
@Marti2203  thanks

Issue #598, opened 2017-04-07
CPP grammar: Creating a reference of typename 'class'


prateekgulati:
When creating a reference variable for classtype, statement is parsed as an expression statement instead of declaration statement
eg 
`string foo = 'Hello';` is parsed as declaration statement 
but `string& fooRef = foo;` is parsed as expression statement and hence the complete subtree after that is flawed.

Marti2203:
Hi, I am currently reworking the C++ grammar. For the given code `string& fooRef = foo;` I get this
![image](https://user-images.githubusercontent.com/7456946/91658979-f30a7c00-eacc-11ea-8347-7acfc7013bf8.png)
Can you please check whether you wanted that?

Marti2203:
Close please @teverett

Issue #490, opened 2016-09-29
Cpp14 parsing error for attributes


DenisM67:
Hi,
I'm trying to parse a C++ header.
I use Cpp14.g grammar.
It seems the C++ attributes are not correctly parsed in a struct.

Thank you for your help.

Denis.
## 

The code to parse:

```c++
struct Aaa {
    [[attr1 ]] int aa; // Ok
    [[attr2(1)]] int bb; // not Ok, 
}
```

The result:

```
[@0,0:5='struct',<59>,1:0]
[@1,7:9='Aaa',<125>,1:7]
[@2,11:11='{',<82>,1:11]
[@3,15:15='[',<80>,2:1]
[@4,16:16='[',<80>,2:2]
[@5,17:21='attr1',<125>,2:3]
[@6,23:23=']',<81>,2:9]
[@7,24:24=']',<81>,2:10]
[@8,26:28='int',<38>,2:12]
[@9,30:31='aa',<125>,2:16]
[@10,32:32=';',<121>,2:18]
[@11,42:42='[',<80>,3:1]
[@12,43:43='[',<80>,3:2]
[@13,44:48='attr2',<125>,3:3]
[@14,49:49='(',<78>,3:8]
[@15,50:50='1',<126>,3:9]
[@16,51:51=')',<79>,3:10]
[@17,52:52=']',<81>,3:11]
[@18,53:53=']',<81>,3:12]
[@19,55:57='int',<38>,3:14]
[@20,59:60='bb',<125>,3:18]
[@21,61:61=';',<121>,3:20]
[@22,76:76='}',<83>,4:0]
[@23,493:492='<EOF>',<-1>,23:0]
line 3:9 no viable alternative at input 'structAaa{[[attr1]]intaa;[[attr2(1'
line 3:9 no viable alternative at input '[[attr2(1'
line 3:11 mismatched input ']' expecting {'(', '[', '+', '-', '*', '/', '%', '^'
, '&', '|', '<', '>', '<<', '==', '!=', '<=', '>=', '&&', '||', '++', '--', ',',
 '->*', '->', ';', '.', '.*'}
(translationunit (declarationseq (declarationseq (declarationseq declaration) (d
eclaration struct)) (declaration (functiondefinition (declarator (ptrdeclarator
(noptrdeclarator (declaratorid (idexpression (unqualifiedid Aaa)))))) (functionb
ody (compoundstatement { (statementseq (statementseq (statementseq (statementseq
[...]

PR #1838, opened 2020-08-30
Major rework of the C++ Grammar

Marti2203:
After checking for issue #490 , I have fully reworked the C++ grammar to  follow a camelCase and (mostly) PascalCase naming convention, a lot of lists are made to not be recursive and alternatives have been combined to make the grammar easier to understand. I have added an example to solve the aforementioned issue and #598 is also solved I think. Currently the only thing left is to make the rules directly left recursive such that the generated trees have a smaller depth, but this will be in another PR


teverett:
@Marti2203  thanks

Issue #1793, opened 2020-07-17
[CPP14 Grammar] Incorrect parsing of consecutive template ids


romeo-cmd:
The function definition `A<T> B<T>::C::Fonction(){}` gives the following AST:

![Modeles_consecutifs_non_reconnus](https://user-images.githubusercontent.com/61050967/87785323-46d04700-c838-11ea-9786-c91215bf19e6.png)

 which is incorrect in the case B and T refer to types, which is the most frequent.

Note that this also happen with double template end operator like in `A<B<T>> B<T>::C::Fonction(){}`, whereas it's not the case when the two consecutive templates are separated by operator `::` like in `int A<T>::B<T>::C::Fonction(){}`.

I don't know how to solve this issue since for me such pattern is actually ambiguous, and that to clear up ambiguity one has to determine whether `B` and `T` are types. Thus maybe actions raised when expanding `templateargument` rule would solve the issue.

mycroft8:
May or may not be related.  I'm trying to decide whether this fragment from the glibc type_traits file is a GCC dialect issue, or a problem in template alias parsing.  The second comma in the fragment is where the recognition error is thrown.

template <typename _ToElementType, typename _FromElementType> 
		using is_array_convertible = is_convertible <_FromElementType(*)[],_ToElementType(*)[]>;

Issue #1754, opened 2020-06-19
Incomplete memberdeclaration rule in Cpp grammar


romeo-cmd:
I think the `explicitspecialization` rule is missing as an alternative within the rule

```
memberdeclaration

: attributespecifierseq? declspecifierseq? memberdeclaratorlist? ';'

| functiondefinition

| usingdeclaration

| static_assertdeclaration

| templatedeclaration

| aliasdeclaration

| emptydeclaration ;
```

I don't know if it was already possible in C++14 standard, but at least now I think it is possible to explicitly specialize (but not explicitly instanciate) a template within some class declaration. Therefore I would add the alternative `explicitspecialization` to the `memberdeclaration` parser rule.

Marti2203:
Yes, this should be added. Will do now. I am making a big PR for the C++ grammar

Issue #1702, opened 2020-04-26
ANTLR CPP14 grammar template supporting


m-zakeri:
The ANTLR grammar for CPP14 does not accept nested template, for example, such code can not be parsed by the current grammar:

```cpp
class A
{
std::vector<std::unique_ptr<Controller>> controller_list_;
};
```

DonMathi:
For reference:
With a space in between > and > it parses, OK
```
class A
{
std::vector<std::unique_ptr<Controller> > controller_list_;
};
```

m-zakeri:
It's ok but how can I resolve this issue within the grammar? I have a huge source code.

DonMathi:
How is it going?
I have been thinking about how it maybe should be done. Two template '<' pushes the lexer into sub rules. But if there could be a way to detect if an expression with a possible '>>' then maybe the lexer should be pushed into a sub-sub set of rules, until the expression ends. If the expression possible includes two template '<' then it should push the lexer into the sub-rule for the second time, and so on.

romeo-cmd:
I got the same issue and solved it by changing lexer rule `RightShift` into parser rule `rightshift : '>' '>'` and change every reference to `RightShift` into a reference to `rightshift`.

KvanTTT:
Yes, there is another issue for reference: https://github.com/antlr/grammars-v4/issues/1753

DonMathi:
Here is a dummy example of the problem:
d is a variable of type a that has template parameter b, that has template parameter c right shifted 1.
`a< b< c>>1 > > d;`

Is there even any possible application for this?

DonMathi:
Maybe this:
`std::array<std::array< int, 3>>1 >, 4>>2 > myArray{{1, 2},{3,4}};`

romeo-cmd:
![antlr4_parse_tree](https://user-images.githubusercontent.com/61050967/84992942-e373cc00-b148-11ea-915b-02437c95a29e.png)
You are right it's an interesting point.

However I think that with enough lookup the parser is able to ditinguish the rightshift operator from the end of the template. Indeed, when parsing the rightshift operator `3>>` the parser find two matching rules in the rule table but then the hypothesis of rightshift operator rule is left because of the next double `>>` .

To be sure I tested your first example and obtained the good parse tree, as you can see in the attached document.

DonMathi:
What is the purpose of the 3rd expansion in 
`memberdeclarator
   : declarator virtspecifierseq? purespecifier?
[...]

Issue #1724, opened 2020-05-20
cql3 error for cpp


cangfengzhs:
java -jar ~/download/antlr-4.8-complete.jar -Dlanguage=Cpp -o gen/  CqlParser.g4 
error(134): CqlParser.g4:539:0: symbol delete conflicts with generated code in target language or runtime
error(134): CqlParser.g4:63:5: symbol delete conflicts with generated code in target language or runtime

KvanTTT:
Rename `delete` to `delete_`.

cangfengzhs:
thanks

Marti2203:
Hi, I am gonna make a pull request for this now, so this can be closed.

KvanTTT:
We are closing issues after pull request acceptance.

Marti2203:
Okay, my bad. Sorry

PR #1846, opened 2020-09-03
WIP: C++ Expression and Template Rework, CQL small fix

### LINKED ISSUE #1753: RightShift lexer rule eclipsing double template end character '>'
I think lexer rule `RightShift : '>>'` unables matching rule `simpletemplateid` in the case where there are more than two levels of imbrication of `simpletemplateid` rule.

For instance, the explicit template instanciation `template class MyTemplate1<MyTemplate2<MyClass>>;` will not be recognized because symbol `>>` is matched with lexeme `RightShift`.

To correct it, I propose to change lexer rule `RightShift : >>` into parser rule `rightshift : '>' '>'` and change every reference to rule `RightShift` into a reference to rule `rightshift`.

Note: This problem might also happen with lexemes `RightShiftAssign`, `LeftShift`, and `LeftShiftAssign` but I think in practice it is impossible.

Marti2203:
Issues related to this : #1793 #1753, #1702 #1754 #1869 
Current state: Not done with Templates


teverett:
The PR notes that it is not done.  Do you prefer to pull this on to MASTER now or wait?

Marti2203:
I thought it was done, but then I found out that it does not catch #1793

Marti2203:
I would prefer I fix that also and then give a light to pull

teverett:
All good.  Just drop a note here when you're ready.

Marti2203:
Fixing #1724

teverett:
@Marti2203 good to pull this?

Marti2203:
I thought I had fixed the templates issue (A<B<C>> not parser correctly) , but I have. I have fixed the code to use Antlr's left recursion....

teverett:
@Marti2203  @KvanTTT  is this ready to pull?

KvanTTT:
Please remove `WIP` prefix at first.

Marti2203:
@teverett it is not ready, I currently only the templates are left and I am not seeing any proggress with them.

PR #1851, opened 2020-09-05
enable Dependabot v2

sullis:
https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/
﻿


teverett:
I like the idea of dependabot, however I am a little concerned that it will automatically update the antlr deps.  There have been times in the past where antlr was updated and had bugs that prevented parsing all grammars in grammars-v4.  I prefer to update the antlr deps manually.

PR #1852, opened 2020-09-05
annotation rule fix: alternative qualified name

mlorek:


KvanTTT:
Thanks! Could you please extend AllInOne7.java or AllInOne8.java files with a test case that covers your fixes?

PR #1857, opened 2020-09-20
MySQL: DATETIME|TIMESTAMP data type column with default value initialization and updating

imac-beep:
Hi all, 
I've encountered a real-life corner case, with null default value on DATETIME|TIMESTAMP datatype that can not be parsed.
```mysql
create table ts_table(
  ts10 TIMESTAMP DEFAULT NULL ON UPDATE CURRENT_TIMESTAMP
);

create table dt_table(
  dt10 DATETIME DEFAULT NULL ON UPDATE CURRENT_TIMESTAMP
)
```
According to official documentation https://dev.mysql.com/doc/refman/8.0/en/timestamp-initialization.html, the case described above will become parse-able with this patch.


KvanTTT:
@jpechane are you ok with this change?

jpechane:
@imac-beep Hi, thanks for the contribution! I checked the linke grammer and I see there is also another form supported
```
CREATE TABLE t1 (
  ts TIMESTAMP DEFAULT 0 ON UPDATE CURRENT_TIMESTAMP,
  dt DATETIME DEFAULT 0 ON UPDATE CURRENT_TIMESTAMP
);
```

Will this work with the current PR? If not could you fix it too? In that case it might make sense to modify the rule as `(ON UPDATE currentTimestamp)?` would be supported by all three cases so it might not be necessary to be repeated in all of them.

imac-beep:
@jpechane Hi,  you're right about the lacking of accuracy and elegance of this patch. It was an urgent patch without thorough consideration.

As you mentioned, all three `defaultValue`s ( NULL / constant / currentTimestamp ) can be prefixed with `ON UPDATE` clause.

```diff
 //    Common Expressons

 defaultValue
-    : NULL_LITERAL (ON UPDATE currentTimestamp)?
-    | constant
-    | currentTimestamp (ON UPDATE currentTimestamp)?
+    : (NULL_LITERAL | constant | currentTimestamp) (ON UPDATE currentTimestamp)?
```

I'd like to to amend this PR as above with test cases covered.  What do you think?

jpechane:
@imac-beep Yes, that'd be spot on, thanks! So just add TC for `constant` case and I belive we are good to go.

PR #1863, opened 2020-10-01
.

Rahul-Mishra-coder:


KvanTTT:
Why? Grammars - it's the correct plural form of grammar.

teverett:
Grammars is the correct work since there are more than one.

PR #1864, opened 2020-10-01
Add top level elements

kayman-mk:
Remove the top level block element and replace it by resource, data, provider, output, local and module. These are the elements supported by Terraform.


kayman-mk:
@teverett Fixed. Should be green in some minutes.

teverett:
@kayman-mk thanks

PR #1870, opened 2020-10-03
Removed grammar ambiguities, lot of rule fixes and additions

siggemannen:
Hi.
I have been making a lot of fixes to TSQL-grammar, and i thought these might be interesting to everyone to have.
Here's a summary of the fixes in this pull request, ordered by importance:

- A lot of ambiguities have been removed (especially around the search_condition and expression predicates). These fixes now allow tsql to be parsed using the quick SLL prediction mode, which removes all performance issues of the original grammar. I have not hit any downsides or any LL degradation after the changes after running against a huge sql repository at my current work
- a LOT of rules have been corrected. A quick summary:
 - a whole bunch of new keywords that can be identifiers were added
 - fixes of batch delimiters like GOs and one statement execution batches
 - better support for index (including table declaration, xml) and set options
 - a lot of fixes in three-four-unlimited name parts of tables, columns etc, and also which kind of names is used in which contexts
 - corrected merge statement when matched
 - collate support for default declaration
 - improved xml method calls
 - at time zone support
 - added support for some freetext functions
 - more window functions
 - support added to for json statement
 - string matching IP address got highjacked by wrong rules
 - added support for changetable version
- restructured keywords in two types, reserved and ok keywords
- added "list" variables for some repeating things like parameters, so it's easier to aggregate for developers writing parsers etc 

So yeah, a lot of work done, i hope this patch can be adopted but if not, maybe it will be useful for someone else.
I'm better at sql than git, so feel free to correct my commit in any way.

Known limitations of these changes:
 - GO {number} isn't supported, Also, i had to do some workarounds regarding support for multiple GO statements which might affect some of the more odd constructs
 - Only tested in SQL Server. I might have inadvertently dropped support for some sybase constructs.

//Cheers


KvanTTT:
Thank you for such a valuable improvements. Are you taking participation in [hacktoberfest](https://hacktoberfest.digitalocean.com/)? This request is more than suitable for accepting.

siggemannen:
Hi, just happy to help! Not sure about hacktoberfest, but i would definitely like to help out with the tsql-grammar if you need any kind of corrections of improvements to the rules

sql/tsql/TSqlLexer.g4
```diff
@@ -551,6 +369,7 @@ GB:                                    'GB';
 GETDATE:                               'GETDATE';
 GETUTCDATE:                            'GETUTCDATE';
 GLOBAL:                                'GLOBAL';
+GO_BATCH:                              [\r\n]* [\t ]* 'GO'[\t\n\r ];
```
KvanTTT, line 372: I don't think such construction is optimal from the performance perspective because it conflicts with `SPACE` token. Can it be replaced with post predicate like this?

```suggestion
GO_BATCH:                              'GO' {isNewLine(-2)}? [\t\n\r ];
```

We usethe same approach for [plsql](https://github.com/antlr/grammars-v4/blob/master/sql/plsql/PlSqlLexer.g4#L2345) and [php](https://github.com/antlr/grammars-v4/blob/master/php/PhpLexer.g4#L46) grammars.

sql/tsql/TSqlLexer.g4
```diff
@@ -104,310 +67,161 @@ CURRENT_TIME:                          'CURRENT_TIME';
...
 INNER:                                 'INNER';
 INSERT:                                'INSERT';
-INSTEAD:                               'INSTEAD';
 INTERSECT:                             'INTERSECT';
 INTO:                                  'INTO';
-IPV4_ADDR:                             [']? IPV4_OCTECT DOT IPV4_OCTECT DOT IPV4_OCTECT DOT IPV4_OCTECT [']?;
-IPV6_ADDR:                              [']?[0-9A-F]?[0-9A-F]?[0-9A-F]?[0-9A-F]?[:][0-9A-F]?[0-9A-F]?[0-9A-F]?[0-9A-F]?[:][0-9A-F]?[0-9A-F]?[0-9A-F]?[0-9A-F]?[:][0-9A-F]?[0-9A-F]?[0-9A-F]?[0-9A-F]?[:][0-9A-F]?[0-9A-F]?[0-9A-F]?[0-9A-F]?[:][0-9A-F]?[0-9A-F]?[0-9A-F]?[0-9A-F]?[:][0-9A-F]?[0-9A-F]?[0-9A-F]?[0-9A-F]?[:][0-9A-F]?[0-9A-F]?[0-9A-F]?[0-9A-F]?[']?;
+IPV4_ADDR:                              IPV4_OCTECT DOT IPV4_OCTECT DOT IPV4_OCTECT DOT IPV4_OCTECT;
+IPV6_ADDR:                              [0-9A-F]?[0-9A-F]?[0-9A-F]?[0-9A-F]?[:][0-9A-F]?[0-9A-F]?[0-9A-F]?[0-9A-F]?[:][0-9A-F]?[0-9A-F]?[0-9A-F]?[0-9A-F]?[:][0-9A-F]?[0-9A-F]?[0-9A-F]?[0-9A-F]?[:][0-9A-F]?[0-9A-F]?[0-9A-F]?[0-9A-F]?[:][0-9A-F]?[0-9A-F]?[0-9A-F]?[0-9A-F]?[:][0-9A-F]?[0-9A-F]?[0-9A-F]?[0-9A-F]?[:][0-9A-F]?[0-9A-F]?[0-9A-F]?[0-9A-F]?;
```
KvanTTT, line 121: Maybe replace it with using of `fragment` tokens for clarity?

sql/tsql/TSqlParser.g4
```diff
@@ -1850,20 +1864,27 @@ create_database
...
     (WHERE where=search_condition)?
     (index_options)?
     (ON id)?
     ';'?
     ;
 
+create_xml_index
+    : CREATE PRIMARY? XML INDEX id ON table_name '(' id ')'
+    (USING XML INDEX id (FOR (VALUE | PATH | PROPERTY)?)?)? 
+    (index_options)?
```
KvanTTT, line 1878: Excess parentheses:

```suggestion
    index_options?
```

sql/tsql/TSqlParser.g4
```diff
@@ -1969,7 +1990,13 @@ update_statistics
 
 // https://msdn.microsoft.com/en-us/library/ms174979.aspx
 create_table
-    : CREATE TABLE table_name '(' column_def_table_constraints ','? ')' (LOCK simple_id)? table_options* (ON id | DEFAULT)? (TEXTIMAGE_ON id | DEFAULT)?';'?
+    : CREATE TABLE table_name '(' column_def_table_constraints  (','? table_indices)*  ','? ')' (LOCK simple_id)? table_options* (ON id | DEFAULT)? (TEXTIMAGE_ON id | DEFAULT)?';'?
+    ;
+
+table_indices
+    :  INDEX id  (UNIQUE | CLUSTERED | NONCLUSTERED)? '(' column_name_list_with_order ')'
+    (index_options)?
```
KvanTTT, line 1998: Excess parentheses.

sql/tsql/TSqlParser.g4
```diff
@@ -2513,16 +2545,36 @@ kill_stats_job
 
 // https://msdn.microsoft.com/en-us/library/ms188332.aspx
 execute_statement
-    : EXECUTE execute_body
+    : EXECUTE execute_body ';'?
+    ;
+
+execute_body_batch
+    :
+    func_proc_name_server_database_schema (execute_statement_arg (',' execute_statement_arg)*)? ';'?
```
KvanTTT, line 2553:
```suggestion
    : func_proc_name_server_database_schema (execute_statement_arg (',' execute_statement_arg)*)? ';'?
```

sql/tsql/TSqlParser.g4
```diff
@@ -2833,18 +2908,25 @@ constant_LOCAL_ID
 expression
     : primitive_expression
     | function_call
+    | expression '.' value_call
+    | expression '.' query_call
+    | expression '.' exist_call
+    | expression '.' modify_call
```
KvanTTT, line 2914: Do all alternatives have different priorities? Is not, merge it to the single alternative:

```suggestion
    | expression '.' (value_call | query_call | exist_call | modify_call)
```

sql/tsql/TSqlParser.g4
```diff
@@ -2879,68 +2961,67 @@ subquery
...
-search_condition_and
-    : search_condition_not (AND search_condition_not)*
+    : pred+=predicate_br (log=(OR | AND) pred+=predicate_br)*
     ;
 
-search_condition_not
-    : NOT? predicate
+predicate_br
+    : NOT* predicate
+    | NOT* '(' search_condition ')'
```
KvanTTT, line 2990:
```suggestion
    : NOT* (predicate | '(' search_condition ')')
```

sql/tsql/TSqlParser.g4
```diff
@@ -2879,68 +2961,67 @@ subquery
...
     : EXISTS '(' subquery ')'
+    | freetext_predicate
     | expression comparison_operator expression
     | expression comparison_operator (ALL | SOME | ANY) '(' subquery ')'
-    | expression NOT? BETWEEN expression AND expression
-    | expression NOT? IN '(' (subquery | expression_list) ')'
-    | expression NOT? LIKE expression (ESCAPE expression)?
+    | expression NOT* BETWEEN expression AND expression
+    | expression NOT* IN '(' (subquery | expression_list) ')'
+    | expression NOT* LIKE expression (ESCAPE expression)?
```
KvanTTT, line 3000: It should be merged to the single alternative since all operators `BETWEEN`, `IN`, and `LIKE` have the same priority, see https://docs.microsoft.com/en-us/sql/t-sql/language-elements/operator-precedence-transact-sql?view=sql-server-ver15

```suggestion
    | expression NOT* (BETWEEN expression AND expression | IN '(' (subquery | expression_list) ')' | LIKE expression (ESCAPE expression)?)
```

sql/tsql/TSqlParser.g4
```diff
@@ -3072,63 +3154,105 @@ table_source
...
     ;
 
 schema_declaration
-    : column_declaration (',' column_declaration)*
+    : xml_col+=column_declaration (',' xml_col+=column_declaration)*
     ;
 
 column_declaration
-    : ID data_type (STRING)?
+    : id data_type (STRING)?
```
KvanTTT, line 3199: Also excess parentheses:

```suggestion
    : id data_type STRING?
```

sql/tsql/TSqlParser.g4
```diff
@@ -3072,63 +3154,105 @@ table_source
...
-    : CHANGETABLE '(' CHANGES table_name ',' (NULL | DECIMAL | LOCAL_ID) ')'
+    : change_table_changes
+    | change_table_version
     ;
 
+change_table_changes
+    : CHANGETABLE '(' CHANGES changetable=table_name ',' changesid=(NULL | DECIMAL | LOCAL_ID) ')'
+    ;
+change_table_version
+    : CHANGETABLE '(' VERSION versiontable=table_name ',' (pk_columns=full_column_name_list) ',' (pk_values=select_list)  ')'
```
KvanTTT, line 3211:
```suggestion
    : CHANGETABLE '(' VERSION versiontable=table_name ',' pk_columns=full_column_name_list ',' pk_values=select_list  ')'
```

sql/tsql/TSqlParser.g4
```diff
@@ -2879,68 +2961,67 @@ subquery
...
+      columns=select_list
       // https://msdn.microsoft.com/en-us/library/ms188029.aspx
-      (INTO table_name)?
-      (FROM table_sources)?
+      (INTO into=table_name)?
+      (FROM from=table_sources)?
       (WHERE where=search_condition)?
       // https://msdn.microsoft.com/en-us/library/ms177673.aspx
-      (GROUP BY (ALL)? group_by_item (',' group_by_item)*)?
+      (GROUP BY (groupByAll=ALL)? groupBys+=group_by_item (',' groupBys+=group_by_item)*)?
```
KvanTTT, line 3024:
```suggestion
      (GROUP BY groupByAll=ALL? groupBys+=group_by_item (',' groupBys+=group_by_item)*)?
```

sql/tsql/TSqlParser.g4
```diff
@@ -3072,63 +3154,105 @@ table_source
...
+    | apply
+    | pivot
+    | unpivot
+    ;
+join_on
+    :
+    (inner=INNER? |
+       join_type=(LEFT | RIGHT | FULL) outer=OUTER?) (join_hint=(LOOP | HASH | MERGE | REMOTE))?
+       JOIN source=table_source ON cond=search_condition
+    ;
```
KvanTTT, line 3228: Formatting:

```suggestion
    ;

join_on
    : (inner=INNER? | join_type=(LEFT | RIGHT | FULL) outer=OUTER?) (join_hint=(LOOP | HASH | MERGE | REMOTE))?
       JOIN source=table_source ON cond=search_condition
    ;
```

sql/tsql/TSqlParser.g4
```diff
@@ -3072,63 +3154,105 @@ table_source
...
+apply
+    : apply_style=(CROSS | OUTER) APPLY source=table_source
+    ;
+
+pivot
+    : PIVOT pivot_clause as_table_alias
+    ;
+
+unpivot
+    :     UNPIVOT unpivot_clause as_table_alias
```
KvanTTT, line 3243:
```suggestion
    : UNPIVOT unpivot_clause as_table_alias
```

sql/tsql/TSqlParser.g4
```diff
@@ -3455,23 +3612,27 @@ ddl_object
...
-    | (table_name '.')? QUOTED_IDENTIFIER
-    | (table_name '.')? ARITHABORT
-    | (table_name '.')? ANSI_WARNINGS
-    | (table_name '.')? ANSI_PADDING
-    | (table_name '.')? ANSI_NULLS
-
+    : server=id? '.' schema=id? '.' tablename=id? '.' column_name=id
+    | schema=id? '.' tablename=id? '.' column_name=id 
+    | tablename=id? '.' column_name=id
+    | column_name=id
```
KvanTTT, line 3618: Maybe optimize it?

```suggestion
    : (((server=id? '.')? schema=id? '.')? tablename=id? '.')? column_name=id
```

sql/tsql/TSqlParser.g4
```diff
@@ -3556,47 +3717,14 @@ send_conversation
...
-    | XML*/
-    : id IDENTITY? ('(' (DECIMAL | MAX) (',' DECIMAL)? ')')?
-    | DOUBLE PRECISION?
-    | INT
-    | TINYINT
-    | SMALLINT
-    | BIGINT
+    : scaled=(VARCHAR | NVARCHAR | BINARY_KEYWORD | VARBINARY_KEYWORD) '(' MAX ')'
+    | ext_type=id '(' scale=DECIMAL ',' prec=DECIMAL ')'
+    | ext_type=id '(' scale=DECIMAL ')'
```
KvanTTT, line 3724:
```suggestion
    | ext_type=id '(' scale=DECIMAL (',' prec=DECIMAL)? ')'
```

sql/tsql/TSqlParser.g4
```diff
@@ -3619,442 +3747,639 @@ sign
     | '-'
     ;
 
+keywords:
```
KvanTTT, line 3750: As I understand it's a single keyword, not multiple. `keywords` -> `keyword`.

Also, indentation is not unified for keyword rule. It should be 4 spaces instead of none.

KvanTTT (changes requested):
It's great in general, but there are some minor issues and a couple of more severe.

siggemannen:
Hi, and thanks a lot for review, i agree with most issues and will try to fix as soon as possible.
A couple of personal comments though:
I unrolled some rules into several alternatives as you noticed, i think it was to help with the parsing, i *think* in some cases i got hit with ambiguities without the unroll, but not sure. In some cases i didn't get an ambiguity but i got a very slow parsing time. The problem is that i was testing against some code which cannot be easily converted into proper testcases, and once i hit a problem i tried to simplify the antlr-rule until the problem went away. It was an iterative process and i don't really remember which fixes fixed what :)
But, i don't think unrolling is necessary a bad thing, plus it makes the rules a bit clearer.
GO_BATCH: this one is a bit tricky. In my opinion, i think it's better to stay clear of language-specific antlr-constructructs like isNewLine, plus i don't have the enviroments to test the changes in all languages. Perhaps i just need some help with this part.

KvanTTT:
> GO_BATCH: this one is a bit tricky. In my opinion, I think it's better to stay clear of language-specific antlr-constructs like isNewLine, plus I don't have the environments to test the changes in all languages. Perhaps I just need some help with this part.

Unfortunately, it's not always possible to get rid of language-specific constructs. Moreover, they also can be unified with using of base classes, see https://github.com/antlr/grammars-v4/tree/master/javascript/javascript implementations for instance. But I can fix it later.

siggemannen:
Hi again. I have made some more indentation fixes and some merges of statements, as well as removed parens and other suggestions. Also, got all changes from recent master commits. I kept some of the unrolled changes due to reasons stated above.

siggemannen:
Actually, i realized that IP for availability_group required a string, so might as well remove it the ip fragment stuff altogether, and can validate it runtime if needed.
If we insist on making a string validation, all the rules requiring strings will need to include the ip string tokens which is ugly

KvanTTT:
> I unrolled some rules into several alternatives as you noticed, I think it was to help with the parsing, I think in some cases I got hit with ambiguities without the unroll, but not sure. In some cases, I didn't get an ambiguity but I got a very slow parsing time. The problem is that I was testing against some code that cannot be easily converted into proper test cases, and once I hit a problem I tried to simplify the antlr-rule until the problem went away. It was an iterative process and I don't really remember which fixes fixed what :)

Ok, I'm merging this request but more thorough investigation is required.

PR #1887, opened 2020-10-08
Allow keywords as identifiers

kayman-mk:
Sometimes keywords are used as identifiers in the HCL. Make sure that the parser doesn't fail.

data "aws_iam_policy_document" "datadog_aws_integration_assume_role" {
...
    condition {
      test = "StringEquals"
      variable = "sts:ExternalId"  //variable is a keyword

      values = [
        "123456"
      ]
    }
  }
}


teverett:
@kayman-mk thanks

PR #1888, opened 2020-10-08
Replace the new RESOURCEREFERENCE by a section.section as it was before

kayman-mk:
It makes no sense to have a separate RESOURCEREFERENCE (was introduced by me). But the section has 
to be enhanced as some keywords are used as identifiers as well.


kayman-mk:
Everything fixed

teverett:
@kayman-mk  thanks

PR #1892, opened 2020-10-15
Add inline index expression

kayman-mk:
aws_acm_certificate.ecs_certificate.domain_validation_options.0.resource_record_name

### DIFF
--- a/terraform/terraform.g4
+++ b/terraform/terraform.g4
@@ -108,7 +108,7 @@ identifierchain
 inline_index
    : NATURAL_NUMBER
    ;
-   
+
 expression
    : section
    | expression operator expression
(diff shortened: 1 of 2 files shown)


terraform/terraform.g4
```diff
@@ -172,10 +177,10 @@ fragment DIGIT
...
    ;
 
-SIGNED_NUMBER
-   : '+' NUMBER
-   | '-' NUMBER
-   | NUMBER
+signed_number
+   : '+' number
+   | '-' number
+   | number
```
KvanTTT, line 183: Can be simplified:

```suggestion
signed_number
   : ('+' | '-')? number
```

kayman-mk: Looks much better.

teverett:
@kayman-mk  thanks

PR #1895, opened 2020-10-20
chore: re-integrate the PL/SQL grammar into the build

Fiouz:
The grammar was there already but unreachable by the build (it was
commented out without any explanation).
This commit re-integrates the grammar into the Maven build.


teverett:
@Fiouz  thanks

PR #1905, opened 2020-10-26
Rename NULL, TRUE, and FALSE literals in SQLite grammar to successfully compile targeting c++

### LINKED ISSUE #1904: SQLite grammar uses "NULL" which causes auto generated c++ code to fail to compile
The SQLite lexer names the NULL sql keyword as "NULL" https://github.com/antlr/grammars-v4/blob/e6b96bde8c3f2b82d404c8daf00162df97566398/sql/sqlite/SQLiteLexer.g4#L131

When ANTLR is run on that grammar to generate code for C++, the lexer names are transformed into to an autogenerated enum defined in SQLiteLexer.h and used in various generated .h and .cpp files. Obviously, SQLiteLexer.h includes "antlr4-runtime.h" (at the top of the file, prior to the enum definition), which transitively includes stdlib.h where NULL is defined as a constant via #define NULL ...

This causes compilation of these c++ files to fail as the compiler ends up substituting identifier NULL in the enum definition (and various other locations it is used in the autogenerated code) with 0 (or whatever stdlib.h defines NULL to be on the particular architecture).

This is a well known issue with several tools that perform code generation, including ANTLR (https://github.com/antlr/antlr4/issues/2029) and protobuf (https://github.com/protocolbuffers/protobuf/issues/6609) among others. The ANTLR team does not seem interested in supporting this (via e.g. automatically rewriting NULL to something else) and instead recommends that the grammar calls such entities NULL_ or something similar.

The MySql grammar does not suffer this, it uses NULL_LITERAL as the name in the lexer instead of NULL.

KinanBab:
ANTLR runtime includes header files that define NULL via preprocessing directive, using NULL as an identifier in grammars (specifically the lexer) causes it to be used
as an identifier in the corresponding auto-generated c++ code which then fails to compile.

[edited:] We change NULL to NULL_  in SQLiteLexer.g4 and update its use in SQLiteParser.g4. We also do the same to TRUE and FALSE to avoid similar issues that may occur when compiling on windows.

This fixes #1904


studentmain (approved):

KvanTTT (commented):
Why do you think that using `NULL_LITERAL` and others instead of `NULL_`, `TRUE_`, `FALSE_` is better? We have such agreements about conflicting identifiers.

KinanBab:
Sorry, I had originally planned to do it with just _, but the MySQL grammar you currently have uses NULL_LITERAL instead and I thought I would be consistent.

All good now, changed my PR to use NULL_, TRUE_, and FALSE_

KvanTTT (approved):

PR #1918, opened 2020-11-10
Fix SQLite grammar parsing column constraints into type name

### LINKED ISSUE #1911: SQLite type name captures too much in column definition
Hello,

I ran into an issue while using the SQLite grammar.

I want to parse a (simple) statement like this:
```sql
CREATE TABLE mytable (col1 int UNIQUE);
```
This is a very simple statement that should parse correctly for the parser to be of use.

The interesting thing is that it is parsed, but the resulting parse tree is in correct. Namely, the type is parsed as "int UNIQUE" rather than just "int" and "UNIQUE" being a constraint.

The culprits are these lines in the grammar:
1. In the column definition rule, things go as you expect: https://github.com/antlr/grammars-v4/blob/39e35769d4e8d66fe5989e458ac0caedb8a92fd2/sql/sqlite/SQLiteParser.g4#L118
2. However, in the type_name rule, you can see that the "name" rule is used with a +. It always the name to span several words / tokens. https://github.com/antlr/grammars-v4/blob/39e35769d4e8d66fe5989e458ac0caedb8a92fd2/sql/sqlite/SQLiteParser.g4#L121
3. Finally, the "any_name" rule ("name" is just an alias for it) is defined to include keywords as well as identifiers https://github.com/antlr/grammars-v4/blob/39e35769d4e8d66fe5989e458ac0caedb8a92fd2/sql/sqlite/SQLiteParser.g4#L759

So all in all, if we expand the column definition rule, it becomes something like this:
```antlr
column_def: column_name <<<<((keyword | identifier | ...)+ ...)>>>>? column_constraint*
```
Where the portion between angular brackets is the expanded type_name.

The problem is now clear: the type name is going to capture essentially everything up to the , that signifies the end of this column definition. There is never a way to capture a constraint, regardless of what the statement being parsed actually is.

Even worse, column definitions like "id UNIQUE" or "id PRIMARY KEY" also parse the constraint as the type name (even though it is omitted).

Furthermore, as a work around, I tried to have the constraint specified separately similar to:
```sql
CREATE TABLE mytable(col1 int, CONSTRAINT c UNIQUE (col1));
```
and even that did not work properly, CONSTRAINT (albeit being a keyword) is matched as the column_name (remember that any_name includes keywords) and then "c UNIQUE (col1)" is matched as the typename, because any_name also allows for parenthesized any_name.

This is probably a result of making the grammar look exactly like this piece of sqlite documentation: 
https://sqlite.org/syntax/type-name.html

However, this is clearly in correct. My understanding is that a keyword can only used as an identifier in sqlite if it is either quoted (using ', ", or `), or surrounded with square brackets ([ ]), and similarly for names containing whitespace.
http://barbra-coco.dyndns.org/sqlite/lang_keywords.html
https://stackoverflow.com/questions/14019275/read-data-from-sqlite-where-column-name-contains-spaces

I am not entirely certain what is the best approach here yet. I am doing some additional reading to see if I can find something. Only allowing keywords in names if they are quoted or square-bracketed sounds like a very helpful improvement, I am unsure if this should be enforced for all names, or just column names and type names.

carlreinke:
Fixes #1911.

### DIFF
--- a/sql/sqlite/SQLiteParser.g4
+++ b/sql/sqlite/SQLiteParser.g4
@@ -118,7 +118,7 @@ create_table_stmt:
 column_def: column_name type_name? column_constraint*;
 
 type_name:
-	name+ (
+	name+? (
 		'(' signed_number ')'
 		| '(' signed_number ',' signed_number ')'
 	)?;
--- a/sql/sqlite/SQLiteParser.g4
+++ b/sql/sqlite/SQLiteParser.g4
@@ -108,7 +108,7 @@ create_table_stmt:
 		schema_name '.'
 	)? table_name (
 		(
-			'(' column_def (',' column_def)* (
+			'(' column_def (',' column_def)*? (
 				',' table_constraint
 			)* ')' (WITHOUT rowID = IDENTIFIER)?
 		)


sql/sqlite/SQLiteParser.g4
```diff
@@ -118,7 +118,7 @@ create_table_stmt:
 column_def: column_name type_name? column_constraint*;
 
 type_name:
-	name+ (
+	name+? (
```
KvanTTT, line 121: Please replace with `name*`

carlreinke: That wouldn't be right.  The grammar should not match `CREATE TABLE foo (bar (1) NOT NULL)`.

carlreinke:
I also added a fix for `table_constraint` being greedily parsed into `column_def`.

teverett:
@carlreinke  @KvanTTT  is this ok to merge?

carlreinke:
@teverett AFAIK, yes.

mike-lischke:
I just opened an new PR for the SQLite grammars. Why is this PR not merged yet? Everyone seems to agree it is ok.

teverett:
@KvanTTT are you ok with this PR?

PR #1924, opened 2020-11-13
make MySqlLexer case insensitive

byronhe:
converted using this little script :
```py
import re
import sys
import io

# cat MySqlLexer.g4  | python3 cvt.py  > MySqlLexer.case_insensitive.g4
def cvt_char(s):
    arr=[  ]
    for c in s:
        if c >= 'A' and c <= 'Z':
            arr.append(c)
        else:
            arr.append("'" + str(c) + "'")
    return " ".join(arr)


r=re.compile("([A-Z0-9_]*:\s*)('[A-Z0-9_]*')(;)")
for line in sys.stdin:
    s=r.match(line)
    if s is None:
        print(line, end='')
        continue

    part=s.group(2)
    part=cvt_char(part.replace("'",""))
    print(s.group(1) + part + s.group(3) )

```


KvanTTT:
Please use `CaseInsensitiveInputStream`, see [Case-Insensitive Lexing](https://github.com/antlr/antlr4/blob/master/doc/case-insensitive-lexing.md).

PR #1935, opened 2020-11-15
Ged

teverett:


pom.xml
```diff
@@ -121,7 +121,8 @@
 				<module>fol</module>
 				<module>fortran77</module>
 				<module>fusion-tables</module>
-				<module>gff3</module>
+                <module>gedcom</module>
+                <module>gff3</module>
```
KvanTTT, line 125: Could you please normalize spaces and tabs in this file?

PR #1955, opened 2020-12-03
Update C# and Java Antlr4 parsers to work out of the box

EwoutWal:
The Antlr4 lexer grammar had a mode called 'Action' which collided with the function of the same name in the Lexer class. It has been renamed to TargetLanguageAction in the lexer grammar.

The C# lexer adaptor was still missing the handleOptionsLBrace method, this has been implemented as a replication of the Java reference.


teverett:
@EwoutWal  thanks!

Issue #1956, opened 2020-12-03
antlr/antlr4/ANTLRv4Parser.g4 has an error in the lexerElements rule


kaby76:
An issue was raised in Antlrvsix ([Significant typing lag in grammar files and wrong colorization #111](https://github.com/kaby76/AntlrVSIX/issues/111)) that uncovered an error in the ANTLRv4Parser.g4 file.

The input grammar for the Antlrvsix issued contained the following lexer rule:

	EMPTY_STR           : -> type(UNTERMINATED_STR), popMode ;

Yes, it was not parsing in Antlrvsix, which is based on the grammar-v4 grammar [ANTLRv4Parser.g4](https://github.com/antlr/grammars-v4/blob/master/antlr/antlr4/ANTLRv4Parser.g4). Per chance, I checked the grammar against the Antlr4 tool itself. While there was a warning, the tool parses the grammar.

I compared [the Antlr tool grammar](https://github.com/antlr/antlr4/blob/0d73fc28e98c1a83f5cc637c3d41f3834c9d0b47/tool/src/org/antlr/v4/parse/ANTLRParser.g#L549) with [antlr/antlr4/ANTLRv4Parser.g4](https://github.com/antlr/grammars-v4/blob/master/antlr/antlr4/ANTLRv4Parser.g4#L231), and indeed, there is an error in the grammars-v4 version for lexerElements:

    lexerElements
       : lexerElement+
       ;

It should be instead:

    lexerElements
        :	lexerElement+
	    |	// empty alt
        ;

I checked the Antlr3 grammar, and that looks fine.

I haven't finished my grammar comparison tool yet for Antlrvsix/Trash, but this is exactly why I need to get a version working.

PR #1959, opened 2020-12-14
Fix for https://github.com/antlr/grammars-v4/issues/1956

kaby76:
The most important change here is to add an empty alternative to the lexerElements rule, bringing the grammar closer to the grammar used by the Antlr4 tool.

I also changed the indentation and spacing of the comments between rule groups because some of the comments were up againts the ';' of the previous rule, e.g., "// Lexer rules". I think that indentation got in from an automated reformater. The indentation of the comments dont' look as odd now.

I added an example for Issue 1956, as described in the issue.


teverett:
@kaby76  thanks!

PR #1965, opened 2020-12-16
php 8 support

studentmain:


php/PhpLexer.g4
```diff
@@ -111,6 +111,8 @@ MultiLineComment:   '/*' .*? '*/' -> channel(PhpComments);
 SingleLineComment:  '//' -> channel(SkipChannel), pushMode(SingleLineCommentMode);
 ShellStyleComment:  '#' -> channel(SkipChannel), pushMode(SingleLineCommentMode);
 
+AttributeStart:     '#[';
```
KvanTTT, line 114: It seems like spaces are allowed between `#` and `[`, see https://3v4l.org/5ZYsq

studentmain: No, `#` is start of comment, so `# [Attribute]` will parse. But it's not interpreted as attribute, can't read by getAttributes() reflection API. see https://3v4l.org/7FPRr

KvanTTT (approved):

KvanTTT:
Thanks!

PR #1966, opened 2020-12-16
Php Improvements

KvanTTT:
@studentmain, please review


php/PhpParser.g4
```diff
@@ -156,22 +152,18 @@ typeParameterWithDefaultsList
...
 genericDynamicArgs
     : '<:' typeRef (',' typeRef)* ':>'
     ;
 
 attributes
-    : attributesGroup*
-    ;
-
-attributesGroup
     : AttributeStart (identifier ':')? attribute (',' attribute)* ']'
```
studentmain, line 167:
```php
#[Attr1]
#[Attr2]
function f(){}
```
works, https://3v4l.org/8tEvX

studentmain (changes requested):

php/PhpParser.g4
KvanTTT: Thanks, I fixed it and extend attribute.php test file.

studentmain (approved):

Issue #1964, opened 2020-12-16
[JSX]Comment in the render block should be put into hidden channel.


Markz666:
```javascript
import React from 'react';
class SummaryContainer extends React.Component {
    render() {
        return (
            <div>
                <p>content</p>
                {/*comment*/}
            </div>
        );
    }
}
```
add the following code to  https://github.com/antlr/grammars-v4/blob/master/javascript/jsx/JavaScriptLexer.g4
```
JsxComment:                     '{/*' .*? '*/}'           -> channel(HIDDEN);
```

Markz666:
Already merged, https://github.com/antlr/grammars-v4/commit/76bcb019b14998d0fdaf3afffd19de67dd5ec2b2

PR #1967, opened 2020-12-17
put jsx comment into hidden channel

Markz666:
see https://github.com/antlr/grammars-v4/issues/1964


teverett:
@Markz666  thanks!

Issue #1526, opened 2019-10-13
can this repository add the  grammar for the ABB Robot Language 


heroinxj:
hello,I find the  grammar for the KUKA robot language in this repository and I need generate the AST of ABB Robot language,so I wish this repository can add the RAPID(ABB robot use) language grammer.THANKS!

teverett:
https://en.wikipedia.org/wiki/RAPID

heroinxj:
> https://en.wikipedia.org/wiki/RAPID

heroinxj:
> https://en.wikipedia.org/wiki/RAPID

so what do you mean

loonydev:
It's just link for somebody who want to add this language.

MSE1188:
[abbParser.zip](https://github.com/antlr/grammars-v4/files/5034505/abbParser.zip)

Hello I attached ABB Parser files, which I recently started to make. They are derived from the kuka grammar [https://github.com/antlr/grammars-v4/tree/master/kuka](url) and are still in progress.

In the current state you are able to parse variable declarations and assignments from ABB RAPID language files.

I am pretty new in ANTLR so sorry, if the grammar is not perfect. Maybe someone can expand the grammar and make a branch and pull request from it.

heroinxj:
> [abbParser.zip](https://github.com/antlr/grammars-v4/files/5034505/abbParser.zip)
> 
> Hello I attached ABB Parser files, which I recently started to make. They are derived from the kuka grammar [https://github.com/antlr/grammars-v4/tree/master/kuka](url) and are still in progress.
> 
> In the current state you are able to parse variable declarations and assignments from ABB RAPID language files.
> 
> I am pretty new in ANTLR so sorry, if the grammar is not perfect. Maybe someone can expand the grammar and make a branch and pull request from it.

heroinxj:
> [abbParser.zip](https://github.com/antlr/grammars-v4/files/5034505/abbParser.zip)
> 
> Hello I attached ABB Parser files, which I recently started to make. They are derived from the kuka grammar [https://github.com/antlr/grammars-v4/tree/master/kuka](url) and are still in progress.
> 
> In the current state you are able to parse variable declarations and assignments from ABB RAPID language files.
> 
> I am pretty new in ANTLR so sorry, if the grammar is not perfect. Maybe someone can expand the grammar and make a branch and pull request from it.

Hello, I would like to know if you have any new progress in ABB grammar file recently, thanks.

[...]

PR #1984, opened 2021-01-08
add abb parser files

MSE1188:
Added ANTLR parser files for abb rapid language.
Reverse Engineered from abb-sys (machine-data) files to parse variable definitions.

Added a short example of a robdata.sys file

References #1526


teverett:
Thanks!

Issue #1976, opened 2021-01-02
C# isn't in CI builds, sql/plsql for C# not targeting NET Standard build of Antlr


kaby76:
It looks like [sql/plsql](https://github.com/antlr/grammars-v4/tree/master/sql/plsql) for C# has a few problems:

* [PlSqlLexerBase.cs]() is written against a specific namespace, while [PlSqlLexerBase.java]() over in the Java side is not. This is inconsistent.
* The C# code is targeted for Harwell's version of Antlr C#--**which hasn't been updated for years**, and isn't supported by the official Net Standard version of Antlr 4.9. Fortunately, I have a shim to fix this.
* I don't see any Appveyor.yml at the top-level directory for the repository. It's likely that CI isn't being done at all for C# targets. This really should be fixed. People check in all sorts of crap.

studentmain:
AFAIK CI is only running with Java runtime (some times we even skipped some grammar). BTW we use GitHub Actions for CI.

kaby76:
Yes, I see the logs [here](https://github.com/antlr/grammars-v4/actions). There are no rules for a Dotnet build and test. It's likely that there is no CI build for C# because there is no "official" MSBuild rule file for C# Antlr programs. Fortunately, I also have a fix for this that's based on Harwell's Antlr4BuildTasks. I forked that code and have been cleaning up for years. But, it is very frustrating to see this go on year after year after year.

kaby76:
Based on the absence of a build for C#, I am enumerating the design requires for CI of the C# target here.

1) The CI for C# must run with Github Actions. G.A. supports any OS environment, but the current CI tests for grammars-v4 test [all environments](https://github.com/antlr/grammars-v4/blob/master/.github/workflows/main.yml#L13). However, as I don't want to write a mess of tools, I will require the Bash toolchain for scripting, and will require xml2 to extract information from a pom.xml file, which is at the center of the current CI testing.
2) C# is platform independent with NET 5. It works on Linux and Windows. It is installed in Github Actions. Therefore, C# CI will use the NET 5 SDK.
[...]

PR #1989, opened 2021-01-17
Add CI of C# target

kaby76:
This PR is a partial fix of https://github.com/antlr/grammars-v4/issues/1976. It includes three file changes: main.yml for Github Actions, .gitignore, and the main testing script csharp-regression-tests.sh.

* In main.yml, I tightened up the irrelevant and confusing blank lines in job "build". I added a new job, "build-csharp", to perform a build and test of the C# target on Ubuntu. Note, it takes 20m runtime for this new job. Due to the time required to run the C# tests, and MS's/Github rules on total CI time per month, and the fact that the test script requires Bash, xml2, etc, I only run it on Ubuntu. I have configured the test script to skip about 70 grammars because they don't build or run yet, but the rest look excellent.
* I added a new Bash script at the top-level directory in the repo, csharp-regression-tests.sh. This script examines all the pom.xml files and recursively goes through and tests grammars, skipping all grammars marked as such. By default, if you include a Java test for the grammar, this script will also test the grammar for the C# target. Tests use NET SDK 5, which works everywhere. The drivers for each grammar are created by a tool in [Antlr4BuildTasks](https://github.com/kaby76/Antlr4BuildTasks), which is a fork of Harwell's excellent tool, but retargeted for Antlr NET Standard 2.0 and the Antlr Java tool. See the design requires in [Issue 1976](https://github.com/antlr/grammars-v4/issues/1976) for the reasoning involved in the CI implementation.
* I updated the .gitignore to not check in the generated files from the C# build script since these files clutter up the grammars with code that's unnecessary for most people. The test script removes these temporary directories afterwards, but the update to .gitignore is included just in case, and to remove confusion when people go to make changes to the repo.

This is a partial fix because I will need to make further changes for the major grammars that are skipped (cpp, antlr, java, etc). But, I think the changes so far are a great addition to the repo.

--Ken Domino


studentmain (approved):

kaby76:
It looks like the grammar [idl](https://github.com/antlr/grammars-v4/tree/master/idl) does not parse identically between Java and C#. It's a combined grammar, with none of the usual overt trip-ups such as newlines in files between different operating systems. The tokens recognized by the lexer are identical between C# and Java, which means the issue is in the parser. Investigating. But, this is exactly why tests have to be run not just on Java.

Also, I don't know why but changes to my forked copy of the repo are being propagated to this PR. That's really, really annoying, and really bad. When I make a PR, Git should note the exact version to merge, not additional crap I check into my forked repo. Geesus.

kaby76:
The parse of idl/examples/helloworld.idl does not work because the parser lookahead sets are different between Java and C#. There's a bug in one of the runtimes, likely C#. I'll submit an issue against the runtime once I figure out which LA set is correct and why the LA sets are different between the two runtimes.

In the meanwhile, please integrate this PR so I can nuke the repo, re-fork, and continue. Ty. --Ken

studentmain:
> In the meanwhile, please integrate this PR so I can nuke the repo, re-fork, and continue. Ty. --Ken

GitHub PR (Git has no PR. You know, the tradition is sending patch file via email) is branch to branch, so your change to branch will affect PR. Just create a new branch.

I suggest merge this PR now, test framework itself is ready and works well, that's enough. Making it perfect and pass all tests is another task.

kaby76:
Yes, that is my plan...get in basic C# regression testing, then add in updates for the skipped grammars. I opened https://github.com/antlr/antlr4/issues/3042 for grammar idl. I debugged it to a diff in the lookahead set computation between C# and Java. It'll likely, though, take months--and thus not one additional PR--to fix all the skipped grammars.

teverett:
@kaby76 thanks

teverett:
@kaby76 I see some build errors in the C# tests.  Do you mind taking a look?

kaby76:
Looks like `Grammars that failed:  kotlin/kotlin`. The script worked, which is good! I thought I added kotlin to the "skip list", but I'll take a closer look....

kaby76:
Right. @gfour added `@header {  package org.antlr.grammars; }`. Remember I pointed out that by default changes to grammars that currently pass in C# are going to be checked again if it builds in C#. The change @gfour made is Java code because it's a Java package statement. The grammar is now specific for Java. Of course, it's not going to build for C#. I think for now just add "kotlin/kotlin" to the skip list [here](https://github.com/antlr/grammars-v4/blob/master/csharp-regression-tests.sh#L27) after "kotlin/kotlin-formal", or removed the `@header`. I thought a pom.xml entry for [package](https://github.com/antlr/grammars-v4/commit/81e0bbab59a052afc98d2cfb6cb7a45cc9ea7e23#diff-462c52a697a70b6eb09b12d873b1c710ec9720121bf5b91c9e5416c07c210d68R47) causes the Antlr parser generator tool to use a "-package" option., which would generate the parser with the package statement.

teverett:
Thanks!

PR #1993, opened 2021-01-22
Bring ANTLR4 grammar closer to spec

EwoutWal:
There are two main differences: moving where doc comments are processed to the lexer (in the same way any other comment is handled), and updating the LexerAdaptor.cs implementation to match the Java reference exactly.

There is one prevailing issue, when a rule has the same name as one of the reserved keywords (and the parser thinks that it has left the preamble) where ANTLR 4.9.1 accepts the input, but this grammar does not. At least, that is the case on the C# implementation now.


teverett:
@EwoutWal  thanks

PR #1995, opened 2021-01-25
Update mdx.g4

JRandallPierce:
The comma is not mandatory, and you cannot have multiple properties after the comma.  Rather, the statement can have multiple pairs of a comma and cell_property after the first cell_property


teverett:
@JRandallPierce  thanks

PR #1997, opened 2021-01-27
Use package for Kotlin parsers.

gfour:
This PR puts the Kotlin parsers ("kotlin"/"kotlin-formal") in a proper Java package ("org.antlr.grammars"). Currently, the parsers use the default Java package and this makes the parsers difficult to call from Java code in packages.


teverett:
@gfour  thanks

gfour:
I can confirm that removing both `@ header` blocks and using the "<arguments>...</arguments>" option in pom.xml still works (i.e. creates a JAR with a proper package).

PR #2002, opened 2021-02-01
Fix Kotlin parser packaging that would break C# builds.

gfour:
This PR fixes PR #1997 which would break C# builds for Kotlin (#1989). Now, Kotlin parsers use the "-package" option in pom.xml, as suggested by @kaby76.

On the C# front, the two Kotlin grammars seem to behave differently when running `bash ../../csharp-regression-tests.sh` in their respective subdirectories but the issues seem unrelated to this PR:
* **kotlin-formal** fails because of the dash in the generated code (`namespace kotlin-formal` in three .cs files). Replacing it with `namespace kotlinFormal` makes the tests pass.
* **kotlin** fails with a message possibly related to the way csharp-regression-tests.sh parses multiple tests in a pom.xml (`find: ‘examples/kotlinFile\nexamples/script’: No such file or directory`).

### DIFF
--- a/kotlin/kotlin-formal/KotlinParser.g4
+++ b/kotlin/kotlin-formal/KotlinParser.g4
@@ -12,10 +12,6 @@
 
 parser grammar KotlinParser;
 
-@ header {
-    package org.antlr.grammars;
-}
-
 options { tokenVocab = KotlinLexer; }
 
 kotlinFile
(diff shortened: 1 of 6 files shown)


kaby76:
Thanks! A dash in the namespace name is not correct. It gets passed directly to the tool to generate the C# program. I'll fix that immediately, but this PR is fine since they're currently bypassed now. I'll submit an issue and PR for adjusting the C# build script for Kotlin as well.

teverett:
@gfour  thanks

Issue #2009, opened 2021-02-05
Kotlin: Failed to parse class with constructor without 'block' statement 


heRHOX:
When I use the kotlin grammar and try parsing the following files, it fails/returns wrong output.

This file fails.
```
internal class NoBody() {
    constructor(secondaryConstructor: Any) : this()
}
```

This file gives me a wrong output (`enumClassBody `instead of `classBody`).
```
internal class NoConstructorDelegationCallAndNoBody {
    constructor(secondaryConstructor: Any)
}
```

PR #2010, opened 2021-02-05
Fix parsing classes with constructor without block statement

heRHOX:


heRHOX:
Fixes #2009

teverett:
@heRHOX  thanks

Issue #2003, opened 2021-02-03
Antlr tool generation of JavaScript with target language JavaScript broken


kaby76:
I'm trying to now generate JavaScript targets with my universal Antlr program driver generator and looking at grammars-v4/javascript.javasript/*.g4. It fails on the parser generator:

    $ java -jar ~/Downloads/antlr-4.8-complete.jar -Dlanguage=JavaScript *.g4
    error(134): JavaScriptParser.g4:501:0: symbol let conflicts with generated code in target language or runtime
    error(134): JavaScriptParser.g4:158:6: symbol let conflicts with generated code in target language or runtime
    error(134): JavaScriptParser.g4:487:6: symbol let conflicts with generated code in target language or runtime

This change broke the JS target:

https://github.com/antlr/grammars-v4/commit/7e489c90d0cdbab7c69f5375e816bf9371477735

The workaround is to rename the non-terminal to something other than the keyword `let`.

KvanTTT:
Yes, it makes sense to rename to `let_`

kaby76:
I found several other cases where there are conflicts in the grammar with JavaScript keywords, e.g., basic/jvmBasic.g4 with "var". I will be making a PR for these grammars and close out this bug. I now have a script that tests all grammars with the JS target. I am happy to report most pass! In a separate PR, I will add the script, and place both the test scripts I wrote, for JS and C# targets, under a directory grammars-v4/scripts/. NB!! I won't be adding the JS test to CI because it takes almost an hour to run! The scripts need to test only what changed from the last CI.

PR #2013, opened 2021-02-06
Fixes for Issue #2003.

kaby76:
This change renames conflicting parser symbols in grammars for JavaScript, i.e., var, function, let. I have also included a script to test the grammars against the JavaScript target, and moved the C# target tests to _scripts/. All tests not skipped in the js-regression-tests.sh pass, but the script has not been added to CI because the JS tests take about an hour to run. https://github.com/antlr/grammars-v4/issues/2003


teverett:
@kaby76  thanks

Issue #2016, opened 2021-02-07
.gitignore for VS IDE temps is probably wrong


kaby76:
I happened to run VS2019 via `devenv .` on one of the grammar sub-directories, and noticed that `git status` indicates unstaged files under the `.vs` directory created by VS2019. If there is a requirement (the [wiki](https://github.com/antlr/grammars-v4/wiki) page is more of a Q+A page than a requirements document) to ignore all editor and IDE temps created in the source tree, then the rules in [.gitignore](https://github.com/antlr/grammars-v4/blob/master/.gitignore) for VS IDE temporaries, and specifically "/.vs", are wrong. Scanning the history, it goes back to https://github.com/antlr/grammars-v4/commit/00496dba334c71077d97380030ad3c4a92b6e227#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947

Instead of:

    /.vs

(a file that ends in .vs)

Replace with:

    .vs/

(a directory that is named .vs, and keep any files that end in .vs around)

Looking at what the general [gitingore master for VS](https://github.com/github/gitignore/blob/master/VisualStudio.gitignore) lists, there are maybe other things that should be added.

But, in any case, I think `/.vs` is one that looks wrong.

kaby76:
Fixed.

PR #2019, opened 2021-02-10
Fix for Issue #2016

kaby76:
VS2019 creates and stores editor configurations in .vs/. The .gitignore file is not correct because after running VS2019 from a sub-directory in the repo, git indicates unstaged files in the .vs/ directory. Further, there are no '.vs' files within the repo--only .vs/ directories created by VS2019. Since the .gitignore comments indicate to ignore VS2019 temps, as well as other IDEs, the fix is to change "/.vs" to ".vs/". https://github.com/antlr/grammars-v4/issues/2016

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,7 @@
 # Visual Studio Generated
 .antlr
 .fetch_time_cache
-/.vs
+.vs/
 .vscode/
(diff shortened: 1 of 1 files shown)


teverett:
@kaby76  thanks

Issue #2007, opened 2021-02-04
Using javascript grammar in Antlr4 and get Maximum call stack size exceeded error


dmaon:
I used antlr version 4.8 and javascript antlr4 runtime version 4.8. grammar(JavaScriptLexer.g4, JavaScriptParser.g4) and JavaScriptLexerBase.js, JavaScriptParserBase.js and i face error 'Maximum call stack size exceeded'.
[https://stackoverflow.com/questions/66031981/using-javascript-grammar-in-antlr4-and-get-maximum-call-stack-size-exceeded-erro](url)
Unfortunately, the structure of the files[JavaScriptLexerBase.js, JavaScriptParserBase.js in official grammar of antlr4] for the syntax is pre - version 4.9 and is not updated. so i can't use antlr4.9.1 with runtime4.9.1 and gets another error.

kaby76:
I rewrote the base classes for the parser and lexer, but I have verified that the bug exists in Antlr 4.9.1 with a Node environment. The stack overflows in AdaptivePredict(). I will continue debugging, but it's likely there is a bug in the Antlr4 runtime, hence another demonstration of why all grammars should be checked against all targets.

kaby76:
I am debugging side by side "i = 1;" between C# and JS, stopping at the first call to ParserATNSimulator.ComputeTargetState, then in predicteDFAState(). For sure, the JS runtime is not computing the DFA states correctly. In ParserATNSimulator.predictDFAState, at one point decisionState has in JS `{<EOF>, 1, 4..5, 7, 9, 11, 18..23, 59..70, 71, 73, 76..77, 80..82, 83..84, 85..87, 87..89, 91..92, 93, 95, 98, 101..105, 104..106, 108..109, 116..119}`, while in the same code in C#, it's `{{<EOF>, 1, 4..5, 7, 9, 11, 18..23, 59..71, 73, 76..77, 80..89, 91..93, 95, 98, 101..106, 108..109, 116..119}}`. You will note that one range overlaps another (`101..105` and `104..106`). The ranges in JS are not coallesced as well, which they are done in C#. There is a lot wrong here with the JS runtime!

kaby76:
I am raising this in Antlr4 issues, where it really belongs.

kaby76:
[...]

Issue #1580, opened 2019-12-11
JavaScript grammar for JavaScript target fails on tree build


aherlihy:
Hi, I'm trying to build and visit a JavaScript (ECMAScript 6+) tree in the JavaScript target, but the tree-building step hangs and eventually fails with "Maximum call stack size exceeded". I have it working for the [ECMAScript](https://github.com/antlr/grammars-v4/blob/master/ecmascript/JavaScript/ECMAScript.g4) and [Python3](https://github.com/antlr/grammars-v4/blob/master/python3-js/Python3.g4) grammars (which is why I'm not using Esprima or another JS parser since I need multi-language support).

I build via:

`java -Xmx500M -cp '/usr/local/lib/antlr-4.7.2-complete.jar:$CLASSPATH' org.antlr.v4.Tool  -Dlanguage=JavaScript -lib antlr -o antlr -no-listener -visitor -Xexact-output-dir antlr/JavaScriptLexer.g4`

then

`java -Xmx500M -cp '/usr/local/lib/antlr-4.7.2-complete.jar:$CLASSPATH' org.antlr.v4.Tool  -Dlanguage=JavaScript -lib antlr -o antlr -no-listener -visitor -Xexact-output-dir antlr/JavaScriptParser.g4`

Which seems to generate the correct files (JavaScriptParser.interp, JavaScriptLexer.interp, JavaScriptLexer.tokens, JavaScriptParser.tokens, JavaScriptParser.js, JavaScriptLexer.js, JavaScriptParserVisitor.js) into the same directory with [JavaScriptBaseParser.js](https://github.com/antlr/grammars-v4/blob/master/javascript/JavaScript/JavaScriptBaseParser.js) and [JavaScriptBaseLexer.js](https://github.com/antlr/grammars-v4/blob/master/javascript/JavaScript/JavaScriptBaseLexer.js).

I try to construct the tree via:

```js
const JavaScriptLexer = require('../antlr/JavaScriptLexer.js').JavaScriptLexer;
const JavaScriptParser = require('../antlr/JavaScriptParser.js').JavaScriptParser;
const JavaScriptVisitor = require('../antlr/JavaScriptParserVisitor.js').JavaScriptParserVisitor;

const input = "1";
const chars = new antlr4.InputStream(input);
const lexer = new JavaScriptLexer(chars);
lexer.strictMode = false;
const tokens = new antlr4.CommonTokenStream(lexer);
const parser = new JavaScriptParser(tokens);
[...]

PR #2020, opened 2021-02-13
Fix for issue #2007. 

kaby76:
Duplicate of #1580. This fix works, but does require an Antlr4 runtime fix https://github.com/antlr/antlr4/pull/3077 (fix to IntervalSet) and https://github.com/antlr/antlr4/pull/3080 (fix to JS semantic predicates, otherwise the predicates are never ever called).

Summary of changes:

* I included updates to the C# code because there are actually two Antlr C# generators and runtimes.
* CSharp/JavaScript{Parser,Lexer}Base.cs targets the Antlr4.Runtime.Standard 4.9.1 Nuget package runtime. This is the so-called "Antlr4" C# target, the code at github.com/antlr/antlr4. Use the standard Antlr4 Tool in Java to generate the parsers for this target.
* Antlr4cs/JavaScript{Parser,Lexer}Base.cs targets Harwell's Antlr4, Antlr4.Runtime 4.6.6 Nuget package runtime. This is the so-called "Alternative C#" runtime.
* The JavaScript/*.js source code has been updated to ES6 because the Antlr 4.9 JavaScript target now outputs ES6.
* All targets assume NO package, and should remain so. I have a template driver app that outputs source to build an app, but it uses the pom.xml file.  Do not change pom.xml with the addition of a -package option, nor add namespace declarations to the grammar or base classes.
* I moved the .bat and support files to _scripts/. This is support code, irrelevant to the grammar and base classes.
* The Java/*.java code has been "untabified". Tabs are evil and should never be used because the indentation level is dependent on your editor.
* In README.md, I corrected some grammar, and removed a couple sections on semantic predicates, which is much better explained in antlr.org.

I have validated the Java, CSharp, and JavaScript targets. I have not checked the other language targets.

--Ken


teverett:
@kaby76  thanks

PR #2022, opened 2021-02-15
Port supporting code to C# for Rust parser.

gfour:
This PR adds C# support for the Rust parser (based on `RustLexerBase.java`/`RustParserBase.java`). The C# CI script has been edited to detect and include the extra base classes (@kaby76 is this a good way to include extra .cs files?).


kaby76:
> The C# CI script has been edited to detect and include the extra base classes (@kaby76 is this a good way to include extra .cs files?).

I think with this version of dotnet-antlr, you can place your extra .cs files in a directory called CSharp/ and leave the .sh file alone. Or, you can roll this change in.

The placement of extra source files I was thinking of would be modeled after grammars-v4/javascript/javascript. There isn't any coding standard per se, which is why everyone is coming up with their own solution.

The current version of dotnet-antlr program looks in the CSharp/ directory for additional .cs. It copies all those files to Generated/. Since I use the dotnet-antlr tool to generate drivers for grammars outside of this repo, the copying of extra source files really should be in dotnet-antlr, not the bash script in this repo for testing C#.

I've been updating the dotnet-antlr program quite a bit though with additional targets and options. The next version I am working on relaxes even further where to place extra source files. Everyone has a different way of organizing the source.

--Ken

gfour:
@kaby76 Thank you for the clarification. I had tried using the "CSharp/" directory but it wasn't automatically included. The problem was dotnet-antlr being too old (1.1.0), I tried with 1.4.0 and now it works. I have updated the PR accordingly.

kaby76 (approved):
Yes, that should work--I hope!

studentmain (approved):

teverett:
@gfour  thanks

PR #2026, opened 2021-02-18
[scss] Support pseudo-classes appearing anywhere in the selector.

nickwalther:
Pseudo-classes are allowed anywhere in selectors while pseudo-elements may only be appended after the last simple selector of the selector.
Double colons were added to the spec to distinguish pseudo-elements from
single colon pseudo-classes, however for backwards compatibilty many
pseudo-elements work with single colons. So our parser wont distinguish
the pseudo type or restrict the positioning.

Spec: https://www.w3.org/TR/CSS21/selector.html#pseudo-elements


nickwalther:
Merge conflicts resolved

teverett:
@nickwalther  thanks

PR #2027, opened 2021-02-18
[scss] Remove Nullable and NotNull annotations from tests.

nickwalther:
This is for compatability with newer versions of antlr.
They were removed from antlr here:
https://github.com/tunnelvisionlabs/antlr4/commit/c84f024676c26ca00e9516599e1b643085571132


teverett:
@nickwalther  thanks

PR #2028, opened 2021-02-19
[scss] Support stand-alone attribute selectors

nickwalther:
`[disabled]` is interpreted as `*[disabled]`, the * isn't required.


teverett:
@nickwalther  thanks

PR #2029, opened 2021-02-19
[scss] Support pseudo-classes with selector parameters.

nickwalther:
E.g. `button:not([disabled])`
At the moment, only value parameters can be parsed.


teverett:
@nickwalther  thanks!

PR #2031, opened 2021-02-20
Hive: fix redundant "DEC" reserved keyword

wyruweso:
It seems that "DEC" should be removed from the keywords. According to the [documentation](https://cwiki.apache.org/confluence/display/Hive/LanguageManual+DDL) there is no such reserved keyword.
Using this type as a synonym to "decimal" cause an error:
Error: Error while compiling statement: FAILED: ParseException line 12:4 cannot recognize input near 'dec' '(' '12' in column type (state=42000,code=40000)


teverett:
@wyruweso thanks

PR #2033, opened 2021-02-21
[Hive] Add grammar for 2.x.x version

wyruweso:
Here is an implementation of Hive 2.3.8 grammar based on the existing 3.1.2 one. I've made it because the 3.0.0 version had breaking changes (ex.: CREATE/ALTER/DROP INDEX statements [were removed](https://issues.apache.org/jira/browse/HIVE-18448)). 

Also, I've added a couple of changes:
1. The root rule parses multiple statements
2. Grammar is case-insensitive as it should be for SQL-like languages

I've left examples the same for both grammars and add the new one.


teverett:
@KvanTTT  good to pull this?

@wyruweso  thanks!

teverett:
@wyruweso thanks

Issue #2032, opened 2021-02-21
Antlr4-maven-plugin does not use "<grammars>" in pom.xml


kaby76:
I have been updating my tool, which generates target-specified driver code for Antlr grammars, to directly read the Maven pom.xml files. I've added code to check for the existence of the grammar files referenced in the pom's, and I've noticed errors in several pom's that contain references to files that do not exist:

    <grammars>fortran77.g4</grammars>
    <grammars>Golang.g4</grammars>
    <grammars>Java8.g4</grammars>
    <grammars>PromQL.g4</grammars>
    <grammars>upnp.g4</grammars>

I then started to wonder how the Antlr4-maven-plugin actually works. After reading the doc and code, I found that the pom files in this repo are wrong: the `<grammars>` element is never used!

In fact, the antlr4-maven-plugin only uses the ["`<includes>` and `<excludes>`"](https://maven.apache.org/plugins/maven-resources-plugin/examples/include-exclude.html) method, with defaults made by the antlr4-maven-plugin, to tell what is included and excluded in a build.

It's misleading to reference non-existent files, let alone use the ignored `<grammars>` element. All pom.xml files should be cleaned up:
`<grammars>` elements should be replaced with `<includes>`

--Ken

PR #2034, opened 2021-02-21
Fix for #2032

kaby76:
This is a collection of changes to pom.xml files. Replaced are the "grammars" elements with "includes" elements, because "grammars" elements are ignored by the Antlr4 tool plugin. I did not change the Mysql pom--it's not in the CI tests, and that will be a separate fix. Believe it or not, "python/tiny-python" was never pushed through the Antlr Tool--"No grammars to process". That's because the sourceDirectory was not set correctly. Geezus folks!

One could actually remove the "includes" elements from most of the grammars. But, since the pom's already list the grammar files, and since it helps to remove any ambiguity of what is actually processed by the Antlr tool, I left them in.

### DIFF
--- a/abb/pom.xml
+++ b/abb/pom.xml
@@ -18,7 +18,10 @@
 				<version>${antlr.version}</version>
 				<configuration>
 					<sourceDirectory>${basedir}</sourceDirectory>
-					<grammars>abbParser.g4</grammars>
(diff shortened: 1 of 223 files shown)


teverett:
@kaby76  thanks

Issue #2035, opened 2021-02-22
Go language grammar not tested


kaby76:
The Go language grammar is not being tested, removed from the CI tests on 8 Sep 2019 with [this commit](https://github.com/antlr/grammars-v4/commit/cd077a9c01e1e08d53758d9f9e6843714de12b92#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8), in response to https://github.com/antlr/grammars-v4/issues/1484.

The issue was corrected, as far as I can tell, but the test was never reinserted into the [top-level pom](https://github.com/antlr/grammars-v4/blob/master/pom.xml). And, [golang/pom.xml](https://github.com/antlr/grammars-v4/blob/master/golang/pom.xml) was never fixed to reflect the correct language name, among other things.

There are three ways I got this test to run cleanly, and probably several more available. One thing is clear, though, and that is to not place explicit Java code in your .g4's. If you do, the grammar becomes target-dependent. Go*.g4 contain "package" declarations. You can do this in a target-independent way using the "-package" option. See [antlr/pom.xml](https://github.com/antlr/grammars-v4/blob/master/antlr/antlr4/pom.xml). Or, you can make do without a package, which is my recommendation. See [haskell/pom.xml](https://github.com/antlr/grammars-v4/blob/master/haskell/pom.xml), or better [rust/pom.xml](https://github.com/antlr/grammars-v4/blob/master/rust/pom.xml).

As far as where to place the base classes for the parser and lexer, you can place them with the pom.xml/*.g4's, or place the files in separate sub-directories named after each target. In either case, you need to indicate in the pom.xml where to pick those sources up--which was also not updated in golang/pom.xml. However, my recommendation is to place the source under a sub-directory named after the target, e.g., CSharp, Java, JavaScript, etc. Note, since the grammars *can* be used with Antlr4cs, I recommend any .cs specific for that runtime placed under Antlr4cs, not CSharp, or all with the .g4's, so as to not confuse people and the compiler.

PR #2036, opened 2021-02-23
Fix for #2035

kaby76:
This fixes the CI testing of the Go grammar, adding Java and C# testing. Changes in this commit:

* removed package declarations;
* fixed pom.xml to copy sources from Java/ for "maven test";
* fixed grammar name ("Golang" => "Go" since the grammars were changed long ago);
* split Antlr4cs and CSharp target source code so they can be independent;
* add golang back to top-level pom.

All `examples/*.go` parse.

--Ken


teverett:
@kaby76 thanks

PR #2045, opened 2021-02-24
[scss] Support combinators at start of selectors.

nickwalther:
This is possible with nesting.
Previously known as 'selector prefix'. Combinator is official name.


teverett:
@nickwalther  thanks

PR #2046, opened 2021-02-25
[scss] "Modernize" scss test suite

nickwalther:
- Bump antlr version
- Remove deprecated stream creation method
- Add DiagnosticErrorListener for easier debugging
- Alll assertEquals calls had their arguments backwards. Instead of flipping everything, replace with assertThat().is...
- Fix 2 tests that were broken
- Remove dead code
- Fix lint errors
- and a little auto-formatting xD

### DIFF
--- a/scss/build.gradle
+++ b/scss/build.gradle
@@ -44,29 +44,16 @@ repositories {
 }
 
 dependencies {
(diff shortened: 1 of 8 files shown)


teverett:
@nickwalther  thanks

PR #2048, opened 2021-02-26
[scss] Add media query support up to level 3.

nickwalther:
This goes up to Media Queries Level 3, so doesn't include Level 4 sytnax such as range context.
Previously, only media types worked.


teverett:
@nickwalther  thanks

PR #2049, opened 2021-02-26
[scss] Add a separate token for !important.

nickwalther:
Currently it is being parsed in the values rule, except the '!' is being dropped.
Add a test for !default too.


nickwalther:
Merge conflicts resolved

teverett:
@nickwalther  thanks

PR #2054, opened 2021-02-27
Fixed #2053 - Define version / default configuration of maven-compile…

### LINKED ISSUE #2053: Define version / default configuration of maven-compiler-plugin only once
I have found several configurations for maven-compiler-plugin for example like `python/tiny-python/pom.xml` where the configuration looks like this:
```xml
<plugin>
	<groupId>org.apache.maven.plugins</groupId>
	<artifactId>maven-compiler-plugin</artifactId>
        <configuration>
                <source>11</source>
                <target>11</target>
        </configuration>       
	<executions>
		<execution>
			<id>default-compile</id>
			<phase>none</phase>
		</execution>
	</executions>
</plugin>
```
which does not makes sense because the compilation is turned off but using JDK 11 as source/target whereas the rest of the build is using JDK 7.

Furthermore the configuration should be done only at a single location.

So defining the version of the maven-compiler-plugin only on the root level of the project as well as the target/source level for the used JDK.

Also found different property definitions which are related (based on the name) to defining the source/target level for the JDK...

khmarbaise:
…r-plugin only once

### DIFF
--- a/_grammar-test/pom.xml
+++ b/_grammar-test/pom.xml
@@ -10,9 +10,6 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <java.version>1.8</java.version>
-        <java.version.source>1.8</java.version.source>
-        <java.version.target>1.8</java.version.target>
         <maven.compiler.source>1.8</maven.compiler.source>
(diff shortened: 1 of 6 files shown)


teverett:
@khmarbaise  thanks

PR #2055, opened 2021-02-28
[scss] Support nested property declarations.

nickwalther:
e.g.
```
.info-page {
  margin: auto {
    bottom: 10px;
    top: 2px;
  }
}
```
Also make sure that pseudo-classes/elements don't parse with spaces between the colon and the name.


teverett:
@nickwalther  thanks

Issue #2037, opened 2021-02-24
Antlr/antlr4 CSharp incorrectly uses NextToken() not Emit(), and other issues


kaby76:
I have been looking at the antlr/antlr4 grammar, since it is currently skipped in CI for the C# target.
### Hacks in the C# target.
I found that for the C# target file [LexerAdaptor.cs](https://github.com/antlr/grammars-v4/blob/master/antlr/antlr4/LexerAdaptor.cs) was made using an override for NextToken() not Emit(), which how it is done for the Java target. The override of NextToken() goes back to the first version of the file with [this commit](https://github.com/antlr/grammars-v4/commit/aac9b8e870fd91cefebabc446e08a61d00968c6a#diff-ee72b1a5e29208cbc8ab7f510208a13faaf3c426b350a67b5c9c81b1e23ad5a0).  In response to https://github.com/antlr/grammars-v4/issues/1108, "to make a working version" of the parser for C#, [a hack was introduced](https://github.com/antlr/grammars-v4/commit/ccdcc46df3f5d77edabce61d071fbfe341641a47#diff-ee72b1a5e29208cbc8ab7f510208a13faaf3c426b350a67b5c9c81b1e23ad5a0) to mutate the type of a token in NextToken(), using low-level C# reflection, but a Token is intended to be read-only.

This is ridiculous. It should not have been done this way. Instead, follow what the Java code does, override Emit() and eliminate the hack.

### Missing targets for antlr2/3.
For the antlr2 and antl3 grammars, there are no implementations for the lexer base class other than for Java.
### Add antlr2/3/4 to CI for C#.
The C# versions for all antlr/* grammars should be integrated into the CI.
### Add JS target.
I have a version of LexerAdaptor for the JS target, and it works.
### Reset() missing.
There is a problem with the current Java and C# LexerAdaptor code: it does not override Reset(). This is required to reset the state of the base class. Otherwise, Issue1567.g4 does not reparse.

PR #2056, opened 2021-02-28
Issue 2037

kaby76:
This is a fix for #2037. The PR includes changes to the antlr/antlr[234] grammars and source code. The override of NextToken() is replaced with Emit()--the correct function to override. The hack that was introduced to mutate a token is removed because it's unnecessary in Emit(). I reorganized the files a bit to follow what was done with golang/. The package/namaspace from the grammar and base class source code was removed because it just clutters the Java source code--Java requires a directory structure corresponding to that of the package name, a really bad idea, but that's how Java works. The reorganized files are now easier to find, and much more flexible if one needs to use a target-specific grammar, such as with Python3. I updated the CI script to now include antlr/antlr[34]. I did not add antlr/antlr2 to C# CI because there appears to be a bug in the 4.9.1 generated parser and C# runtime. The CI script has been rewritten and renamed to take advantage of the latest version of the "dotnet-antlr" program, which generates target-specific driver code, and reads the pom.xml files.

--Ken


teverett:
@kaby76  thanks

PR #2058, opened 2021-02-28
[scss] Improve support for mixins and includes.

nickwalther:
Improvements such that all examples on this page are parseable
https://sass-lang.com/documentation/at-rules/mixin 

- Arguments with default values for mixins
- Named arguments and arbitrary arguments in includes
- Support @content blocks, with arguments using @using
- @for accepts 'to' as well as 'through'
- Numbers in @each and @for can be function calls and variables, possibly just 'commandStatement' or 'expression' but we'll get there eventually


teverett:
@nickwalther  thanks

Issue #2038, opened 2021-02-24
Algol60 grammar - arithmetic_expression rule is infinite loop


matevosmehrabyan:
Algol60 grammar's rule `arithmetic_expression` produces infinite loop:

`arithmetic_expression
    : 'step' arithmetic_expression 'until' arithmetic_expression
    | arithmetic_expression 'while' boolean_expression
;`

kaby76:
The Antlr tool doesn't check for this. It should.

kaby76:
I have a fix for this. I started from the ISO 1538 Spec, typing it in by hand, and adding various examples scrounged over the Internet. Details and PR at some point soon.

kaby76:
Please review the PR change for this bug. @matevosmehrabyan

matevosmehrabyan:
Hi, I added a comment on your PR.

PR #2059, opened 2021-03-02
Fix for #2038.

kaby76:
This fix is for #2038, an "infinite loop" in a grammar rule. It's clear the grammar was adapted from algol60.txt, with a misinterpretation of [line 60](https://github.com/antlr/grammars-v4/commit/8aa9a896a61d50454ce42d8948297db57bbe7fca#diff-bd2cff7c242dd151256cccc57887627e83acf0e2785ef3995f028808097a9c63R60). I tried to figure out the origin of the algol60.txt grammar and determine whether it was correct, but I couldn't definitively. It is likely that it was copied from a scanned and OCRed manual for an Algol 60 compiler Instead, I decided to start over with the ISO 1538 specification.

The new grammar is scraped visually from the ISO 1538 specification. The text underlying the PDF for the spec, which also appears to be OCR'ed, is wrong. The rules appear in the same order as in the spec, with comments for each section. This Antlr4 grammar has a more typical lexer/parser boundary for identifiers, numbers, etc., while still keeping mostly to the original BNF grammar in the spec.

Examples are been added. The old "example1.txt" was removed because it wasn't a valid Algol 60 program. It did not have a "begin" and "end" surrounding the snippet. The examples I included are from various sources on the Internet.

I added a readme.md to explain the origin of the grammar and examples.

I reviewed previous commits to the grammar. Some of the changes were optimizations to EBNF. Some changes, such as [this](https://github.com/antlr/grammars-v4/commit/79c2cf173fb63716b45d288c25bc515bc7510bd3#diff-d0f4576b38a32986984692e62fdeebd2d2c67764f5872447049644948c619523L380) didn't really fix the significant problem of the lexer/parser boundary in the original grammar. For optimizations, I suggest using a tool to perform the refactorings in a repeatable and automated way.

The examples have been added to the Maven Java and C# CI testing.

I should add that Algol W, derived from Algol 60, was one of the first languages I learned.

The grammar should be rechecked for "infinite loops".


kaby76:
@matevosmehrabyan please review.

matevosmehrabyan:
Hi, I have checked the grammar that you have modified. Now there is another infinite loop:
`fragment Open_string : Proper_string? (ULCorner Open_string URCorner)+ ;`

As you see, from `Open_string` you produce at least one `Open_string`, which produces infinite loop

kaby76:
@matevosmehrabyan  Thank you for the check. I'm not sure why I had to rewrite those rules at all, deviating from the spec. It's now fixed. I will open a Github Issue on Antlr to add the check.

matevosmehrabyan:
ok

teverett:
@kaby76  thanks

PR #2061, opened 2021-03-03
powershell based tester

studentmain:
I want add git diff support to test script, to bring JavaScript target into CI. But I can't write bash script, and I want to make it cross platform, so I decided to create a new one. PowerShell looks good to me, but I'm not sure if it is the best choice.


studentmain:
cc @kaby76 @teverett

kaby76:
I don't see a problem with this. If you want a script to call dotnet-antlr, that's fine. However, understand I'm still working on dotnet-antlr, so the latest is unstable. You should "dotnet tool uninstall -g dotnet-antlr" then "dotnet tool install -g dotnet-antlr --version 2.1.0" rather than work with the latest. (I think 2.1 should be fine.) Also, the generate code for JavaScript uses Make. So you will need to use WSL, or MSYS2 with mingw32-make.exe installed. Right now, dotnet-antlr doesn't generate much other than a makefile because that's the most basic "build" system I could think of. And, the makefile hardwires some things. I am planning to fix this after working on the rest of the targets, which is currently CSharp, Java, and JavaScript.

The JavaScript target isn't in the regression tests because (1) it takes an hour or two, and (2) there's a major bug in Antlr 4.9.1 for the JavaScript target that I found a couple weeks ago. So, dotnet-antlr generates code to use a hardwired, private version. I don't know if anyone has placed the latest 4.9.1+ code up on NPM. If you could fix that, I'll fix my code right now so you can use dotnet-antlr JavaScript.

I'm adding Python3 to the "dotnet-antlr" program right now, and I plan to add all others in the next few weeks. After I see how all targets work, I plan to rewrite the dotnet-antlr program to use StringTemplates because it's nasty WriteLine's right now. People should be able to use their own templates for driver source code and build systems.

dotnet-antlr in this repo works off the poms, but it can also work without the pom files. But, you do have to pass in some information via options to generate a good program. But, I do have Trash--a tool that reads Antlr4 grammars--and plan to add more code to figure out all the things the pom encodes from just your source files.

--Ken

kaby76:
Thinking about this more, it's probably better to use Powershell than Bash. Powershell is portable to other OSes and Bash has an annoying habit of not working with Windows CRLF end of lines.

kaby76:
I added Python3 and Dart to the dotnet-antlr generator. I'll be adding a Make file for CSharp because I figure people may not understand how to use "dotnet", or forget how to use the target-specific toolchain flipping between seven or more targets for Antlr. Make is an ancient, low, common denominator in calling the target-specific toolchain to build and run, and even set up one's environment with the toolchain. But, it's not necessary.

studentmain:
Interesting, when I try to redirect stderr, GitHub Actions exited...

studentmain:
It seems work now, I'll run a full test.

studentmain:
Technically this script can be used with Java target (with grun) to reduce test time as well.

kaby76:
On this end, this works, generating the Arithmetic grammar and tests. Antlr works!

    for i in CSharp Java JavaScript Dart Python3
    do
        echo $i
        rm -rf Generated
        dotnet-antlr -t $i
        pushd Generated
        ls
        make
        make run RUNARGS="-input 1+2 -tree"
        make clean
        ls
        popd
    done

teverett:
@studentmain  thanks

PR #2062, opened 2021-03-04
[scss] Add explicit parser rules for maps and lists.

nickwalther:
- Able to use maps outside of @each
- Nested lists and nested maps
- Function identifier namespaces
- Spaces between an identifier and lparen will always be parsed as a
function
- Move prefixed '-' and '+' to lexer rules to avoid ambiguity with list
- Explicit rules for list, entry, key, value, etc. for cleaner visitor/listeners


teverett:
@nickwalther  thanks

PR #2070, opened 2021-03-07
improve javascript family grammar

studentmain:
move typescript into jsparent
add missing pom.xml
use unicode category to replace hard coded character list


teverett:
@studentmain  thanks

Issue #2069, opened 2021-03-06
Symbol conflicts, missing target-specific code in Part 1 of N


kaby76:
Dotnet-antlr now generates parsers for CSharp, Java, JavaScript, Dart, Python3, and Go. The following Bash script is what I used to test the first set of grammars in the repo. The script has detected a number of problems in symbol conflicts, and I list these below. But, this is just the first grammars alphabetically in the repo, and I expect more. In the notes below, I list other issues that should be fixed.

Notes: (1) The dotnet-antlr version is 2.2.0. (2) The Go and Dart runtimes need work because there are missing APIs. (3) The runtime for the Dart target is slow. But when the program is compiled, it works better. (4) I have compared the six targets against grammars-v4/html, one of the slowest in the CI test suite. I have tested the first five examples in grammars-v4/html/examples, and have some interesting results. In general, Java is the fastest, Python3 the slowest, often times 4 times slower, sometimes 50 times slower. For Go, I had to kill the parse of "cnn1.html". It's likely that some of the runtimes implement some data structures that do not scale because they deviate non-linearly from Java for the larger tests like "cnn1.html". In other words, for example, it's not a constant "5 times slower" for Python3. (5) There are several bugs in 4.9.1 Antlr which are starting to become an issue in testing grammars-v4, e.g., idl and parse failures in C#, Python3, Go targets. Looks like we're going to get a new release soon!

    #
    shopt -s expand_aliases
    ls -l ~/.bash_profile
    if [ -f ~/.bash_profile ]
    then
        echo sourcing
        . ~/.bash_profile
    fi

    for i in CSharp Java JavaScript Dart Python3 Go
    do
        echo $i
        rm -rf Generated
        dotnet-antlr -t $i -m
        if [[ "$?" != "0" ]]
        then
            rm -rf Generated
            continue
        fi
        pushd Generated
        ls
        make
        if [[ "$?" != "0" ]]
        then
            rm -rf Generated
            continue
        fi
        if [[ -d ../examples ]]
        then
            for file in `find ../examples -type f`
            do
    # NOTE: Python3 has a tendency to just hang with compile errors.
    # So, we avoid reading from stdin for Python3, but may as well do for
    # others.
    #           cat $file | make run
                make run RUNARGS="-file $file"
            done
        else
            make run RUNARGS="-input 1+2 -tree"
        fi
        popd
        rm -rf Generated
    done

* abb/abbParser.g4, type - Python3
* algol60/algol60.g4, type
[...]

PR #2073, opened 2021-03-09
Changes for #2069

kaby76:
This PR fixes only a few parser symbol conflicts in grammars to not conflict with various targets (Python3, Dart, Go). This PR deals with just the grammars starting with the letter "a". I only fixed the symbol conflicts, and haven't rewritten the grammar for multiple targets. Folks can address this by creating a new issue to carve out what they want to work on, or comment on an old issue.

I am now working on the CI tool dotnet-antlr so that it works as much as possible with StringTemplates.

### DIFF
--- a/abb/abbParser.g4
+++ b/abb/abbParser.g4
@@ -78,10 +78,10 @@ functionParameter
     ;
 
 declaration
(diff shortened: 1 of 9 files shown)


teverett:
@kaby76  thanks

Issue #2064, opened 2021-03-04
bcpl grammar - mutually left-recursive rules [vectorE] and [multE] and [repeatable_command, repeated_command]


matevosmehrabyan:
ANTLR failed to generate lexer and parser for bcpl grammar, with the following error message:

`error(119): grammars-v4/bcpl/bcpl.g4::: The following sets of rules are mutually left-recursive [vectorE] and [multE] and [repeatable_command, repeated_command]`

ANTLR supports only direct left recursion.

kaby76:
I will be working on this. I will be starting from the original reference, the Bcpl book, which I have had since 1983, and had worked on. I will provide a concise example of how to start from the "spec" grammar, and apply transformations via my grammar tool Trash to remove MLR.

kaby76:
I have a PR for this. The three MLR were:

* vectorE. This rule contained a __useless grouping__ (i.e., parentheses) around an "alt", i.e., `vectorE : (vectorE '!' primaryE) | primaryE ;`. The grouping is unnecessary since the "|"-operator has lower priority than symbol concatenation.
* multE. This has the same problem as vectorE.
* repeatable_command/repeat_command. This MLR was in the original grammar. It's easily fixed with one unfold transform + one ungrouping.

--Ken

kaby76:
@matevosmehrabyan In your convenience, please check the new BCPL grammar ([grammars-v4/bcpl/bcpl.g4](https://github.com/antlr/grammars-v4/blob/master/bcpl/bcpl.g4)) that we updated. It should be free of MLR now. The within the source are two other grammars for reference: the original grammar with the MLR in "[grammars-v4/bcpl/ignore/bcpl.g4](https://github.com/antlr/grammars-v4/blob/master/bcpl/ignore/bcpl.g4)" and the grammar from Richands' BCPL book, which I re-entered by hand, in "[grammars-v4/bcpl/origbcpl.g4](https://github.com/antlr/grammars-v4/blob/master/bcpl/origbcpl.g4)". Those two grammars both contain MLR, but are not used. Thanks! --Ken

matevosmehrabyan:
I checked [grammars-v4/bcpl/bcpl.g4](https://github.com/antlr/grammars-v4/blob/master/bcpl/bcpl.g4), it doesn't contain MLR, Thanks!

PR #2079, opened 2021-03-14
Changes for #2064.

kaby76:
Replaced grammar for BCPL with grammar from "BCPL: the language and its compiler". Added Bash script that uses Trash to transform the grammar to remove MLR and define a new lexer/parser boundary. (Note, the file grammars-v4/bcpl/bcpl.g4 is the end-modified grammar.) Add bcpl/ (as well as antlr2 and algol60) to CI tests. Added examples/ and readme.md.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -53,4 +53,5 @@ jobs:
             $After = "${{ github.event.after }}"
         }
         dotnet tool install -g dotnet-antlr --version 3.0.6
+        [System.Environment]::SetEnvironmentVariable("Path", [System.Environment]::GetEnvironmentVariable('Path', [System.EnvironmentVariableTarget]::Machine) + ";$($env:HOME)\.dotnet\tools")
         _scripts/test.ps1 CSharp $Before $After
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -52,5 +52,5 @@ jobs:
             $Before = "${{ github.event.before }}"
             $After = "${{ github.event.after }}"
         }
-        dotnet tool install -g dotnet-antlr --version 3.0.5
+        dotnet tool install -g dotnet-antlr --version 3.0.6
         _scripts/test.ps1 CSharp $Before $After
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -35,7 +35,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [macos-latest, windows-latest]
     steps:
     - name: Checkout
       uses: actions/checkout@v2
@@ -53,5 +53,4 @@ jobs:
             $After = "${{ github.event.after }}"
         }
         dotnet tool install -g dotnet-antlr --version 3.0.6
-        [System.Environment]::SetEnvironmentVariable("Path", [System.Environment]::GetEnvironmentVariable('Path', [System.EnvironmentVariableTarget]::Machine) + ";$($env:HOME)\.dotnet\tools")
         _scripts/test.ps1 CSharp $Before $After
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -35,7 +35,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-latest, windows-latest]
     steps:
     - name: Checkout
       uses: actions/checkout@v2
@@ -53,4 +53,6 @@ jobs:
             $After = "${{ github.event.after }}"
         }
         dotnet tool install -g dotnet-antlr --version 3.0.6
+        Get-Childitem "/home" -Recurse -force -Filter 'dotnet-a*'
+        echo "$env:PATH"
         _scripts/test.ps1 CSharp $Before $After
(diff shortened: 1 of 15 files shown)


kaby76:
I do not understand Powershell. I cannot stop Github Actions. There's a race condition between Powershell/Github Actions/multiple builds/multiple machines that I do not understand. This is preventing my simple, clean, reliable Bash script _script/regtest.sh from producing a result that can then be merged. I will wait several hours for the processes to stop, then try this all over.

kaby76:
Well, I just think there are some differences between OSes for Powershell. In build-csharp-pwsh ubuntu, dotnet-antlr is installed, but Powershell cannot find the tool. In build-csharp-pwsh macos, dotnet-antlr starts to run, but fails because I didn't expect it to be run on MacOSX. in build-csharp-pwsh Windows, the dotnet-tool runs, and produces a driver for the grammar, but the test fails because test.ps1 doesn't construct the name correctly (I may have broken the code, but it wasn't working on my Ubuntu WSL2 machine until I did). Will take care of dotnet-antlr first.

studentmain:
> In build-csharp-pwsh ubuntu, dotnet-antlr is installed, but Powershell cannot find the tool.

Shouldn't happen, we already know older dotnet-antlr (and tester) works on all platform.

kaby76:
> > In build-csharp-pwsh ubuntu, dotnet-antlr is installed, but Powershell cannot find the tool.
> 
> Shouldn't happen, we already know older dotnet-antlr (and tester) works on all platform.

Yes, it did work, but it does not now. Dotnet says the dotnet-antlr tool is installed, says it needs to restart (you can see the output). When Powershell tries to exec "dotnet-antlr -m true .......", Powershell says it cannot find the executable: 

    "Write-Host (dotnet-antlr -m true -t $Target --template-sources-di …
     |                  ~~~~~~~~~~~~
     | The term 'dotnet-antlr' is not recognized as a name of a
     | cmdlet, function, script file, or executable program. Check
     | the spelling of the name, or if a path was included, verify
     | that the path is correct and try again.".

 The test.ps1 works find on WSL2 on my machine, works fine on MacOSX and Windows in Github Actions. It is not working in Github Actions Ubuntu.

I would check in a version that does a "find . -name 'dotnet-antlr'", or an "ls -R / | grep dotnet-antlr", but I don't understand what the equivalent of "find" and "ls" are in Powershell. It's all different.

kaby76:
This is now starting to work. The problem is that dotnet-antlr is an executable that must be on PATH. Using Write-Host in test.ps1, we see that /home/runner/.dotnet/tools--the directory that contains the tool--is not on the path for Ubuntu. Once I added that directory, test.ps1 now works fine on Ubuntu. ":" (colon) as the path separator seems to work on all three target systems--or should I say, it at least doesn't break OSX or Windows, which were already working fine. The other test script--_scripts/regtest.sh--should have a similar problem, but for whatever reason, it finds the tool. I will add similar echos to display path in that script. I do not understand why "echo" does not work in Powershell in test.ps1. It works fine on a command-line prompt. Instead, I had to use "Write-Host".

kaby76:
This PR is good to go.

teverett:
@kaby76  thanks

studentmain:
> we see that /home/runner/.dotnet/tools--the directory that contains the tool--is not on the path for Ubuntu.

They said they added ~/.dotnet/tools to $PATH (https://github.com/actions/virtual-environments/issues/213)

kaby76:
> They said they added ~/.dotnet/tools to $PATH ([actions/virtual-environments#213](https://github.com/actions/virtual-environments/issues/213))

Yes, that sure explains why the script wasn't working. But the strange part is why that day when I made this PR it wasn't working when the day before it was! All I think I did was just change the parameters to the dotnet-antlr tool. I had no idea how to change PATH in Powershell. It feels like there was a change to the standardized machine env over on Github Actions just that day. It was definitely driving me insane!

gfour:
@kaby76 The current GitHub Actions workflow uses "-latest" images, so the strange behavior could it be due to one of these images silently upgrading?
```
os: [ubuntu-latest, macos-latest, windows-latest]
```
Recently I also had a GitHub Action suddenly fail (on an unrelated repo) due to `ubuntu-latest`, I am now pinning versions (`ubuntu-18.04`).

studentmain:
OT, but here's a bad news: ANTLR v4.9.2 JS target, C.g4, stucked when parsing. Still needs investigation, but I'll skip it for now, if thing is worse than I expected, I'll skip the whole JS target. Besides of this, everything seems working. I decide always use latest ANTLR, by doing so, the runtime itself can be tested.

kaby76:
> OT, but here's a bad news: ANTLR v4.9.2 JS target, C.g4, stucked when parsing. Still needs investigation, but I'll skip it for now, if thing is worse than I expected, I'll skip the whole JS target. Besides of this, everything seems working. I decide always use latest ANTLR, by doing so, the runtime itself can be tested.

@studentmain Yep, JS still messed up. For Antlr 4.9.2, the grammars-v4/java/java grammar failed on input that contained an expression '"str1" +"str2"', I found the graph representation of [a DFA being mutated during the parse](https://github.com/antlr/antlr4/issues/3126). Another bad bug related to IntervalSets! The fix still has not been merged, and it's unclear when it might be.

These fixes should be integrated into main as they come, ASAP, so people can work with them. But the process seems to hold off on PRs until a release is decided upon. The PRs are then accepted in a flurry of last-minute integrations months, or in my case, a year, after someone wrote the code. By then, the person who wrote the PR forgot all about the fix and needs to now try to integrate to something that has completely changed. There are PRs that have been there for years, not reviewed, rejected, or integrated. The process for antlr/antlr4 is broken.

--Ken

PR #2085, opened 2021-03-17
Support negative precision scale values

Naros:
Oracle allows negative precision scale values, this patch supports such SQL statements.


KvanTTT:
Please resolve conflicts before merging.

PR #2086, opened 2021-03-17
Add inline JSON documents for jsonencode()

kayman-mk:
Add support for heredoc syntax for jsonencode(...)

jsonencode({"hello"="world"})


teverett:
@kayman-mk  thanks

PR #2094, opened 2021-03-20
Consolidation of the SQLite parser and lexer grammars

mike-lischke:
- Instead of appending an underscore on some keywords to avoid conflicts with the target language, name all of them consistently by appending "_SYM".
- Replaced all lexer literals in the parser grammar by their defined lexer symbol.
- Removed a number of superfluous parentheses (mostly around alternatives in rules. They make the grammar harder to read.
- Consistent formatting with Visual Studio Code, using the format settings added to the files.

I also removed things that are not of common interest:
- A runtime error reference, which doesn't work in most targets. In fact handling of an invalid character happens automatically, by capturing it in a dedicated lexer rule. Also, this allows for better handling in custom error handlers.
- Accepting an unclosed multiline comment is bad. That must be reported as error like any other syntax error.

Otherwise no functionality change was brought in.


mike-lischke:
@bkiers, @Marti2203 please review

bkiers (approved):
LGTM @mike-lischke 👍

mike-lischke:
@Marti2203 I would like to get your approval, as my patch touched something you added.

Marti2203:
Hi, yeah, sorry for the late response. I have just fixed up the grammar to follow the specification, I have no issues with the namings and I see eveyrthing is correct.

Marti2203 (approved):

mike-lischke:
@KvanTTT I think this PR is ready to be merged.

KvanTTT:
> Instead of appending an underscore on some keywords to avoid conflicts with the target language, name all of them consistently by appending "_SYM".

Why did you decide to use this suffix? We use `_` suffix in all grammars over repository, it would be nice to unify the syntax.

mike-lischke:
> Why did you decide to use this suffix? We use _ suffix in all grammars from this repository, it would be nice to unify the syntax.

@KvanTTT Doesn't look that's case actually. I just checked the SQL languages:

- Hive uses a KW_ prefix.
- The Positive Technologies lexer grammar uses no suffix at all.
- My MySQL grammar uses a _SYMBOL appendix for keywords.
- PLSQL doesn't use suffixes.
- SQLite didn't use suffixes, except for 2 language specific keywords.
- TSQL doesn't use suffixes.

So one case there is no such case like a common suffix. Additionally, a single underscore suffix can easily be missed and in some languages underscore-only prefixes and suffixes have a special meaning.

So I believe adding _SYM to keyword rules is the better choice.

teverett:
@mike-lischke I agree that there are exceptions.  However the underscore that @KvanTTT  refers to has been an informal standard for quite a while.

KvanTTT:
I mean we use underscore if only token name conflicts with one of the target keywords. I think suffixes everywhere are excess and suggest using suffixes if only it's needed (`_` or `SYM` or whatever).

mike-lischke:
Having an addition only on some keywords makes things even worse. It destroyes consistency in the grammars and you can miss it even more easy, because most of the time you don't see any. I'm ready to step back and convert the `_SYM` suffix to just `_`, but having it only on very few keywords is not acceptable for me.

Side note why this is important for me: in code completion I process the keyword rule names and not having the same structure for all of them makes things unnecessarily complex.

mike-lischke:
I converted the appendices.

PR #2095, opened 2021-03-20
match PL/SQL grammar superClass

basicarrero:
Matching superClass name in .g4 files


KvanTTT:
Why have you decided to change the name? Other targets use `PlSqlParseTree` also we have the following section in grammar:

```antlr
options {
    tokenVocab=PlSqlLexer;
    superClass=PlSqlParserBase;
}
```

basicarrero:
Building the parser for java was pretty straightforward. But trying to archive the same for python I hit some issues: 

As you show, superclass name expected in g4 grammar is "PlSqlParserBase" instead "PlSqlBaseParser" so I had to rename it in the same way the other languajes are

Later I also noticed that this grammar needs to be targeted introducing "self" reference in order get it running properly in python. So I made a replica including it (prefixed with "py" to avoid breaking the current one)

However I missed this last change in the pull request.
Should I relaunch the request including it?

KvanTTT:
> Should I relaunch the request including it?

No, you can just add a new commit to the current request.

basicarrero:
That's the full set

barscm:
I found the same issue with the PlSqlBaseParser.py vs PlSqlParserBase.py (and Lexer), but hadn't got around to doing a pull request. My (very basic) tests so far consist of just changing the file names to be the correct way around and renaming the classes, although PyCharm does highlight that the import of PlSqlParser as unresolved, the code does run. This pull appears to be the rename, plus adding new `.g4` files in the `python` subdirectory, but I don't see a change that refers to `self`, which file is this in?

KvanTTT:
Please remove duplicated `.g4` lexer and parser files in Python3 subdirectory. It should be only a single lexer and parser.

KvanTTT (changes requested):
Please remove duplicated `.g4` lexer and parser files in Python3 subdirectory. It should be only a single lexer and parser.

Patriciamh (approved):

teverett:
@KvanTTT  @basicarrero  is this ready to pull?

KvanTTT:
No, my note is still unresolved:

> Please remove duplicated .g4 lexer and parser files in Python3 subdirectory. It should be only a single lexer and parser.

kaby76:
I don't this PR is needed at this point because [the grammar](https://github.com/antlr/grammars-v4/tree/master/sql/plsql) is in "target agnostic" form. In other words, it uses "self" in the actions in the parser and lexer grammars, which I changed in August 5 '21 for one of the other targets. The commit that this PR is working off of is really quite old, May '21. The current skip lists, which aren't even in the branch for this PR, bypass plsql for Cpp, Dart, Go, and JavaScript, so [the grammar](https://github.com/antlr/grammars-v4/tree/master/sql/plsql) works for CSharp, Java, Python3. We don't test Antlr4cs, PHP, Swift. I have a PR to fix the grammar for Go but it forks the grammar because the Antlr tool codegen template pumps out "l" and "p" parameters for actions for Go, not "self". A change was made to provide the alias "this" for "self" ([for Go](https://github.com/antlr/antlr4/commit/b915242cd638fbb3ff5c14a05600abc3b2e2533c#diff-193c5191f826bb5176962ef28518b3e27a3d9484dbb591e00347aa56188e9f8d), but I didn't know about that until recently (4.9.3 was released in November).

However, after working on making the grammar "target agnostic" for a year now, I can say that a shared grammar for all targets cannot be realized with Antlr 4.9.3. We really need to address the issue in the Antlr4 tool, starting with a discussion about modifying all the .stg templates for Antlr tool codegen. We can then decide what needs to be done in the meanwhile.

KvanTTT:
> However, after working on making many grammars in this repo "target agnostic" (for about a year now), I can say that a shared grammar for all targets cannot be realized with Antlr 4.9.3. We really need to address the issue in the Antlr4 tool, starting with a discussion about modifying all the .stg templates for Antlr tool codegen. We can then decide what needs to be done in the meanwhile.

I agree with you. At least we should have universal syntax for method calls, it's the simplest but useful. I'm tired of duplicated grammars. I'll create a discussion about it in the ANTLR repository.

kaby76:
@KvanTTT Excellent.  I think we are just about at that point of having enough information to make some decisions on where best to go with target agnostic grammars.

KvanTTT:
BTW, there was a topic [Unified Actions Language](https://github.com/antlr/antlr4/issues/1045) on the ANTLR repository but such a language is really very hard to implement. Universal method calls are much simple and even they can significantly improve target-agnostic grammar development.

PR #2096, opened 2021-03-21
golang: use unicode property in identifier rule

studentmain:
Use `\p` to replace hard-coded unicode table, to reduce size and improve accuracy, especially make sure all CJK character is included.

I suggest reject grammar which included a hard-coded unicode table in the future unless it has a strong reason to do that.

By the way, the grammar seems a little outdated (<1.13?), I'm trying to figure out which version it is exactly.


teverett:
@studentmain  thanks

PR #2103, opened 2021-03-24
Move open/close brace from method/function definition to functionBody itself

Vivecccccc:
In accordance with the usage in JavaParser, moving the braces "{" and "}" from the declarations and definitions of method and function to functionBody.


teverett:
@Vivecccccc  thanks

Issue #2102, opened 2021-03-24
Java9 contains Java-target code, needs lexer base classes for other targets 


kaby76:
Despite not being optimized for performance, the Java9 grammar is an important addition to this repository. The grammar is derived directly from the Java Spec, and thus contains similar rules to the Spec, and more importantly, to the later versions of the Java language. Thus, getting this grammar straightened out is important.

Unfortunately, the Java9 grammar contains Java code in semantic actions in several lexer rules, thus limiting its use to the Java target. In order for it to be used in other targets, the grammar must be rewritten with the target-specific code replaced with "target agnostic" code: The action code is extracted from the grammar and placed in a lexer base class method; In its place, a method call is made.

I have rewritten the code so that there are four targets now supported: Java, CSharp, Cpp, Dart. I am now at the stage of creating a PR for these targets.

--Ken

PR #2104, opened 2021-03-24
Fix for #2102

kaby76:
This is a change to the java/java9/ grammar and code so that it works for Java, CSharp, Cpp, and Dart. To do this, the target-specific code was replaced with "agnostic function calls" in the semantic predicates of the grammar. The original grammar was targeted for Java, so there was a bit of reorganization involved.

(1) I had to split the grammar. In order to create a base class for the lexer, the grammar needs to be split.
(2) The semantic predicates have been moved over to target-specific code in Java/, CSharp, Cpp/, and Dart/.
(3) The Antlr tool does not generate a #include "base-class" for the Cpp target, so this had to be manually added with an `@header` statement in the lexer grammar. This makes the grammar Cpp specific, so a copy is placed here.
(4) The pom.xml file had to be updated so that the sources for Java/ are now copied into the standard Maven CI test.
(5) I have added additional-examples/ so that these files are not tested with the CI tests. These examples take an enormous time to run because the Java9 grammar is un-optimized.
(6) .dotnet-antlr-ignore has been added so that additional-examples/ is not copied into Generated/ and misinterpreted as being part of the generated parser for Java.
(7) I added a readme.

Additional targets will still need to be added, but this PR is a start.

--Ken


teverett:
@kaby76  thanks

studentmain:
> The Antlr tool does not generate a #include "base-class" for the Cpp target

Let's open an issue, it's supposed to do that to support a base class.

kaby76:
@studentmain I will later today. I have one already open for Dart for the same reason--Dart requires the import of any source it uses, including code in a file in the same package--terrible. Not related, I would also like to see the templates for the tool be opened up where one can specify the templates used for a target, and a filter in between the tool and the templates that can be called (so I can work pass bugs quicker, and tnink about a target-neutral language for semantic predicates). As nice as the StringTemplate model is, it's not enough.

Issue #2099, opened 2021-03-22
Cpp target--cannot use "NULL", "register" as grammar symbol names


kaby76:
I now have templates for the Cpp target (v3.0.9 dotnet-antlr with templates that I haven't checked in yet as I haven't quite abstracted out the OSes), and have made a 1st pass through the grammars for the target. Grammars cannot have "NULL" or "register" as symbol names.

Occurrences of NULL:

	./cobol85/Cobol85.g4:2600:   : ALL literal | HIGH_VALUE | HIGH_VALUES | LOW_VALUE | LOW_VALUES | NULL | NULLS | QUOTE | QUOTES | SPACE | SPACES | ZERO | ZEROS | ZEROES
	./cobol85/Cobol85.g4:2935:NULL : N U L L;
	./csharp/CSharpLexer.g4:88:NULL:          'null';
	./csharp/CSharpParser.g4:1011:	| NULL
	./csharp/CSharpParser.g4:1096:	| NULL
	./csharp/CSharpSharwell/CSharpLexer.g4:86:NULL:          'null';
	./informix/informix.g4:258:   | (sqlExpression IS NOT? NULL) sqlExpression IS NOT? NULL
	./informix/informix.g4:315:   : unsignedConstant | string | NULL | FALSE | TRUE
	./informix/informix.g4:430:   | NULL
	./informix/informix.g4:584:   | INITIALIZE variable (COMMA variable)* (TO NULL | LIKE expression (COMMA expression)*)
	./informix/informix.g4:825:   : constantIdentifier (dataType | (BYTE | TEXT) (IN (TABLE | constantIdentifier))?) (NOT NULL)?
	./informix/informix.g4:1025:   : (ACCEPT | ASCII | COUNT | CURRENT | FALSE | FIRST | FOUND | GROUP | HIDE | INDEX | INT_FLAG | INTERRUPT | LAST | LENGTH | LINENO | MDY | NO | NOT | NOTFOUND | NULL | PAGENO | REAL | SIZE | SQL | STATUS | TEMP | TIME | TODAY | TRUE | USER | WAIT | WEEKDAY | WORK)
	./informix/informix.g4:2144:NULL
	./less/LessLexer.g4:10:NULL
	./objc/ObjectiveCParser.g4:790:    | NULL
	./objc/one-step-processing/ObjectiveCLexer.g4:87:NULL:                     'NULL';
[...]

PR #2105, opened 2021-03-24
Fix for 2099

kaby76:
This change is for #2099, symbol conflicts for the Cpp target.

(1) The Cpp target build reports that it has symbol conflicts for NULL, register, operator, extern, namespace, delete, signed, and, or, not, union, struct, typedef, typename, bool, template. Some are not checked by the Antlr tool (NULL), but most are. The symbols are renamed in the usual manner--with a trailing underscore.
(2) There are two grammars that define symbols that I think are unused ([sharc, NULL](https://github.com/antlr/grammars-v4/blob/master/sharc/SHARCLexer.g4#L2138); [icalendar, bool](https://github.com/antlr/grammars-v4/blob/master/icalendar/ICalendar.g4#L923)). These might be errors in the grammars, but they should be investigated.

--Ken


teverett:
@kaby76  thanks

PR #2110, opened 2021-03-26
Fix Erlang check error caused by TestRig package name change

airstone42:
When I tried the Erlang grammar, [check.sh](https://github.com/antlr/grammars-v4/blob/master/erlang/check.sh) reported errors. It seems that the captured message "Warning: TestRig moved to org.antlr.v4.gui.TestRig; calling automatically"  was recognized as an error output.

Since the package name change updated in docs as https://github.com/antlr/antlr4/commit/978851f7034bb882ac546d85bf925293cd3643ac, I just simply replace them in Erlang and other directories. Hope it does not break compatibility.

### DIFF
--- a/erlang/Makefile
+++ b/erlang/Makefile
@@ -5,7 +5,7 @@ compile:
 	javac Erlang*.java
 
 debug: compile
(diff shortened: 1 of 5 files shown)


teverett:
@airstone42  thanks!

Issue #2106, opened 2021-03-24
Updates for CI build regtest.sh


kaby76:
I have a few updates for the _scripts/regtest.sh CI testing script:

* The readme requires some updates to have a better description of how to use the scripts, how to update the templates, and new attributes that I had to add for the Cpp target.
* Add the Cpp target to _scripts/regtest.sh, add date for timing information, use the latest version of dotnet-antlr, required for Cpp.
* Update the templates for CPP in _scripts/templates/Cpp. Update the _scripts/templates/files.
* Note, the Cpp target is *not* run in CI because it takes 6 hours to do on my unloaded AMD Ryzen 2 machine. That compares to an hour or less for the CSharp target. This is using MinGW 64-bit on Windows. But, it's provided so we can eventually add it. The time is pretty much the same for GNU gcc, Windows C++ 2019. And, contrary to popular opinion, it is *not* faster than the other targets. In fact, it is in many cases slower than all the other targets, even in "release" mode.

--Ken

PR #2111, opened 2021-03-26
Changes for #2106

kaby76:
This is a change to the CI scripts in order to use the latest version of dotnet-antlr, for the Cpp target. Unfortunately, the changes are breaking changes because the -package namespace argument to the Antlr tool must be passed onto the generated code, specifically for kotlin/kotlin-formal. Grammar files are now placed in a directory structure corresponding to the name space. Yeah, unfortunately Java requires this terrible source code structure, and it's carried over to all the other targets. The solution is to just stop using packages for grammars in this repo. It would make life much much easier.

Also changed is _scripts/readme.md to contain more information on the CI test.


kaby76:
Regressions with mysql. Will take care in the morning.

kaby76:
Terrible. I rely on a consistent experience of the Antlr Tool JAR when generating files. [It turns out, it's not consistent](https://github.com/antlr/antlr4/issues/3138). I will need to think about this.

gfour:
The "terrible source code structure" is not a Java thing, rather a Maven assumption that can be overridden in pom.xml, I think it's worth it fixing it this way.
Having proper Java packages is necessary or other projects will not be able to use the parsers here as-is but fork them and add the package there. The problematic Java "feature" at play here is that if the parsers have no package then other code that is organized in packages (probably everything but toy programs) cannot see their classes. See for example: https://stackoverflow.com/questions/283816/how-to-access-java-classes-in-the-default-package

kaby76:
For sure, the tool should be fixed. If I say "put the output _here_", it should go "_here_". The `-o` option should be simple, transparent, unconfusing: just put the output files where I say to put the files. No exceptions.

kaby76:
@gfour Please review the change I made to kotlin/kotlin-formal pom. I modified the pom so that the antlr tool is not run over UnicodeClasses.g4, generating a separate tokens file, and java file. Imported grammars are generally not run through the tool directly, e.g. antlr/antlr4 and LexBasic.g4. If it is intended to be a standalone lexer, I will remove the change. That's the only change I made to kotlin-formal; I didn't change the package/ namespace.

gfour (approved):
@kaby76 This change looks good.

kaby76:
Almost there. Long story short:
* The Antlr tool observes the -o option most of the time, but on Linux not always.
* The Dart target supposedly supports namespaces/packages, but it doesn't due to code generation problems.
* The Go target always generates to some package whether you have a -package option or not on the Antlr tool call. For now, I now hardwire it to "parser", but do it in a "correct" way in the dotnet-antlr tool.
* In _scripts/test.ps1, the Sql/mysql wasn't being tested properly. No tests were parsed using the generated parser. When I fixed that, I come to find that it requires the "case insensitive parser" stream hack. So, I added that.
* The templates required a bit of updating, but _scripts/test.ps1 isn't standalone: to run "pwsh _scripts/test.ps1 Go", it requires an environment set up contained in .github/workflows/main.yml. It would be good to make it standalone so one can do the testing on a local machine rather than use Github Actions to do the testing, which is never reliable due to occasional download failures.

--Ken

kaby76:
Miracles happen. The PR now works. -K

teverett:
@kaby76  @gfour  thanks

studentmain:
> It would be good to make it standalone so one can do the testing on a local machine

I installed antlr following the https://github.com/antlr/antlr4/blob/master/doc/getting-started.md , so `antlr` is already in my PATH and antlr.jar is already in my CLASSPATH.

Maybe add `-AntlrLocation` parameter to specify antlr.jar 's location?

PR #2112, opened 2021-03-28
Update Erlang grammar

airstone42:
Update Erlang grammar rules to OTP 23.3 and add some small fixes.

I mainly added two new grammar rules with their test cases.

- Funs with Names
- Maps

I have run [check.sh](https://github.com/antlr/grammars-v4/blob/master/erlang/check.sh) to check against the whole Erlang/OTP. It seems that most of them passed except a few due to known issues such as character set problems mentioned in #24.

As described in #24, this grammar works on preprocessed Erlang code. However, I found that the test cases currently are not preprocessed, and there is even an ```.errors``` file for a legal Erlang file. It would be better to test preprocessed Erlang code only.

```testFileExtension``` in antlr4test-maven-plugin does not work now as https://github.com/antlr/antlr4test-maven-plugin/pull/36, so I have to add a subdirectory for preprocessed `.P` files. It may be optimized in the future.


airstone42:
The subdirectory for preprocessed files was removed, and only preprocessed ```.P``` files were left in examples.

teverett:
@airstone42 thanks

PR #2113, opened 2021-03-29
Allow to check colmun

daikoz:
In this case:

```sql
SELECT 1 AS toto, table.Id
FROM table
GROUP BY toto
```

To check the column in select, we must parse the first FROM clause. But then parse FromClause, in the previous g4, group by clause is included in from clause and we cannot know the alias in select. Thus, we must separate from clause and group by clause to simply the SQL parsing.


KvanTTT:
Could please add a test sample that covers new changes?

daikoz:
Done. Sample added.

Issue #2077, opened 2021-03-12
C++ parameter types


gfour:
I believe the C++ parser (CPP14Parser) does not correctly parse function parameters. Assume this piece of code (from [Boost](https://github.com/boostorg/interprocess/blob/develop/include/boost/interprocess/xsi_key.hpp)):
```
xsi_key(const char *path, boost::uint8_t id) {}
```
But, as shown in the screenshot, "id" is considered a "className" (while "path" is not).

![cpp-parseTree](https://user-images.githubusercontent.com/1374661/110969403-51a89080-8361-11eb-8f5b-f936599d647f.png)

kstallard:
I also think there is something wrong with the CPP14Parser.g4.  

I've made some modifications that seem to fix it, but it still needs to be vetted with a wide variety of function parameters.

Disclaimer: I am not an Antlr expert...

[CPP14.zip](https://github.com/antlr/grammars-v4/files/6290649/CPP14.zip)

kstallard:
Just made another commit that fixes the proper parsing of signdness and length modifiers.

kstallard:
@gfour The general changes that are currently in a pull request now produce this using your example.

![antlr4_parse_tree_gfour_example](https://user-images.githubusercontent.com/5996043/114288580-80805800-9a2e-11eb-9cd0-1fa20246d090.png)

kaby76:
The problem you found is caused by the ambiguity of the derivation of `parameter-decl`attribute-specifier-seqopt decl-specifier-seq declarator` vs. `attribute-specifier-seqopt decl-specifier-seq abstract-declaratoropt` in the [spec, page 1214](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3797.pdf). In "const char* path”, the sequence “*path” must be derived from `declarator` because it involves a pointer. But, for "boost::uint8_t id", “id” is ill-considered as a class-name.

[...]

PR #2124, opened 2021-04-10
Fix for parsing function prototypes correctly

kstallard:
The parse tree was classifying variables as types.  CPP14Parser was using an overly broad rule that was intended for non-function prototype variable declarations.  Made a more limited rule for function/method prototypes.

I'm not 100% sure this covers all cases, but is an improvement from what it was doing previously.  There are some CPP parsing requirements I have on a project I am working on so it will be tested more.  But I think this at least gets us going down the right road.

I also noticed that some simple types are not being parsed correctly, for example, unsigned short, unsigned int, unsigned long long, for example.  I think I also have a fix for that, but it needs more testing and will do that as a separate commit.

### DIFF
--- a/cpp/CPP14Parser.g4
+++ b/cpp/CPP14Parser.g4
@@ -377,10 +377,10 @@ simpleTypeSpecifier:
 	| nestedNameSpecifier Template simpleTemplateId
 	| simpleTypeSignednessModifier
 	| simpleTypeSignednessModifier? simpleTypeLengthModifier+
-	| Char
-	| Char16
-	| Char32
-	| Wchar
+	| simpleTypeSignednessModifier? Char
+	| simpleTypeSignednessModifier? Char16
+	| simpleTypeSignednessModifier? Char32
+	| simpleTypeSignednessModifier? Wchar
 	| Bool
 	| simpleTypeSignednessModifier? simpleTypeLengthModifier* Int
 	| Float
--- a/cpp/CPP14Parser.g4
+++ b/cpp/CPP14Parser.g4
@@ -330,12 +330,7 @@ declSpecifier:
 	| Typedef
 	| Constexpr;
 	
-paramDeclSpecifier:
-	typeSpecifier;
-	
-declSpecifierSeq: declSpecifier+ attributeSpecifierSeq?;
-
-paramDeclSpecifierSeq: cvqualifierseq? paramDeclSpecifier attributeSpecifierSeq?;
+declSpecifierSeq: declSpecifier+? attributeSpecifierSeq?;
 
 storageClassSpecifier:
 	Register
@@ -567,7 +562,7 @@ parameterDeclarationList:
 	parameterDeclaration (Comma parameterDeclaration)*;
 
 parameterDeclaration:
-	attributeSpecifierSeq? paramDeclSpecifierSeq (
+	attributeSpecifierSeq? declSpecifierSeq (
 		(declarator | abstractDeclarator?) (
 			Assign initializerClause
 		)?
--- a/cpp/CPP14Parser.g4
+++ b/cpp/CPP14Parser.g4
@@ -329,8 +329,13 @@ declSpecifier:
 	| Friend
 	| Typedef
 	| Constexpr;
-
+	
+paramDeclSpecifier:
+	typeSpecifier;
+	
 declSpecifierSeq: declSpecifier+ attributeSpecifierSeq?;
+
+paramDeclSpecifierSeq: cvqualifierseq? paramDeclSpecifier attributeSpecifierSeq?;
 
 storageClassSpecifier:
 	Register
@@ -556,7 +561,7 @@ parameterDeclarationList:
 	parameterDeclaration (Comma parameterDeclaration)*;
 
 parameterDeclaration:
-	attributeSpecifierSeq? declSpecifierSeq (
+	attributeSpecifierSeq? paramDeclSpecifierSeq (
 		(declarator | abstractDeclarator?) (
 			Assign initializerClause
 		)?
@@ -816,3 +821,4 @@ literal:
 	| BooleanLiteral
 	| PointerLiteral
 	| UserDefinedLiteral;
+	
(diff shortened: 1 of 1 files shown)


kstallard:
Issue linking: #2077

kstallard:
Here is an example method/function prototype that would fail:

int foo( short s );

The before Parse Tree:
antlr4_parse_tree_CPP14_incorrect_function_parsing


After the fix the parse tree looks like this:
antlr4_parse_tree_CPP14_correct_function_parsing

kstallard:
Well...looks like I need to improve my workflow with GitHub.  I meant this second commit to be a separate pull request. 
In any case, this second commit resolves issues with signdness or length modifiers while at the same time allowing those modifiers to be used a simple types.  The following test prototypes now work:

int foo( short s );
int foo( short int si );
int foo( signed short ss );
int foo( signed short int ssi);
int foo( unsigned short us );
int foo( unsigned short int usi );
int foo( int i);
int foo( signed s );
int foo( signed int si );
int foo( unsigned u );
int foo( unsigned int ui );
int foo( long l );
int foo( long int li );
int foo( signed long sl );
int foo( signed long int sli );
int foo( unsigned long ul );
int foo( unsigned long int uli );
int foo( long long ll );
int foo( long long int lli );
int foo( signed long long sll );
int foo( signed long long int slli );
int foo( unsigned long long ull );
int foo( unsigned long long int ulli );
int foo( double d );
int foo( long double ld );
int foo( unsigned long );
int foo( unsigned long int );
int foo( long long );
int foo( long long int );

kstallard:
Final commit for this series of changes.  Realized that I had forgotten the char data type modifiers.  Tested with:

int foo( char c );
int foo( unsigned char c );

int foo( char16_t c );
int foo( unsigned char16_t c );

int foo( char32_t c );
int foo( unsigned char32_t c );

int foo( unsigned long char32_t c );	//This one should fail

int foo( wchar_t c );
int foo( unsigned wchar_t c );

kstallard:
Final note:  Wanted to say thank you to @Camiloasc1  and @Marti2203 for doing such a fantastic job on this grammar.  It is really helping us with a project we are working on and has made things SO much easier.  I mention them here so that they can hopefully critique these changes.  I am out of my element here and while they appear to be okay changes, someone else with more experience (like them) having a look at it would be great.

cpp/CPP14Parser.g4
```diff
@@ -359,21 +364,27 @@ typeSpecifierSeq: typeSpecifier+ attributeSpecifierSeq?;
...
+
 simpleTypeSpecifier:
 	nestedNameSpecifier? theTypeName
 	| nestedNameSpecifier Template simpleTemplateId
-	| Char
-	| Char16
-	| Char32
-	| Wchar
+	| simpleTypeSignednessModifier
+	| simpleTypeSignednessModifier? simpleTypeLengthModifier+
```
Marti2203, line 379: This will catch `short long short long short long`. Is that possible?

Marti2203 (changes requested):

cpp/CPP14Parser.g4
kstallard: It certainly will.  But should that be limited by the grammar itself or should that be sorted out by higher level code? I am new to this and so I not quite sure where to draw the line.

cpp/CPP14Parser.g4
```diff
@@ -359,21 +364,27 @@ typeSpecifierSeq: typeSpecifier+ attributeSpecifierSeq?;
...
+	| simpleTypeSignednessModifier? Char16
+	| simpleTypeSignednessModifier? Char32
+	| simpleTypeSignednessModifier? Wchar
 	| Bool
-	| Short
-	| Int
-	| Long
-	| Signed
-	| Unsigned
+	| simpleTypeSignednessModifier? simpleTypeLengthModifier* Int
```
kstallard, line 385: @Marti2203 This would also be related to your comment above....it would also match long long long long long int here....or even long short long short int.

Again, I'm not sure if the grammar should limit it or if should be done at a higher level. For example in the long long long int case, I could see it being valid on a 256 bit processor sometime in the future? (just like long long int on a 64 bit processor is valid).

Obviously long short long short is bogus for everyone...again...not sure where we should try to catch this.

cpp/CPP14Parser.g4
Marti2203: I really am not sure for this, I prefer being able to go with a more conservative approach.

cpp/CPP14Parser.g4
Marti2203: In my opinion, it should be caught, but I would like somebody else to give their opinion.

kstallard: I'm included to go with your suggestion @Marti2203. I'll work on change and if no one comments in the next day or so, I'll push it up.....

teverett:
Is this ready to pull?

kstallard:
> Is this ready to pull?

I would say yes.  I'll follow up with the long short long short thing a bit later.

cpp/CPP14Parser.g4
```diff
@@ -329,8 +329,8 @@ declSpecifier:
 	| Friend
 	| Typedef
 	| Constexpr;
-
-declSpecifierSeq: declSpecifier+ attributeSpecifierSeq?;
+	
+declSpecifierSeq: declSpecifier+? attributeSpecifierSeq?;
```
KvanTTT, line 333: Why just not `declSpecifier*` instead of `declSpecifier+?`

cpp/CPP14Parser.g4
KvanTTT: It should be limited by grammar if possible.

cpp/CPP14Parser.g4
kstallard: Because we need at least one.

cpp/CPP14Parser.g4
kstallard: It is better than it was (ie short int, long int and such declarations as unsigned long were not captured).  We can follow up with something that restricts it to specific patterns of unsigned, signed, short and long later.

cpp/CPP14Parser.g4
kstallard: I'm going to leave it as is for right now.  We can follow up with something better.  It at least recognizes these patterns, as it didn't do that before.

teverett:
@kstallard  thanks

kstallard:
@teverett Thanks for accepting this pull request.  How do I close the issue associated with the pull request?  It is issue #2077 

Thanks
Kevin

PR #2132, opened 2021-04-13
Cleanup the C grammar, update the go grammar and fix #2122 

### LINKED ISSUE #2122: Go Grammar cannot parse constants with comments correctly and lineTerminatorAhead is a bad name
Hello,
Yesterday I began playing around with the go grammar to see how close it is to spec. After fixing some small things I found a case where it does not work correctly. I used [Hashicorp's Vault](https://github.com/hashicorp/vault) as a playground to test the power of the grammar and I found that [for this example, the grammar does not parse correctly.](https://github.com/hashicorp/vault/blob/master/vendor/github.com/Azure/go-ntlmssp/negotiate_flags.go). 
After some experimenting I found that because of the way the comments are placed - if you remove the `/*B*/` or add a `;` at the end of the first definition of T , the grammar parses correctly. The faulty method is `lineTerminatorAhead`, which sounds like an incorrect name for me - should it not be `lineTerminatorBefore` as the method looks in the previous tokens?
Small reproducible example.
```go
package A
const (
	/*A*/ T = 1 << 0
	/*B*/ T = 1 << 1
)
```

Marti2203:
I was reading the C grammar as a reference to fix the objective C one and I saw that at times it was a little too unoptimized with the tendency to create deep trees. I have mostly condensed things and not added anything new.


Marti2203:
As I am still not good enough with github pull requests, I mistakenly pushed the commit I have for #2122 and some small updates  for the go grammar.

PR #2134, opened 2021-04-13
T-SQL - Logical Operator Precedence in search_condition

Silvenga:
This pr restores building of the search condition syntax tree removed by 67d66f1e24474d4531f7bb1c4c29b57b93d9280b and restores operator operator precedence of `AND` and `OR`.

(also removes extra trailing whitespace)

I'm not sure if there's a performance difference in the original syntax, but I would suggest that building the tree is required to make a listener that doesn't need to implement something like shunting-yard during tree walking. @siggemannen do you remember if there was a reason for this change?

### DIFF
--- a/sql/tsql/TSqlParser.g4
+++ b/sql/tsql/TSqlParser.g4
@@ -674,7 +674,7 @@ alter_availability_group_options
 ip_v4_failover
     : STRING
     ;
-    
+
 ip_v6_failover
     : STRING
     ;
@@ -760,7 +760,7 @@ drop_database_audit_specification
 drop_database_encryption_key
     : DROP DATABASE ENCRYPTION KEY
     ;
-    
+
 // https://docs.microsoft.com/en-us/sql/t-sql/statements/drop-database-scoped-credential-transact-sql
 drop_database_scoped_credential
    : DROP DATABASE SCOPED CREDENTIAL credential_name=id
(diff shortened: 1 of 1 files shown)


siggemannen:
Hi @Silvenga. Let me check that, i got a test case that parses through a couple of thousand of procedure definitions. The changes was (i think?) to simplify for the parser to resolve ambiguity between some search conditions especially larger nested ones. Before the changes, the larger procedures were unparseable due to too slow speed or not being able to be done in the simpler SLL mode.
Unfortunately, i'm kind of a n00b so mostly likely messed something up for those using the parser for other things like actual language processing / ASTs =)

siggemannen:
Hi again. I have checked and i don't see any regressions or speed downsides of this change

sql/tsql/TSqlParser.g4
```diff
@@ -2977,15 +2977,14 @@ update_elem_merge
...
     ;
 
+// https://docs.microsoft.com/en-us/sql/t-sql/queries/search-condition-transact-sql
 search_condition
-    : pred+=predicate_br (log=(OR | AND) pred+=predicate_br)*
-    ;
-
-predicate_br
     : NOT* predicate
     | NOT* '(' search_condition ')'
```
KvanTTT, line 2983: Maybe merge these alternatives?

```suggestion
    : NOT* (predicate | '(' search_condition ')')
```

Silvenga: Thanks!

Silvenga:
Perfect @siggemannen - thanks for checking!

PR #2135, opened 2021-04-13
T-SQL - Improves Parentheses Handling in table_source's

Silvenga:
This pr improves handling of parentheses in table_source parsing.

The [grammar](https://docs.microsoft.com/en-us/sql/t-sql/queries/from-transact-sql) suggests these parentheses aren't valid (or that I'm blind). That said, SQL Server does parse these queries just fine, and Microsoft's Entity Framework generates these types of queries via Linq to SQL.

### DIFF
--- a/sql/tsql/TSqlParser.g4
+++ b/sql/tsql/TSqlParser.g4
@@ -3143,14 +3143,15 @@ table_sources
     : source+=table_source (',' source+=table_source)*
     ;
 
-// https://msdn.microsoft.com/en-us/library/ms177634.aspx
+// https://docs.microsoft.com/en-us/sql/t-sql/queries/from-transact-sql
 table_source
     : table_source_item_joined
-    | '(' table_source_item_joined ')'
+    | '(' table_source ')'
     ;
 
 table_source_item_joined
     : table_source_item joins+=join_part*
+    | '(' table_source_item_joined ')' joins+=join_part*
     ;
 
 table_source_item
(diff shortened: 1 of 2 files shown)


sql/tsql/TSqlParser.g4
```diff
@@ -3143,14 +3143,15 @@ table_sources
     : source+=table_source (',' source+=table_source)*
     ;
 
-// https://msdn.microsoft.com/en-us/library/ms177634.aspx
+// https://docs.microsoft.com/en-us/sql/t-sql/queries/from-transact-sql
 table_source
     : table_source_item_joined
-    | '(' table_source_item_joined ')'
+    | '(' table_source ')'
```
Silvenga, line 3149: Handles nested parentheses e.g.

```
SELECT * FROM ((A INNER JOIN B ON 1 = 1))
```

sql/tsql/TSqlParser.g4
```diff
@@ -3143,14 +3143,15 @@ table_sources
...
     ;
 
 table_source_item_joined
     : table_source_item joins+=join_part*
+    | '(' table_source_item_joined ')' joins+=join_part*
```
Silvenga, line 3154: Handles cases like:

```
SELECT * FROM (A INNER JOIN B ON 1 = 1) LEFT JOIN C ON 1 = 1
```

siggemannen:
This change also doesn't seem to affect perf badly. Nice work, @Silvenga !

Issue #1745, opened 2020-06-06
Smiles build fails on Windows


teverett:

Marti2203:
Hi, I am going to make a PR to fix the Smiles language as there was an issue with some elements missing and I flattened one of the rules.

PR #2137, opened 2021-04-14
Rework smiles grammar - fix for #1745

Marti2203:
Yesterday i saw #1745  and began fixing it to match the specification. i saw that a couple of elements were missing, there were some typos and the definition of the chain rule made it a very right-leaning tree.


smiles/smiles.g4
```diff
@@ -1,561 +1,369 @@
...
+	'[' isotope? symbol chiral? hcount? charge? class_? ']';
+
+isotope: DIGIT+;
+
+symbol: element_symbols | aromatic_symbol | '*';
+
+element_symbols: (UH)
+	| (UH LE)
+	| (UL LI)
+	| (UB LE)
```
KvanTTT, line 49: It would be better to remove useless parentheses:

```suggestion
element_symbols: (UH)
	| UH LE
	| UL LI
	| UB LE
	...
```

smiles/smiles.g4
```diff
@@ -1,561 +1,369 @@
...
 
-UF
-   : ('F')
-   ;
+terminator: SPACE TAB | LINEFEED | CARRIAGE_RETURN;
 
-UG
-   : ('G')
-   ;
+LA: ('a');
```
KvanTTT, line 254: Also useless parentheses for lexer rules:

```suggestion
LA: 'a';
```

smiles/smiles.g4
Marti2203: Fully agree, will add it

smiles/smiles.g4
```diff
@@ -1,561 +1,369 @@
...
+
+isotope: DIGIT+;
+
+symbol: element_symbols | aromatic_symbol | '*';
+
+element_symbols: (UH)
+	| UH LE
+	| UL LI
+	| UB LE
+	...
```
KvanTTT, line 50: Not only here but in all alternatives. Moreover, now this grammar is not valid due to ellipsis.

Marti2203: Yeah, I fixed that just now.

smiles/smiles.g4
```diff
@@ -1,561 +1,368 @@
...
 
-UA
-   : ('A')
-   ;
+bond: '-' | '=' | '#' | '$' | ':' | '/' | '\\';
 
-UB
-   : ('B')
-   ;
+ringbond: bond? DIGIT | bond? '%' DIGIT DIGIT;
```
KvanTTT, line 244: Simplify?

```suggestion
ringbond: bond? (DIGIT | '%' DIGIT DIGIT);
```

KvanTTT (approved):

teverett:
@Marti2203  thanks

Issue #1223, opened 2018-09-09
Develop STL grammar


teverett:
https://en.wikipedia.org/wiki/STL_(file_format)

Marti2203:
Hi,
Just the ASCII STL grammar right?

PR #2138, opened 2021-04-14
Adding STL

Marti2203:
I saw the issue #1223 and decided to add the ASCII STL grammar. I think the binary one is not needed.


teverett:
@Marti2203  thanks

Issue #1373, opened 2019-03-09
noreturn keyword in CPP14 grammar


sonicning:
I checked there is noreturn keyword since C++11， but there isn't in CPP14.g4

7.6.3 in this PDF

http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4296.pdf

Would you please check and add this keyword into CPP14 grammar ? Thanks

Marti2203:
Hi,
This issue is fixed. The grammar can correctly parse it now.

Issue #1369, opened 2019-02-13
ANTLR4 stops parsing in the middle of the code, because it expects EOF


YugoCode:
I'm using ANTLR to parse all function names defined in a C++ file. 

While using the latest grammar (_CPP14.g4_), I got many errors parsing this file: https://android.googlesource.com/platform/system/bt/+/master/stack/avrc/avrc_api.cc

This are the errors:
```
line 517:10 token recognition error at: '"Received a CONTINUE/END without no corresponding START \\n'
line 518:77 token recognition error at: '");\n'
line 519:19 no viable alternative at input 'if(pkt_type!=AVRC_PKT_SINGLE&&cr==AVCT_RSP){if(pkt_type==AVRC_PKT_START){p_rcb->rasm_pdu=*p_data;p_rcb->p_rmsg=(BT_HDR*)osi_malloc(BT_DEFAULT_BUFFER_SIZE);memcpy(p_rcb->p_rmsg,p_pkt,sizeof(BT_HDR));memcpy((uint8_t*)(p_rcb->p_rmsg+1),(uint8_t*)(p_pkt+1)+p_pkt->offset,p_pkt->len);p_rcb->p_rmsg->offset=p_rcb->rasm_offset=0;osi_free(p_pkt);*pp_pkt=p_rcb->p_rmsg;p_rcb->p_rmsg->offset+=p_rcb->p_rmsg->len;req_continue=true;}elseif(p_rcb->p_rmsg==NULL){AVRC_TRACE_DEBUG((orpreviousfragmentedresponsewasdropped)drop_code=5;'
line 484:53 missing ';' at '{'
line 519:19 no viable alternative at input 'if(p_rcb->p_rmsg==NULL){AVRC_TRACE_DEBUG((orpreviousfragmentedresponsewasdropped)drop_code=5;'
line 513:38 missing ';' at '{'
line 519:19 no viable alternative at input 'AVRC_TRACE_DEBUG((orpreviousfragmentedresponsewasdropped)drop_code=5;'
line 522:6 mismatched input 'else' expecting '}'
line 573:2 mismatched input 'if' expecting <EOF>
```

I can detect all the functions defined until line 573, but all following functions are not detected.

The first two errors are based on a multiline string:

```
      AVRC_TRACE_DEBUG(
          "Received a CONTINUE/END without no corresponding START \
                                (or previous fragmented response was dropped)");
```

Can someone help with fixing this and the other errors? 

[...]

PR #2140, opened 2021-04-14
Added two files that show correctness of the grammar.

Marti2203:
I saw that #1373 #1369 can be closed and adding files to provide examples for correctness.


teverett:
@Marti2203  thanks

Issue #2063, opened 2021-03-04
[JSX] url inside html tag are mistakenly put into the hidden channel


Markz666:
```javascript
import React from 'react';
class SummaryContainer extends React.Component {
    render() {
        return (
            <div>Content：<a target="_blank" href="https://www.github.com">https://www.github.com</a>
            </div>
        );
    }
}
```
content after "//" are missing due to the following rule in lexer:
```
SingleLineComment:              '//' ~[\r\n\u2028\u2029]* -> channel(HIDDEN);
```
Related file: https://github.com/antlr/grammars-v4/blob/master/javascript/jsx/JavaScriptLexer.g4

Marti2203:
Hi, 
A simple solution is to move all the comment definitions after the string definitions. I am going to make a PR for that.

Marti2203:
Another issues is that https://www.github.com is also not caught and that I will include in the PR as well.

PR #2147, opened 2021-04-17
Fix for #2063

Marti2203:
Moves the comment defintion rules below the string defintiion such that they are not "cut away" from the text. Also adds a link literal that catches hypertext references.


javascript/jsx/JavaScriptLexer.g4
```diff
@@ -189,6 +185,8 @@ StringLiteral:                 ('"' DoubleStringCharacter* '"'
              |                  '\'' SingleStringCharacter* '\'') {this.ProcessStringLiteral();}
              ;
 
+LinkLiteral: ('http' | 'https' | 'ftp' | 'file') '://' [a-zA-Z0-9./?=]+; // TODO Could be more precise
```
studentmain, line 188: LinkLiteral url scheme needn't be precise, we can't enumerate every possible url scheme.

```html
<a href=mailto:example@example.com>e mail</a>
```
<a href=mailto:example@example.com>e mail</a>

Marti2203: Yes, that is true, but there are maybe more protocols needed to that list. I just added the default ones I could think of on the top of my head and there are URLs which have `@` which I just remembered so I want to point out that it is known that the pattern is not perfect.

KvanTTT, line 188: Minor, but can be simplified:

```suggestion
LinkLiteral: ('http' 's'? | 'ftp' | 'file') '://' [a-zA-Z0-9./?=]+; // TODO Could be more precise
```

Issue #2153, opened 2021-04-20
Symbol conflicts in grammars


kaby76:
I am nearly done with [trrename](https://github.com/kaby76/Domemtech.Trash/tree/main/trrename), which takes a grammar and a list of symbols to rename and generates a new grammar with the renamed symbols. And, I have made an initial pass over all the grammars in the repo. There are many, many changes still required to remove all symbol conflicts. Most of them are now due to the Swift target, which I haven't yet added to the [trgen](https://github.com/kaby76/Domemtech.Trash/tree/main/trgen) template processor.

Below is a script that does a complete rename of the symbols in all grammars. The runtime for it took less than I initially estimated, 9 minutes to fully rename all symbol conflicts in all grammars. There are still bugs in the trrename tool, so I don't know exactly how many grammars changed, but it's likely well over half, probably close to 3/4 of the grammars that require some symbol renaming. If this was to be done manually, it would in all likelihood require months of work.

Most of the script just sets up the list of symbols that are to be renamed. There are 9 targets for Antlr needed to be specified, a merge of each into one list, and a sort of the list. Only a few lines are required to find a grammar file and do the renaming.

Since the script is so quick and easy to apply, I'll likely supply a script to rename symbols back without the underscores. This will allow people who don't want to change to still use the grammars.

	# Symbol conflicts arise over the use of a target-specific reserved
	# word being used in a grammar. Up to now, this list was derived
	# experimentally. The list can actually be derived from the sources in
	# Antlr here:
	# https://github.com/antlr/antlr4/tree/master/tool/src/org/antlr/v4/codegen/target
	# The list here is a copy/paste of the source code reserved words,
	# modified for consumption in Bash.

	all_sym=("alignas")

	sym_csharp=( \
			"alignas" "alignof" "and" "and_eq" "asm" "auto" "bitand" \
			"bitor" "bool" "break" "case" "catch" "char" "char16_t" \
			"char32_t" "class" "compl" "concept" "const" "constexpr" \
			"const_cast" "continue" "decltype" "default" "delete" "do" \
			"double" "dynamic_cast" "else" "enum" "explicit" "export" \
			"extern" "false" "float" "for" "friend" "goto" "if" \
			"inline" "int" "long" "mutable" "namespace" "new" \
			"noexcept" "not" "not_eq" "nullptr" "operator" "or" \
			"or_eq" "private" "protected" "public" "register" \
			"reinterpret_cast" "requires" "return" "short" "signed" \
[...]

PR #2155, opened 2021-04-23
Fix for Issue 2153

kaby76:
This is a fix for #2153, which is about "symbol conflicts". I'll first go over why I'm making this change and then go over the details.

## Background on "symbol conflicts"
Antlr can generate a parser for a dozen or so different targets. In the past, people noticed that using reserved words from a target language in a grammar would cause Antlr to generate a parser with build issues for that target.

For example, if you had a rule in your grammar such as "int : INT;", the Antlr4 Tool would have generated methods, functions, or variables with the name "int", which clearly would not build for C# or Java. This problem was called a symbol conflict.

To work around symbol conflicts, it was decided to add a check to the tool to prevent people from including symbols in a grammar that would cause build errors. The Tool would stop and output "symbol _foo_ conflicts with generated code in target language or runtime".

This was fine because old grammars would not be impacted when using a new version of Antlr. People writing a new grammar would just "get used to" the preventive check if they tried to use a symbol labeled _haram_.

The problem with that design decision is that the grammars in this repo are assumed to work on different targets. People who pick up the grammar and try to use it for a different target than what it was developed for would run into symbol conflicts. And, the CI scripts for this repo now test grammars across the various targets.

While a change in the Antlr4 Tool to [generate unique names in the generated target code](https://github.com/antlr/antlr4/issues/1851#issuecomment-302924469) would be the right way to solve this issue, it's not possible because "Antlr is in maintenance mode". _Painfully, the only practical solution is to rename all conflicting symbols in all grammars. However, in my opinion, this is the wrong way to solve the problem. For this and other reasons, we should start to look beyond Antlr, towards a new parser generator tool._

### Grammar file and base class changes
This change was created using the Trash program [trrename](https://github.com/kaby76/Domemtech.Trash/tree/main/trrename), and deals with almost all possible "symbol conflicts" that remain.

This change is essentially the result of applying the following two lines of Shell code to every grammar file:

    renameList="#available,#available_;#column,#column_;#else,#else_;#elseif,#elseif_;#endif,#endif_;#file,#file_;#function,#function_;#if,#if_;#line,#line_;#selector,#selector_;False,False_;Function,Function_;None,None_;Protocol,Protocol_;Self,Self_;True,True_;Type,Type_;_,__;__CLASS__,__CLASS___;__COLUMN__,__COLUMN___;__DIR__,__DIR___;__FILE__,__FILE___;__FUNCTION__,__FUNCTION___;__FUNCTION____LINE__,__FUNCTION____LINE___;__LINE__,__LINE___;__METHOD__,__METHOD___;__NAMESPACE__,__NAMESPACE___;__TRAIT__,__TRAIT___;__halt_compiler,__halt_compiler_;__import__,__import___;abs,abs_;abstract,abstract_;alignas,alignas_;alignof,alignof_;all,all_;and,and_;and_eq,and_eq_;any,any_;apply,apply_;array,array_;as,as_;asm,asm_;assert,assert_;associatedtype,associatedtype_;associativity,associativity_;async,async_;auto,auto_;await,await_;bin,bin_;bitand,bitand_;bitor,bitor_;bool,bool_;boolean,boolean_;break,break_;buffer,buffer_;byte,byte_;bytearray,bytearray_;callable,callable_;case,case_;catch,catch_;chan,chan_;char,char_;char16_t,char16_t_;char32_t,char32_t_;chr,chr_;class,class_;classmethod,classmethod_;clone,clone_;coerce,coerce_;compile,compile_;compl,compl_;complex,complex_;concept,concept_;const,const_;const_cast,const_cast_;constexpr,constexpr_;continue,continue_;convenience,convenience_;covariant,covariant_;debugger,debugger_;declare,declare_;decltype,decltype_;def,def_;default,default_;defer,defer_;deferred,deferred_;deinit,deinit_;del,del_;delattr,delattr_;delete,delete_;dict,dict_;didSet,didSet_;die,die_;dir,dir_;divmod,divmod_;do,do_;double,double_;dynamic,dynamic_;dynamicType,dynamicType_;dynamic_cast,dynamic_cast_;echo,echo_;elif,elif_;else,else_;elseif,elseif_;empty,empty_;enddeclare,enddeclare_;endfor,endfor_;endforeach,endforeach_;endif,endif_;endswitch,endswitch_;endwhile,endwhile_;enum,enum_;enumerate,enumerate_;eval,eval_;except,except_;exec,exec_;execfile,execfile_;exit,exit_;explicit,explicit_;export,export_;extends,extends_;extension,extension_;extern,extern_;external,external_;factory,factory_;fallthrough,fallthrough_;false,false_;file,file_;filter,filter_;final,final_;finally,finally_;float,float_;for,for_;foreach,foreach_;format,format_;friend,friend_;from,from_;frozenset,frozenset_;func,func_;function,function_;get,get_;getattr,getattr_;global,global_;globals,globals_;go,go_;goto,goto_;guard,guard_;hasattr,hasattr_;hash,hash_;help,help_;hex,hex_;hide,hide_;id,id_;if,if_;implements,implements_;import,import_;in,in_;include,include_;include_once,include_once_;indirect,indirect_;infix,infix_;init,init_;inline,inline_;inout,inout_;input,input_;instanceof,instanceof_;insteadof,insteadof_;int,int_;interface,interface_;intern,intern_;internal,internal_;is,is_;isinstance,isinstance_;isset,isset_;issubclass,issubclass_;iter,iter_;lambda,lambda_;lazy,lazy_;left,left_;len,len_;let,let_;library,library_;list,list_;locals,locals_;long,long_;map,map_;max,max_;memoryview,memoryview_;min,min_;mixin,mixin_;mutable,mutable_;mutating,mutating_;namespace,namespace_;native,native_;new,new_;next,next_;nil,nil_;noexcept,noexcept_;none,none_;nonlocal,nonlocal_;nonmutating,nonmutating_;not,not_;not_eq,not_eq_;null,null_;nullptr,nullptr_;object,object_;oct,oct_;on,on_;open,open_;operator,operator_;optional,optional_;or,or_;or_eq,or_eq_;ord,ord_;override,override_;package,package_;part,part_;pass,pass_;postfix,postfix_;pow,pow_;precedence,precedence_;prefix,prefix_;print,print_;private,private_;property,property_;protected,protected_;protocol,protocol_;public,public_;raise,raise_;range,range_;raw_input,raw_input_;reduce,reduce_;register,register_;reinterpret_cast,reinterpret_cast_;reload,reload_;repeat,repeat_;repr,repr_;require,require_;require_once,require_once_;required,required_;requires,requires_;rethrow,rethrow_;rethrows,rethrows_;return,return_;reversed,reversed_;right,right_;round,round_;select,select_;self,self_;set,set_;setattr,setattr_;short,short_;show,show_;signed,signed_;sizeof,sizeof_;slice,slice_;sorted,sorted_;static,static_;static_assert,static_assert_;static_cast,static_cast_;staticmethod,staticmethod_;str,str_;strictfp,strictfp_;struct,struct_;subscript,subscript_;sum,sum_;super,super_;switch,switch_;sync,sync_;synchronized,synchronized_;template,template_;this,this_;thread_local,thread_local_;throw,throw_;throws,throws_;trait,trait_;transient,transient_;true,true_;try,try_;tuple,tuple_;type,type_;typealias,typealias_;typedef,typedef_;typeid,typeid_;typename,typename_;typeof,typeof_;unichr,unichr_;unicode,unicode_;union,union_;unowned,unowned_;unset,unset_;unsigned,unsigned_;use,use_;using,using_;var,var_;vars,vars_;virtual,virtual_;void,void_;volatile,volatile_;wchar_t,wchar_t_;weak,weak_;where,where_;while,while_;willSet,willSet_;with,with_;xor,xor_;xor_eq,xor_eq_;xrange,xrange_;yield,yield_;zip,zip_"
    trparse -f FooBar.g4 | trrename -r "$renameList" | trprint > NewFooBar.g4

The arg to [trrename](https://github.com/kaby76/Domemtech.Trash/tree/main/trrename) was derived from reading the source code of the Antlr Tool [starting here](https://github.com/antlr/antlr4/blob/master/tool/src/org/antlr/v4/codegen/target/CppTarget.java#L59).

[trrename](https://github.com/kaby76/Domemtech.Trash/tree/main/trrename) requires two input: a semicolon-separated list of comma-separated symbol names that specify what the original symbol name is, and what to rename it to in a new grammar. [trrename](https://github.com/kaby76/Domemtech.Trash/tree/main/trrename) also requires for input the parse tree data produced by [trparse](https://github.com/kaby76/Domemtech.Trash/tree/main/trparse). Parse tree data is passed via stdin and stdout, following a pure functional approach towards modifying the grammar. The serialization of parse tree data has been optimized for speed, so the overhead to read and write parse tree data is negligible.
[trprint](https://github.com/kaby76/Domemtech.Trash/tree/main/trprint) reads the parse tree data and outputs the associated text.

While it should be easy to write steps to reverse the symbol renaming, there is additional code that must be hand tweaked. [trrename](https://github.com/kaby76/Domemtech.Trash/tree/main/trrename) does not rename symbols in target code, but I am considering an enhancement.

### Updates to "skip lists"
With the symbol renaming changes I outline above, I also changed the "skip lists" of the CI testing--list of grammars that should not be tested. Presumably, tests that used to fail would now pass. To update the skip lists, I removed the "?!" character string in [this line of Bash code](https://github.com/antlr/grammars-v4/blob/master/_scripts/regtest.sh#L36), then ran the test suite.

In addition, I am changing a few of the build script files:

* _scripts/templates/*/test.sh, which are used by "regtest.sh". We only want an oracle to state whether or not a grammar works or not. The behavior is now to give up on further parses if one parse fails for a grammar.
* _scripts/templates/{Go,Dart}/makefile. The tester.psm1 and makefiles did not agree on the steps for building the code. The default step for "make" should perform "export GO111MODULE=auto; go build Program.go" because the code should be checked for syntax errors before running the parser. For Dart, the default step for "make" should perform "dart compile exe cli.dart" for the same reason.


studentmain:
remember to update skip list

kaby76:
@studentmain At your convenience, please review this PR. Thanks! --Ken

_scripts/templates/Dart/makefile
```diff
@@ -10,6 +10,7 @@ SOURCES = $(GENERATED) \
 default: classes
 classes: $(SOURCES)
 	dart pub get
+	dart compile exe cli.dart
```
studentmain, line 13: These build steps generate executable file. You can use it in `run` target and add `classes` to `run`'s dependency. Same for Go makefile.

studentmain (approved):

_scripts/templates/Dart/makefile
kaby76: I need to update the makefiles to work with real targets and dependencies and to this right. I created [this issue](https://github.com/antlr/grammars-v4/issues/2159) for general makefile-related issues. I didn't program the makefile very well because I was just following a "how to do Dart/Go" somewhere, and really just wanted to get something to work. The [ANTLR4 Runtime for Dart](https://github.com/antlr/antlr4/blob/master/doc/dart-target.md#antlr4-runtime-for-dart) didn't help much. For now, let's push the current changes in, and fix the makefiles in another step. -K

kaby76:
@KvanTTT or @teverett This PR (2153) is ready to merge, with [other work separately tracked](https://github.com/antlr/grammars-v4/issues/2159). Thanks! -Ken

teverett:
@kaby76  thanks

Issue #2171, opened 2021-04-30
Disabling JavaScript CI


kaby76:
The Antlr 4.9.2 JavaScript target is currently not stable. In particular, a recent build of PR2157 resulted in a hung parse of grammars-v4/agc/, which then had to be manually canceled. The tests for JavaScript should be disabled until the JS target is more stable.

* We know of at least two issues that have been created describing problems with the target:  [3135](https://github.com/antlr/antlr4/issues/3135) and [3150](https://github.com/antlr/antlr4/issues/3150).
* The build for [PR2157](https://github.com/antlr/grammars-v4/pull/2157) had to be killed because it wasn't terminating.
* Parsing at worse case is known to be exponential, which could mean that even for the other targets, we could see an "excessive" parse time.

There are a few things to implement:
* We should wait till the Antlr JS target is updated.
* To combat excessive parse times, the template driver code could be multi-threaded, one thread containing a timer, and another the parse. After a certain period, the parsing thread could be canceled and the program exit in a controlled manner.

kaby76:
@ericvergnaud fixed a bug with the JavaScript target with https://github.com/antlr/antlr4/pull/3275 . So far, the results look promising. I've re-tested a number of the grammars for the JavaScript target, and it is now working much better--although not perfectly. The following are a few of the grammars in the repo I tested. I compared the test with straight Antlr4.9.2+runtime versus the test with a patched runtime with all the changes up to this point in time (Sep 25 2021).

At some point, CI with JavaScript should be added back in. I'm reopening this issue to serve as a placeholder on reintegrating the JavaScript target.

### antlr/antlr2

	../examples/ada.g2
	[@481,2636:2645='IDENTIFIER',<87>,76:18] line 76, col 18: extraneous input 'IDENTIFIER' expecting {'grammar', 'tree', TOKEN_REF, RULE_REF}
	[@485,2663:2666='SEMI',<87>,76:45] line 76, col 45: missing {'grammar', 'tree', TOKEN_REF, RULE_REF} at 'SEMI'
	(Parse driver program outputs the above errors, then does not terminate.)

All tests now pass.

### arithmetic

Failed previously with:

	../examples/paren1.txt
	[@5,15:15='+',<5>,1:15] line 1, col 15: missing {'+', '-', '*', '/'} at '+'
	Parse failed.

All tests now pass.

### basic

Failed on first test previously:

	bash test.sh
	../examples/a/data.bas
[...]

PR #2172, opened 2021-04-30
Fix for 2171

kaby76:
This is a fix for #2171. The Antlr 4.9.2 JavaScript runtime is unstable and should be disabled until the next release of the runtime. For a full explanation, see #2171.


studentmain (approved):

teverett:
@kaby76  thanks

PR #2176, opened 2021-05-02
Fixing JavaScript with Go target

mitar:
I made instructions more general, so that they do not get obsolete so quickly.


teverett:
@mitar  thanks

PR #2184, opened 2021-05-06
Adding MySQL 8 privileges

tommyk-gears:
Added the privileges listed at https://dev.mysql.com/doc/refman/8.0/en/privileges-provided.html to the MySQL parser (note that Table 6.3 on that page is incomplete, there are privileges not listed in that table but documented further down in the document, e.g. SERVICE_CONNECTION_ADMIN)


KvanTTT:
Welcome to the grammars repository and thank you for the request!

PR #2185, opened 2021-05-06
Fix mistakes in Swift grammar and add more examples

Marti2203:
I just saw the new Swift grammar and began testing it. I have not had issues for now, except the problems with the lexer tokens.


studentmain:
It's weird, why some test passed and other reported parse error.

Marti2203:
I really do not know enough about the testing capabilities of the repo to give a solid reason. I just changed 3 rules and I am using the same grammar now locally.

studentmain:
> It's weird, why some test passed and other reported parse error.

Now I know why. Those succeed tests are marked as skip in previous PR, they didn't run at all as intended... Let's just look at failed test.

Marti2203:
> > It's weird, why some test passed and other reported parse error.
> 
> Now I know why. Those succeed tests are marked as skip in previous PR, they didn't run at all as intended... Let's just look at failed test.

I tried running the tests, but the runtime needs the SwiftSupport class

Marti2203:
Okay, I have found an edge case in one the examples I have put. I will find a fix for it now.

Marti2203:
I have made a huge rework - had to split into lexer and parser grammars as lexer modes are needed for interpolation. Currently I am trying to catch small edge cases.

Marti2203:
I have made string interpolation work fully, captured some old swift constructs and added more keywords as identifiers.

teverett:
@Marti2203 thanks

Issue #2182, opened 2021-05-05
CI requires a watchdog program for runaway tests


kaby76:
Github Actions does not have a good quota system to handle problems with CI tests that exceed a reasonable time or resource limit. Go and Javascript based parsers have been problematic, requiring an administrator to kill the CI build.

CI tests must have a watchdog program to kill runaway tests. This must be added asap.

What I am thinking about is a command-line C# dotnet program that takes as args a command to exec. It starts a timer then process with args. If the timer expires, kill the process tree and return error. Otherwise return the exit code from spawned process. I don't know what C# offers to query user time vs system time, memory size, etc., but I would assume it has all that, and in a OS independent way. It should be a dotnet tool on nuget for easy installation. And it shouldn't have to be run in SU mode.

When adding to CI, the default timeout should be T seconds per parse, something that if worse case happened, T times N parsers run for all tests on one machine shouldn't be more than an hour. In other words, the CI build should take max one hour even if it is run parallel on X machines.

kaby76:
I have a preliminary version of a command-line tool that functions as a watchdog. It is [here](https://github.com/kaby76/Domemtech.Trash/tree/main/trwdog). There are problems in finding programs because the environment PATH variable isn't used correctly for MSYS2/Cygwin so `trwdog find / -name '*.g4'` does not find the right "find" program.

kaby76:
The program isn't ready. Dart on Windows could be a Bash or Cmd script. ProcessCreate() does not know how to spawn "dart". For regtest.sh, Bash likely spawns "dart", a Bash script. In Cmd or Powershell, "dart" likely invokes "dart.bat". Even if fed "trwdog dart.bat" from within a Cmd shell, the program fails with "The system cannot find the path specified.".

PR #2186, opened 2021-05-07
Fix for Issue 2182, CI requires a watchdog program for runaway tests

kaby76:
## Introduction
This is a fix to #2182, for CI builds that "hang", or just simply take way too long. Runaway tests have happened several times: [CI 742](https://github.com/antlr/grammars-v4/actions/runs/813077171), [CI 789](https://github.com/antlr/grammars-v4/actions/runs/811386045), [CI 764](https://github.com/antlr/grammars-v4/actions/runs/794538366), and typically occur for one of the slower targets, such as Go or Javascript. Hangs could happen for a number of reasons: a bug in the runtime or a poor choice of a data structure. Until now, the "skip lists" were used to manage this problem, which would involve testing first on the developer's machine before submitting the PR. Some changes [were made](https://github.com/antlr/grammars-v4/commit/7830e465420ce433fca22f2d50cc8de1e76c39c9#diff-b384a886f95430e252ae85c366cc5984d0a8d75701640a2caeb2eee3da67315dR19) to stop testing all parses once one parse fails, but it doesn't fix the "hang" problem.

Unfortunately, even if a developer can test the changes locally, testing is incomplete until the PR is actually submitted:
* The developer's machine supports only one OS.
* The developer's machine has different hardware than what is used for Github Actions.
* The developer's machine doesn't have enough resources, or the right resources installed.
* Github Actions is not implemented on a developer's machine.
* If the grammar is misspelled in the list, the regular expression generated from the skip list won't match the actual name of the grammar.
* When the grammar is added, the default behavior for CI is to test across all targets. Modifying the skip is an afterthought.

## Overall solution
This change adds a [watchdog program](https://github.com/kaby76/Domemtech.Trash/tree/main/trwdog) wrapper to the tests in the templates for each target. The watchdog program is very simple: spawn a process of a program specified by the command-line arguments to the program. At the same time, start a timer for 5 minutes. If the program does not terminate within the timeout period, kill the process tree and return an error. Otherwise, return the result of the spawned process.

## Implementation details
Changes were made to:
* The .github actions, to set up the environment by installing the watchdog program. Two files were modified to install the newest version of trgen and trwdog, the watchdog program, when the build starts.
* "trwdog" is invoked with the command line args to call the target toolchain parser. About two dozen files required the change.  I removed tabs in the source code since my editors don't handle tab expansion correctly when programs are debugged.
* The timeout period for the watchdog program defaults to 5 minutes. While one can pass an option to trwdog for a different timer, I felt it would be too verbose to pass the parameter on each trwdog call in all ~20 or so files. Instead, if one wants to set a different timeout period, create a JSON file for the option in ~/.trwdog.rc. It is likely I will add other options.
* The watchdog program currently only performs time-outs, so it doesn't check high memory usage as I mentioned in the issue. Usually, the two are interrelated.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -66,8 +66,8 @@ jobs:
     - name: Install trgen
       shell: pwsh
       run: |
-        dotnet tool install -g trgen --version 0.6.5
-        dotnet tool install -g trwdog --version 0.6.2
+        dotnet tool install -g trgen --version 0.7.0
+        dotnet tool install -g trwdog --version 0.7.0
         if ("${{ matrix.os }}" -eq "ubuntu-latest") {
             echo "$HOME/.dotnet/tools" >> $env:GITHUB_PATH
         }
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -66,8 +66,8 @@ jobs:
     - name: Install trgen
       shell: pwsh
       run: |
-        dotnet tool install -g trgen --version 0.6.4
-        dotnet tool install -g trwdog --version 0.6.1
+        dotnet tool install -g trgen --version 0.6.5
+        dotnet tool install -g trwdog --version 0.6.2
         if ("${{ matrix.os }}" -eq "ubuntu-latest") {
             echo "$HOME/.dotnet/tools" >> $env:GITHUB_PATH
         }
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -67,6 +67,7 @@ jobs:
       shell: pwsh
       run: |
         dotnet tool install -g trgen --version 0.5.3
+        dotnet tool install -g trwdog --version 0.6.0
         if ("${{ matrix.os }}" -eq "ubuntu-latest") {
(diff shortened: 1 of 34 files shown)


kaby76:
Excellent! Ubuntu/Python3/Fasta failed because it took over 60s. The test was _apparently_ killed 

    2021-05-07T12:21:23.5388231Z ./fasta test /home/runner/work/grammars-v4/grammars-v4/fasta/examples/NC_009925.fna failed
    2021-05-07T12:21:23.5413488Z Test completed, time: 00:02:16.0514236
    2021-05-07T12:21:23.5440623Z ./fasta failed

The output "Test complete, time" is an aggregate time for the test, and it would not have failed if the test completed in the timeout period. In past CI builds, it would parse cleanly.

But, I also do see "Parse succeeded" nor "Parse failed", as I see in the CSharp tests. It seems that this may be because the CSharp driver outputs to stderr, while all others output to stdout the completion line. This must be fixed.

kaby76:
Excellent. All looking good. I will now adjust fasta time out period, then it will be complete.

teverett:
@kaby76 thanks

PR #2187, opened 2021-05-07
Rename and Update to ESM the sql/plsql/JavaScript Lexer and Parser Base Files

cofad:
There are currently (2) compatibility issues when trying to build and use the lexer/parser for the [sql/plsql/JavaScript](https://github.com/antlr/grammars-v4/tree/master/sql/plsql/JavaScript) target:

1. The parser/lexer base files names do not match what the antlr4 generated parser/lexer expect
2. The parser/lexer base files are written in CommonJS, but the antlr4 generated parser/lexer is written in ESM

This PR addressses these (2) issues with the following changes:

1. Rename the base files as follows:
    1. ["PlSqlBaseLexer.js"](https://github.com/antlr/grammars-v4/blob/master/sql/plsql/JavaScript/PlSqlBaseLexer.js) -> "PlSqlLexerBase.js" 
    1. ["PlSqlBaseParser.js"](https://github.com/antlr/grammars-v4/blob/master/sql/plsql/JavaScript/PlSqlBaseParser.js) -> PlSqlParserBase.js"
2. Convert the lexer/parser base files from CommonJS to ESM.
    1. The conversion is based on the TypeScript files

These changes will bring the file naming in line with the file names for the other language's base files and allow the generated parser/lexer to find and import the base files properly.


KvanTTT:
Welcome to grammars and thank you for the request!

Issue #2156, opened 2021-04-24
Java8, Java9, and Apex all need to be fixed


kaby76:
Java8, Java9, and Apex all need fixing for target-independent coding.

* The Java9 grammar is currently not strictly target-independent because it contains semantic predicates.
* The rules with semantic predicates (e.g., [here](https://github.com/antlr/grammars-v4/blob/master/java/java9/Java9Lexer.g4#L487)) are based on the Java Language Spec. There are implementations in grammars-v4/java/java/ and grammars-v4/java/java8 that are predicate-free (see the rules in the Java8 grammar starting [here](https://github.com/antlr/grammars-v4/blob/e53d7a1228505bfc80d8637808ef60e7eea92cc2/java/java8/Java8Lexer.g4#L469)). It's possible to rewrite this without action code.
* The apex/ grammar is essentially a derivative of Java9. It too can be rewritten to not use action code.
* Java8 is a predicate-free solution, but it can be improved: Antlr supports a 32-bit code point value in the grammar and runtime (see [this](https://github.com/antlr/antlr4/pull/1633)).
* Not all targets or runtimes can handle the 32-bit code point value notation in Antlr4 grammars. Some targets will need to have a separate grammar and source code. The Antlr4cs tool does not accept 32-bit code points. The encoding feature in the Dart runtime doesn't seem to work.
* The CI driver scripts should be updated to be able to create a stream that can handle an encoding, although it wouldn't be used in any tests.
* '<>' is implicitly defined in a parser rule. It needs to be explicitly defined in the lexer.
* Java8 has commented out rules [here](https://github.com/antlr/grammars-v4/blob/e53d7a1228505bfc80d8637808ef60e7eea92cc2/java/java8/Java8Lexer.g4#L449). The old rules should be deleted for readability.
* The Java Language Spec defines JavaLetter and JavaLetterOrDigit operationally. A question should be asked of the authors of the Java Language Specification as to why it's necessary to use an operational definition.

studentmain:
StringTemplate grammar is using Java identifier rule too.

kaby76:
[...]

PR #2189, opened 2021-05-08
Fix for Issue 2188, Apex requires changing in order to make it target independent

kaby76:
This PR changes the apex/ grammar for target independence.
* I replaced the rules for JavaLetter and JavaLetterOrDigit, from ones that contained semantic predicates, to ones that use 32-bit Unicode code points and general categories. These rules were computed using a program that queries the Java runtime.
* I added Antlr4cs-specific files since Antlr4cs does not support 32-bit Unicode code points and general categories. Antlr4cs is old, several years behind the current version of Antlr4, but still requires support here.
* I updated the builds to use the latest trgen tool. This is required so that Antlr4cs builds and tests work. Note, the CI build does not test JavaScript nor Antlr4cs targets, but I expect that to change at some point. I've included this update so that if we were to do that, it would be easy to just flip a switch.
* I removed apex from the skip lists because it now works for all targets (tested on my machine, including Antlr4cs). To be clear though, the JavaScript target builds but does not work because JavaScript 4.9.2 has a serious bug in the runtime (what I call "mutating transitions of an ATN"). But, I have a fix for the 4.9.2 runtime, and it will work whenever I check that in and get a new release of Antlr4.


kaby76:
@KvanTTT Ok, let's start here, a much cleaner (and correct) PR. Can you check this over? My plan is to work on each grammar and submit separately.

studentmain:
I suggest use \uxxxx instead of \u{xxxx}, when xxxx < 0xffff, they're same. Maybe remove Antlr4cs' own g4 too.

Antlr4cs has incomplete Unicode support is not  a big problem. As long as nobody use some really rare Chinese characters (They're too rare, as a native Chinese speaker, I can't read them) in identifiers, it will works. It's behavior is not correct, but it's very unlikely we meet the edge case in the wild. And it's a runtime's problem, it affects every grammars, and inconsistent with official runtime.

KvanTTT:
> Maybe remove Antlr4cs' own g4 too.

I also suggest removing duplicated g4 grammars from Antlr4cs subdirectory. It should be possible to use grammars from `apex` directory and superclasses from `apex/CSharp`, `apex/Java`, and other runtime-specific directories.

kaby76:
> I also suggest removing duplicated g4 grammars from Antlr4cs subdirectory. It should be possible to use grammars from `apex` directory and superclasses from `apex/CSharp`, `apex/Java`, and other runtime-specific directories.

Well, I think that would be moving backwards, so I wouldn't recommend it. But, we could give and take on what would be best going forward.

First, let's just go over the discussion on how I got to this PR. This discussion is main in [PR 2157](https://github.com/antlr/grammars-v4/pull/2157) for issue #2156. Here's a recap.

1) The current apex/apex.g4 contains Java predicates in the grammar, and corresponds to what the Java Language Spec (JLS) says in [section 3.8](https://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.8). What seems like many weeks ago, I was going to just change the apex grammar to make it "target-independent" format, i.e., replace the inlined Java code with a call to a method in a base class. I also added base classes to the grammar for some targets. As I said, I actually did do a bit of this work, and started exactly with the apex/ grammar, which you see [here](https://github.com/antlr/grammars-v4/pull/2157/commits/d07afa11abfe7590ef25acb1e2123931f1b8db07). Understand, of course, this doesn't actually make the grammar "target-independent", and in order to do that, you have to add "superClass" for the lexer. And, as soon as you add a superClass option, the grammar has to be split, which I did.
2) April 25, @studentmain [suggested](https://github.com/antlr/grammars-v4/pull/2157#discussion_r619599488) that we move towards a predicate-free solution, and pointed out his code change in java/java8. This solution removes the predicates completely. How? A critical and wonderful observation was made: why is the JLS operationally defined? Can't one pre-compute the character sets in the predicates? I thought--and still do thnk--that this is the right way to proceed. It gets rid of a lot of base class support code, and the grammar can remain a combined grammar. 
3) Later on that day, on April 25, you reviewed the code and [suggested changes](https://github.com/antlr/grammars-v4/pull/2157#pullrequestreview-644034661). However, I felt at that point it was worth investigating and finding a predicate-free solution instead.
4) On April 26, I noted [here](https://github.com/antlr/grammars-v4/pull/2157#issuecomment-826939487) that the current solution in java/java8/ does not handle code points >16-bits. The current solution with predicates I thought would, but @studentmain mentioned that it wasn't.
5) But, on April 26, @studentmain suggested [here](https://github.com/antlr/grammars-v4/pull/2157#issuecomment-826980619) to use a new feature in Antlr4 that solves this 32-bit code point problem! We'd just never had taken advantage of it before.
6) On April 26, I posted [code](https://github.com/antlr/grammars-v4/pull/2157#issuecomment-826986932) to compute the sets for JavaLetter and JavaLetterOrDigit, and rewrote the rules to be predicate-free.
7) I then tested this combined, predicate free grammar on 50k Java source files (I assume 16-bit code points, standard "Unicode") over a day or two. The solution to have no performance degradation. I also tested it on encodes utf-8, utf-16, utf-32 and found the grammar worked with C# and Java drivers. Note, I had to modify the templates to create an CharStream with different encodings. But, you don't need to change the driver--the tests and 50k files mostly pass.
8) On April 28 [I pointed out](https://github.com/antlr/grammars-v4/pull/2157#issuecomment-828884848) that Antlr4cs did not accept the reworked grammars. Antlr4cs is several years behind Antlr4 but people still use it. As a fact, I have been very vocal in telling telling people to stop using Antlr4cs: it is old; it is not maintained, it is missing features that exist in Antlr4. We don't actually test it in the CI builds, but I have everything to do that, and it's checked in. Critically, it *cannot* support 32-bit code points!

Antlr4 does support 32-bit code points; Antlr4cs does not. We can implement a combined grammar for all targets of the standard Antlr4 using a combined, 32-bit Unicode code point grammar. Antlr4cs fails to parse the grammar.

The question is how best to move forward?

As I said, the JLS is operationally defined. I have sent an email to the Standards Committee for the language, but I'm not hopeful that I will get an answer, let alone have an impact on getting the JLS changed from an operational def to a static def. The current grammar follows the JLS; the PR does not, but it is predicate-free, and works with the standard Antlr4 toolchain.

Antlr4cs is unsupported. I don't see a long-term viability to that code because it is missing features in Antlr4, and presents a different API surface for C#. I don't see Harwell making any changes here. So, I see it more as an exception. We should not be hammering the code to fit backwards into Antlr4cs. That's why I chose to fork the old grammar for apex/ and place it into apex/Antlr4cs. That's why there are no apex/CSharp, no apex/Java, etc. support files.

Possible solutions:
* Keep with the old semantic predicate solution, but just make them "target-independent" (i.e., refactor the predicate into a base class method that returns a bool).
* Move forward with the predicate-free implementation.
a) Port the old predicate-containing grammar and place it into apex/Antlr4cs. I can then leave the grammar as a combined grammar, and it makes it easier for people to understand. Or I could combine the grammars, and just directly add in the base class code into the .g4.
b) Use the incomplete, 16-bit Unicode solution that currently exists in java/java (or java/java8, you need to pick) and apply that here. As @studentmain notes above, it's not completely correct, but it does work typically. Who is going to be using Antlr4cs and Chinese character encodings? Goodness, I agree.

Votes and opinions please. --Ken

studentmain:
Just delete Antlr4cs folder. When using Unicode category, if runtimes support 32 bit Unicode, they'll recognize 32 bit characters. Otherwise they can still recognize 16 bit characters.

In case some old tools not happy, replace \u{0000xxxx} to \uxxxx. This change has no effect to 32 bit Unicode support, it's always enabled and can't be disabled.

kaby76:
I'm going to change this back to draft, and work on other things. We don't have a consensus, and don't want this accidentally merged in until we do. No hurries. Ken

kaby76:
I'm going to close out this PR and bug for now. I thought I had an alternative solution, but it's far more complicated than I could have ever imagined because of the changes post-Antlr4.7, and Unicode support per target in various states. I'll open a new bug--likely a week or so from now--when I have a thorough description of all the Java-based grammars and Unicode support across all targets, all tools, all grammars.

PR #2190, opened 2021-05-08
declarationStatement being interpreted as an expression

kstallard:
In the following code:

```c++
int f()
{
    MyClass *test= new MyClass;
    test->foo();
}
```

`MyClass *test` was being interpreted as a `multiplicativeExpression`.  

I've included the parse tree before and after to illustrate

parseTreeBefore

parseTreeAfter

I also added some test files. There are a couple of .cpp files there, one I don't think is a really good test, it is way too big for Grun to absorb and it is too complex. I think it should be broken down into more simpler snippets.

Also the raw string example breaks the grammar. I'll take a look and see where the problem might be, I think it is a lexer rule that needs some TLC. So I didn't create .tree files for those.


KvanTTT:
Could you also please add your test sample to the examples directory?

kstallard:
> Could you also please add your test sample to the examples directory?

Is this what you had in mind? 

Let me know if you need anything else....

Thanks

studentmain:
Add a .tree file (`grun -tree`'s output) into test, to make sure parser generate correct parse tree.

kstallard:
> Add a .tree file (`grun -tree`'s output) into test, to make sure parser generate correct parse tree.

I've got the tree output, but I'm not sure where to put it.  The CPP folder doesn't have a test folder or anything similar.  I found the TestCPP.java, but doesn't seem to reference any kind of test files. Can you point me in the right direction?

Thank you,
Kevin

studentmain:
> > Add a .tree file (`grun -tree`'s output) into test, to make sure parser generate correct parse tree.
> 
> @studentmain So it appears that some grammars have *.tree files in their examples folders.  There are current none for CPP.  Looks like I need to place the tree file in the examples folder along with the test .cpp file.  It appears that the name needs to match the associated .cpp file that contains the example?

Yes, you are right.

kstallard:
> > > Add a .tree file (`grun -tree`'s output) into test, to make sure parser generate correct parse tree.
> > 
> > 
> > @studentmain So it appears that some grammars have *.tree files in their examples folders.  There are current none for CPP.  Looks like I need to place the tree file in the examples folder along with the test .cpp file.  It appears that the name needs to match the associated .cpp file that contains the example?
> 
> Yes, you are right.

Done

kstallard:
Hmm...nice...I just broke everything...looking....

kstallard:
Looks like I could make tests for the other example code and I should probably add a test for a fix I submitted a few weeks ago...

kstallard:
Added some test files.  There are a couple of .cpp files there, one I don't think is a really good test, it is way to big for Grun to absorb and it is too complex.  I think it should be broken down into more simpler snippets.

Also the rawstring example breaks the grammar.  I'll take a look and see where the problem might be, I think it is a lexar rule that needs some TLC.  So I didn't create .tree files for those...

kstallard:
@KvanTTT  @studentmain  I think git, when retrieving test2.cpp is adding a <cr> to the <lf> in the test.  It generates different grammar than on unix.

I guess it is possible there is a problem with the grammar.  Either way, I think I will remove it for now.

kstallard:
I think this is ready, I don't have anything to add to this PR (unless someone needs me too).

csharp build seems stuck...perhaps it just needs more time, but has been stuck on the setup step for 20 minutes?

KvanTTT:
Thanks!

PR #2193, opened 2021-05-10
Remove duplicate alternatives from the `theOperator` in CPP grammar.

chengniansun:
There are two `Assign` in the `theOperator` parser rule in the CPP grammar.


KvanTTT:
Welcome to the grammars repository and thank you for the pull request!

PR #2205, opened 2021-05-21
Fix for Issue 2204, "CI build fails in Dice"

kaby76:
There's a problem with the build in that the Antlr tool isn't being run on both grammars correctly. Dice is a split grammar, but after reading over this, I noticed a problem: the lexer grammar was labeled as a lexer grammar, but the parser grammar was actually a "combined" grammar, not a proper "parser grammar". I changed the name of the "combined" grammar to be a "parser grammar", changed the file name to reflect the parser name, and updated the POM file. I'm hoping this takes care of the build problem.

### DIFF
--- a/dice/DiceNotationParser.g4
+++ b/dice/DiceNotationParser.g4
@@ -19,7 +19,7 @@
  * 
  * This is the notation which RPGs and other tabletop games use to represent operations with dice.
  */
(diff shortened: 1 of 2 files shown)


teverett:
@kaby76  thanks

Issue #1555, opened 2019-11-07
Remaining problem of JS grammar


studentmain:
Unfinished part of #1553

these syntax shoudn't work:
  - `1_2___3;` (only one `_` allowed , e.g.:`1_2_3`)
  - `await a = b;` (await expression can't be lvalue)
  - `var [...a,...b] = c;` (array destruction assign is kinda different with array literal)

these syntax should work:
  - <pre>`aaa${`bbb`}ccc`</pre> (nested template literal)
  - ~~`for (a of b){}`~~ (problem not in grammar itself?)
  - `function async(){}` (allow _some_ keyword as identifier)
  - `with(a){}` (legacy `with`, only works in sloppy mode)

need improve:
  - `classElement` rules, as class field proposal add some new modifier into class elements
  - grammar perfomance, 5x slower than hand written JS parser (not very important)
  - generate more ESTree-like parse tree (not very important)
  - keep track with current standard (long term)

yanmingsohu:
**Should I separate the different versions of the javascript syntax description files?**
ES6 one file, ES7 one file, and so on.

Because, whenever the grammar description file changes, the generated header file also changes, and the dependent project needs a lot of modifications to be compiled.

Sometimes a lower version of the grammar description file is required.

studentmain:
I think no. JS changes very quickly. We'll soon have 20+ description file.

You can just copy paste g4 file and never update it. Or find old version in git history.

PR #2206, opened 2021-05-22
Added support for nested template strings

bkiers:
Added support for nested template strings:

```js
`aaa${`bbb`}ccc`
```

Mentioned in https://github.com/antlr/grammars-v4/issues/1555


teverett:
@bkiers  thanks

PR #2214, opened 2021-05-26
Fix JavaScript newExpression parsing. 

### LINKED ISSUE #2178: new Class(argument) is not parsed correctly for JavaScript
So current grammar parses `new Class(argument)` not as `NewExpression`with `singleExpression` and `arguments`, but as `NewExpression` with only `singleExpression`, where then in turn `singleExpression` is parsed as `singleExpression arguments`.

The relevant bits of grammar are here:

```g4
arguments
    : '('(argument (',' argument)* ','?)?')'
    ;

argument
    : Ellipsis? (singleExpression | identifier)
    ;

singleExpression
    : anoymousFunction                                                      # FunctionExpression
    | Class identifier? classTail                                           # ClassExpression
    | singleExpression '[' expressionSequence ']'                           # MemberIndexExpression
    | singleExpression '?'? '.' '#'? identifierName                         # MemberDotExpression
    | singleExpression arguments                                            # ArgumentsExpression
    | New singleExpression arguments?                                       # NewExpression
    | New '.' identifier                                                    # MetaExpression // new.target
    ...
```

So the issue is that `NewExpression` has `?` for `arguments` in `New singleExpression arguments?`, so a different parsing path is chosen, where `New singleExpression` is followed, and then `singleExpression` expands back into `singleExpression arguments`.

Is there a way to fix grammar to prefer `NewExpression` case? If I change things to `New singleExpression arguments` then parsing is done correctly, but of course `new` with any arguments does not succeed anymore.

studentmain:
close #2178

It seems `arguments?` is "non-greedy", it prefer not to match `arguments`. So I use a dumb solution to make it greedy.

cc: @mitar


mitar (approved):

mitar:
Awesome! Thanks!

teverett:
@studentmain  thanks

PR #2215, opened 2021-05-26
Fix a small and a major bug in the Go grammar

Marti2203:
The lineTerminatorAhead() method did not consider LINE_COMMENT which meant that it could get stuck in an infinite loop.
A label could be statementless which is fully valid.


Marti2203:
Yes, I have changed the indentations, VS Code is at fault, if it is a problem, I would like someone with the proper indentation setup to fix it.

kaby76:
@Marti2203 It looks like the grammars-v4/golang/GoParser.g4 and grammars-v4/golang/Go/GoParser.g4 are out of sync. This is why the CI build failed:

	Parse succeeded.../examples/statementlessLabel.go
	line 35:1 extraneous input '}' expecting {'break', 'func', 'interface', 'select', 'defer', 'go', 'map', 'struct', 'chan', 'goto', 'switch', 'const', 'fallthrough', 'if', 'type', 'continue', 'for', 'return', 'var', 'nil', IDENTIFIER, '(', '{', '[', ';', '!', '+', '-', '^', '*', '&', '<-', DECIMAL_LIT, BINARY_LIT, OCTAL_LIT, HEX_LIT, FLOAT_LIT, IMAGINARY_LIT, RUNE_LIT, RAW_STRING_LIT, INTERPRETED_STRING_LIT}
	line 58:0 no viable alternative at input 'func() {\n\tfor {\n\t\tif true {\n\t\t\tprintln()\n\t\t}\n\t\tbreak // want `the surrounding loop is unconditionally terminated`\n\t}\n}\n'
	line 58:0 no viable alternative at input '<EOF>'


 	cd grammars-v4/golang
 	diff GoParser.g4 Go/
	135c135
	< labeledStmt: IDENTIFIER COLON statement?;
	---
	> labeledStmt: IDENTIFIER COLON statement;
 	...
	274c274
	< 		| slice_
	---
	> 		| slice
 	...
	352c352
	< slice_:
	---
	> slice:

The reason why there are two GoParser.g4 is because the actions for the Go target have to be different from the other targets: a "p." is required. When tje grammar is tested, [trgen](https://github.com/kaby76/Domemtech.Trash/tree/main/trgen) is called to create a driver program in Go. It first creates grammars-v4/golang/Generated, then copies into Generated/ the files grammars-v4/golang/*, then copies grammars-v4/golang/Go/* overwriting GoParser.g4, since it's Go specific, and finally instantiates all the template files in grammars-v4/_scripts/templates/Go/* into grammars-v4/golang/Generated/.

If you'd like to test the Go target, the procedure is:

	cd grammars-v4/golang
	trgen -t Go --template-sources-directory ../_scripts/templates/
	cd Generated
	make
	make test

--Ken

Marti2203:
Hi @kaby76 ,
Thanks! I needed that info but forgot your username.😅
I will fix the PR in an hour, cheers!

Marti2203:
@kaby76 Thanks for reminding me about the difference of the grammars... That question mark was the important thing...
@teverett the PR ready to be accepted

kaby76 (approved):

teverett:
@Marti2203 thanks

PR #2216, opened 2021-05-27
Flatbuffers.g4 update: fixed size arrays, escape sequences in strings, inf/nan for floats, keywords as identifiers,...

stefan301:
- support fixed size arrays (size can be specified as decimal or hexadecimal integer)
- support escape sequences in strings
- allow keywords as identifiers like flatc (e.g. field named attribute)
- support aliases in union values
- add inf and nan to floating point literals
- add native_include

-> with this grammar all .fbs in the flatbuffers tests folder can be parse except from union_vector.fbs because of the union alias for string.

### DIFF
--- a/flatbuffers/FlatBuffers.g4
+++ b/flatbuffers/FlatBuffers.g4
@@ -1,81 +1,197 @@
-
 grammar FlatBuffers ;
 
 // Parser rules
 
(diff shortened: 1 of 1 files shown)


teverett:
@stefan301 thanks

Issue #2222, opened 2021-06-06
sql/plsql for Python3 target doesn't work


kaby76:
This issue describes what [PR#2095](https://github.com/antlr/grammars-v4/pull/2095) attempts to solve.

1) The Python3 target for sql/plsql does not run:

	$ make test
	bash test.sh
	../examples/aggregate01.sql
	Traceback (most recent call last):
	  File "Program.py", line 7, in <module>
	    from PlSqlLexer import PlSqlLexer;
	  File "C:\Users\kenne\Documents\GitHub\issue-2219\sql\plsql\Generated/PlSqlLexer.py", line 14, in <module>
	    from PlSqlLexerBase import PlSqlLexerBase
	ModuleNotFoundError: No module named 'PlSqlLexerBase'

2) The base class for the parser and lexer in the Python3 target are currently incorrectly named. If we want to use the grammar, the base class for the parser must be [PlSqlParserBase](https://github.com/antlr/grammars-v4/blob/master/sql/plsql/PlSqlParser.g4#L25).

PR #2223, opened 2021-06-06
Ready for review--Fix for issue #2222

kaby76:
This is a solution for #2222, which allows sql/plsql to work across targets. This supersedes [PR#2095](https://github.com/antlr/grammars-v4/pull/2095). So after merging this PR (PR#2223), the old [PR#2095](https://github.com/antlr/grammars-v4/pull/2095) should be closed. That PR#2095 has been in the queue for several months and hasn't been updated in response to a request from @KvanTTT to remove the forked grammar from the Python3 sub-directory. While this PR removes that specific copy of the grammar, my understanding is to merge all forked copies of grammars across all targets. But, I don't think that is possible.

In fact, as far as I can tell, it is impossible to provide complete cross-target support with one lexer/parser separated grammar, due to bugs in Antlr. The best solution, presented in this PR, is to fork a copy of the grammar for particular targets that we know contain a bug (Dart), and when the bug is fixed, just remove the forked grammar.

In particular, the generated Dart code should include the parser base class, but it doesn't because the Dart target .stg file does not do so. Most other targets generate the appropriate code. For the Cpp, there is a [special-case include syntax for the Cpp target only](https://github.com/kaby76/grammars-v4/blob/issue-2222/sql/plsql/PlSqlLexer.g4#L27), which is a terrible solution, but it works. The bug in the Dart target has been noted [here](https://github.com/antlr/antlr4/issues/3124).

Alternatively, a preprocessor for grammars, something like what @studentmain suggested (see [PR#2234](https://github.com/antlr/grammars-v4/pull/2234)), could process the grammar to deal with the include problem with Antlr.

Ideally the dart.stg file should be fixed. But, Antlr hasn't been updated for over a month, and I'm not confident that there will be any more releases and bug fixes!

1) This solution uses a universal parser and lexer grammar that does not require grammar forking (except for Dart due to a bug in Antlr, see below) or a preprocessor step to convert "this." to "self." Similar problems exist in other grammars: [golang](https://github.com/antlr/grammars-v4/blob/master/golang/Go/GoParser.g4), [ecmascript](https://github.com/antlr/grammars-v4/blob/master/javascript/ecmascript/Go/ECMAScript.g4), [python3](https://github.com/antlr/grammars-v4/blob/master/python/python3-py/Python3.g4), [antlr4](https://github.com/antlr/grammars-v4/blob/master/antlr/antlr4/Python3/ANTLRv4Lexer.g4), and others. The solution here is to just declare "self" for certain targets, and use that in the semantic predicates in the grammar.

2) The PR adds sql/plsql to the CI build for Python3 since the grammar now works for that target.

3) Although the Dart target base class code works in most cases, there is a crash parsing examples/analyze.sql. [I've reported the bug](https://github.com/antlr/antlr4/issues/3198). Consequently, it's still on the skip list for Dart, but I provide the base class code in case someone wants to use the Dart target. Note, Dart requires a forked grammar for the header includes. That is a bug in Antlr because the base class needs to be included in the generated parser, but Antlr doesn't do that. A similar problem exists for the Cpp target. In that case, the solution is even worse, providing a special-case include just for Cpp. Honestly, it should just generate the right code for the right target. It has all the info it needs with `superClass=PlSqlParserBase`.

    @parser::postinclude {
    #include <PlSqlParserBase.h>
    }

4) The Go target should work, but it doesn't. That's because the grammar is just friggin' huge. It crashes the Go compiler. Consequently, it's still on the skip list for Go. I haven't included the code for Go because I'm not sure how to work past the compiler crash to test the super class code.


kaby76:
Converting this PR back to a draft--not ready yet. It turns out that for the Go target, my environment is messed up. The compiler seems to be crashing for whatever reason on anything. So, I will try to fix that, and update the Go target code here, and check that in.

kaby76:
Converting back to draft--"expressions" was changed to "expressions_", but I don't recall why at all it was done. It obviously is due to a symbol conflict. "expressions" is also used in other grammars. This must be resolved and explained fully, which target has the problem.

kaby76:
It's Dart. When I compile, I get:

    PlSqlParser.dart:148399:22: Error: 'expressions' is already declared in this scope.
    ExpressionsContext expressions() => getRuleContext<ExpressionsContext>(0);

Looking over the code that's generated, it looks like the Antlr 4.9.2 tool generates two expressions(). I think it's a bug.

For now, I'll keep the rename in the Dart version for the grammar, but roll back the "universal" grammar renaming of expression.

I'll also raise a bug in the Antlr4 tool.

kaby76 (commented):
Looks good. The only changes in the grammars are actions with the inclusion of "self." references, and, of course, the fork for Dart because it needs a header include statement and "expression" rename bugs. When those are fixed in Antlr tool, then get rid of the forked copy.

teverett:
@kaby76  is this ready to pull?

kaby76:
> @kaby76 is this ready to pull?

@teverett Yes, it's ready for review and pull (since June 10).

teverett:
@kaby76  thanks

PR #2224, opened 2021-06-06
[PL/SQL] Change into_clause to accept the same targets that are valid in assignment_statement

felipebz:
In PL/SQL there are many expressions that can be used as "assignment target" like simple variables, records, collections and host variables. Those expressions are correctly recognized in the assignment_statement rule but they aren't accepted in the into_clause rule. This changes fixes the into_clause.


kaby76:
There are a slew of other "into_clause"-like structures in the language. Have you considered any of these (see below, a partial list)?

* [into_clause1](https://github.com/antlr/grammars-v4/blob/030c60aef6003c2fdf6dd18dc1a4cc6c5e05eaf7/sql/plsql/PlSqlParser.g4#L999)
* [insert_into_clause](https://github.com/antlr/grammars-v4/blob/030c60aef6003c2fdf6dd18dc1a4cc6c5e05eaf7/sql/plsql/PlSqlParser.g4#L3616)
* [dependent_exceptions_part](https://github.com/antlr/grammars-v4/blob/030c60aef6003c2fdf6dd18dc1a4cc6c5e05eaf7/sql/plsql/PlSqlParser.g4#L390) "INTO tableview_name"
* [exceptions_clause](https://github.com/antlr/grammars-v4/blob/030c60aef6003c2fdf6dd18dc1a4cc6c5e05eaf7/sql/plsql/PlSqlParser.g4#L2578) "INTO tableview_name"

It seems everyone gets a shot at throwing something onto this grammar, but there's no cogent guiding view. Unfortunately, [Oracle's doc](https://docs.oracle.com/en/database/oracle/oracle-database/21/lnpls/database-pl-sql-language-reference.pdf), page 14-158, is all we have.

felipebz:
> Have you considered any of these (see below, a partial list)?

Yes, I have. I found the same error in the [fetch_statement](https://github.com/antlr/grammars-v4/blob/030c60aef6003c2fdf6dd18dc1a4cc6c5e05eaf7/sql/plsql/PlSqlParser.g4#L3218) but I didn't included the fix here yet.

> It seems everyone gets a shot at throwing something onto this grammar, but there's no cogent guiding view.

I agree with you. Btw, it's not clear to me if the grammars in this repository are meant to be just examples or if there's really an effort to implement a "complete" parser.

> page 14-158, is all we have, and that's for RETURNING INTO, yet describes into_clause used elsewhere. Geezus Oracle.

That's true. The Oracle documentation seems to be great at first, but when we start to compare the syntax described there to the real implementation becomes very clear that there are many omissions. I can say this from my experience. I've been developing a PL/SQL parser since 2015 and, gosh, the Oracle docs aren't that good.

I have lots of test case on my project and I'm currently checking if I could replace my own parser by this ANTLR grammar, but I'm having many issues parsing some real-world code. I'm not sure if you folks prefer that I report and/or fix these issues here or I should simply copy the grammar and maintain my own fork.

kaby76:
@felipebz Sounds great.

Many of the grammars in the repo are examples, but a few try otherwise, e.g., java/java. The problem, of course, is that most programming languages don't start from the grammar but from implementation, and then just document the grammar afterwards.

KvanTTT:
> I'm not sure if you folks prefer that I report and/or fix these issues here or I should simply copy the grammar and maintain my own fork.

I prefer the unification of all efforts as much as possible. Thus, fix issues here if possible.

Thank you for your contribution and welcome to the grammars repository!

Issue #1983, opened 2021-01-07
Scala grammar testability issues and possible fix


Marti2203:
Hello,
Two weeks ago I started to play around with the Scala Grammar. 
There were a couple of issues - the grammar does not support XML patterns(will have to add that) and the current implementation did not work(it passed tests,but it was not made to try to reach EOF so the parser stopped at the first moment it could not be sure about the input, which was just before the opening brace I suppose).

After a lot of tweaking I have made it working, but with one caveat - I had to make a custom TokenStream as the parser by specification needs to be contextually aware to know whether to emit newline tokens or not (this is just for Scala 2, probably for Scala 3 another rework will be needed as it is becoming an indentation based language). I am not very well versed in the current testing facilities and would like to know what is the proper approach if I were to try making a Pull Request for this?

Probably more work would be needed or something better could be suggested as a replacement such that the TokenStream need not exist.

studentmain:
For now, I'm putting all necessary tools (except pom.xml) here, and hope next developer remember to run it. It's not intergrated into CI, and only java runtime is tested, but at least it works on our local machine.

If a new test framework is too far, maybe we can write test scripts when current test framework not works, and run it in CI.

Marti2203:
But this doe not make it possible to properly test the grammar through the CI/CD, I would like to have that.

aryx:
@Marti2203 is your tokenstream available somewhere? from what you said it seems like the Scala ANTLR grammar can not parse correctly Scala files right?

Marti2203:
I can send it privately. I am still not okay with including it in the repo.

Marti2203:
@kaby76 , you are currently developping the testability suite for this repo right?

kaby76:
@Marti2203 Yes, I have been writing the trgen testing tool, but it's generally driven by the existence of the maven pom file and templates in grammars-g4/_scripts/templates/. We can still test it with some custom templates for code generation. If one can write a Java driver that does work, it can be plugged into testing as a separate step. -Ken

Marti2203:
@kaby76  Okay, that is good to know. I am still working on making the grammar able to parse the Scala repo itself and another Scala project that is  popular - XML is almost fully parseable, even with nestedness of interpolated strings. It is not possible to make it work without the custom token stream, not sure if I can integrate it into the lexer, will try that later on.

Vladimir-Chan:
[...]

PR #2228, opened 2021-06-08
Updating the Scala grammar [WIP]

Marti2203:
This is the PR containing the work mentioned in  #1983.
@kaby76 @Vladimir-Chan @aryx


Marti2203:
I am still thinking on whether to make the ScalaParser encapsulate the TokenStream it gets and puts it inside of a Scala Token Stream, but I am not sure that is good design, I would like to make the testing infrastructure understand this. @kaby76 iirc you are the one that works on it currently/

kaby76:
@Marti2203 I'm checking out the code. So, for CI builds, you're going to need an updated pom.xml file. Rogo*.g4 should be replaced with Scala*.g4 in the antlr4-maven-plugin spec of the pom.xml, otherwise the Antlr tool isn't even run to generate a parser.

Marti2203:
I wrote the wrong names...

kaby76:
And, all that `@header` and action code should be yanked into a base class, and placed in grammars-v4/scala/Java. I'll write the base class code for every other target, but I gather there's going to need to be a special case driver written. I'll need that code soon so I can change the CI build to trgen a special case for this grammar.

kaby76:
I moved around some stuff and got it to this point where it's started to go in the direction of target independence. See the attached zip. 
[scala.zip](https://github.com/antlr/grammars-v4/files/6616127/scala.zip) The actions should all go into "agnostic style", i.e., action code stuffed into a method on the base class, and replaced with "{ self.whatever_action(); }". (See the [Python doc](https://github.com/antlr/antlr4/blob/master/doc/python-target.md#target-agnostic-grammars)).

Marti2203:
The grammar works, but now I need to make it target agnostic... This will take a day or two...

teverett:
@Marti2203  are you planning to continue work on this PR?

teverett:
@Marti2203  I am closing this for now.  You are welcome to re-open it with merge fixes.

Vladimir-Chan:
I don't know scala so I can't help this issue.

Vladimir-Chan:
I don't know scala so I can't help solve this issue.

PR #2230, opened 2021-06-10
Godot GDScript

studentmain:
https://docs.godotengine.org/en/stable/development/file_formats/gdscript_grammar.html

They have an official grammar, but it's pretty outdated and lexer is not included.


teverett:
@studentmain  thanks

PR #2234, opened 2021-06-14
translate g4 in CI (demo)

studentmain:
Translate function call syntax in g4 between different target language to make itself more "target independent".

Just a demo, don't merge.

cc @kaby76


kaby76:
After porting grammars-v4/csharp to the targets, I think this solution is still needed.

For Dart, an `@header` must be added to the grammar because it's not in the generated parser or lexer file, a bug I think in the Dart .stg file for the Antlr tool. Dart does not "import" the declarations from one file to another automatically even if both files are located in the same directory. For Java and C#, the compiler looks around for the sources. There must be an explicit import for Dart in the generated parser file. Either the Dart.stg file is fixed (slow to accept changes, unknown when or even if), or this tool is used to "fix" the grammar for Dart by adding `@lexer::header{ import 'CSharpLexerBase.dart'; }`. There's a similar problem for Cpp.

Also, I'm not sure why "l" and "p" were picked instead of just "self" because you can't use "l" in parser actions for the Go target. So, for target agnostic grammars, the tool should be used to translate "self" to "l" in lexer actions and "p" in parser actions.

studentmain:
> For Dart, an @header must be added to the grammar

OK, https://github.com/studentmain/SempredTranslation/issues/1

> Also, I'm not sure why "l" and "p" were picked instead of just "self"

It's Golang tradition... I remember we already fixed it in ANTLR?

kaby76:
> > Also, I'm not sure why "l" and "p" were picked instead of just "self"
> 
> It's Golang tradition... I remember we already fixed it in ANTLR?

Yes, it was changed (antlr/antlr4#3174). As I said, you have to know that l and p are declared in the func's that are generated. "self" would be just easier to remember. But, c'est la vie.

In any case, Go target is broken for superClass option used in a lexer grammar. That's because it's declared wrong in the generated lexer file by the Antlr tool. I can "fix" this by modifying the Antlr tool-generated lexer. But, the Go.stg file for the Antlr tool should be fixed.

studentmain:
@kaby76 Is this ok?

图片

kaby76:
@studentmain Yes, that is correct. An example is [the Java9 grammar](https://github.com/antlr/grammars-v4/blob/9e449d012348fdb143a3375bed73ee01ba562196/java/java9/Cpp/Java9Lexer.g4#L74). The plsql grammar is another example. The lexer contains a [@header::postinclude](https://github.com/antlr/grammars-v4/blob/9e449d012348fdb143a3375bed73ee01ba562196/sql/plsql/PlSqlLexer.g4#L28). (The parser grammar has a similar include.) I pointed out that this hack is only observed by the Cpp target.

studentmain:
> The lexer contains a `@header::postinclude`. (The parser grammar has a similar include.)

https://github.com/antlr/antlr4/blob/master/doc/cpp-target.md#named-actions

Order is important in C++, seems I should use this one.

kaby76:
Yes, "postinclude" for Cpp. Honestly, a properly written C include file should work in any order. Another example of how the .stg files need to be rewritten and should be passed to the Antlr tool so they can be customized.

kaby76:
Also to consider is the translation of Antlr4's Unicode to the old style for Antlr4cs. I never wrote up this fiasco (which took weeks of work), but we had a wonderful, unified, combined, no-base class solution for the lexer issue for Java-based grammars. The [PR I wrote](https://github.com/antlr/grammars-v4/pull/2189) modified Apex](https://github.com/kaby76/grammars-v4/tree/issue-2188/apex) but was [pushed back](https://github.com/antlr/grammars-v4/pull/2189#issuecomment-835385328) by @KvanTTT because it required a grammar specifically for Antlr4cs--the ancient tool and runtime for C# only supported the old 16-bit version '\u1234' Unicodes, not true 24-bit significant '\u{123456}' syntax Unicodes, nor the named sets that we use for Java. https://github.com/antlr/antlr4/blob/master/doc/unicode.md#unicode-code-points-in-lexer-grammars. Note, _we don't actually test Antlr4cs in the CI build._ Also, I did write the Java Standards Committee as to why the character set is not defined apriori rather than operationally. But, I never did get a response.

studentmain:
@kaby76 Translate unicode escape syntax is possible but I think it's unnecessary. Only a few grammars require new syntax in a few place (I remember less than 10 places in whole grammars-v4 repo), most grammars can just use \p, and it will only break an old target which replaced by official target. Consider how people use this repo (copy paste g4 file into their project), in those rare case, user can manually change the syntax.

Another reason is the translate tool is not really designed to do that, I designed it to translate function call syntax specifically. But if there are more requirements, I'll consider redesign the tool.

For now, the most important part is how to use the translated grammar. If we let user running the translate tool themselves, maybe I need write the tool in Java. If we provide them to user, how? Maybe put them in repo, use CI to update them and ensure nobody accidentally touched them?

studentmain:
closed, translator are now a part of #2443

PR #2236, opened 2021-06-15
Add nest generic type  support, such as ’NSDictionary<NSNumber *, NSA…

czqasngit:
Add nest generic type  support, such as ’NSDictionary<NSNumber *, NSArray<NSString *> *>’.


teverett:
@czqasngit  thanks

Issue #2237, opened 2021-06-15
plsql exception_declaration doesn't work


IlyaRyzhov:
I am using PlSqlParser.g4 to build AST and noticed that when building the tree for the next procedure:

```sql
CREATE OR REPLACE PROCEDURE PROC(PARAM NUMBER)
IS
LOCAL_PARAM NUMBER;
USER_EXCEPTION EXCEPTION;
BEGIN
SELECT C INTO V FROM T
WHERE P = PARAM;
END;
 ```

AST:
 ![ast_for_procedure_with_current_rule](https://user-images.githubusercontent.com/55021692/122079085-c0383700-ce05-11eb-9bb6-1a6f44ae4875.jpg)
`USER_EXCEPTION` is in the node `variable_declaration` and not in the node `exception_declaration`. Think it should be fixed by changing order of `exception_declaration` and `variable_declaration` in rule `declare_spec` like this:

```g4
declare_spec
    : pragma_declaration
    | exception_declaration
    | variable_declaration
    | subtype_declaration
    | cursor_declaration
    | type_declaration
    | procedure_spec
    | function_spec
    | procedure_body
    | function_body
    ;
```

that is rule `exception_declaration` must go before rule `variable_declaration`.

KvanTTT:
Please fix it if you can and suggest a pull request.

PR #2238, opened 2021-06-16
plsql exception_declaration swapped with variable_declaration

IlyaRyzhov:
#2237 fix for issue


IlyaRyzhov:
This issue also occurred in rule package_obj_body and in rule package_obj_spec and was corrected in the same way

KvanTTT:
It is possible to add a test sample that covers these changes? Ideally with `.tree` file.

PR #2249, opened 2021-06-27
Adding the Lark parser generator grammar.

kaby76:
This is a grammar for Lark, another parser generator, derived from the sources for [Lark in github](https://github.com/lark-parser/lark/blob/master/lark/grammars/lark.lark).


teverett:
@kaby76 thanks

PR #2256, opened 2021-07-12
flatbuffers grammar fix

### LINKED ISSUE #2252: Flatbuffer grammar rule recognation order problem.
As we know if two rules match the same token at the same time, the type of token will be the rule's type who appears first in the list.
In grammar flatbuffers the rule `INTEGER_CONSTANT : [-+]? [0-9]+ | 'true' | 'false' ;` appears after rule `IDENT : [a-zA-Z_] [a-zA-Z0-9_]* ;`, so the tokens `true` and `false` are matched by the rule `IDENT`, but it should not.

1artashes97:
flatbuffers grammar fixes #2252


teverett:
@1artashes97  thanks

PR #2259, opened 2021-07-16
Support Mysql INVISIBLE COLUMN syntax

tinntsea:


sql/mysql/Positive-Technologies/MySqlParser.g4
```diff
@@ -350,6 +350,7 @@ columnDefinition
 columnConstraint
     : nullNotnull                                                   #nullColumnConstraint
     | DEFAULT defaultValue                                          #defaultColumnConstraint
+    | (VISIBLE | INVISIBLE)                                         #invisibleColumnConstraint
```
KvanTTT, line 353: I suggest using the first level alternatives if possible (it may improve performance):

```suggestion
    | VISIBLE                                             #invisibleColumnConstraint
    | INVISIBLE                                           #invisibleColumnConstraint
```

KvanTTT (commented):
Welcome to the grammars repository and thank you for the contribution!

sql/mysql/Positive-Technologies/MySqlParser.g4
tinntsea: Updated as your suggestion

sql/mysql/Positive-Technologies/MySqlParser.g4
```diff
@@ -350,6 +350,8 @@ columnDefinition
 columnConstraint
     : nullNotnull                                                   #nullColumnConstraint
     | DEFAULT defaultValue                                          #defaultColumnConstraint
+    | VISIBLE                                                       #visibleColumnConstraint
+    | INVISIBLE                                                     #invisibleColumnConstraint
```
KvanTTT, line 354: FYI, It's possible to use the same label here `visibilityColumnConstraint` if it's convenient for you:

```g4
    | VISIBLE                                                       #visibilityColumnConstraint
    | INVISIBLE                                                     #visibilityColumnConstraint
```

ANTLR merges such labels into a single alternative.

tinntsea: sure, updated.

PR #2260, opened 2021-07-16
PostgreSQL pgplsql implementation

OleksiiKovalov:
**What does this PR do?**
Implements Postgresql pgplsql grammar with tests and sandbox c# app
**Why?**
There is no pgplsql grammar in this repository
**How is it checked?**
Tested using unit tests over the tests scripts from original postgresql repository


sql/postgresql/.gitignore
```diff
@@ -0,0 +1,353 @@
+## Ignore Visual Studio temporary files, build results, and
```
KvanTTT, line 1: Does it make sense to have a nested `.gitignore` file? Is it possible to merge needed items with root `.gitignore`?

sql/postgresql/README.md
```diff
@@ -0,0 +1,10 @@
+# postgresql grammar
```
KvanTTT, line 1: Use correct title names in docs:

```suggestion
# PostgreSQL grammar
```

sql/postgresql/pggrammar/README.md
```diff
@@ -0,0 +1,5 @@
+# postgresql grammar
```
KvanTTT, line 1: Does it seem like it should be something else?

sql/postgresql/pggrammar/pggrammar.sandbox/App.config
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
```
KvanTTT, line 1: Please remove files not related to grammar, examples, or tests scripts (seems like all files from `sql/postresql/pggrammar/pggrammar.sandbox`).

sql/postgresql/pggrammar/pggrammar.test/parsertest/aggregates.sql
```diff
@@ -0,0 +1,1218 @@
+--
```
KvanTTT, line 1: Move all example files from `sql/postresql/pggrammar/pggrammar.test/parsetest/*.sql` to `sql/postresql/examples`.

sql/postgresql/pggrammarBaseListener.cs
```diff
@@ -0,0 +1,9686 @@
+//------------------------------------------------------------------------------
```
KvanTTT, line 1: Remove generated files (listeners, visitors, lexers, parsers).

KvanTTT (changes requested):
Thank you for the contribution, but there are a lot of unrelated files in this request. Also, there are a lot of changes, and GitHub even can't display all of them. Please remove:

* Generated files (parsers, lexers, visitors, listeners)
* Windows forms files (from sandbox)
* Binary files (antlr.jar)

I'll take a look at the grammar after removing this unrelated info.

OleksiiKovalov:
@KvanTTT any updates?

KvanTTT:
I'll try to review it this week, closer to the weekend.

taozuhong:
Thanks awfully.

I try them and rename them as: 
`PgSqlLexer.g4 PgSqlParser.g4`

Good news to pass the build.

sql/postgresql/pggrammar.g4
```diff
@@ -0,0 +1,4294 @@
+parser grammar pggrammar;
```
KvanTTT, line 1: It's better to rename to `PostgreSQLParser`. We use `Lexer` and `Parser` suffixes in all grammars in this repository.

```suggestion
parser grammar PostgreSQLParser;
```

sql/postgresql/pggrammar.g4
```diff
@@ -0,0 +1,4294 @@
...
+
+options {
+	tokenVocab =pglexer;
```
KvanTTT, line 4:
```suggestion
	tokenVocab = PostgreSQLLexer;
```

sql/postgresql/pggrammar.g4
```diff
@@ -0,0 +1,4294 @@
...
+    }
+        private readonly IList<ParseError> m_ParseErrors = new List<ParseError>();
+        public IList<ParseError> ParseErrors => m_ParseErrors;
+        public IParseTree Root => root();
+        public IParseTree PlSqlRoot => plsqlroot();
+}
+
+root:
+	stmtblock EOF
+	;
```
KvanTTT, line 139: Could you please use the following format for all parser rules?

```g4
rule
    : alternative1
    | alternative2
    ...
    | alternativen
    ;
```

```suggestion
root
	: stmtblock EOF
	;
```

sql/postgresql/pggrammar.g4
```diff
@@ -0,0 +1,4294 @@
...
+                 | SYSID iconst
+                 | ADMIN role_list
+                 | ROLE role_list
+                 | IN_P ROLE role_list
+                 | IN_P GROUP_P role_list
+;
+createuserstmt: CREATE USER roleid opt_with optrolelist;
+
+alterrolestmt: ALTER ROLE rolespec opt_with alteroptrolelist
+             | ALTER USER rolespec opt_with alteroptrolelist
```
KvanTTT, line 323: It's better to merge common subparts for performance and clarity.

```suggestion
alterrolestmt
    : ALTER ( ROLE rolespec opt_with alteroptrolelist
            | USER rolespec opt_with alteroptrolelist )
    ;
```

sql/postgresql/pggrammar.g4
```diff
@@ -0,0 +1,4294 @@
...
+ a_expr: c_expr
+       | a_expr TYPECAST typename
+       | a_expr COLLATE any_name
+       | a_expr AT TIME ZONE a_expr
+       | PLUS a_expr
+       | MINUS a_expr
+
+       | a_expr STAR a_expr
+       | a_expr SLASH a_expr
+       | a_expr PERCENT a_expr
```
KvanTTT, line 2759: Not sure it's correct. [The documentation](https://www.postgresql.org/docs/7.2/sql-precedence.html) says these operators have the same precedence, so they should be merged into one alternative. Also, please take a look at how [left recursive](https://github.com/antlr/antlr4/blob/master/doc/left-recursion.md) rules work.

```suggestion
       | a_expr (STAR | SLASH | PERCENT) a_expr
```

It's actual for other operators too.

sql/postgresql/pglexer.g4
```diff
@@ -0,0 +1,1045 @@
...
+ *    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ *    DEALINGS IN THE SOFTWARE.
+ * 3. Except as contained in this notice, the name of Tunnel Vision
+ *    Laboratories, LLC. shall not be used in advertising or otherwise to
+ *    promote the sale, use or other dealings in this Software without prior
+ *    written authorization from Tunnel Vision Laboratories, LLC.
+ */
+
+
+lexer grammar pglexer;
```
KvanTTT, line 32: Rename to `PostreSQLLexer`

```suggestion
lexer grammar PostreSQLLexer;
```

sql/postgresql/pglexer.g4
```diff
@@ -0,0 +1,1045 @@
...
+	;
+
+//
+// KEYWORDS (Appendix C)
+//
+
+//
+// reserved keywords
+//
+ALL					: [Aa][Ll][Ll];
```
KvanTTT, line 163: I suggest using fragment rules for clarity:

```suggestion
ALL					: A L L;
...
fragment A : [Aa];
fragment L : [Ll];
```

Also, [Case insensitive stream](https://github.com/antlr/antlr4/blob/master/doc/case-insensitive-lexing.md) allows writing lexer rules in the simple style:

```suggestion
ALL					: 'ALL';
```

KvanTTT (changes requested):
Please fix at least file names. Other issues can be fixed later after merging.

Also, it makes sense to include this grammar in CI. For that, a `pom.xml` file should be added to the grammar directory (see for example [pom.xml](https://github.com/antlr/grammars-v4/blob/master/sql/tsql/pom.xml) from TSql).

taozuhong:
thanks for reviewing.

and I want to do something if I have permission.

sql/postgresql/pggrammar.g4
OleksiiKovalov: @KvanTTT what plugin/tool can you advise for grammar reformating? Currently, I use Intellij Plugin for ANTLR 4.

sql/postgresql/pggrammar.g4
OleksiiKovalov: For Phase 1 I'm trying to keep code as close to original grammar as possible, it makes it easy to keep track of original grammar
I think at 1st it should be proven that grammar  works correctly and only after that - optimization / etc

sql/postgresql/pggrammar.g4
OleksiiKovalov: the same consideration as above - keep grammar closer to original, even if it may be optimized.
currently, diffs are only to respect precedences and to avoid mutual left recursion problems in simple_select rule

KvanTTT: It's not about optimization but about correctness.

```sql
a / a * a
```

For the old grammar this input is parsed as follows (`STAR` has higher priority than `SLASH`):

```
 div
 /  \
a    *
    /  \
   a    a
```

Is it correct? I think the valid output is (`STAR` has the `SLASH` priority):

```
       *
     /   \
   div    a
 /     \
a       a
```

Could you point to your reference documentation?

sql/postgresql/pggrammar.g4
KvanTTT: Unfortunately, I don't for now. Ok, let it be as is. Just pay attention to our grammar formatting rules for the future.

sql/postgresql/pggrammar.g4
OleksiiKovalov: fixed precedences according to https://www.postgresql.org/docs/12/sql-syntax-lexical.html#SQL-PRECEDENCE
@KvanTTT

sql/postgresql/PostgreSQLParser.g4
```diff
@@ -0,0 +1,4323 @@
...
+       | PLUS a_expr
+       | MINUS a_expr
+
+        //left	exponentiation
+       | a_expr CARET a_expr
+
+        //left	multiplication, division, modulo
+       | a_expr STAR a_expr
+       | a_expr SLASH a_expr
+       | a_expr PERCENT a_expr
```
KvanTTT, line 2769: As I already wrote, it should be merged because they have the same priority correspondent to precedence from doc.

```suggestion
       | a_expr (STAR | SLASH | PERCENT) a_expr
```

It's related to other operators with the same priority as well.

KvanTTT (changes requested):

KvanTTT:
It seems like there is an error with tests: https://github.com/antlr/grammars-v4/pull/2260/checks?check_run_id=3652328184#step:8:1524

I suggest fixing it (preferably) or disable tests for PostgreSQL for now.

OleksiiKovalov:
@KvanTTT pls direct me how to fix test for pg
thx

kaby76:
OK, looking at it.

kaby76:
Long story short: just add sql/postgressql to the [C# skip list](https://github.com/OleksiiKovalov/grammars-v4/blob/master/_scripts/skip-csharp.txt) as you did for the other skip lists.

Your grammar is C# specific. And, not only that, it's not compatible with the Antlr4cs target--which we don't test anyway. I did check this with Antlr4cs, and after many minutes, the Antlr4cs tool crashed!

The code does compile for the "normal", "standard", "official" Antlr4. And it does work, although it is slow because it requires optimizations. For example, `examples/arrays.sql` takes 20s.

But, what happened here is that you removed the grammar information from the pom.xml file, with [this change](https://github.com/OleksiiKovalov/grammars-v4/commit/001ed65cdb3bdb24b6723d0978a7af65ea17cfdc#diff-5589d2e8e15ff265421e34a063ad0e475bc10d690c669eaeb29174df44c41cbe) because you didn't want to the Maven plugin to generate a driver and try to compile it for the Java target.
trgen was designed to use the pom.xml file to gather information about the grammar.
Since there is no information about the grammar, I can't actually generate a C# driver program, compile, and run it. I could modify the tool to parse the grammar first, then look it over and generate code from that.

studentmain:
> Thx, kaby76!
> Added PostgreSQL to skip file
> worth noting that when I've tried to make the test work, I've generated test exe/dll file and it successfully parsed script files, which I've passed to it manually
> yup, right now grammar is dramatically slow :( I plan at 1st phase to make grammar correct even if it is slow and refactor/optimize it on 2nd phase, having tests/baseline.

My opinion is having test first. It will help you make it correct and fast.

We can merge this with all CI test skipped, refactor and add necessary Java code to make CI test works. Since you already tested manually, merge it will not cause problem.

OleksiiKovalov:
@studentmain of course I have tests but implemented as nUnit tests, not a tests used by this repository.
Scripts used in these tests included in PR
Not sure how grammar can be refactored to be compatible with both Java/C# as it uses actions in lexer and a custom utility code in a grammar. However, I'm pretty sure that code from grammar can be moved to "partial" class.

kaby76 (approved):
Looks good. When I get a chance, I'll try to convert this to "standard" target-independent code with a base class. But, it looks fine right now. Good work.

PR #2265, opened 2021-07-22
[scss] Support imports using @use, including as and with clauses.

nickwalther:
Grammar defined here
https://github.com/sass/sass/blob/main/accepted/module-system.md#use-1


nickwalther:
@teverett review please

teverett:
@nickwalther  thanks

PR #2273, opened 2021-08-04
Added scss @forward keyword

tristanlins:
This is new since Dart Sass 1.24.0


KvanTTT:
Could you also please add a test sample to https://github.com/antlr/grammars-v4/blob/master/scss/examples/example1.txt that covers this change?

tristanlins:
I am in the process of revising the whole grammar.
The current one does not support a lot.
I close this PR, in favor of a new one.
But my revised grammar does not yet support all (special) cases, I still need a little time  🙄

PR #2276, opened 2021-08-07
TSql: Review rule create_or_alter_ddl_trigger for multiple ddl_trigger_operation

AMagistroni:
In the `create_or_alter_ddl_trigger` rule there is the possibility to put many `ddl_trigger_operation`, not a `dml_trigger_operation`


KvanTTT:
Thanks, could you please also add a test sample that covers this change?

AMagistroni:
I've added in the triggers.sql an example

KvanTTT:
Thanks and welcome to the world of ANTLR grammars!

PR #2284, opened 2021-08-14
TSqlParser : Alter table

AMagistroni:
Alter table could have CHECH or NOCHECK.

For example with this two tables:  

```
CREATE TABLE [dbo].[Table](	
	[ID] [int] IDENTITY(0,1) NOT NULL,
	[StateId] [int]	
GO

CREATE TABLE [dbo].[TableState](	
	[ID] [int] IDENTITY(0,1) NOT NULL,
	[Description] [nvarchar](10)
GO
```
you could have a constraint with CHECH or NOCHECK

```
ALTER TABLE [dbo].[Table]  WITH NOCHECK ADD  CONSTRAINT [FK_NAME] FOREIGN KEY([StateId])
REFERENCES [dbo].[TableState] ([ID])
ON DELETE CASCADE
```

I also add the clause ON CASCADE and ON_DELETE

Perhaps there is the possibility to rewrite the grammar with the rule 'table_constraint'
but it's too difficult for me -:)

[https://docs.microsoft.com/it-it/sql/t-sql/statements/alter-table-transact-sql?view=sql-server-ver15](url)

and 

[https://docs.microsoft.com/it-it/sql/t-sql/statements/alter-table-column-constraint-transact-sql?view=sql-server-ver15](url)


KvanTTT:
Could you please add a sample that covers suggested changes?

AMagistroni:
Hi
how works the examples?
They are executed automatically to check if there is an error when the project is compiled?

I'm not sure how to do, so I added the example in the comment above.

I can add in the ddl_table.sql at the end of the file this example?

Thank you

KvanTTT:
> They are executed automatically to check if there is an error when the project is compiled?

Yes, they are being parsed during CI build.

> I can add in the ddl_table.sql at the end of the file this example?

Yes, it's exactly the thing I meant.

PR #2293, opened 2021-08-21
JavascriptParser/TypeScriptParser: No trailing-comma for empty object

SchildersChris:
# Problem
I was digging through the `JavaScriptParser.g4` and came across the following rule.

```antlr
objectLiteral
    : '{' (propertyAssignment (',' propertyAssignment)*)? ','? '}'
    ;
````

Which means that the following syntax is correct:
```js
const o = {,};
```
antlr4_parse_tree_incorrect

## Steps to create syntax tree
```
1. cd ./grammars-v4/javascript
2. mvn clean compile
3. cd ./javascript/target/classes
4. grun JavaScript -gui
5. const o = {,};
6. ^D
```

According to the ECMAScript 2020 specifications ObjectLiterals should follow the following specification.
```
ObjectLiteral[Yield, Await]:
   {}
   {PropertyDefinitionList[?Yield, ?Await]}
   {PropertyDefinitionList[?Yield, ?Await],}
```
Source: https://262.ecma-international.org/11.0/#prod-ObjectLiteral

# Solution
I decided to create a PR, bringing the  `','?` (trailing-comma) inside the capture group.
```antlr
objectLiteral
    : '{' (propertyAssignment (',' propertyAssignment)* ','?)? '}'
    ;
````

Which produces the following syntax tree:
antlr4_parse_tree_correct

### DIFF
--- a/javascript/javascript/JavaScriptParser.g4
+++ b/javascript/javascript/JavaScriptParser.g4
@@ -366,7 +366,7 @@ assignable
     ;
 
 objectLiteral
-    : '{' (propertyAssignment (',' propertyAssignment)*)? ','? '}'
+    : '{' (propertyAssignment (',' propertyAssignment)* ','?)? '}'
     ;
 
 anonymousFunction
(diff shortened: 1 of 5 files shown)


studentmain (approved):
Can you add a test?

SchildersChris:
> Can you add a test?

I will right away.

KvanTTT:

PR #2294, opened 2021-08-22
Set statement could have many keyboards on the same line

AMagistroni:


sql/tsql/TSqlParser.g4
```diff
@@ -3078,6 +3078,32 @@ fetch_cursor
...
+    | NOCOUNT 
+    | NOEXEol 
+    | NUMERIC_ROUNDABORT 
+    | PARSEONLY 
+    | REMOTE_PROC_TRANSACTIONS 
+    | SHOWPLAN_ALL 
+    | SHOWPLAN_TEXT 
+    | SHOWPLAN_XML 
+    | XACT_ABORT
+    ;
```
KvanTTT, line 3106: It's always better to place definitions after usings if possible (similar to methods in source code). Exceptions are recursive definitions where it's not possible.

AMagistroni: Do you mean after set_special at line 3084?

KvanTTT: Yes.

PR #2300, opened 2021-08-29
Fix runtime errors of JavaScript JavaScript parser

hanohrs:
JavaScript parser written in JavaScript throws "method is not defined" errors.
Also, parsing examples fails because of its incorrect "strict mode" handling.
This PR fixes these issues along with some minor changes.


teverett:
@hanohrs  thanks

hanohrs:
@teverett tank you too

Issue #2320, opened 2021-09-05
Merge SystemVerilog grammars


KvanTTT:
Now we have two versions of such grammar: https://github.com/antlr/grammars-v4/tree/master/systemverilog and https://github.com/antlr/grammars-v4/tree/master/verilog/system_verilog

I suggest removing the latest one (it seems old and incomplete) and moving the first to that place (`verilog/system_verilog` subdirectory).

@msagca what do you think? If you agree, could you please perform it?

msagca:
Hi @KvanTTT,
I think it would be better to keep it in its own folder since it's a different language standard even though Verilog is a subset of SystemVerilog. I suggest removing the `verilog/system_verilog` directory, moving `Verilog2001.g4` up one level and keeping `systemverilog` folder in the `root` directory.

msagca:
@KvanTTT  I've created a pull request (#2321) regarding this issue.
I removed the `verilog/system_verilog` folder and moved Verilog2001 files up one level. I also fixed some rules in `SystemVerilogParser.g4` and made format changes to `Verilog2001.g4` without altering the rules (except the incorrect `udp_declaration` rule which was commented out anyway). I removed a failing Verilog example.

msagca:
I've closed #2321 and created a new pull request (#2325).

PR #2321, opened 2021-09-05
#2320 [Verilog/SystemVerilog] removed verilog/system_verilog folder, moved Verilog2001.g4 and re…

msagca:
…lated files up one level, updated pom.xml files, fixed some bugs in SystemVerilogParser.g4


msagca:
Runner configuration needs to be updated since Verilog files have been relocated.

KvanTTT:
I was thinking again and came up to the conclusion that it's better not to place these grammars into different directories.

Rationalization: we are trying to group grammars by some characteristics for better navigation over the repository. SystemVerilog is an ancestor of Verilog, and it can be placed in the same directory. The repository already contains some grammars that don't closely relate to each other (javascript and typescript, sql dialects).

msagca:
Then, I will close this pull request and open another one after making the necessary changes.

PR #2324, opened 2021-09-09
DBZ-3984 Mysql-Connector fails parsing invalid decimal format DDL statement

harveyyue:
refer [DBZ-3984](https://issues.redhat.com/browse/DBZ-3984)
fix the invalid decimal format with create table ddl
create ddl:

```sql
CREATE TABLE foo (c1 decimal(19), c2 decimal(19.5), c3 decimal(0.0), c4 decimal(0.2), c5 decimal(19,2));
```


KvanTTT:

PR #2336, opened 2021-09-17
External name assembly for functions

AMagistroni:
Example External assembly


AMagistroni (commented):
I forgot the GO statement

PR #2341, opened 2021-09-19
Attempted to fix a crash of the Python3-Cpp on the first token. Becau…

3walks:
…se m_nLastToken is undefined, a NULL-pointer is dereferenced which causes a crash. The fix seems to help for the problem, but has not been tested further.


teverett:
@3walks thanks

PR #2342, opened 2021-09-25
Alter index rebuild partition

AMagistroni:


sql/tsql/TSqlParser.g4
```diff
@@ -1926,7 +1926,11 @@ create_index
     ;
     
 alter_index
-    : ALTER INDEX id_ ON table_name (DISABLE | PAUSE | ABORT)
+    : ALTER INDEX id_ ON table_name (DISABLE | PAUSE | ABORT | rebuild_partition )
```
KvanTTT, line 1929: Excess whitespace.

```suggestion
    : ALTER INDEX id_ ON table_name (DISABLE | PAUSE | ABORT | rebuild_partition)
```

PR #2344, opened 2021-09-25
Union on sql insert

AMagistroni:
.

### DIFF
--- a/sql/tsql/TSqlParser.g4
+++ b/sql/tsql/TSqlParser.g4
@@ -3254,8 +3254,8 @@ predicate
 // Changed union rule to sql_union to avoid union construct with C++ target.  Issue reported by person who generates into C++.  This individual reports change causes generated code to work
 
 query_expression
-    : (query_specification | '(' query_expression ')')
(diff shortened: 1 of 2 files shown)


sql/tsql/TSqlParser.g4
```diff
@@ -3254,8 +3254,8 @@ predicate
 // Changed union rule to sql_union to avoid union construct with C++ target.  Issue reported by person who generates into C++.  This individual reports change causes generated code to work
 
 query_expression
-    : (query_specification | '(' query_expression ')')
-    |  query_specification order_by_clause? unions+=sql_union+ //if using top, order by can be on the "top" side of union :/
+    : (query_specification | '(' query_expression ')' (UNION ALL? query_expression)? )
+    |  query_specification order_by_clause? (unions+=sql_union)* //if using top, order by can be on the "top" side of union :/
```
KvanTTT, line 3258: It seems like parentheses are redundant here:

```suggestion
    |  query_specification order_by_clause? unions+=sql_union* //if using top, order by can be on the "top" side of union :/
```

PR #2349, opened 2021-10-04
Support for lowercase keywords

RupertAvery:
For completeness lowercase and mixed case keywords should be supported.


KvanTTT (commented):
Thank you for the contribution but please use [case-insensitive char stream](https://github.com/antlr/antlr4/blob/master/doc/case-insensitive-lexing.md#custom-character-streams-approach) for your runtime. At the present time, I'm thinking about how to use case insensitiveness without fragment rules (because it looks ugly).

RupertAvery:
I see. I thought it would affect the text, but I guess I didn't read all the way through to the code. The text on that page implies that the entire stream would be affected (which of course sounds completely ridiculous as it would ruin the integrity of the original data)

KvanTTT:
> which of course sounds completely ridiculous as it would ruin the integrity of the original data

Yes, actually, it does not corrupt the original data. The changed stream is only used during lexing.

PR #2365, opened 2021-10-25
Improve grammar's performance

OleksiiKovalov:
**What does this PR do?**
Improves grammar's performance using grammar simplification and merging alternatives for frequently used rules (ref https://tomassetti.me/improving-the-performance-of-an-antlr-parser/ )
**Why?**
Grammar is about 3 times faster compared with base/original (original grammar test pass time 138 seconds, optimized 43 seconds)
**How is it checked?**
Tested using unit tests over the tests scripts from the original PostgreSQL repository
**Pros**
Grammar is significantly faster than original
**Cons**
Result parsed tree will significantly differ in affected areas.

### DIFF
--- a/sql/postgresql/PostgreSQLParser.g4
+++ b/sql/postgresql/PostgreSQLParser.g4
@@ -2786,19 +2786,19 @@ opt_escape:
                     |DOCUMENT_P
                 )
 ;
- c_expr: columnref # c_expr_expr
+ c_expr: PARAM opt_indirection # c_expr_expr
+       | EXISTS select_with_parens # c_expr_exists
+       | ARRAY (select_with_parens|array_expr) # c_expr_expr
+       | GROUPING OPEN_PAREN expr_list CLOSE_PAREN # c_expr_expr
+       | columnref # c_expr_expr
        | aexprconst# c_expr_expr
        | plsqlvariablename# c_expr_expr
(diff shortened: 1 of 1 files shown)


sql/postgresql/PostgreSQLParser.g4
```diff
@@ -2865,55 +2779,55 @@ not_la: NOT;
...
+                    DISTINCT FROM b_expr
+                    |OF OPEN_PAREN type_list CLOSE_PAREN
+                    |DOCUMENT_P
+                )
+;
+ c_expr: PARAM opt_indirection # c_expr_expr
+       | EXISTS select_with_parens # c_expr_exists
+       | ARRAY (select_with_parens|array_expr) # c_expr_expr
+       | GROUPING OPEN_PAREN expr_list CLOSE_PAREN # c_expr_expr
+       | columnref # c_expr_expr
```
tisonkun, line 2793: If we match `columnref` before `func_expr`, won't `f(1)` matches `f` as columnref first and remains `(1)` as illegal?

cc @OleksiiKovalov @KvanTTT

tisonkun: I know that ANTLR 4 uses LL(*) to de-ambiguous now. I'm using a parser combinator (nom) so I may need to peek and look ahead by myself. Or properly ordering the matching order.

PR #2376, opened 2021-10-29
Minor TOML 1.0 compatibility fixes

valderman:
This PR addresses the following discrepancies with the TOML spec:
* using a single space as time separator is allowed in offset and local date-time see second paragraph of [offset date-time spec](https://toml.io/en/v1.0.0#offset-date-time))
* adding one or more newlines between array elements and their following commas is allowed (second paragraph of [array spec](https://toml.io/en/v1.0.0#array))
* parser should not silently ignore trailing garbage


teverett:
@valderman thanks

PR #2381, opened 2021-11-05
JS/go: Add template depth support

gordon-cc-run:
This fixes compilation error for missing functions


kaby76:
I'm issuing a [bug](https://github.com/antlr/grammars-v4/issues/2382) for the Cpp, Go, and Python3 targets, which are all missing the template depth calculation. It needs to be tracked and addressed. Piecemeal fixes are fine, but not when the bug hasn't been tracked and then forgotten.

teverett:
@gordon-cc-run thanks

PR #2387, opened 2021-11-07
Issue 2386

kaby76:
This change moves the "test.pm1" build and associated files to Antlr 4.9.3.

The PR includes changes for hardwired version numbers in the templates and in the scripts. The change also includes an update to the templates for Python3 and Go to force the 4.9.3 version of the runtime to be used.

Also in this PR is a change for unicode/graphemes. Antlr 4.9.3 does not handle the grammar, and gives a cryptic error message which is explained elsewhere.

This PR just modifies the version of Antlr4 for the build. A separate PR should be performed to: (a) move to a  more recent version of trgen; (b) add the JavaScript target; (c) update the skip lists. One variable at a time.

### DIFF
--- a/_scripts/templates/Dart/cli.dart
+++ b/_scripts/templates/Dart/cli.dart
@@ -9,8 +9,8 @@ import 'dart:convert';
 <if (case_insensitive_type)>
 
 class CaseChangingCharStream extends CharStream {
-        CharStream stream;
-        bool upper;
+    CharStream stream = InputStream.fromString('');
+    bool upper = false;
 
     CaseChangingCharStream(CharStream str, bool up)
     {
@@ -40,7 +40,7 @@ class CaseChangingCharStream extends CharStream {
 
     @override
     int LA(int offset) {
-            int c = stream.LA(offset);
+            int c = stream.LA(offset)!;
 
             if (c \<= 0)
             {
--- a/_scripts/templates/Dart/cli.dart
+++ b/_scripts/templates/Dart/cli.dart
@@ -8,97 +8,64 @@ import 'dart:convert';
 
 <if (case_insensitive_type)>
 
+/// This class supports case-insensitive lexing by wrapping an existing
+/// {@link CharStream} and forcing the lexer to see either upper or
+/// lowercase characters. Grammar literals should then be either upper or
+/// lower case such as 'BEGIN' or 'begin'. The text of the character
+/// stream is unaffected. Example: input 'BeGiN' would match lexer rule
+/// 'BEGIN' if constructor parameter upper=true but getText() would return
+/// 'BeGiN'.
 class CaseChangingCharStream extends CharStream {
-    CharStream stream = InputStream.fromString('');
-    bool upper = false;
+  final CharStream stream;
+  final bool upper;
 
-    CaseChangingCharStream(CharStream str, bool up)
-    {
-        stream = str;
-        upper = up;
+  /// Constructs a new CaseChangingCharStream wrapping the given [stream] forcing
+  /// all characters to upper case or lower case depending on [upper].
+  CaseChangingCharStream(this.stream, this.upper);
+
+  @override
+  int? LA(int i) {
+    int? c = stream.LA(i);
+    if (c == null || c \<= 0) {
+      return c;
     }
-
-    @override
-    int get index {
-        return stream.index;
+    String newCaseStr;
+    if (upper) {
+      newCaseStr = String.fromCharCode(c).toUpperCase();
+    } else {
+      newCaseStr = String.fromCharCode(c).toLowerCase();
     }
-
-    @override
-    int get size {
-        return stream.size;
+    // Skip changing case if length changes (e.g., ß -> SS).
+    if (newCaseStr.length != 1) {
+      return c;
+    } else {
+      return newCaseStr.codeUnitAt(0);
     }
+  }
 
-//  @override
-//  void reset() {
-//      stream.reset();
-//  }
+  @override
+  String get sourceName => stream.sourceName;
 
-    @override
-    void consume() {
-        stream.consume();
-    }
+  @override
+  void consume() => stream.consume();
 
-    @override
-    int LA(int offset) {
-            int c = stream.LA(offset)!;
+  @override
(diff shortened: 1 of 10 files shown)


unicode/graphemes/Graphemes.g4
```diff
@@ -12,7 +12,7 @@ fragment TextPresentationSequence: EmojiPresentationCharacter VS15;
 fragment EmojiPresentationSequence: TextPresentationCharacter VS16;
 fragment EmojiCharacter: [\p{Emoji}];
 fragment EmojiModifierSequence:
-    [\p{Grapheme_Cluster_Break=E_Base}\p{Grapheme_Cluster_Break=E_Base_GAZ}] [\p{Grapheme_Cluster_Break=E_Modifier}];
+    . [\p{Grapheme_Cluster_Break=E_Modifier}];
```
KvanTTT, line 15: What's the need for the dot here? I'd just removed not supported grapheme.

studentmain: https://www.unicode.org/reports/tr29/#Default_Grapheme_Cluster_Table

Seems E_Base, E_Modifier, E_Base_GAZ are all obsoleted.

KvanTTT: Yes, I know, I asked about the dot, not about these graphemes.

kaby76: Removing a specific character set is not the same as allowing a single character of any value. The rule accepted two characters before, it still does. The change you propose says that two characters are not acceptable, that the fragment denotes now one character. Are you sure you are correct? It really doesn't not sound correct. `[\p{Grapheme_Cluster_Break=E_Base}\p{Grapheme_Cluster_Break=E_Base_GAZ}]` != ''.

kaby76: I don't understand this grammar, or the other in `unicode/`. I'm reading `unicode/readme.md` and I can't understand what is the purpose. Is it just to test Antlr4 unicode expressions in a lexer set? It seems it's a test suite of some kind, but it doesn't say so *anywhere* in the readme. If it is, it should say so.

The change to get it to work for 4.9.3 shouldn't even be here since it has nothing to do with a build. If someone wants to modify this, by my guest. I'll roll whatever changes into the PR.

KvanTTT: The rule `EmojiModifierSequence` was removed entirely from the similar grammar from the official repository: https://github.com/antlr/antlr4/pull/3323/files I suggest doing the same.

> I don't understand this grammar, or the other in unicode/. I'm reading unicode/readme.md and I can't understand what is the purpose. Is it just to test Antlr4 unicode expressions in a lexer set? It seems it's a test suite of some kind, but it doesn't say so anywhere in the readme.

It seems like so.

kaby76: According to [Unicode.org](https://unicode.org/emoji/charts/full-emoji-modifiers.html), an emoji modifier sequence is a sequence of two code points. And from [The Unicode® Standard, Version 14.0 – Core Specification](https://www.unicode.org/versions/Unicode14.0.0/UnicodeStandard-14.0.pdf), page 884, "Emoji Modifiers. The emoji modifiers U+1F3FB..U+1F3FF designate five different skin tones based on the Fitzpatrick scale. These may be displayed in isolation as color or halftone swatches, or they may form a ligature with a preceding emoji character representing a person or body part in order to specify a particular appearance for that character." Using Emacs, Epsilon, and VSCode, I've verified using the emoji sequences in [this table in wiki](https://unicode.org/emoji/charts/full-emoji-modifiers.html) that the modifier sequences in the table are actually two code points, not one. But, "." does not seem correct because it could be not just any preceding emoji, but a "body part". I still do not know where this grammar came about.

kaby76: grapheme.g4 implements a grapheme cluster breaker. It segments the Unicode code point stream into graphemes. See [Table 1c](https://unicode.org/reports/tr29/#Regex_Definitions) of [Unicode Standard Annex #29](https://unicode.org/reports/tr29/). In addition, the grammar was extended to include Emoji's, which are defined by regular expressions in [Section 1.4.9 EBNF and Regex](https://www.unicode.org/reports/tr51/#EBNF_and_Regex) of [Unicode® Technical Standard #51 UNICODE EMOJI](https://www.unicode.org/reports/tr51/). All this syntax is old-style Java regular expression naming, and it should all be rewritten. I will do with this PR.

kaby76:
Ops. Forgot to upgrade pubspec.yml in templates for Dart.

_scripts/templates/Dart/cli.dart
```diff
@@ -9,8 +9,8 @@ import 'dart:convert';
 <if (case_insensitive_type)>
 
 class CaseChangingCharStream extends CharStream {
-        CharStream stream;
-        bool upper;
+    CharStream stream = InputStream.fromString('');
+    bool upper = false;
```
KvanTTT, line 13: Such dummy initialization looks weird.

I think it makes sense to copy this class from https://github.com/antlr/antlr4/blob/master/doc/resources/case_changing_char_stream.dart

kaby76: I did copy the code, but I see, they changed the code in the last two months.

When the build settles down, I'll make the change.

KvanTTT:
Unfortunately, Dart tests are still failing.

kaby76:
Right, there's a problem with java/java9--they [removed the default constructor for a non-nullable List of non-nullable int's](https://github.com/dart-lang/language/issues/746). I have to review what the "new", so-called "right way" to do things is now. Perhaps the signature for `base64.encode(encoded)`, which might now mean a whole new API I need to use instead. Will take a while.

kaby76:
List<int> encoded = List.filled(2, 0, growable: false);

kaby76:
OK, the PR seems to finally be working. While I'm not super happy with all the hardwired versions sprinkled throughout the code, I'll work on that in a separate Issue/PR, which will likely mean updating trgen a little bit. I just would like to get the build to work with Antlr 4.9.3 in the smallest set of changes.

KvanTTT:
Thanks!

PR #2392, opened 2021-11-10
Fix for Issue 2390--Add JavaScript testing.

kaby76:
This PR adds JavaScript testing back into the CI GitHub Actions build. 

* Updates _scripts/regtest.sh so that one can do an "invert" run of the skip list (test all in the skip list itself rather than test everything except the skip list grammars);
* Add JavaScript back to GitHub main.yml;
* Remove selected skipped grammars because the grammars now work;
* Update the test.sh template script for JavaScript: output is truncated so tests of grammars like sql/plsql don't clog the logs with too much output;
* Add file "skp-javascript.why", which is just documentation on why some of the grammars are actually skipped. Generally, the problems fall into (a) not ported to JavaScript, (b) parsing errors still (bug in Antlr or maybe the driver), (c) bug in the driver code `ErrorCaptureStackTrace(err) undefined`, (d) stack overflow, (e) doesn't contain a functioning grammar, or incomplete.


studentmain:
we should use a specified node.js version in test.
I'm using node 16.13.0, for now, github action is using 14.18.1 by default.

my skip list is
```
_grammar-test
apex
asn/asn_3gpp
cql3
csharp
fortran77
gdscript
golang
haskell
html
hypertalk
java/java9
javadoc
javascript/jsx
javascript/typescript
kotlin/kotlin
logo/ucb-logo
pgn
php
powerbuilder
python/python2
python/python3
python/python3-js
python/python3-py
python/python3-ts
python/python3-without-actions
python/tiny-python
rego
rexx
rust
sql/hive
sql/plsql
sql/postgresql
sql/sqlite
sql/tsql
stringtemplate
swift-fin
swift/swift2
swift/swift3
swift/swift5
tcpheader
v
velocity
webidl
xpath/xpath31
z
```

kaby76:
@studentmain Good idea. I will try to do that.

Not exactly sure why regtest.sh is running cleanly on my machine, while test.ps1 here in GitHub Actions is reporting errors. Looks like the skip list isn't being handled consistently between regtest.sh and test.ps1. Will fix.

kaby76:
The difference in what is reported as an error is caused by a simple difference in interpretation of the results of trgen. regtest.sh considers a grammar as "failed" if trgen returns the name in stderr; test.ps1 considers a grammar as "failed" if trgen didn't actually generate a "Generated/" directory. In some of the grammars, the pom.xml has not the usual two plug-in specifications (one for running the Antlr4 tool on the grammar(s), the other for running the generated parser on input files), but just one--only test whether the Antlr4 tool compiles the grammar cleanly. When there isn't the test plug-in spec in the pom.xml, trgen doesn't know what the grammar is, what the start symbol is for the grammar. Without that information, it doesn't know what to do and doesn't generate anything at all. What should happen is to at least generate a makefile that runs the Antlr4 tool. For now, I will add in these grammars to the skip list. In case when trgen really fails on the pom, do not generate partially filled Generated/ directories. I am planning to update the build with the latest trgen, but after I fix this build.

kaby76:
All set. Javascript testing is back.

teverett:
@kaby76  thanks

PR #2393, opened 2021-11-11
OBJ parser to pass all the documented examples

hakanai:
The grammar for OBJ files didn't really pass any of the real examples, so I have copied a bunch of the real examples from the documentation and rewritten the grammar to cope with all those examples.

Resubmitting in the hope that the checks pass on this one.


KvanTTT (commented):
I think it makes sense to use `.obj` file extension for samples instead of `.txt`.

teverett:
@hakanai  thanks

PR #2398, opened 2021-11-18
Initial commit of power query grammar

dmosesson:
This is a first cut of the grammar for Power Query (https://docs.microsoft.com/en-us/powerquery-m/m-spec-consolidated-grammar).

There are a few issues which prevent it from being 100% complete:

- AVAILABLE_IDENTIFIER should exclude all keywords. I don't know how to do that in a portable way
- For some reason I can't seem to figure out, DEMILITED_COMMENT doesn't seem to work
- Potentially, the grammar's performance is terrible (when testing, it seems to be something in variable rule, potentially the very deep way all the expressions are parsed). When parsing files in VS Code using the ANTLR extension things are very fast (< 1 second), but when I try linking the listener in C# to actually use it, it can take 40 seconds to parse a file.
- I do not know how to comprehensively test the grammar

This is my first grammar, so apologies if there are things that are done in a non-standard/wrong way. Assistance getting over the finish line with the last issues would also be appreciated.


teverett:
Could you add this into the root pom.xml pls?

dmosesson:
Do you know what the error could be? I tested it locally with the extension and it seemed to work. Is there some configuration that I missed?

dmosesson:
> Do you know what the error could be? I tested it locally with the extension and it seemed to work. Is there some configuration that I missed?

If it helps, I ran an `mvn test` in the `powerquery` directory, and it looks like everything worked (generated the classes, and printed success)

dmosesson:
@teverett any chance you can approve this to run?

teverett:
@dmosesson  thanks

PR #2400, opened 2021-11-23
ORCL PL/SQL  grammar converted to case-insensitive; lexer simplified;

OleksiiKovalov:
**What does this PR do?**
1. Makes grammar case insensitive
2. simplifies lexer by removes literals which are not directly used by production rules (except lists of "literals which can be used as identifiers")
**Why?**
1. according to https://docs.oracle.com/cd/A97630_01/appdev.920/a96624/02_funds.htm#10519 "PL/SQL is not case sensitive, so lower-case letters are equivalent to corresponding upper-case letters except within string and character literals." so it is good to follow this rule. also, using "uppercase charstream" will break case of strings/literals - it may be vital for code analysis
2. code size is less and slightly faster. much faster compilation.
not-used literals removed;
lexer size    1056179 ->   343687 bytes
parser size 3096431 -> 2752576 bytes

**How is it checked?**
using standard post-commit checks/ci


OleksiiKovalov:
hi @KvanTTT , @kaby76 
i need to re-generate .tree file for one of the example script ( [sql/plsql/examples/create_procedure.sql.tree](https://github.com/antlr/grammars-v4/blob/master/sql/plsql/examples/create_procedure.sql.tree) )
How can I do that? Please advise
thanks!

kaby76:
> hi @KvanTTT , @kaby76 i need to re-generate .tree file for one of the example script ( [sql/plsql/examples/create_procedure.sql.tree](https://github.com/antlr/grammars-v4/blob/master/sql/plsql/examples/create_procedure.sql.tree) ) How can I do that? Please advise thanks!

Yes, that's a good question. The way I normally do this is to generate a parser using trgen -t Java and call the executable with the -tree option. But, I don't know the official way to generate the .tree files.

OleksiiKovalov:
> The way I normally do this is to generate a parser using trgen -t Java and call the executable with the -tree option

Works for me, thanks!

kaby76 (approved):

studentmain:
> > hi @KvanTTT , @kaby76 i need to re-generate .tree file for one of the example script ( [sql/plsql/examples/create_procedure.sql.tree](https://github.com/antlr/grammars-v4/blob/master/sql/plsql/examples/create_procedure.sql.tree) ) How can I do that? Please advise thanks!
> 
> Yes, that's a good question. The way I normally do this is to generate a parser using trgen -t Java and call the executable with the -tree option. But, I don't know the official way to generate the .tree files.

I use grun with -tree option, you may need to remove trailing newline to get it works.

OleksiiKovalov:
hi @KvanTTT , 
please take a look

KvanTTT:
Also postponed until status resolving of case insensitivity feature.

KvanTTT:
[`options { caseInsensitive=true; }`](https://github.com/antlr/antlr4/pull/3399) will be available since the next version of ANTLR. That's why such transformation will not be needed.

PR #2402, opened 2021-11-26
Fix for Issue #2401 https://github.com/antlr/grammars-v4/issues/2401

kaby76:
This grammar is an incorrect split grammar. Even though the issue raised was solved using another grammar, it's important that the current grammar here be brought up to a minimal working standard.

(1) rename the files as a combined grammar with include (SystemVerilogParser.g4 => SystemVerilog.g4, SystemVerilogLexer.g4 => Includes.g4);
(2) fix the grammar declarations for the two grammars (`parser grammar` => `grammar`, `lexer grammar` => `grammar`);
(3) fix pom.xml (only run Antlr tool on SystemVerilog.g4);

I chose to make it a combined grammar because there are string literals in parser rules that are not declared in a corresponding lexer grammar (which you cannot do, Antlr tool will fail). As the lexer rules in SystemVerilogLexer.g4 was incorrectly included in the parser grammar, I kept with this. Otherwise, it would have been much more work to make it a split grammar. 

@ogheri Could you please test this grammar for the Cpp target? To build, make sure to run the Antlr tool on the SystemVerilog.g4 itself, *not the include.g4 grammar*. `java -jar ~/Downloads/antlr-4.9.2-complete.jar -Dlanguage=Cpp SystemVerilog.g4`. I don't have a functioning trgen for Cpp. Thanks.

--Ken


kaby76:
Can someone tell me how to add a "reviewer" to this PR? I cannot understand GitHub whatsoever. If I make the PR with the default "ready for review", I cannot add a reviewer at the right-hand side of the window. If I make the PR a draft, I still cannot add a reviewer. If I click on "Ready for review", Github does not offer a way to add a reviewer. Where is the UI to add a reviewer for a PR?

kaby76:
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review :

"Note: Pull request authors can't request reviews unless they are either a repository owner or collaborator with write access to the repository."

What a ridiculous, ineptly designed system. In order to have a "review" as a PR author, the PR author has to subvert the system by actually referencing other users in a Conversation.

KvanTTT:
> Can someone tell me how to add a "reviewer" to this PR?

Seems like you advanced permissions is required for adding a reviewer. For instance, I can do it (I'm a member of this repository who can merge). But reviewer list consists of members of this repository, not any user.

verilog/systemverilog/Includes.g4
```diff
@@ -1,7 +1,7 @@
 // Author: Mustafa Said Ağca
 // License: MIT
 
-lexer grammar SystemVerilogLexer;
+grammar Includes;
```
KvanTTT, line 4: Why just don't use the common structure with `lexer` and `parser` grammars that we have in other grammars? With `tokenVocab` but without `import`.

KvanTTT:
> I chose to make it a combined grammar because there are string literals in parser rules that are not declared in a corresponding lexer grammar (which you cannot do, Antlr tool will fail).

In this case it makes sense to introduce new tokens for this string literals and split grammar on lexer and parser. It also will make lexer more clear.

verilog/systemverilog/Includes.g4
kaby76: There's a choice here with either making the grammar split or combined. The problem is that the current situation has elements of both, but it's done wrong for either. There are many string literals in the parser rules in the "parser grammar", but these are not actually noted in the "lexer grammar". To make this a split grammar, I'd have to enter all the string literals into the lexer, and I just want to get this to work in under a minute of work--as I said, I don't have a lot of time working on scraping and rewriting the C++ grammar. Since there's another grammar that will replace this grammar, I felt this an okay decision.

kaby76:
Right. "1542 Error(s)" starting out as a split grammar. There are a ton of string literals missing. Will use trparse|trxgrep|trtext to fix this mess.... Hang on a minute.....

KvanTTT:
Actually, I think ANTLR should be smart enough to deduce tokens automatically if they are not defined at least for simple string literals. I've created the issue for that: https://github.com/antlr/antlr4/issues/2361

kaby76:
After renaming back and then doing a `trparse SystemVerilogParser.g4 | trxgrep " //STRING_LITERAL" | trtext | sort -u`, there are 341 string literals that I will have to check against what is currently in the SystemVerilogLexer.g4 (not a lot). Then I suppose you want me to come up with real names other than the Antlr tool generated names "T_1", etc? I don't have a Trash tool to come up with good names--but I have thought of it several times before. This is a lot of work, and it will all be thrown away.

kaby76:
Okay, changes made to now make this a split grammar. There was a copy/paste error in the grammar--Unicode character for for ">" in  "->" in the original grammar. "Copy and paste" is not a solution to the problem of "scraping" a grammar from a spec.

kaby76:
(Note, on the performance improvement with this change, I'm not sure what is happening. But, I have looked at why on my machine async_fifo.sv takes almost 8s. According to Antlr Profiling, it looks like rule "generate_item" has to parse the entire text of the clause in order to make a decision. Commenting out the input "generates" section in the input, it parses in less than half the time. So, this in one problem with the grammar. But, there are others. Antlr is an excellent tool in order to figure out how to fix a grammar, but it takes a little bit of looking around to figure out what the counters all mean.)

studentmain:
> (Sorry it is the "-" sign in "->" for rule "constraint_expression". I just looked it up in a version of the Spec [here](http://courses.eees.dei.unibo.it/LABMPHSENG/wp-content/uploads/2016/02/SystemVerilog_3.1a.pdf). It should be code point U+002d not U+2013. The file should now be encoded as straight-up ascii, not UTF8-No-BOM! Really, why in the world UTF8-No-BOM rather than have a damn BOM to let us know what's going on here? That's like leaving an IED in the grammar. Geesuz.)

related (?) issue: https://github.com/microsoft/vscode/issues/136437

kaby76:
> > (Sorry it is the "-" sign in "->" for rule "constraint_expression". I just looked it up in a version of the Spec [here](http://courses.eees.dei.unibo.it/LABMPHSENG/wp-content/uploads/2016/02/SystemVerilog_3.1a.pdf). It should be code point U+002d not U+2013. The file should now be encoded as straight-up ascii, not UTF8-No-BOM! Really, why in the world UTF8-No-BOM rather than have a damn BOM to let us know what's going on here? That's like leaving an IED in the grammar. Geesuz.)
> 
> related (?) issue: [microsoft/vscode#136437](https://github.com/microsoft/vscode/issues/136437)

I think you may be right.

I found an illegal but official-looking copy in "pdfroom.com". (I assume the person who wrote the Antlr version bought the copy, but there's no direct link of the Spec in the [readme.md file](https://github.com/antlr/grammars-v4/blob/abfc9ec45da36da09fdb19b7264d5c76d8885555/verilog/systemverilog/README.md).) Using Adobe Reader and Google Chrome (it has a built in reader that is different from Adobe Reader), I copied and pasted page 1067, which contains the production with "->" (which is colored red), and pasted it into Epsilon (an Emacs editor) and VSCode, on Windows, and then saved the file. While Epsilon copied and saved the text as Ascii, VSCode created the Unicode codepoint U+2013 for the "-" minus sign and created a BOM-less file as the default, not even a warning! I have verified that the character and BOM-less file using `od -t x1 -c`. 

I've noticed that the copy value depends on what tool you use to open the PDF. But, now we see here that the "paste" program also can mutate the value. The conclusion is the same: copy and paste is not a substitute for "scraping", which involves correcting the pasted text as well as modifying it to fit Antlr syntax, and finally optimization. This is why we are still in the stone age: you can't do this by hand very well.

KvanTTT:
Thanks. So, seems like this PR is ready to merge?

kaby76:
> Thanks. So, seems like this PR is ready to merge?

@KvanTTT Yes, this PR is ready to merge. Good, non-critical things to work on elsewhere, so it was a good bug. --Ken

Issue #2382, opened 2021-11-05
javascript grammar: Add template depth support for all targets (Cpp, Go, Python3)


kaby76:
There's a [fix](https://github.com/antlr/grammars-v4/pull/2381) for the javascript grammar to provide a template depth calculation. For whatever reason, this is being done piecemeal in separate PRs. It needs to be addressed for all targets, including Cpp, Go, Python3. When fixed, make sure to update the "skip lists".

gordon-cc-run:
This was added to the `.g4` files in https://github.com/antlr/grammars-v4/pull/2206 but only appears to have been added to the Java target. My PR also is only for `javascript/javascript` and the original change adds it to `javascript/typescript` as well (but not the other dialects, not sure if they're also applicable or not).

studentmain:
see also #2257

PR #2404, opened 2021-11-30
Fix for Issue 2382

kaby76:
This PR is a fix for #2382. The files changed here are minimal:

* Add the missing methods.
* Move the constructor in the Cpp target to the header because it doesn't compile with Gnu in anticipation of adding "self". I have not added "self" field.
* Rename the Python directory to Python3. There is no "Python" target; it is Python3 or Python2. I only bother to write the Python3 target, and most people don't care for Python2 anymore.
* Rename the names of several methods in the Python3 base classes to be capitalized. Simplified the rewrite script to remove all these crazy rewrites. Let's keep it simple folks!

With these changes, I was able to get Cpp, CSharp, Java, Python3 to work.

The goal of ["target agnostic" grammar](https://github.com/antlr/antlr4/blob/master/doc/python-target.md#target-agnostic-grammars) is not satisfiable. Something is going to have to give in before this is solved properly.

* The Antlr4 tool generates recognizers for Python3 with a parameter "self", not "this". I can't change the templates that the Antlr4 tool uses. In lieu of that, run the editing script to fix the grammar for Python3.
* The Antlr4 tool generates recognizers for Go with parameters "l" and "p", not "this". Why in the world would you have two different parameter names used for lexer and parser rule actions? I can't change the templates that the Antlr4 tool uses, even less convince that "l" and "p" parameter names are just wrong. In lieu of that, run the editing script to fix the grammar for the Go target.
* Antlr4 tool does not take an arbitrary [option "options { language=TargetAgnostic; }" in a grammar file. Although we could define a "universal action language", no one has done that yet.
* The Antlr4cs target exists but is completely broken. That's because the grammar now references Unicode sets in Antlr4 Standard tool, which is not available in the Antlr4cs tool. The grammars before [this change](https://github.com/antlr/grammars-v4/commit/39adcce6a863917a494842651bebd3e7a4ce3806#diff-71e0c45e5f8c3dac3d71cfcf70e959e36369cc5f22786a59448b3a81caa789f3) could be used, but I haven't done so.
* I haven't bothered testing the JavaScript, Dart, Swift targets.

--Ken


teverett:
@kaby76  thanks

Issue #2363, opened 2021-10-22
Verilog hang jdk 16 or 17


quantrpeter:
Hi
   Parsing this file using Verilog2001.g4 will hang, forever looping. It only happen in Mac.

```verilog
/* verilator lint_off UNUSED */
module id(input wire rst,
          input wire[`INST_WIDTH] inst_i,
          input wire[`MXLEN] reg1_data_i,
          input wire[`MXLEN] reg2_data_i,
          output reg reg1_read_o,
          output reg reg2_read_o,
          output reg[`REG_WIDTH] reg1_addr_o,
          output reg[`REG_WIDTH] reg2_addr_o,
          output reg[`AluOpBus] aluop_o,
          output reg[`AluSelBus] alusel_o,
          output reg[`MXLEN] reg1_o,
          output reg[`MXLEN] reg2_o,
          output reg[`REG_WIDTH] wd_o,
          output reg wreg_o);
    
    wire[6:0] opcode = inst_i[6:0];
    wire[2:0] funct3 = inst_i[14:12];
    wire[6:0] funct7 = inst_i[31:25];
    
    reg[`MXLEN] imm;
    reg instvalid;
    
    always @(*) begin
        if (rst == `RESET) begin
            aluop_o     = 16'h0;
            alusel_o    = 3'b000;
            wd_o        = 5'b00000;
            wreg_o      = 1'b0;
            instvalid   = 1'b0;
            reg1_read_o = 1'b0;
            reg2_read_o = 1'b0;
            reg1_addr_o = 5'b00000;
            reg2_addr_o = 5'b00000;
            imm         = `ZeroDWord;
            end else begin
            aluop_o     = 16'h0;
            alusel_o    = 3'b000;
            wd_o        = 5'b00000;
            wreg_o      = 1'b0;
            instvalid   = 1'b1;
            reg1_read_o = 1'b0;
            reg2_read_o = 1'b0;
            reg1_addr_o = inst_i[19:15];
            reg2_addr_o = inst_i[24:20];
            imm         = `ZeroDWord;
            
            if (opcode == 7'b0110011 && funct3 == 3'd6 && funct7 == 7'd0) begin
                //OR
                wreg_o      = 1'b1;
                reg1_read_o = 1'b1;                
                aluop_o     = `EXE_I_OR;
                reg2_read_o = 1'b1;
                alusel_o    = `EXE_RES_LOGIC;               
                imm         = {48'h0, inst_i[15:0]};
[...]

PR #2413, opened 2021-12-06
Verilog 2001 -> 2005 + compiler directive parser

msagca:
- Verilog 2001 grammar has been upgraded to 2005 revision (IEEE Std 1364-2005).
- Added VerilogPreprocessorParser grammar.
- Included the fix given for #2363.
- Tested the grammar on 3000+ sources (https://github.com/msagca/verilog_grammar_tests).
- There is a small issue regarding nested compiler directive parsing but none of the tested files had such structures, so it's not high priority.


teverett:
@msagca  thanks

PR #2414, opened 2021-12-06
Allow omitting semicolon before closing parenthesis

nicoberling:
The Go specification allows omitting semicolons before closing parentheses. This is already implemented for closing curly braces, but not for parentheses. The most common use case for this is multi-imports like:

`
import ("firstimport"; "secondimport")
`

See also examples/imports.go


teverett:
@nicoberling  thanks

PR #2417, opened 2021-12-07
sql:tsql - grammar converted to case-insensitive

OleksiiKovalov:
**What does this PR do?**
Makes grammar case insensitive
minor fixes to avoid c# warnings (EMPTY literal renamed to EMPTY_)

**Why?**
By standard SQL is case-insensitive when parsing keywords.
shame on me but I was unable to find SQL Server documentation that directly confirms that.

**How is it checked?**
using standard post-commit checks/ci


KvanTTT:
Hi! Thank you for the request but I postpone it for now. I don't like transformation to such fragments in all our case insensitive grammars (T-Sql, MySql, PlSql, Pascal ant others). I would like to ask @parrt to consider again introducing the new option `options { caseInsensitive=true }` that will get us rid of big amount of useless and ugly changes. I have such improvements but they need to be updated. There are a lot of problems for users because of absence of such an option: https://github.com/antlr/grammars-v4/issues?q=label%3Acase-insensitive

OleksiiKovalov:
hi @KvanTTT ,
good, let it be postponed till decision. 
I'll proceed with other changes in another branch (eg - making grammar "wider" 'cause actual parser does not follow documentation and current grammar fails on some real-world scripts)

kaby76:
@OleksiiKovalov  @KvanTTT  I should note that I just added the transform in Trash called "[trull](https://github.com/kaby76/Domemtech.Trash/tree/main/trull)" ("upper/lowercase literal") that can rewrite the lexer rules with case insensitivity using lexer character sets. One would have to change the pom.xml to remove the case folding specification to have a complete, consistent rewrite, but it's something I will probably add to the transform since the pom.xml is part of the grammar. The transform doesn't introduce the fragments for individual letters, instead uses the lexer set directly, eg, 'abc' => [aA] [bB] [cC], nor the inverse. Again, something I'll eventually get around to implementing. So, making a private copy with the transform for yourself is completely automatable, quick, and easy.

KvanTTT:
Also, waiting for the next version of ANTLR with `caseInsensitive` option. But anyway thank you for the contribution. Please create separated PR with the rest issues.

parrt:
we have merged the case insensitive option by @KvanTTT

Issue #2418, opened 2021-12-07
Haskell grammar missing CSharp target


kaby76:
I was looking into a question in StackOverflow.com and realized that the Haskell grammar isn't quite target independent. It only has the Java target.

PR #2419, opened 2021-12-07
Fix for Issue #2418 (haskell)

kaby76:
* Changed the haskell grammar to be target agnostic.
* Added the CSharp target.
* Updated the CI building and testing to include the haskell grammar for CSharp.


teverett:
@kaby76  thanks

Issue #2423, opened 2021-12-12
GDScript missing CSharp target


kaby76:
Currently, gdscript is targeted only for Java. But, I made a port of the GDScriptLexerBase.java to CSharp, and it seems to work fine.

Issue #2422, opened 2021-12-11
GDScript: Only up to Two Parameters Allowed


ogrady:
The parser grammar for GDScript [allows only for exactly one or two parameters](https://github.com/antlr/grammars-v4/blob/master/gdscript/GDScriptParser.g4#L74). Replacing `?` with `*` should fix this issue.

kaby76:
Yep, that was "scraped" wrong from the "official grammar". But, there's lot's more wrong: expression alts are in the wrong order. For example, for this input:

	const FOO = 1+2/3

the parse tree is:


	( program
	  ( topLevelDecl
	    ( constDecl
	      ( CONST i=2 txt=const tt=9 DEFAULT_TOKEN_CHANNEL
	      ) 
	      ( IDENTIFIER i=8 txt=FOO tt=47 DEFAULT_TOKEN_CHANNEL
	      ) 
	      ( ASSIGN i=12 txt== tt=56 DEFAULT_TOKEN_CHANNEL
	      ) 
	      ( expression
		( expression
		  ( expression
		    ( literal
		      ( INTEGER i=14 txt=1 tt=51 DEFAULT_TOKEN_CHANNEL
		  ) ) ) 
		  ( ADD i=15 txt=+ tt=90 DEFAULT_TOKEN_CHANNEL
		  ) 
		  ( expression
		    ( literal
		      ( INTEGER i=16 txt=2 tt=51 DEFAULT_TOKEN_CHANNEL
		) ) ) ) 
		( DIV i=17 txt=/ tt=93 DEFAULT_TOKEN_CHANNEL
		) 
		( expression
		  ( literal
		    ( INTEGER i=18 txt=3 tt=51 DEFAULT_TOKEN_CHANNEL
	      ) ) ) ) 
	      ( NEWLINE i=19 txt=\r\n tt=46 DEFAULT_TOKEN_CHANNEL
	  ) ) ) 
	  ( EOF i=21 txt= tt=-1 DEFAULT_TOKEN_CHANNEL
	) ) 

That's `(1+2)/3`. Usually, it's `1+(2/3)`. The grammar was rewritten for expressions.

Also, from the original grammar https://docs.godotengine.org/en/stable/development/file_formats/gdscript_grammar.html `arrayDecl` and `dictDecl` have the comma right before the closing bracket. The grammar mistakenly places `expression` last https://github.com/antlr/grammars-v4/blob/master/gdscript/GDScriptParser.g4#L238 .

[...]

PR #2424, opened 2021-12-12
Fix for Issue #2423 (gdscript)

kaby76:
This fix consists of:
* gdscript/CSharp/GDScriptLexerBase.cs -- this is a port of GDScriptLexerBase.java for Java.
* _scripts/skip-csharp.txt -- added gdscript testing for the CSharp target.

There are still other bugs with gdscript not addressed here. https://github.com/antlr/grammars-v4/issues/2422


teverett:
@kaby76  thanks

PR #2426, opened 2021-12-13
Fix for Issue 2411 (java grammars, ...)

kaby76:
This change eliminates the Go target symbol conflict "emptyStatement". I changed all grammars that had the conflict (java/java8, java/java9, javascript/..., kirikiri-tjs, pascal, php, protobuf3, sql/...), but I only added testing of the Go target for java/java8. The java/java target now works with Antlr 4.9.3, so I also added that for the Go target testing as well.

Eventually, when we have a new version of Antlr that implements symbol conflict resolution properly, we can rename all these symbol conflicts back (without the underscore).

### DIFF
--- a/_scripts/skip-go.txt
+++ b/_scripts/skip-go.txt
@@ -21,8 +21,6 @@ hypertalk
 icalendar
 inf
 informix
-java/java
-java/java8
 java/java9
 javadoc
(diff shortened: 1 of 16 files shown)


teverett:
@kaby76  thanks

PR #2429, opened 2021-12-16
Asl parser

oldpeculiar:
This is a grammar for the Action Specification Language.

I am new to GitHub and the grammars-v4 repository so apologise in advance for any errors in implementing this change. I have given it my best shot but realise that my lack of familiarity probably means that there are errors.


oldpeculiar:
I have committed a change to correct an error in my pom.xml. Hopefully this fixes the problem.

oldpeculiar:
I have renamed "struct" to "struct_" as suggested. Thanks for the help with this, I would have struggled with that.

teverett:
`
Info: Compiling with sound null safety
ASLParser.dart:6408:17: Error: Can't declare a member that conflicts with an inherited one.
  StartContext? start() => getRuleContext<StartContext>(0);
                ^^^^^
../../../../../.pub-cache/hosted/pub.dartlang.org/antlr4-4.9.3/lib/src/parser_rule_context.dart:46:10: Context: This is the inherited member.
Build failed
  Token? start, stop;

         ^^^^^
Build completed, time: 00:00:05.0332100
`

oldpeculiar:
I am assuming that the reason for the failure this time is because I have a parser rule called "start". I have renamed this to "start_time" so hope that this fixes the problem.

oldpeculiar:
I can't work out why this pull request is failing. It seems to be reporting an error with the vb6 grammar which I haven't changed. Any help would be appreciated as I am stumped.

studentmain:
> I can't work out why this pull request is failing. It seems to be reporting an error with the vb6 grammar which I haven't changed. Any help would be appreciated as I am stumped.

Golang runtime is recently updated at https://github.com/antlr/antlr4/pull/3421 and we are using the "latest" golang runtime version, seems like they meet some edge case in VB6 grammar. I already reported to them.

oldpeculiar:
It looks like the issue with the Golang runtime has been resolved. Could this pull request be checked again please to see how it gets on now.

teverett:
@oldpeculiar thanks!

Issue #2430, opened 2021-12-17
Mismatched issue in Csharp "error out --> Missing bracket where it is not required"


DivyaSrivastava45:
**E.g my code :-** 
#region connection_f
My method 
{
implementation 
}

Error in o/p ==> mismatched input '#' expecting { 'as' , ''is ,'}'}
**Eg in my code** 
int a;

In Parse tree i got --> missed ' } ' after 'a' **grammar is expecting '}' where it is not required** .

Each time After **#**  **parse Tree shows mismatched input 
or missed Bracket**

Issue #1652, opened 2020-02-29
csharp lexer grammar bug


jimmy-the-eek:
The lexer grammar contains the production:

`SHARP:         '#'                                -> mode(DIRECTIVE_MODE);`

but the parser doesn't parse compiler directives, hence any input that contains a compiler directive will cause the parser to bail due to the unrecognized SHARP token.

Fortunately, there is a simple fix:

`SHARP:         '#'                                -> channel(DIRECTIVE),mode(DIRECTIVE_MODE);`

This will place the SHARP token on the DIRECTIVE channel where the rest of the directive details are already diverted. (An alternative would be to ignore the token instead, diverting to HIDDEN, etc.)

PR #2431, opened 2021-12-17
Fix for Issue 2430, 2329, 1652, 328 (csharp)

kaby76:
This is a fix for several issues with the "csharp" grammar.
* Issue 2430: This issue is about the lack of default macro processing in the grammar. In the "cpp" grammar, macros are "parsed" in so far as just placing them on the hidden channel. Ideally, one would run a preprocessor on the code. But, it seems reasonable to assume the same thing here for parsing C#: just ignore macros. The fix adds the lexer rule "Directive: '#' ~ [\n]* -> channel (HIDDEN);"
* Issue 2329: This issue is about the parser being able to accept "var a=1,b=2;" multiple declarations with var. This is illegal with the C# compiler, resulting in an error message. But, the complaint is that it's allowed here with the Antlr grammar. There are, of course, two ways to implement this: either in the grammar or as a semantic predicate. Since the [generated grammar from the Roslyn compiler](https://github.com/dotnet/roslyn/blob/main/src/Compilers/CSharp/Portable/Generated/CSharp.Generated.g4) itself does not introduce rules to distinguish the problem, I decided against a syntactic solution. The solution adds a semantic predicate for local_variable_declaration. After inserting the check, I found out that the examples/AllInOneNoPreprocessor.cs file contains the very multiple var declaration problem!
* The grammar was not in "target agnostic" structure, and actually had Antlr "members" and actions all in Java in the grammar. I moved all the code to base classes, updated the Java and Antlr4cs targets (we don't test Antlr4cs in the CI, but I checked it), and added the CSharp target.


kaby76:
Related, #1652. I should probably removed the SHARP rule altogether, then note that this PR fixes that issue.

csharp/Antlr4cs/CSharpLexerBase.cs
```diff
@@ -0,0 +1,106 @@
...
+        : base(input)
+    {
+    }
+
+    protected int interpolatedStringLevel;
+    protected Stack<bool> interpolatedVerbatiums = new Stack<bool>();
+    protected Stack<int> curlyLevels = new Stack<int>();
+    protected bool verbatium;
+
+    protected void Action1()
```
KvanTTT, line 20: I think it should be more meaningful names for actions. `ActionN` does not say what action really does. For instance, `IncrementInterpolatedStringLevel`.

kaby76: Yeah, I agree. I was just lazy. I'll do something like "On<Rule name>" instead.

kaby76:
@KvanTTT I found out after removing the lexer rule SHARP that doing so opened up a can of worms. The lexer grammar is shared between the normal parser grammar and the preprocessor grammar. This seems ok as that's what the ISO C++ Spec says to do over in C++. I will try to fix #2430 but may just back out rule "Directive" if it takes too long (next day or two). Switching to draft mode to put this on ice until I have a resolution to the problem.

kaby76:
This PR is replaced with this PR https://github.com/antlr/grammars-v4/pull/2444 so that the required dependency updated and the commits all squashed.

PR #2438, opened 2021-12-23
use specific runtime version in golang test

studentmain:
Fix runtime version issue mentioned in https://github.com/antlr/grammars-v4/pull/2429#issuecomment-999423065

Although run test with latest ANTLR seems beneficial, we'd better do that as a test of ANTLR itself. That require a tester designed to work with local package repo, just remove version constraint is not the correct way.

### DIFF
--- a/_scripts/templates/Go/Program.go
+++ b/_scripts/templates/Go/Program.go
@@ -6,9 +6,9 @@ import (
     "os"
     "io"
     "github.com/antlr/antlr4/runtime/Go/antlr"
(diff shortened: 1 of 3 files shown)


studentmain:
This PR enabled go module in template, I didn't found a better solution.

Using tagged version is not supported when GO111MODULE=off, and relative path is not allowed when GO111MODULE=on.

I'll rebase after #2439 merged.

teverett:
@studentmain thanks

PR #2439, opened 2021-12-23
Fix for Issue 2435 -- updating to latest trgen, add Cpp target

kaby76:
This is a fix to use the updated [trgen](https://github.com/kaby76/Domemtech.Trash/tree/main/trgen) for parser driver generation. Specifically, it is needed in order to solve various problems where we have multiple grammars specified in the Antlr Maven test pom.xml file. It is a prerequisite for updating the csharp grammar. In that grammar, there is a preprocessor grammar that has been ported to the CSharp target (see [this](https://github.com/antlr/grammars-v4/pull/2431)), but it cannot compile because there is a dependency in the base class for the preprocessor grammar--which isn't being generated with the old, May 2021 version of trgen. The Antlr Maven test plugin specifies grammars that are compiled by the Antlr tool with a `<include>` element. This PR makes sure each and every grammar is compiled by the tool.

The changes include:
* Update to build to use version 0.13.2 of trgen.
* Update the Github Actions to set the version of Go and Cpp being used now with Antlr 4.9.3. There is no guidance in the [Go target doc](https://github.com/antlr/antlr4/blob/master/doc/go-target.md) and the [Cpp target doc](https://github.com/antlr/antlr4/blob/master/doc/cpp-target.md) is clearly out of date.
* Update the templates to be compatible with the new version of trgen.
* Update the skip lists to including skipping the "r" grammar for most targets because the grammar is Java specific. The "skip lists" are handled exclusively via a skip list regular expression, so the code in _scripts/regtest.sh were changed to follow the new arguments to trgen.
* Update the skip list processing in _scripts/regtest.sh, where grammars containing a `'-'` character were not being escaped.
* Fix the asm/masm and kirikiri-tjs grammars so that they work across targets.
* Finally, I added the Cpp target testing! Right now, I only test a few targets.


studentmain:
https://github.com/studentmain/0638f611-3374-495c-a27d-07c4c1d9243b/pull/7

seems we need merge #2438 first to "stabilize" it

kaby76:
> [studentmain#7](https://github.com/studentmain/0638f611-3374-495c-a27d-07c4c1d9243b/pull/7)
> 
> seems we need merge #2438 first to "stabilize" it

Correct. Your work should go in. I will need to add other information, e.g., "go env", "go version", "export | grep GO", to figure out why this environment does not work, while on my machine (Windows and Ubuntu) work fine. I do not have GOROOT nor GOPATH set, and as far as I know, cannot be set 1.16 and newer. Go is such a terrible language and environment.

studentmain:
> while on my machine (Windows and Ubuntu) work fine

just saying, on my machine, nothing works unless it's a module.

kaby76:
> > while on my machine (Windows and Ubuntu) work fine
> 
> just saying, on my machine, nothing works unless it's a module.

@studentmain Turns out make on my machine Bash shell and make get different environments. So, in the makefile, it was using the "old" way, and somehow worked. Obviously, it's different on the Github Actions machines. Using go.mod seems unavoidable at this point.

There is another issue with Go in which there are multiple grammars compiled by the Antlr tool. Antlr generates in a .go file a bunch of declarations that are global in the package declared at the top. Unfortunately, with multiple grammars, the generated code will have duplicate variables in the scope of the package. Since Go doesn't have classes, now I see why there is a "package parser" at the top of the generate .go file: it corrects for the name collision problems. I'll have to redo the trgen tool and templates for Go a little bit, but after I get this stabilized.

kaby76:
Closing out this specific PR and recreating a clean PR that has one commit instead of all this nonsense. I cannot get Git/Github/git rebase -i/git reset --soft/git diff to do what I want: one set of changes in the branch so I know exactly what was changed.

kaby76:
Replaced by https://github.com/antlr/grammars-v4/pull/2442

Issue #2435, opened 2021-12-20
trgen-based CI build and test does not test "preprocessor" grammars


kaby76:
The current version of trgen in the GitHub Actions is old (version 0.7.0, May 2021), and does not test some grammars correctly. 
* cobol85: This grammar contains a [preprocessor grammar](https://github.com/antlr/grammars-v4/blob/master/cobol85/Cobol85Preprocessor.g4) that is not run through the Antlr4 tool.
* csharp: This grammar contains a [preprocessor grammar](https://github.com/antlr/grammars-v4/blob/master/csharp/CSharpPreprocessorParser.g4) that is not run through the Antlr4 tool. In fact, if it did, it would find that it's not target agnostic!
* stringtemplate: There are actually two sets of gramars (ST, STG) for a total of six grammar files. All of the tests skip (except Cpp) stringtemplate. The pom.xml has an "excludes" tag that list the imported grammars via other grammars, but it's not actually used in testing: only the "includes" tag is required, and it's not used elsewhere in the repo! While skipped by trgen, the Maven test plugin does work.
* z: This grammar contains ZOperatorParser.g4 and a Java driver that first calls ZOperatorParser, walks the tree to find "associations", rewrites the token stream with type information from the walk, and finally calls ZParser. The trgen-based testing does not use the ZOperatorParser code, but the Maven-based test does.

A newer version of trgen must be used as a prerequisite for https://github.com/antlr/grammars-v4/pull/2431, which updates the csharp/ grammar to be "target agnostic", among other bug fixes.

PR #2442, opened 2021-12-26
Fix for Issue 2435 -- updating to latest trgen, add Cpp target

kaby76:
This is a fix for #2435, which updates trgen and added testing of the Cpp target to the GitHub Actions CI.

The tool [trgen](https://github.com/kaby76/Domemtech.Trash/tree/main/trgen) is used in this repository for testing, generating a parser driver and build files for a grammar. The newest version of trgen fixes several problems with the old version 0.7.0 from in May 2021.

The main problem with the old generator tool was that it was not testing all grammars marked with `<include>` elements in a pom.xml file. For example, the [csharp grammar](https://github.com/antlr/grammars-v4/tree/master/csharp) contains two grammars, [one for preprocessing](https://github.com/antlr/grammars-v4/blob/3d0c5f07b139ac95c50b29d79ccc7d7b6be2c920/csharp/pom.xml#L22), and a [second, a split grammar, for post-processed C#](https://github.com/antlr/grammars-v4/blob/3d0c5f07b139ac95c50b29d79ccc7d7b6be2c920/csharp/pom.xml#L23). In order to write a target agnostic grammar for csharp, the preprocessor must have base classes for each target, which cannot not be written because the preprocessor grammar would not be compiled by the Antlr tool.

The changes include the following:
* Update use of trgen to version 0.13.2.
* Update the templates to work with the new version of trgen. This is where most of the changes are.
* Add Cpp testing. For now, we only test a small number of the grammars until we see how this goes.
* Update to the skip lists to skip the Java-specific grammar "r" for most targets.
* Update _scripts/regtest.sh to handle correctly the skip lists.
* Fix kirikiri-tjs to be target agnostic.
* Fix asm/masm to be target agnostic.
* Remove package information in kotlin/kotlin-formal because this does not work in JavaScript, Go, etc.
* Fix incorrect `<include>` entry in oncrpc for an included grammar file--don't run the Antlr tool on imported grammars.

This replaces the previous, bloated PR https://github.com/antlr/grammars-v4/pull/2439


teverett:
@kaby76  thanks

PR #2444, opened 2021-12-27
Fix for Issue 2430

kaby76:
This is a fix for Issue 2430 (parsing C# code with preprocessor directives), and a few other issues that have been lingering around for a while.

* Issue 2430: This issue is about parsing macros used in the grammar. The fix changes the lexer rule `SHARP` so that the token is not passed to the parser. It isn't passed either to the preprocessor grammar, so the change is okay. I don't know why, but the poster, @DivyaSrivastava45 , closed the bug, but it was a legit bug that needed to be fixed. Note, preprocessing is not handled, but at least files with preprocessing directives doesn't just crash.
* Issue 2329: This issue is about the parser being able to accept "var a=1,b=2;" multiple declarations with var. This is illegal with the C# compiler, resulting in an error message. The issue is that it should not be allowed in this grammar. Since the [generated grammar from the Roslyn compiler](https://github.com/dotnet/roslyn/blob/main/src/Compilers/CSharp/Portable/Generated/CSharp.Generated.g4) itself does not introduce rules to distinguish the problem, I decided to implement the solution as a semantic predicate. After fixing the bug, I found out that the examples/AllInOneNoPreprocessor.cs file contains the very multiple var declaration problem, so I changed the test.
* The grammar was not in "target agnostic" structure, and actually had Antlr "members" and actions all in Java in the grammar. I moved all the code to base classes, updated the Java and Antlr4cs targets (we don't test Antlr4cs in the CI, but I checked it), and added the CSharp target. All three grammars have been ported to "target agnostic" and all three are now processed through the Antlr tool.

This PR replaces https://github.com/antlr/grammars-v4/pull/2431 to squash the commit history and to roll forward with a dependency with trgen changed with https://github.com/antlr/grammars-v4/pull/2442


teverett:
@kaby76  thanks

PR #2446, opened 2021-12-28
Fix for Issue 2433 -- Rust is case sensitive

kaby76:
Although not explicitly stated in the Rust "Spec", Rust is case sensitive. I checked this against the compiler against a few examples (changing keywords to mixed case, referencing variables in a different case from the definition). Anyway, "true" for the "caseInsensitiveType" in the Antlr Maven Plugin is not a valid value, and the element "caseInsensitive" as far as I know is not even a valid element! So, it is best to remove the element.


teverett:
@kaby76  thanks

Issue #2440, opened 2021-12-25
Potential bug in Java grammar


mike-lischke:
I'm currently using [this Java grammar](https://github.com/antlr/grammars-v4/tree/master/java/java) in a project and found something which I believe is a bug. Look at these interface method rules:

```g4
interfaceMethodDeclaration
    : interfaceMethodModifier* (typeTypeOrVoid | typeParameters annotation* typeTypeOrVoid)
      IDENTIFIER formalParameters ('[' ']')* (THROWS qualifiedNameList)? methodBody
    ;

genericInterfaceMethodDeclaration
    : typeParameters interfaceMethodDeclaration
    ;
```

As you can see the generic interface method decl uses the normal one. However, looking closer you can see that `interfaceMethodDeclaration` actually also parses type parameters (so it's actually generic too).

However `genericInterfaceMethodDeclaration` **also** parses type parameters, allowing so 2 sets of parameters, which is wrong IMO. But since I'm not so fluent in Java, it could be I'm wrong here.

PR #2447, opened 2021-12-28
fix for #2440

mlorek:
fix for #2440


KvanTTT:
@mike-lischke could please check if the fix resolves your problem?

mike-lischke (approved):
Looks good, thanks.

PR #2453, opened 2021-12-29
DBZ-4497 Parse multiple signed/unsigned keyword from ddl statement failed

harveyyue:


sql/mysql/Positive-Technologies/MySqlParser.g4
```diff
@@ -2107,13 +2107,13 @@ dataType
         TINYINT | SMALLINT | MEDIUMINT | INT | INTEGER | BIGINT
         | MIDDLEINT | INT1 | INT2 | INT3 | INT4 | INT8
       )
-      lengthOneDimension? (SIGNED | UNSIGNED)? ZEROFILL?            #dimensionDataType
+      lengthOneDimension? (SIGNED | UNSIGNED | ZEROFILL)*?          #dimensionDataType
```
KvanTTT, line 2110: Maybe extract `SIGNED | UNSIGNED | ZEROFILL` to subrule since it's used in several places? Also, `?` is redundant if `*` is used:

```suggestion
      lengthOneDimension? (SIGNED | UNSIGNED | ZEROFILL)*          #dimensionDataType
```

harveyyue: Yes, I will remove `?`.
Debezium implemented some logic dependence on these as token type, so I didn't extract these token as subrule.

Issue #1972, opened 2020-12-26
Java 10/11 grammar request


doctorjei:
I was wondering if anyone is considering working on a Java 10/11 grammar. I will also look into helping with this where I can.

PR #2456, opened 2021-12-29
java 11 upgrade - #1972

mlorek:


java/java/JavaParser.g4
```diff
@@ -377,6 +410,12 @@ blockStatement
 
 localVariableDeclaration
     : variableModifier* typeType variableDeclarators
+    | variableModifier* VAR identifier '=' expression
```
KvanTTT, line 413: Alternatives can be merged:

```suggestion
    : variableModifier* (typeType variableDeclarators | VAR identifier '=' expression)
```

It's important because the first part can match the empty string, which may affect performance.

java/java/JavaParser.g4
```diff
@@ -426,7 +465,7 @@ resources
     ;
 
 resource
-    : variableModifier* classOrInterfaceType variableDeclaratorId '=' expression
+    : (variableModifier* ( (classOrInterfaceType variableDeclaratorId) | (VAR identifier) ) '=' expression) | identifier
```
KvanTTT, line 468: Simplify parentheses and fix formatting:

```suggestion
    : variableModifier* (classOrInterfaceType variableDeclaratorId | VAR identifier) '=' expression
    | identifier
```

java/java/JavaParser.g4
```diff
@@ -377,6 +410,12 @@ blockStatement
 
 localVariableDeclaration
     : variableModifier* typeType variableDeclarators
+    | variableModifier* VAR identifier '=' expression
+    ;
+
+identifier
+    : IDENTIFIER
+    | VAR
```
KvanTTT, line 418: It looks like all newly added keywords should be added to `identifier` as well. The following code is valid (you can check it online at [repl.com](https://replit.com/@IvanKochurkin/JavaKeywordsAsIdentifiers#Main.java)):

```java
class Main {  
  public static void main(String args[]) { 
    var module = "module";
    var open = "open";
    var requires = "requires";
    var exports = "exports";
    var opens = "opens";
    var to = "to";
    var uses = "uses";
    var provides = "provides";
    var with = "with";
    var transitive = "transitive";
    var var = "var";
    System.out.println(module);
    System.out.println(open);
    System.out.println(requires);
    System.out.println(exports);
    System.out.println(opens);
    System.out.println(to);
    System.out.println(uses);
    System.out.println(provides);
    System.out.println(with);
    System.out.println(transitive);
    System.out.println(var);
  } 
}
```

```suggestion
identifier
    : IDENTIFIER
    | MODULE
    | OPEN
    | REQUIRES
    | EXPORTS
    | OPENS
    | TO
    | USES
    | PROVIDES
    | WITH
    | TRANSITIVE
    | VAR
```

Also, I suggest adding this test to `examples`.

KvanTTT (changes requested):
Thank you for your effort! But some issues should be fixed before merging.

studentmain:
we should rename folder to java11

KvanTTT:
It's an update of existing grammar that does not follow the official documentation. It always was just `java`. We have other grammars that follow official documentation: [java8, java9](https://github.com/antlr/grammars-v4/tree/master/java) but they slow and barely usable.

java/java/JavaParser.g4
```diff
@@ -32,9 +33,31 @@ parser grammar JavaParser;
 options { tokenVocab=JavaLexer; }
 
 compilationUnit
-    : packageDeclaration? importDeclaration* typeDeclaration* EOF
+    : (packageDeclaration? importDeclaration* typeDeclaration*)
```
KvanTTT, line 36: Why do you use parentheses here? They are redundant since `|` operator has the lowest priority.

```suggestion
    : packageDeclaration? importDeclaration* typeDeclaration*
```

java/java/JavaParser.g4
```diff
@@ -426,7 +475,8 @@ resources
     ;
 
 resource
-    : variableModifier* classOrInterfaceType variableDeclaratorId '=' expression
+    : (variableModifier* ( classOrInterfaceType variableDeclaratorId | VAR identifier ) '=' expression)
```
KvanTTT, line 478: Also redundant parentheses:

```suggestion
    : variableModifier* ( classOrInterfaceType variableDeclaratorId | VAR identifier ) '=' expression
```

KvanTTT (approved):

KvanTTT:
Thanks again!

KvanTTT:
Also, please add info about Java 11 to README: https://github.com/antlr/grammars-v4/tree/master/java/java#supported-java-versions (also you can add yourself as the main contributor as well).

Issue #2353, opened 2021-10-12
Java 17 support


Clashsoft:
Java 17 is the latest long-term support release. A number of language features were added that should be supported in the grammar.

- [ ] Switch Expressions
- [ ] Text Blocks
- [ ] Records
- [ ] Pattern Matching for `instanceof`
- [ ] Sealed Classes (`sealed` and `non-sealed` keywords)
- [ ] Pattern Matching for `switch` expressions and statements

https://docs.oracle.com/en/java/javase/17/language/java-language-changes.html

omegaui:
When is this likely to get done!

mlorek:
> When is this likely to get done!

soon

esoftGithub:
hi mlorek i dont know how github works for now.
is java 17 grammar completed where can i see it?
thanks a lot.

mlorek:
> hi mlorek i dont know how github works for now. is java 17 grammar completed where can i see it? thanks a lot.

https://github.com/antlr/grammars-v4/blob/master/java/java/JavaParser.g4#L6

PR #2464, opened 2021-12-31
java 17 upgrade

mlorek:
#2353

### DIFF
--- a/java/java/JavaLexer.g4
+++ b/java/java/JavaLexer.g4
@@ -2,6 +2,8 @@
  [The "BSD licence"]
  Copyright (c) 2013 Terence Parr, Sam Harwell
  Copyright (c) 2017 Ivan Kochurkin (upgrade to Java 8)
+ Copyright (c) 2021 Michał Lorek (upgrade to Java 11)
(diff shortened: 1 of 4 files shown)


teverett:
@mlorek thanks

java/java/JavaLexer.g4
```diff
@@ -97,6 +99,17 @@ TRANSITIVE:         'transitive';
...
 
+// Switch Expressions
+YIELD:              'yield';
+
+// Records
+RECORD:             'record';
+
+// Sealed Classes
+SEALED:             'sealed';
+PERMITS:            'permits';
```
KvanTTT, line 110: They all also can be identifiers. Please all `yield` `record`, `sealed`, `permits` to `identifier` rule.

mlorek: > They all also can be identifiers. Please all `yield` `record`, `sealed`, `permits` to `identifier` rule.

it's already there

KvanTTT: Sorry, I missed that. But `yield` is still missing and it's an acceptable identifier as well.

PR #2465, opened 2022-01-01
python3: add JS baseclass, remove python3-js

studentmain:
since we now have JS CI to make sure it works


studentmain:
There are still two grammars remaining.

python3-py is blocked by grammar translator, without that, we still need to copy paste most of the code, not much better than current situation.
python3-ts is blocked by typescript target test. I'm waiting for it become an official supported target.

teverett:
@studentmain  thanks

PR #2468, opened 2022-01-02
Add XPath2.0 grammar.

kaby76:
This is the XPath2.0 grammar that I adapted from the xpath31/ grammar, and updated with the rules from the XPath2.0 Spec at https://www.w3.org/TR/xpath20/.


teverett:
@kaby76  thanks

PR #2471, opened 2022-01-05
tsql - Add Azure Synapse Analytics specific options for CREATE TABLE

grschneider:
Current T-SQL lexer/parser fails to parse certain Azure Synapse Analytics specific CREATE TABLE options (doc: https://docs.microsoft.com/en-us/sql/t-sql/statements/create-table-azure-sql-data-warehouse?view=aps-pdw-2016-au7).
I added these and created an example file for testing purposes.


KvanTTT:
Thanks for the contribution and welcome to ANTLR grammars repostory!

Issue #2448, opened 2021-12-28
Invalid antlr4test-maven-plugin values


kaby76:
I added a check to [trgen](https://github.com/kaby76/Domemtech.Trash/tree/main/trgen) to test all [antlr4test-maven-plugin](https://github.com/antlr/antlr4test-maven-plugin) values in the pom.xml files for each grammar. It looks like there are several entries that should be changed in keeping up to date with the latest Anltr Maven Plugins:

* Invalid element "//plugins/plugin[artifactId='antlr4test-maven-plugin']/configuration/caseInsensitive. In [dice](https://github.com/antlr/grammars-v4/blob/715f1a948f549c60282493c1a22b36154fad5251/dice/pom.xml#L44) -- should this be in the antlr4-maven-plugin section?
* Invalid element "//plugins/plugin[artifactId='antlr4test-maven-plugin']/configuration/caseInsensitive. In [kotlin/kotlin-formal](https://github.com/antlr/grammars-v4/blob/715f1a948f549c60282493c1a22b36154fad5251/kotlin/kotlin-formal/pom.xml#L44) => remove.

There is not much documented on the acceptable element types for the antlr4-maven-plugin [here](https://www.antlr.org/api/maven-plugin/latest/index.html), or [anything under the source in github](https://github.com/antlr/antlr4/tree/master/antlr4-maven-plugin). So, I can't be sure of what to implement for checking invalid entries. But, after examining all the grammars, I've noticed only a few additional items that should be changed:

* //plugins/plugin[artifactId='antlr4-maven-plugin']/configuration/grammars in [alloy](https://github.com/antlr/grammars-v4/blob/715f1a948f549c60282493c1a22b36154fad5251/alloy/pom.xml#L21). Change "grammars" to "includes" and "include" in keeping with the structure in all the other grammars in the repository.
* //plugins/plugin[artifactId='antlr4-maven-plugin']/configuration/outputDirectory in [golang](https://github.com/antlr/grammars-v4/blob/715f1a948f549c60282493c1a22b36154fad5251/golang/pom.xml#L45). This probably isn't needed because the code is now in target agnostic format (i.e., grammar in the base directory and target-specific code in CSharp/, Java/, etc). Remove.
[...]

PR #2473, opened 2022-01-09
Fixes for #2448

kaby76:
This is a collection of several fixes to the pom.xml in the repo. This is being done now because the next version of trgen will flag errors for invalid pom.xml.

* Invalid (or just old, non-standard) elements in several pom.xml are removed. Trgen had been warning about some of the entries up to this point.
* Reorganize sql/hive to be like other multi-version grammars (eg, java, antlr). With the reorganization, I discovered that sql/hiv/v2 pom.xml was invalid (grammarName entry), thus never actually tested by Maven!
* Update skip lists to avoid hive/v1 and hive/v2.


teverett:
@kaby76 thanks

Issue #2452, opened 2021-12-29
PLSQL/Golang：need a demo how to build a parser for plsql using Golang


wangzhanbing:
I want to parser sql from oracle. and get some information, for example: db/table/column/index and so on.
then I refer to some introduction from web like following, but it doesn't work
can someone support a demo for me?

 `func main() {
	sql := "select colA from tblname where id in (1,2,3);"
	input := antlr.NewInputStream(strings.ToUpper(sql))

	lexer := parser.NewPlSqlLexer(input)

	stream := antlr.NewCommonTokenStream(lexer, 0)

	p := parser.NewPlSqlParser(stream)
	p.AddErrorListener(antlr.NewDiagnosticErrorListener(true))
	p.AddParseListener(NewTreeShapeListener())
	p.BuildParseTrees = true

	parent := antlr.NewBaseParserRuleContext(nil, -1)
	p.SetParserRuleContext(parent)

	p.Consume()

	return
}`

studentmain:
https://github.com/antlr/grammars-v4/blob/7d7a5051dd17f3b4c09015f45a255809d577e2df/_scripts/skip-go.txt#L75

https://github.com/antlr/grammars-v4/commit/cb406a9e42c4fbdcf1cadd9cd2853655fdf404f6

We don't know how to build it either now. I'll take a look later.

By "it doesn't work", what happened? Any reproduce steps?

wangzhanbing:
as expected, the function EnterSelect_only_statement & EnterFrom_clause & EnterWhere_clause should be called. 
but nothing happend.

kaby76:
The Go/ target files is more aspirational than functional. The problem is that the Go target generates actions that pass "l" and "p" pointers, not "self". The use of "self" in the grammar action rules cannot work. There was some discussion where I tried to convince folks who wrote the Go target using "l" and "p" are really really bad names, because it makes "target agnostic" grammars impossible.

The first thing to fix is to make a functioning set of grammar files. You'll need to change the use of "self" in the grammars. In the lexer grammar, change "self" to "l". In the parser grammar change "self" to "p". Then, you will need to write all the base class code, which currently doesn't even compile.

kaby76:
It looks like 4.9.3 has changed the names of the generated parameter to actions. It's now all "p". So something may have clicked in the Go target authors.

So, change "self." to "p." in all the grammar actions.

studentmain:
[...]

PR #2474, opened 2022-01-10
Fixes for #2452

kaby76:
This is a fix to the sql/plsql grammar for Go.

It also uses an updated trgen, v0.13.3, which includes a change to the "-s" option for @studentmain.

The changes had to dance around quite a few limitations in the Antlr tool 4.9.3 for Go ("golang").

* Believe or not, the Antlr tool 4.9.3 does not generate code to initialize base classes! This means that "_isVersion12" which is true must be "inverted" for the Go target. Base class are written for Go with "zero defaults" because that is the default initialization for a Go struct.
* The Antlr tool generates action code to use parameter "p" for Go, not "self" or "this". So, since actions must use "p", grammars have  to be duplicated in the Go-specific code directory with the actions changed. Presumably when @studentmain updates the build to include translation, we can remove these duplicate files.
* The Antlr tool ALWAYS creates a package for the generated files! "trgen" now generates code to a specific package, the default being "parser" since whether you like or not, that's what Antlr tool will generate. The base classes have to be placed in Go/parser/ because that's how it works in all the other targets--the package and directory are one and the same. In addition, the Antlr tool generates several global variables in the package for Go, so this is yet another reason why grammars have to be placed in their own package/directory for Go!
* Other changes: rename the base class (and file) for the Go target to be identical to how it's done for all the other targets. We need to be consistent in the naming of the base classes. Flipping the order of "parser base" vs "base parser" per target is ridiculous!

Note: It's really apparent that whoever wrote the Go target did not test the Go target with many grammars. While there are many things good with the port to Go, there are many things that need to be changed to have a consistent experience with Go compared to the other targets.


kaby76:
Oh yes, there were changes that I had for golang/ grammar, but didn't add them to the commit.

kaby76:
Finally, I found out that trgen was not working, and templates for Go were wrong. Updating to latest trgen, then will try again. This is why I was never getting any output. Likely will add comments with generated .psm1 (.ps1?, what is the diff) that outputs everything for future debugging without being an expert in pwsh ( or powershell.exe? There is a diff.) Many things confusing about pwsh because it's one big monolithic shell tather than separate executables that can be composed together.

kaby76:
Closing this PR and moving the changes into one nice commit and new PR. https://github.com/antlr/grammars-v4/pull/2476

PR #2476, opened 2022-01-12
Fix for #2452 -- updates to sql/plsql

kaby76:
This is a fix to the sql/plsql grammar for Go.

It also uses an updated trgen, v0.13.8, which includes a change to the "-s" option for @studentmain.

The changes had to dance around quite a few limitations in the Antlr tool 4.9.3 for Go ("golang").

* Believe it or not, the Antlr tool 4.9.3 does not generate code to initialize parser or lexer base "classes" (structs)! This means that "_isVersion12", which is initialized in constructors to true in the CSharp and Java targets, must be "inverted" for the Go target. Base "classes" (structs) for Go must be written with "zero defaults" because that is how initialization works for the moment with the Go target. Go doesn't have constructors per se, using the New...() coding standard to initialize a struct. But we cannot initialize the nested  struct ( i.e. "the base class for the parser") because that func call is not generated by the Antlr tool in the code for the parser or lexer.
* The Antlr tool generates action code to use parameter "p" for Go, not "self" or "this". So, since actions in the grammar must use "p", grammars have to be duplicated in the Go-specific code directory with the actions changed. Presumably, when @studentmain updates the build to include translation, we can remove these duplicate files. The other way to fix this would be to change the go.stg template, but that won't happen quickly.
* The Antlr tool always generates a "package" declaration at the top of the generated files. "trgen" now generates code to a specific package, the default being "parser" since that's what Antlr tool generates. The base classes have to be placed in Go/parser/. In addition, the Antlr tool generates several global variables in the package for Go, so this is yet another reason why grammars have to be placed in their own package/directory for Go.
* The parser and lexer base class files were renamed to be similar to the file names for the other targets.
* Several pom.xml contained illegal or useless elements.

Ideally, there should be one grammar for all targets. However, Python3 requires "self" and Go requires "l" or "p" or "this". The parameter should be standardized across targets so we can achieve target agnostic grammars. Again, this can be fixed by adding a translation phase to change those in the grammar after trgen, or it can be fixed by adjusting the .stg templates for the Antlr tool. Unfortunately, there's quite a bit of resistance for either, and forking grammars, so who knows.


kaby76:
Ready for review and merge. I was making sure trgen would work in other environments besides the grammars-v4 build here. It was a bit of tweaking.

_scripts/templates/Go/Test.go
```diff
@@ -7,7 +7,7 @@ import (
     "io"
     "time"
     "github.com/antlr/antlr4/runtime/Go/antlr"
-    "example.com/myparser/parser"
+    "example.com/myparser/<package_name>"
```
kaby76, line 10: Whiel the default "package" generated by the Antlr4 tool is "parser", trgen now accepts a package spec in the pom. For now, CI just let's it be "parser", so the generated code is the same. However, there is a big problem in the Antlr4 tool for Go: one cannot have multiple parsers or lexers in one directory. trgen does not determine how to partition the parser grammars and associated source code into separate directories, and I would prefer that Antlr just handle this by qualifying the names of global variables in the code that it generates with the name of the grammar.

_scripts/templates/Go/makefile
```diff
@@ -8,9 +8,9 @@ SOURCES = $(GENERATED) Test.go
 default: classes
 classes: setup $(SOURCES) program
 setup:
-	export | grep GO
-	go env
-	go version
+	-export | grep GO
+	-go env
+	-go version
```
kaby76, line 13: By default, "make" will stop when a command reports an error. It is possible that one does not have any "GO" environmental variables set, so the "grep" may fail. I put the '-' sign on all these lines so that it fails on the "main" build step.

_scripts/templates/Go/tester.psm1
```diff
@@ -1,14 +1,20 @@
 # Template generated code from trgen <version>
 function Build-Grammar {
-<tool_grammar_files:{x |
-    $g = antlr <x> -Dlanguage=Go <antlr_tool_args:{y | <y> } >
+<tool_grammar_tuples:{x |
+    $g = antlr <x.GrammarFileName> -encoding <antlr_encoding> -Dlanguage=Go <x.AntlrArgs> <antlr_tool_args:{y | <y> } >
```
kaby76, line 4: Some of the attributes were changed. I now have all the grammars in a set of tuples, which describe the name of the grammar, the location of the grammar file in the original grammars-v4 directory, the target location (in "Generated/"), the name of the function to call to create the parser, etc. Also, I notice that the tester.psm1 script wasn't setting the encoding of the input grammar. Most of the time this is empty, but there are a couple of grammars in grammars-v4 where it is not. For "Go", I added "AntlrArgs" which is a per-grammar set of options. Unfortunately, Windows and Linux differ, where -o/-lib are required for Windows, and nothing for Linux. I noted this problem in Antlr, and someone recently ran across the problem, but likely will never be fixed.

_scripts/templates/Go/tester.psm1
```diff
@@ -1,14 +1,20 @@
...
             Success = $false
         \}
     \}
 }>
+    # Output version of pwsh.
+    #Get-Host | Select-Object Version | Write-Host
+    # Output environmental variables.
+    #dir env: | Out-String | Write-Host
+    # Output go version
+    #go version | Write-Host
```
kaby76, line 17: I'm adding all this "debugging" code in comments. Powershell is extremely difficult to use. Years of working with it have taught me that. Unless one programs in it every day, it's best to save the code to get basic debugging information out. I spent days trying out various codes to print debugging info. For example, in Bash, "export" just prints the list of environmental variables. It's been that way for ~30 years. What is the equivalent on Powershell? Well, if you look that up, you get dozens of answers in Stackoverflow. "dir env:" seems to be the most favored, but it never worked. The problem turned out to be the fact that Windows11 installs an ancient version of Powershell in c:/Windows/ that does not support "dir env:". YOU HAVE TO INSTALL THE LATEST PWSH.EXE!

golang/Go/parser/go_parser_base.go
```diff
@@ -2,7 +2,6 @@ package parser
 
 import (
 	"strings"
-
```
kaby76, line 5: I never understood why people have this blank line between "system" packages, and "local". Is it a "Go" requirement??? For God's sake, a package is a package! Stop this needless blank line crap. There are *so* many things wrong with Go in forcing formatting (requiring "{" to be in hanging style), we need to stop adding syntax that adds to the confusion.

kotlin/kotlin/pom.xml
```diff
@@ -45,7 +45,6 @@
 					<verbose>false</verbose>
 					<showTree>false</showTree>
 					<grammarName>Kotlin</grammarName>
-					<caseInsensitive>true</caseInsensitive>
```
kaby76, line 48: This element is illegal. "true" is not a valid value for caseInsensitive. The Maven plugin doesn't check for bad values, but trgen now does, forcing the user to correct bad information before it gets copied and pasted around into other grammars by people who don't know what they are doing.

kaby76: For Antlr 4.10, it seems that "caseInsensitive=true" is only available as an option in the grammar options section. _It isn't available as a command-line option on the call for the tool (nowhere [here](https://github.com/antlr/antlr4/blob/1b144fa7b40f6d1177c9e4f400a6a04f4103d02e/tool/src/org/antlr/v4/Tool.java#L121))_. Therefore, "caseInsensitive=true" should be removed from the pom.xml file, and the rest of the occurrences of caseInsensitive will need to be changed. https://github.com/antlr/antlr4/issues/3436 https://github.com/antlr/antlr4/commit/d8e152592fb4dbdd6c9fda2c169031cca17dfc4f

kaby76:
I will be replacing this PR since the PR for PHP builds will conflict with this one.

Issue #2483, opened 2022-01-21
Limited support for subqueries in PromQL grammar


antonio-cuomo:
While using the PromQL grammar, I realised that it only (partially) supports a special case of [subquery](https://prometheus.io/docs/prometheus/latest/querying/basics/#subquery), the one where the subquery is applied to a matrixSelector. i.e.:

```
http_requests_total[5m:]
```
Such support is managed though a special case of TIME_RANGE lexing. 

In the general case, subqueries can apply to any instant query, as in one of the official doc [examples](https://prometheus.io/docs/prometheus/latest/querying/examples/#subquery): 

```
rate(http_requests_total[5m])[30m:1m]
```

The above subquery fails with the current grammar.

PR #2484, opened 2022-01-21
Fix for #2483 - Support PromQL subqueries

antonio-cuomo:
Addresses issue #2483


teverett:
@antonio-cuomo  thanks

PR #2485, opened 2022-01-24
Added a grammar for Unreal Angelscript.

Temaran:
A grammar for the Unreal variant of the Angelscript scripting language (https://angelscript.hazelight.se/)

Based on the C++ grammar made by Camilo Sanchez (Camiloasc1) and Martin Mirchev (Marti2203) above.
Adapted to Unreal Angelscript by Embark Studios AB (Fredrik Lindh).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -292,6 +292,7 @@
 				<module>turtle</module>
 				<module>turtle-doc</module>
 				<module>unicode</module>
+				<module>unreal_angelscript</module>
 				<module>upnp</module>
 				<module>url</module>
 				<module>useragent</module>
--- a/unreal_angelscript/README.md
+++ b/unreal_angelscript/README.md
@@ -1,0 +1,6 @@
+# Unreal Angelscript
+
+A grammar for the Unreal variant of the Angelscript scripting language (https://angelscript.hazelight.se/)
+
+Based on the C++ grammar made by Camilo Sanchez (Camiloasc1) and Martin Mirchev (Marti2203) above.
+Adapted to Unreal Angelscript by Embark Studios AB (Fredrik Lindh).
--- a/unreal_angelscript/UnrealAngelscriptLexer.g4
+++ b/unreal_angelscript/UnrealAngelscriptLexer.g4
@@ -24,8 +24,6 @@ StringLiteral:
 		| Escapesequence
 	)* '"';
 
-BooleanLiteral: False | True;
-
 UserDefinedLiteral:
 	UserDefinedIntegerLiteral
 	| UserDefinedFloatingLiteral
@@ -102,7 +100,7 @@ Enum: 'enum';
 
 Export: 'export';
 
-False: 'false';
+False_: 'false';
 
 Final: 'final';
 
@@ -136,7 +134,7 @@ Switch: 'switch';
 
 This: 'this';
 
-True: 'true';
+True_: 'true';
 
 Virtual: 'virtual';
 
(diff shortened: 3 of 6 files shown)


teverett:
[INFO] --- antlr4-maven-plugin:4.9.3:antlr4 (default) @ unreal_angelscript ---
[INFO] ANTLR 4: Processing source directory /home/runner/work/grammars-v4/grammars-v4/unreal_angelscript
[INFO] Processing grammar: UnrealAngelscriptParser.g4
Error:  error(114): UnrealAngelscriptParser.g4:30:14: cannot find tokens file ./UnrealAngelscriptLexer.tokens
Error:  /home/runner/work/grammars-v4/grammars-v4/UnrealAngelscriptParser.g4 [30:14]: cannot find tokens file ./UnrealAngelscriptLexer.tokens

Temaran:
Hm, how strange, I thought the tokens file was generated as part of the workflow. Going to investigate a bit :)

Temaran:
Maybe it was just missing an example file..? If it doesn't have one, maybe it cannot generate the token file etc?

Temaran:
Hm, not sure how to proceed here. Do I need to check in the tokens file? I don't see it in any other package though.
Is there some way I can test this locally to debug further?

The grammar works without problems in my application, so I'm not sure what might be missing here. 🤔

kaby76:
The pom.xml is malformed. [trgen](https://github.com/kaby76/Domemtech.Trash/tree/main/trgen) assumes that you placed each and every grammar that must be run through the Antlr4 .jar tool. The pom.xml `<includes>` should not contain any "import" grammars whatsoever because the Antlr4 .jar tool does not process those files.

The problem is [here](https://github.com/Temaran/grammars-v4/blob/70d041985484027fbddd1432a1dd022e8d0b651b/unreal_angelscript/pom.xml#L21). Add `<include>UnrealAngelscriptLexer.g4</include>`, before the UnrealAngelscriptParser.g4 file line. (It can probably go after the Parser.g4 line, but I you may need to test that. Trgen reads all the .g4 files and performs an analysis of the import and vocabulary dependencies, then orders the call to the Antlr tool.

Temaran:
Oh, thank you for the insight!

I just installed maven and tried running it locally, which worked, but still gives the same error. I've been suspecting the POM since it's the only thing that I don't use in my project, but excellent to have confirmation that that is the probably issue!

I'll see if I can figure it out <3

kaby76:
The other problem is the grammar name in the testing of the parser. It cannot find the parser using reflection because you set the name to `UnrealAngelscriptParser`. Change [this line](https://github.com/Temaran/grammars-v4/blob/70d041985484027fbddd1432a1dd022e8d0b651b/unreal_angelscript/pom.xml#L42) to use `UnrealAngelscript`. With these changes, it all now works.

Temaran:
a-ha! I just hit this as well when testing locally, excellent! Thank you! <3

Temaran:
<3 It's running for me now locally as well! Thank you for all the help @kaby76 ! <3

kaby76:
* There are "symbol conflicts" in the grammar with the Go and Python3 targets. Antlr 4.10 will have a fix for this, but for now, v4.9.3, you will have to rename a few symbols. `newInitializer` => `newInitializer_`; `emptyDeclaration` => `emptyDeclaration_`. `True` => `True_`; `False` => `False_`.
* `BooleanLiteral` is a lexer token that hides `True` and `False`. Fortunately, you don't refer to `True` and `False` in the parser grammar. But, you should move `BooleanLiteral` to the parser (rename to `booleanLiteral`) or make `True` and `False` fragment (in addition to renaming with '_').

teverett:
@Temaran  thanks

Issue #2488, opened 2022-01-25
PHP target of grammars-v4 missing, new class of "symbol conflicts"


kaby76:
I am now implementing the PHP target for trgen for CI builds. I have detected a number of issues.

* PHP has symbol conflicts when a grammar uses the same symbol name but using different case, one for a lexer symbol and the other for a parser symbol. Unlike previous "symbol conflicts" for all other targets, the Antlr tool has this new restriction, making many of the grammars here not functional for PHP. E.g., defining a grammar with `nCName` and `NCName` (in xpath/xpath1) will result in a compilation error due to the Antlr tool generating declarations that conflict with each other. "BLOCK" in lexer and "block" in parser. grammars-v4/wat.
* "CLASS" as a lexer rule name is illegal. I don't know if this is caused by PHP being case insensitive.

There are other issues as well--timeouts and crashes. I haven't worked with PHP in a while,  and I don't recall if there is case sensitivity. E.g., "True" and "true" are the same in PHP.

kaby76:
https://github.com/antlr/antlr4/issues/3462

Broken in:

* acme: identifier
* antlr/antlr2: class
* antlr/antlr3: requires base class
* antlr/antlr4: requires base class
* apex: class
* asn/asn_3gpp: not target agnostic
* cobol85: class
* cool: class
* cpp: class
* csharp: class
* cto: Undefined offset: 0 in /mnt/c/Users/Kenne/Documents/GitHub/grammars-kaby76/cto/Generated/vendor/antlr/antlr4-php-runtime/src/Atn/ATNDeserializer.php on line 731
* fortran77: not target agnostic
* gdscript: class
* golang: requires base class
* haskell: class
* html: Undefined offset: 0 in /mnt/c/Users/Kenne/Documents/GitHub/grammars-kaby76/cto/Generated/vendor/antlr/antlr4-php-runtime/src/Atn/ATNDeserializer.php on line 731
* calendar: action
* Informix: attribute
* iri: Undefined offset: 0 in /mnt/c/Users/Kenne/Documents/GitHub/grammars-kaby76/cto/Generated/vendor/antlr/antlr4-php-runtime/src/Atn/ATNDeserializer.php on line 731
* java/java: class
* java/java8: class
* java/java9: class
* javadoc: not target agnostic, _input.
* javascript/javascript: class
* javascript/jsx: class
* javascript/typescript: class
[...]

PR #2489, opened 2022-01-26
Add testing for PHP target

kaby76:
This is a partial solution to #2488. The code "skips" the grammars that I mentioned in the comments for the issue. The PR uses a new version of trgen required for PHP and Go, templates for PHP, updated skip lists (removing some bugs). Although the Powershell version of the CI works, I chose to use the regtest.sh version because the Antlr4 PHP runtime has a restriction in PHP 7.4 that I did not anticipate, and so limited the testing of PHP just to Linux (as the older version doesn't seem available on other OSes).


kaby76:
$skip = Get-GrammarSkipList -Target $Target | Resolve-Path -Relative

test.ps1 requires the skip list to be hard-core paths in the file system, while regtest.sh allows them to actually be patterns. Unfortunately, when I made the list online, "autocorrecting" mutated the names, which I transcribed to the actual skip-php.txt file.

kaby76:
The Antlr 0.5.1 PHP runtime is compatible with PHP8.1 so I'm not sure why there is a problem. with composer. https://github.com/antlr/grammars-v4/runs/4960091050?check_suite_focus=true#step:9:5437. I will try to address this later.

teverett:
@kaby76  is this ready to pull?

kaby76:
@teverett Yes. This is ready to merge. If we want the grammars that are currently on the skip list for PHP to actually work with PHP, we will need to either change symbols in the grammars that cause conflicts (like "Class: 'class'" in a lexer grammar) or wait for Antlr 4.10. In the meanwhile, let's make sure people are aware of symbol conflicts for PHP. @KvanTTT 

On "target agnostic grammars", I think we are actually very close to finding a solution. I found an undocumented action attribute "$parser" that works for most targets. Antlr would need to be extended for "$lexer" to access the lexer class or "$self" to access either lexer or parser. We'd also need a change to Cpp .stg for rewrite "$parser" into "(*this)" and extending for PHP with a rewrite of "$parser." to "$this->".

teverett:
@kaby76  thanks

PR #2492, opened 2022-01-31
Hive: incorrect grammar version

wyruweso:
There is a little confusion with versions after [this PR](https://github.com/antlr/grammars-v4/issues/2448) for Hive grammar. "v1" should be actually "v3" as we have in this repository only 3.1.2 and 2.3.8 versions of Hive grammar.


teverett:
@wyruweso  thanks

Issue #2496, opened 2022-02-03
Golang: Missing semicolons not properly induced


nicoberling:
The parser currently doesn't induce semicolons correctly:
The statementList
```golang
i := a
+b
```
is parsed as a single line `i := a + b`:
![semi](https://user-images.githubusercontent.com/15216033/152435940-58a40c5b-3b99-4930-bc8b-c358fa6a9e20.png)
but a semicolon should have been inserted after the identifier `a`, giving
```
i := a;
+b;
```
where the second statement would give us an error in Go because the value is never used, but is syntactically correct, and should get parsed.

Issue #2285, opened 2021-08-16
A possible parsing error related to type conversion of Golang


enochii:
Given the below code(code snippet from [here](https://github.com/json-iterator/go/blob/master/any.go#L260)):

```go
var anyType = reflect2.TypeOfPtr((*Any)(nil)).Elem()
```

if I'm not wrong, `(*Any)(nil)` should be a conversion expression. But it seems it's parsed as an invocation(in which the `(nil)` has been parsed as `ArgumentContext`):

<img src=https://user-images.githubusercontent.com/33364190/129570630-64bbb936-8547-4980-8b2b-4b87a0c389d3.png height=70% width=70%>

kaby76:
How can you distinguish between the two without static semantics because they are syntactically identical? Consider `(*ppfunc)("hello")`.

	package main

	import "fmt"

	func hello(m string){
	       fmt.Println(m)
	}

	func main(){
	       pfunc := hello
	       ppfunc := &pfunc
	       (*ppfunc)("hello")
	}

studentmain:
https://play.golang.org/p/2V_kauSg8CF

Their official parser said type conversion is function call.

kaby76:
But, the grammar--which is scraped from https://golang.org/ref/spec#Conversions--includes a production for Conversion.

	primaryExpr:
		...
		| conversion
		...

	conversion: type_ L_PAREN expression COMMA? R_PAREN;

And, type conversion is parsed as a function call, even through `conversion` is provided.


### Code

	package main
	import "fmt"
	func main(){
		x := bool(true)
		fmt.Println(x)
	}

cat code | trparse | trxgrep ' //expression' | tree:
 
	( expression
	  ( primaryExpr
	    ( primaryExpr
	      ( operand
		( operandName
		  ( IDENTIFIER i=48 txt=bool tt=27 DEFAULT_TOKEN_CHANNEL
	    ) ) ) )
	    ( arguments
	      ( L_PAREN i=52 txt=( tt=28 DEFAULT_TOKEN_CHANNEL
	      )
	      ( expressionList
		( expression
		  ( primaryExpr
		    ( operand
		      ( operandName
			( IDENTIFIER i=53 txt=true tt=27 DEFAULT_TOKEN_CHANNEL
	      ) ) ) ) ) )
	      ( R_PAREN i=57 txt=) tt=29 DEFAULT_TOKEN_CHANNEL
	) ) ) )

[...]

PR #2497, opened 2022-02-03
Move the grammar closer to the actual grammar used  in parser.go

### LINKED ISSUE #2495: Golang: unaryExpr has incorrect precedence
Unary Expressions currently precede over binary expressions:
`-a + b` is currently parsed as:
`(expression (unaryExpr - (expression (expression (primaryExpr (operand (operandName a)))) + (expression (primaryExpr (operand (operandName b)))))))` 
(meaning `-(a + b)`), but it should be parsed as `(-a) + b`.

nicoberling:
This pull request aims to bring the grammar more closely in line with the grammar used for the actual go compiler ([parser.go](https://cs.opensource.google/go/go/+/refs/tags/go1.17.1:src/cmd/compile/internal/syntax/parser.go)) by fixing some issues:
Semicolons are now induced  during lexing, like in [scanner.go](https://cs.opensource.google/go/go/+/refs/tags/go1.17.1:src/cmd/compile/internal/syntax/scanner.go), which fixes expressions spanning multiple lines in cases where they shouldn't , and removes almost all semantic predicates from the grammar, but slows down the lexing stage a little bit.

It also removes ambiguities, and thus increases performance a little bit, by removing rule productions that are in the Go specifications, but not in the actual parser. This mostly changes where `IDENTIFIER DOT IDENTIFIER`-elements are matched. 

The changes are summarized here:

- Inline unary expressions into primary expressions to fix #2495.
- No longer parse named types under the `type_` rule in contexts where this leads to ambiguities with non-type identifiers (see also #2285 ). This means conversions to named types are now invocations, and method expressions with named types are selections, just like in [parser.go](https://cs.opensource.google/go/go/+/refs/tags/go1.17.1:src/cmd/compile/internal/syntax/parser.go).
- Parse qualified identifier-operands as selection primary expressions instead
- Parse named keys as named operands instead of directly as identifiers.

Because these changes are quite large, I understand if you also want to look at other ways to fix this. I opened issues for the problems I found, in case someone else find a better way (#2495, #2496).


teverett:
@nicoberling  thanks

kaby76:
The [Antlr4cs base class](https://github.com/antlr/grammars-v4/tree/master/golang/Antlr4cs) is now out of synch with the current code. The base class is almost identical to the CSharp target.

Should Antlr4cs be added to CI? Are these ports to be maintained?

studentmain:
> Should Antlr4cs be added to CI? Are these ports to be maintained?

And antlr4ts and antlr4rust.

kaby76:
> And antlr4ts and antlr4rust.

My opinion is no. These ports use an older version of the Antlr tool jar and aren't being validated against the latest tool. Or, are they? Plus, I don't know if the target-specific codegen classes can be packaged in another jar and added to the classpath so as to use the codegen for the target with the current Antlr4 tool jar.

KvanTTT:
I also don't think they are necessary. I suggest integrating antlr4rust into official ANTLR but later.

studentmain:
I think no either. We already have "antlr4cs" in home, that's the main reason to not include it. About antlr4ts and antlr4rust and maybe other third party target, they are not official yet, thus may not always have same feature set and stability, we are here to ensure grammar is correct, not the target (We've found some bug with targets since we start running test over them, but it's because targets are not well tested at the beginning, it should be done when coding them). But if (for example) antlr4rust got enough popularity and quality, and still not supported by official impl, then we'll have practical reason to support it.

New tester need to support custom target in order to use grammars-v4 as e2e test for antlr4. So adding these shouldn't cause additional complexity for it.

kaby76:
OK, I will make a PR to remove the Antlr4cs directories here.

I would prefer if the targets could be moved to there own repos and loosely bound to the tool. Antlr4ts is there with just a few changes, and you could then just call the tool with a -classpath with the target you're interested in:

    java -cp "c:/msys64/home/Kenne/antlr4ts/tool/target/antlr4-typescript-4.9.3-SNAPSHOT.jar;c:/users/kenne/downloads/antlr-4.9.3-complete.jar" org.antlr.v4.Tool -Dlanguage=TypeScript Arithmetic.g4

I didn't realize that it's pretty easy to customize the codegen templates via classpath adjustments. The main problem is that the tool codegen model keeps changing. It was a bit of work updating the antlr4ts target for 4.9.3.

PR #2500, opened 2022-02-07
2 small fixes

bkiers:
2 small fixes:

- `\x` (without hex digits) is valid per PCRE, and will match the literal `x`
- control char can be followed by an optional ascii char


KvanTTT:
Any chance to add an example that covers suggested changes?

bkiers:
Sure @KvanTTT

PR #2501, opened 2022-02-08
Fix xpath1 and/or/div/mul names

rovarga:
https://www.w3.org/TR/REC-xml-names/#NT-NCName allows for NCName to
contain plain 'and'/'or'/'div'/'mul', but our implicit lexer definitions
prevent them from being caught by NCName token. Define them explicitly
and include them as alternatives to NCName token.


rovarga:
This is a port of a downstream fix here: https://github.com/opendaylight/yangtools/commit/1a7e93a50d22944771fe2ea4fc18c24fcf3aae11

PR #2504, opened 2022-02-09
Fix typo in SQLite parser that caused an unexpected recursive rule

renatahodovan:


KvanTTT:
Could you please add a sample that was caused by an unexpected recursive rule?

KvanTTT:
Ok then.

PR #2505, opened 2022-02-09
Add the missing ADMIN and OPTIONAL tokens to MySqlLexer

renatahodovan:


KvanTTT:
Could they be id? Also, please add a test example that covers the changes.

KvanTTT:
There are still erorrs on CI: https://github.com/antlr/grammars-v4/runs/5129063471?check_suite_focus=true#step:4:12382

KvanTTT:
Not sure it's the best solution for resolving `ADMIN` token, but ok, will fix it later if necessary.

Issue #2508, opened 2022-02-12
vb6 grammer MINUS issue


adrishdas:
a=a -1 (no space before 1) throws an error whereas a=a - 1 (a space before 1) is handled correctly

kaby76:
https://github.com/antlr/grammars-v4/blob/7bfee06e5bf167a92340d3e7f2228c63fa805803/vb6/VisualBasic6Lexer.g4#L431

You're not supposed to implement unary operators in the lexer grammar. It was in there since day one. https://github.com/antlr/grammars-v4/blob/4ecb19054ce6cfc07efb9050be5976aac7f6828f/vb6/VisualBasic6.g4#L868

kaby76:
@adrishdas Please check out the updates and see if they solve your problem. --ken

adrishdas:
The solution worked. Sorry, am quite late catch-up

PR #2519, opened 2022-02-16
Issue 2508

kaby76:
Fix for #2508, parsing expressions with and without spaces, e.g., "a -1" vs "a- 1" vs "a - 1", etc. Unary operators belong in the parser, not the lexer.


teverett:
@kaby76  thanks

PR #2521, opened 2022-02-18
Rewritten Verilog and SystemVerilog Grammars

msagca:
- Converted `Formal Syntax` sections in standard documents to ANTLR4 with the help of [a tool](https://github.com/msagca/syntax_scraper) I've written.
- Fixed several parser rules in the process and removed every addition I made to the original grammars.
- Re-enabled all rules except one rule in SystemVerilog grammar that caused too many mutual left recursion errors.
- Fixed all the rules that caused a warning or an error.
- Fixed major issues in Verilog preprocessor grammar and added a preprocessor grammar for SystemVerilog.


msagca:
Can someone tell me what went wrong with the `Go` builds? I don't see it.

kaby76:
> Can someone tell me what went wrong with the `Go` builds? I don't see it.

The Go target code that is generated contains global variables such as parserATN. Since you have two parsers within the same directory, these get generated to the same parser/ directory with the same package name. This is why there is a build error. All the other target languages fix this with classes and namespace, which allow more than one package and/or class to exist in a directory. The solution is to add this grammar to the [skip list for Go](https://github.com/antlr/grammars-v4/blob/master/_scripts/skip-go.txt). I have code to fix this for Antlr, https://github.com/antlr/antlr4/pull/3486, but it is low on my list of things to do. What PRs I have done in the past for Antlr were integrated a year after submission, long after I forgot every detail of the PR. And then, was required to add unit tests. As I like to repeat, the development process for Antlr is broken (although slightly less so with separate branches "dev"/"master" to distinguish between ongoing development vs. released versions).

kaby76:
Looks better. Could you add a few more tests for the examples/ directory (both verilog and systemverilog)? I'm looking into [Grammarinator](https://github.com/renatahodovan/grammarinator) and other implementations to test grammars, but there should be some hand written code too.

msagca:
I've discovered some issues while parsing the test files on [this repo](https://github.com/chipsalliance/sv-tests). I'll fix them and copy those files to `examples` folder.

kaby76:
Grammarinator generates some tests files that look pretty vicious! So, this might yield something. But, half the generated files are completely empty files.

kaby76:
Although it has been like pulling teeth, I have Grammarinator outputting trees and test cases, so I can compare when there is a problem in the parse.

There is a problem in that some tests that are not parsed cleanly. I'm trying to figure out why. But, I have noticed a problem with `constant_primary`.

In `constant_primary`, the rule can derive a sentential form that contains either `parameter_identifier` or `specparam_identifier`. Sentential forms derived from either result in `identifier`, so it is ambiguous. Presumably the two different derivations are disambiguated by semantics. Unfortunately, I don't have the spec.

msagca:
> Although it has been like pulling teeth, I have Grammarinator outputting trees and test cases, so I can compare when there is a problem in the parse.
> 
> There is a problem in that some tests that are not parsed cleanly. I'm trying to figure out why. But, I have noticed a problem with `constant_primary`.
> 
> In `constant_primary`, the rule can derive a sentential form that contains either `parameter_identifier` or `specparam_identifier` (second and third alternative of the rule). Sentential forms derived from either result in `identifier`, so it is ambiguous. Presumably the two different derivations are disambiguated by semantics. Unfortunately, I don't have the spec.

The spec can be downloaded from [here](https://ieeexplore.ieee.org/document/8299595) without a subscription (you still need to login).

kaby76:
Thanks for the like to the spec.

Actually, I'm looking at the spec and verilog/verilog/VerilogParser.g4, for `constant_primary`, 2nd alt:

Spec:

    constant_primary ::= ... | ps_parameter_identifier constant_select | ...

.g4

    constant_primary : ... | parameter_identifier ( '[' constant_range_expression ']' )? | ...

If one unfolds `constant_select` from the spec `constant_select ::= [ { '.' member_identifier constant_bit_select } '.' member_identifier ] constant_bit_select [ '[' constant_part_select_range ']' ]`) into the 2nd alt for `constant_primary` in the spec, that should be:

    constant_primary ::= ... | ps_parameter_identifier  [ { '.' member_identifier constant_bit_select } '.' member_identifier ] constant_bit_select [ '[' constant_part_select_range ']' ]`

(Adding in quotes for literals.)

Instead we see the 2nd alt of `constant_primary` in VerilogParser.g4 `parameter_identifier ( '[' constant_range_expression ']' )?`, which doesn't look anything like a proper unfold refactoring.

msagca:
The standard document I extracted the formal syntax from is [this copy](http://staff.ustc.edu.cn/~songch/download/IEEE.1364-2005.pdf#page=535). I don't have access to https://ieeexplore.ieee.org/document/1620780. I don't know if these documents are different.

kaby76:
I see.

* verilog:
    * IEEE 1364-2005
    * https://ieeexplore.ieee.org/document/1620780
* system verilog:
    * IEEE 1800-2017
    * https://ieeexplore.ieee.org/document/8299595

Sorry, I got the specs mixed up. I how have both specs in front of me. constant_primary is in both grammars, but completely different.

I was looking into why the 2nd and 3rd alts in verilog/verilog/VerilogParser.g4 for constant_primary was always producing the 2nd alt when Grammarinator produced a tree/test case via the 3rd alt.

All ok, but a little confused for the two alts. I need to read the spec (the correct one at that).

msagca:
I think both grammars are ready to be merged.  
There are some issues with the `SystemVerilog` grammar but there seem to be some missing pieces in the spec.  
- The following rule is not referenced anywhere in the grammar.
```antlr
interface_class_declaration
	: 'interface' 'class' class_identifier parameter_port_list? ( 'extends' interface_class_type ( ',' interface_class_type )* )? ';' interface_class_item* 'endclass' ( ':' class_identifier )?
	;
```
- The parser fails to recognize the following structure.
```antlr
packet = packet_out::type_id::create("packet", this);
```
It's legal but the spec does not allow multiple scope resolution operators (`::`) before a function call (`tf_call`), and I haven't modified anything in the grammar other than disabling `cast` and `constant_cast` rules (to prevent mutual left recursion) and unfolding some alternatives (for the same reason).

teverett:
@msagca  thanks!

PR #2523, opened 2022-02-21
first commit

teverett:


KvanTTT:
It should be probably placed into README. At least ref to the doc from README.

PR #2534, opened 2022-03-03
Added Grammar for KLEE KQuery Language.

lahiri-phdworks:
**ANTLRv4 Grammar** for **KLEE KQuery** will save the time and lives of those working with **KLEE** for their own **symbolic execution** projects.


teverett:
Could you please add this to the root pom.xml?   That will enable the tests to run on this PR.

lahiri-phdworks:
Added `pom.xml` file.

teverett:
hey @lahiri-phdworks  thanks for this addition.  You'll need to add it to the root pom.xm for the tests to run. 

https://github.com/antlr/grammars-v4/blob/master/pom.xml

lahiri-phdworks:
@teverett I apologize, completely forget at the last minute. I have added the entry now.

teverett:
@lahiri-phdworks  oh no big deal and thanks for the new grammar

teverett:
@lahiri-phdworks  in your pom.xml consider replacing this

<entryPoint>file_</entryPoint>

with the top level rule in your grammar.  perhaps:

<entryPoint>kqueryExpression</entryPoint>

lahiri-phdworks:
Did that. Thanks.

kquery/KQuery.g4
```diff
@@ -0,0 +1,357 @@
...
+OctConstant 
+    : OctId (OCTAL_DIGIT)+  
+    ;
+    
+HexConstant 
+    : HexId (HEX_DIGIT)+
+    ;
+
+FloatingPointType 
+    : FP DIGIT ((.).*?)?
```
kaby76, line 240: This rule is wrong. I looked at the original Spec and it says `floating-point-type = "fp[0-9]+([.].*)?"` 

`(.)` is not the same as `[.]`--which is exactly what you should be using for your rule.

Further, you have `fp` as a token, but the definition "shields" `Identifier`. Can `fp` be an identifier? If so, you need to define a rule `identifier: Identifier | fp | .......`. The same thing may apply to other rules, e.g., "INT: 'i';".

kquery/KQuery.g4
```diff
@@ -0,0 +1,357 @@
...
+    | SREM
+    ;
+    
+domain : widthOrSizeExpr ;
+rangeLimit : widthOrSizeExpr ;
+arrName : Identifier ;
+
+numberList
+    : number
+    | number numberList
```
kaby76, line 198: You really should write this rule in EBNF. `numberList: number+;`. This isn't just cosmetic, it's essential for a fast implementation--which you don't have. It's really, really slow.

kquery/KQuery.g4
```diff
@@ -0,0 +1,357 @@
...
+    
+arrayInitializer
+    : Symbolic 
+    | LeftBracket numberList RightBracket
+    ;
+    
+expression
+    : varName                                                                               #VariableName
+    | namedConstant Colon expression                                                        #NamedAbbreviation
+    | LeftParen widthOrSizeExpr number RightParen                                           #SizeQuery
```
kaby76, line 82: This needs to be refactored, e.g., `expression: varName | ... | LeftParen (withOrSizeExpr number | arithmeticExpr widthOrSizeExpr leftExpr rightExpr  | .... RightParen | ....;` Your grammar is quite inefficient.

kquery/KQuery.g4
```diff
@@ -0,0 +1,357 @@
...
+    : arrayDeclaration
+    | queryCommand
+    ;
+
+queryCommand 
+    : LeftParen Query evalExprList queryExpr RightParen 
+    ;
+        
+queryExpr 
+    : expression                                #SingletonQueryExpr
```
kaby76, line 49: `queryExpr : expression ( evalExprList evalArrayList? )?;`

kquery/KQuery.g4
```diff
@@ -0,0 +1,357 @@
...
+    : expression
+    ;
+    
+namedConstant
+    : Identifier
+    ;
+    
+updateList 
+    : expression Equal expression COMMA updateList
+    | expression Equal expression
```
kaby76, line 157: `updateList : expression Equal expression ( COMMA expression Equal expression )* ;`

kaby76: There's likely other problems. The issue is that this grammar is awfully slow. That is why it fails for the PHP target--which is terribly slow to begin with. Antlr really is much faster if you use EBNF.

kaby76: BTW, the alternative is to add this to the [skip list for PHP](https://github.com/lahiri-phdworks/grammars-v4/blob/master/_scripts/skip-php.txt) and just say in the readme.md that it's "unoptimized". People will know what you mean.

kquery/KQuery.g4
lahiri-phdworks: Yeah I had this rule `numberList: (number)+` earlier. I shall put that back. The recursive structure does make it slow for large list of numbers.

kquery/KQuery.g4
lahiri-phdworks: I will make this change. Thanks

lahiri-phdworks:
Made the proposed chages. Thanks for the comments.

teverett:
@lahiri-phdworks  thanks

Issue #2539, opened 2022-03-08
grammars-v4/python/python is not being tested


kaby76:
[The grammar](https://github.com/antlr/grammars-v4/tree/f7f1eb1b612c08a942d178ca2224a830982264ed/python/python) is not being tested; it should be.

This grammar was added with a flurry of changes that started with [this commit](https://github.com/antlr/grammars-v4/commit/0c630e36f989c023cb330496fee028dc0c5349ab). Afterwards, the root pom.xml file was never updated, so it was never tested in the CI builds.

In addition, the "CSharp/" directory isn't really Antlr4's "CSharp" target code--it's for the Tunnelvision forked Antlr4 C# target. If we do keep the code that is currently in CSharp/, the files should be moved to an "Antlr4cs/" directory so that there is zero confusion as to what this code is for. Note, we don't even test the Antlr4cs "target" in CI builds.

PR #2540, opened 2022-03-08
Fix for #2539

kaby76:
* Add python/python to testing, ergo the reason for all the pom.xml and skip lists changed.
* Rewrite CSharp files to use official Antlr4 code in Antlr4.Runtime.Standard. As per [this conversation](https://github.com/antlr/grammars-v4/pull/2497#issuecomment-1036224128), the consensus is that Antlr4cs should no longer to be supported--the Tunnelvision fork from github.com/antlr/antlr4 is not maintained.
* Remove nasty package declarations from code. __It's not portable across targets!__ If you need them, add them into your private copy.

### DIFF
--- a/_scripts/skip-cpp.txt
+++ b/_scripts/skip-cpp.txt
@@ -159,6 +159,7 @@ propcalc
 properties
 protobuf3
 prov-n
+python/python
 python/python2
 python/python2-js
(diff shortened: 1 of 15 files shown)


teverett:
@kaby76 thanks

PR #2556, opened 2022-03-23
Ported the JavaScript version of ECMAScript.g4 to TypeScript

montyz:


teverett:
Thanks @montyz .   @KvanTTT , @kaby76 all good to pull this PR?

KvanTTT:
I have no objection but I suggest improving JavaScript.g4 since ECMAScript.g4 is outdated.

kaby76:
I think this is fine.

We've tried pretty hard to move the grammars to a "target agnostic" syntax for the grammars, but this is a grammar that was never updated. Instead, target specific code is place in the grammar using `@members` methods/functions that are emitted in the generated parser itself.

But, in reality, Antlr4 doesn't really have a "target agnostic" solution in which we have one common grammar that wraps all the target specific code in a function call, which is then placed in a base class (or functions) for each target, outside the grammar. It's really "pseudo-target agnostic" because it doesn't work for all targets.

I've tried to explain this in [Antlr4 Issue 3508](https://github.com/antlr/antlr4/issues/3508). In my opinion, this is a fundamental, high-priority problem with Antlr that needs to be fixed.

Serendipitously, I stubbled on a grammar that had a "$parser.foobar()" referenced in an action, and found that Antlr already converts "$parser." into the target specific syntax for all but Cpp and PHP! There is currently no equivalent for a "$lexer" reference, so that would also be needed. However, I would prefer to see references to the parser or lexer as just "$this", dispensing with the distinction. I sometimes find I need to move rules around between lexer and parser.

PR #2557, opened 2022-03-24
Added base classes to parse Python code with `python/python` grammar in Python

rindPHI:
I added the classes `PythonLexerBase` and `PythonParserBase`, by which the `python/python` grammar can be used to parse Python from within Python. Since the grammar is not target agnostic, some actions / semantic predicates have to be slightly changed when these base classes are used; e.g., `{CheckVersion(3)}` gets `{self._check_version(3)}`.

### DIFF
--- a/python/python/Python/PythonLexerBase.py
+++ b/python/python/Python/PythonLexerBase.py
@@ -1,0 +1,193 @@
+import sys
+from enum import Enum
+from pkgutil import get_data
+from typing import List, Optional, Dict, TextIO
+
+from antlr4 import Lexer, Token, InputStream, Parser, TokenStream
+from antlr4.Token import CommonToken
+
+# To use the PythonLexer in Python,
+#
+# 1. Add the code
+#    ```antlr
+#    @header {
+#      from .PythonLexerBase import PythonLexerBase
+#    }
+#    ```
+#    to PythonLexer.g4, directly below `lexer grammar PythonLexer;`.
+# 2. Make sure that the generated lexer class, the file `PythonLexer.tokens`, and this class are in the same directory.
+class PythonLexerBase(Lexer):
+    tab_size = 4
+
+    def __init__(self, input_stream: InputStream, output: TextIO = sys.stdout):
+        super().__init__(input_stream, output)
+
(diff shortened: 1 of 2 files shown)


KvanTTT:
Thanks for the contribution and welcome to grammars repository!

kaby76:
Thanks for porting the grammar to Python3, but this code doesn't work. It seems there are some missing files, and the initialization does not work.

Note, this code requires Python 3.10.x; it fails to parse "|" on line 186 of PythonLexerBase.py. We really should have minimum requirements listed out for a grammar, maybe the readme file. After update Python, I now get to the real problems.

It fails in:
```
  File "C:\msys64\home\Kenne\g-current\grammars-v4\python\python\Generated\PythonLexerBase.py", line 33, in __init__
    tokens_file: str = get_data(__package__, 'PythonLexer.tokens').decode('utf-8').strip()
AttributeError: 'NoneType' object has no attribute 'decode'
```
I don't know why it failed on `get_data()`, but I believe the code isn't in a package. (The concept does not exist across all targets, so there is no package declaration.)

Working past the failing __init__ code that tries to read the tokens file:

        # tokens_file: str = get_data(__package__,
        # 'PythonLexer.tokens').decode('utf-8').strip()
        text_file = open('PythonLexer.tokens', 'r')
        tokens_file: str = text_file.read().strip()

the program then fails on HandleNewLine:
```
  File "C:\msys64\home\Kenne\g-current\grammars-v4\python\python\Generated\PythonLexer.py", line 636, in NEWLINE_action
    HandleNewLine();
NameError: name 'HandleNewLine' is not defined
```

The function does exist, but the name is actually `_handle_new_line(self) -> None`.

This code is incomplete. You obviously forked the grammar files, but you did not check them in, within the Python/ directory.

Note, when you place a copy of the grammar files in the Python/ directory (which should be named Python3 because this is not Python2 code!), the files are overlayed on top of all the other files.

Could you please add the forked .g4 files into Python/?

rindPHI:
Oh, thank you very much @kaby76! I did not find the time to fix the issue myself so far. Great that you have done it.

PR #2559, opened 2022-03-25
support for new listagg syntax

### LINKED ISSUE #2558: PL/SQL multiple LISTAGG problem
When parsing a query with multiple `LISTAGG` functions after another, the parsing process breaks at the second `LISTAGG` function call.

For example, when parsing the below query

```sql
select name
     , listagg(UNIQUE ename, ',' ON OVERFLOW TRUNCATE) within group (order by hiredate) over (partition by deptno) as employees
     , listagg(UNIQUE edepartment, ',' ON OVERFLOW TRUNCATE) within group (order by hiredate) over (partition by deptno) as edepartments
from emp;
```

the parsing of query fails after the second `LISTAGG`.

It also looks like the syntax of `LISTAGG` is changed with newer versions of Oracle (Ref: [PL/SQL Language Reference](https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/LISTAGG.html#GUID-B6E50D8E-F467-425B-9436-F7F8BF38D466)).

rifatx:
As per documentation found at [PL/SQL Reference](https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/LISTAGG.html#GUID-B6E50D8E-F467-425B-9436-F7F8BF38D466) by Oracle, `LISTAGG` analytic function now supports a new syntax.

This PR provides support for this new syntax and resolves #2558.

### DIFF
--- a/sql/plsql/PlSqlParser.g4
+++ b/sql/plsql/PlSqlParser.g4
@@ -3918,6 +3918,10 @@ numeric_function
    | GREATEST '(' expressions ')'
    ;
 
+listagg_overflow_clause
+    : ON OVERFLOW (ERROR | TRUNCATE) (CHAR_STRING)? ((WITH | WITHOUT) COUNT)?
+    ;
+
 other_function
     : over_clause_keyword function_argument_analytic over_clause?
     | /*TODO stantard_function_enabling_using*/ regular_id function_argument_modeling using_clause?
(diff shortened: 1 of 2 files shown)


sql/plsql/PlSqlParser.g4
```diff
@@ -3918,6 +3918,10 @@ numeric_function
    | GREATEST '(' expressions ')'
    ;
 
+listagg_overflow_clause
+    : ON OVERFLOW (ERROR | TRUNCATE) (CHAR_STRING)? ((WITH | WITHOUT) COUNT)?
```
KvanTTT, line 3922: Excess parentheses:

```suggestion
    : ON OVERFLOW (ERROR | TRUNCATE) CHAR_STRING? ((WITH | WITHOUT) COUNT)?
```

PR #2573, opened 2022-04-13
Update to Antlr 4.10

kaby76:
* Fixing the most egregious problems for "mvn test". There are warnings still.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -58,7 +58,7 @@ jobs:
       with:
         python-version: '3.8'
     - name: Install Nodejs
-      uses: actions/setup-node@v3.1.0
+      uses: actions/setup-node@v3.1.1
       with:
         node-version: '16.13.0'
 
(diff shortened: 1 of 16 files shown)


kaby76:
* mvn tests are looking better.
* Need to add -fae option to mvn so we get the whole picture of all grammars that may fail.
* Getting "out of disk space" for Antlr4BuildTasks. Need to add code to place JRE and Antlr4 tool jar in a common place across different Antlr4BuildTasks invocations.
* Need to add explicit .NET version install. We need version 6.0.x. [Version 5.x is no longer supported at the end of next month](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core#lifecycle).
* Need to fix [_scripts/templates/Cpp version](https://github.com/antlr/grammars-v4/blob/06a69f69d2fef19295c8c48dbfc209622d1bc76d/_scripts/templates/Cpp/CMakeLists.txt#L9)--that's the wrong runtime.

PR #2574, opened 2022-04-13
Partial upgrade to Antlr 4.10 -- Maven tests

kaby76:
I'm upgrading the build to Antlr 4.10 for the standard Maven/Java tests. The other targets require a little more work, so they will be done separately.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,7 @@
 
 # Ignore C# regression temporaries.
 **/Generated/
+
+# Ignore Emacs backup files.
+*~
+\#*\#
--- a/antlr/antlr3/ANTLRv3Lexer.g4
+++ b/antlr/antlr3/ANTLRv3Lexer.g4
@@ -111,7 +111,7 @@ DOUBLE_QUOTE_STRING_LITERAL
 // This seems to be available in Antlr3.
 
 DOUBLE_ANGLE_STRING_LITERAL
-   : '<<' ()*? '>>'
+   : '<<' .*? '>>'
    ;
 
 fragment ESC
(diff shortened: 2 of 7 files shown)


kaby76:
[get-antlr.ps1](https://github.com/antlr/grammars-v4/blob/06a69f69d2fef19295c8c48dbfc209622d1bc76d/_scripts/get-antlr.ps1#L52) reads the xml file [here](https://repo1.maven.org/maven2/org/antlr/antlr4/maven-metadata.xml) to determine the latest version to test, which means this goes "live" the moment when the Antlr tool is released. This differs from the [pom.xml](https://github.com/antlr/grammars-v4/blob/06a69f69d2fef19295c8c48dbfc209622d1bc76d/pom.xml#L13) specification of the version to test, where the grammars aren't tested on the latest until things are ready.

The main problem with the "live" release is that the templates that trgen uses are for an old version. There is no "latest" concept for the runtimes--some runtimes are extracted through git, others through a package library that [has not been updated](https://github.com/antlr/antlr-php-runtime/issues/26). This is why the [test.ps1](https://github.com/antlr/grammars-v4/blob/master/_scripts/test.ps1) tests cannot run.

KvanTTT:
Maybe it makes sense to merge upgrading to 4.10 for the main tool and plugin at the beginning? I mean merging this MR as is.

kaby76:
> Maybe it makes sense to merge upgrading to 4.10 for the main tool and plugin at the beginning? I mean merging this [PR] as is.

Sure, the Maven builds are working (with warnings), so at least you can then modify the grammars to stop using the crazy case changing hacks we currently use and address the warnings that you added to the tool. I'll continue working on the other builds (test.ps1 and regtest.sh).

PR #2575, opened 2022-04-13
Partial upgrade to Antlr 4.10 --part 2

kaby76:
**Udated 16 Apr '22** This PR sets the build to Antlr 4.10.
* It uses the latest of Antlr4BuildTasks and trgen.
* It has a few improvements in outputting information useful for debugging builds.
* There are a number of version strings that have to be modified for a new release. These should be combined and centralized, but this PR does not make those changes. That will be done in another PR.


kaby76:
OK, that's looking better. But, it's dying on JavaScript on `const foobar = require("fu");` when it should be `import foobar from "fu";`. It's a mixture of both. How the hell did this every work? Changing to the `import` form, the parsers work fine.

kaby76:
I've had some luck in updating to 4.10 for all targets except Cpp and Dart. I'm testing the changes so far.

kaby76:
Almost there.
* For whatever reason, Powershell interprets the command `$antlrPath = _scripts/get-antlr.ps1 4.10` as the parameter in get-antlr.ps1 is `4.1`. This is of course bogus. Adding double quotes around the argument is required, e.g., `"4.10"`.
* Up to now, I've really, really been avoiding running the script locally--because I did not understand how to call the thing, what Powershell to use, etc. I now have it installed on an Ubuntu VM so I am now testing it locally. Already I see a number of issues.
* One of those issues is that `_scripts/get-antlr.ps1` leaves the Antlr tool jar right there in the root directory. Since this is a git directory, the .ignore has to be set to ignore these .jar's.
* There is little output of the status for `_scripts/test.ps1`. Critically, we need to know how `trgen` is called, how the antlr tool is called, how `javac` is called. Otherwise, we don't know why builds fail.

kaby76:
* CSharp -- requires some mods to Antlr4BuildTasks and trgen to use shared java/Antlr tool jar on path or specified. This is taking a bit of time to get it right.
* Go -- not yet picking up right runtime.
* Java -- "server error"? Working on the other targets first.
* Cpp -- not sure why, but I absolutely don't like, parallel cmakes going off at the same time. I just want the runtime to be built once for all different grammars.

KvanTTT:
`System.IO.IOException: There is not enough space on the disk.` at least for C#. Maybe it's related to other targets as well.

kaby76:
> `System.IO.IOException: There is not enough space on the disk.` at least for C#. Maybe it's related to other targets as well.

Yeah, for the C# target, I'm not sure what I was smoking. Lot's to fix in Antlr4BuildTasks--which should go into github.com/antlr at some point when the code is nice, tight, and tidy.

* The .dll contains embedded resources JRE.zip and antlr4-4.10-complete.jar. To boot, it's already in the .nupkg, when *also* gets unpacked into ~/.nuget/packages/Antlr4BuildTasks/xxxx/build/...
* But, better yet, if it doesn't find the jar or java, it tries to do a download--per grammar into the .../obj/Debug/net6.0/.... area. No sharing.
* The JRE.zip gets unpacked into .../obj/Debug/net6.0/..., so there's another copy.
* Even though Antlr4BuildTasks is referenced with PrivateAssets="all", MS says that really don't apply except when "published". Thanks MS for not explaining. So, it really doesn't help doing that.

I'm fixing all this BS now, and adding some code to look for Java on PATH.

The Cpp target was picking up not 4.10, but 4.1! I thought I took care of this but need to look at it carefully.

kaby76:
* CSharp target correct, now building fine.
* Cpp -- `-- Found ANTLR: /tmp/antlr4-4.10-complete.jar (found version "4.1")`. CMake seems to be outputing this. I don't know why it's `found version "4.1"`, but that seems suspicious. For older builds, it's `-- Found ANTLR: /tmp/antlr-4.9.3-complete.jar (found version "4.9.3") `, which isn't truncated. [Added lines to CMakeLists.txt](https://github.com/antlr/antlr4/issues/3418#issuecomment-1100684217), which corrects the build.

kaby76:
* All targets -- I'm not moving forward to 4.10.1 which was released a day or two ago. This is because not all target runtimes are being released at the same time, not currently there, and I'm not going to beg again for them to be released. The release process is broken. No 4.10.1 runtimes https://pub.dev/packages/antlr4 https://packagist.org/packages/antlr/antlr4-php-runtime

KvanTTT:
Is it possible to use ANTLR 4.10.1 tool and ANTLR 4.10 runtimes? There is almost no difference between 4.10 and 4.10.1.

kaby76:
Maybe. Some of the runtimes check for an exact match with the tool. It
would require some changes in  github/workflows/main.yml to make each
target have separate tool/runtime version specs. I was going to fix this in
"part 3" changes because as I noted earlier version strings are all over
the place. But first want this to work with 4.10, so you can update the
grammar case folding changes, and others can make PRs without a bad build.

kaby76:
* Go -- wrong version of Antlr runtime in a script.

This should be the last to get the build working again. But, spreading version strings all over is a mess. That to do in "part 3".

kaby76:
The build is finally working. Ready to go.

KvanTTT:
What merging strategy should I choose: rebase or squash?

kaby76:
I think squash-- the details are not that important.

KvanTTT:
Thanks for the great work!

kaby76:
Thanks for committing this. Will now work on what we talked about (merging version strings, but allowing separate versions for runtime).

kaby76:
* It's not clear why JavaScript/Ubuntu had a problem. Something to watch out for.
```
2022-04-17T11:04:42.6125102Z Copying source file from /home/runner/work/grammars-v4/grammars-v4/asm/masm/MASM.g4 to Generated/MASM.g4
2022-04-17T17:02:12.3263700Z ##[error]The operation was canceled.
```

PR #2576, opened 2022-04-13
Use caseInsensitive option for case insensitive grammars (ANTLR >= 4.10)

KvanTTT:
@parrt @mike-lischke FYI here I've simplified grammars using `caseInsensitive` option :)


mike-lischke:
I need to change my MySQL grammar yet, but switching to 4.10 will take a bit. Lots of other work and I also need legal approval ¯\_(ツ)_/¯.

Additionally, antlr4ts would have to be updated for the changed ATN serialization. Not sure if Burt or Sam will do that in foreseeable time (and my Java converter tool isn't ready yet).

But thanks for the reminder @KvanTTT.

KvanTTT:
> Additionally, antlr4ts would have to be updated for the changed ATN serialization. Not sure if Burt or Sam will do that in foreseeable time (and my Java converter tool isn't ready yet).

Ideally they also should be merged into official ANTLR repository to simplify synchonization.

> But thanks for the reminder @KvanTTT.

I'd like just to demonstrate how `caseInsensitive` option works on practice :)

parrt:
@KvanTTT wow! Great job!!!!

KvanTTT:
@kaby76 I don't competely understand why all tests are failing. It looks like they use the previous version of ANTLR that doesn't support `caseInsensitive` option.

kaby76:
@KvanTTT I'll take a look.

kaby76:
1st point of fail is grammars-v4/php -- on Java!? Really?
```
../examples/identifiers.php
line 4:3 no viable alternative at input '$? ='
```
In fact, it fails for what's already in github.com/antlr/grammars-v4. So, the build with 4.10 isn't working because it should have failed long before.

Not sure what's the diff between 4.10 and 4.9.3 here.

KvanTTT:
> 1st point of fail is grammars-v4/php -- on Java!? Really?

It also fails for me but I thought it's related to machine-specific issues with encoding. It fails on:

```php
$𠮷 = "𠮷";
```

> So, the build with 4.10 isn't working because it should have failed long before.

In this case we should have encountered the problem before but tests are green for the current master.

kaby76:
Yes, this is puzzling because it fails for the "mvn clean test" run, which has nothing to do with trgen, and all that stuff. At a point a few months ago where it's working. Stepping forward to see when it should have been failing.

kaby76:
* First, I ran this script to compare what's in the main repo vs your repo.
```
git clone https://github.com/KvanTTT/grammars-v4.git g-mod
cd g-mod
git checkout use-case-insensitive-option
cd php
mvn clean test
cd ..
cd ..

git clone https://github.com/antlr/grammars-v4.git g-ori
cd g-ori
cd php
mvn clean test
cd ..
cd ..
```

* The php/ grammar works for main repo, but doesn't for your repo.
* However, running `trgen -t Java; cd Generated; make; make test` fails with the same reason in examples/identifiers.php in the main repo, too.
* FYI, examples/identifiers.php is a utf-8/no bom file.
* Tokens for the trgen parser shows the problem:
```
java -classpath C:/Users/Kenne/Downloads/antlr-4.9.3-complete.jar\;. Program -file ../examples/identifiers.php -tokens
[@0,0:4='<?php',<4>,channel=4,1:0]
[@1,5:6='\n\n',<39>,channel=4,1:5]
[@2,7:17='$??????_???',<219>,3:0]
[@3,18:18=' ',<39>,channel=4,3:11]
[@4,19:19='=',<216>,3:12]
[@5,20:20=' ',<39>,channel=4,3:13]
[@6,21:22='42',<222>,3:14]
[@7,23:23=';',<215>,3:16]
[@8,24:24='\n',<39>,channel=4,3:17]
[@9,25:25='$',<204>,4:0]
[@10,26:26='?',<231>,channel=3,4:1]
[@11,27:27=' ',<39>,channel=4,4:2]
[@12,28:28='=',<216>,4:3]
[@13,29:29=' ',<39>,channel=4,4:4]
[@14,30:30='"',<228>,4:5]
[@15,31:31='?',<234>,4:6]
[@16,32:32='"',<228>,4:7]
[@17,33:33=';',<215>,4:8]
[@18,34:34='\n',<39>,channel=4,4:9]
[@19,35:34='<EOF>',<-1>,5:0]
```
The id at line 4 is an error token. I also see that the string literal isn't being lexed right.
* ***There's no "show tokens" in the [antlr4test-maven-plugin](https://github.com/antlr/antlr4test-maven-plugin), so I can't exactly tell what it's doing right.***

* But, on a hunch, I tried it with an explicit utf-16 arg, and it "works"!
```
java -classpath C:/Users/Kenne/Downloads/antlr-4.9.3-complete.jar\;. Program -file ../examples/identifiers.php -encoding utf-16
Parse succeeded.
```

But, this is why it "works". Geeeeeesuz!
```
java -classpath C:/Users/Kenne/Downloads/antlr-4.9.3-complete.jar\;. Program -file ../examples/identifiers.php -encoding utf-16 -tokens
[@0,0:24='?????????????????????????',<2>,1:0]
[@1,25:24='<EOF>',<-1>,1:25]

Parse succeeded.
```
@teverett Is there a flag in the antlr4test-maven-plugin that I can add to the pom.xml to see the tokens recognized?

kaby76:
* The Antlr4 Maven test plugin works fine. The tree it outputs for php/examples/identifiers.php is `(htmlDocument (phpBlock (topStatement (statement (expressionStatement (expression (assignable (chain (chainOrigin (chainBase (keyedVariable $привет_мир))))) (assignmentOperator =) (expression (constant (literalConstant (numericConstant 42))))) ;))) (topStatement (statement (expressionStatement (expression (assignable (chain (chainOrigin (chainBase (keyedVariable $𠮷))))) (assignmentOperator =) (expression (string " (interpolatedStringPart 𠮷) "))) ;)))) <EOF>)`.
* The parser template isn't right for Java. Not sure why yet, but likely because the trgem template driver isn't setting up encodings somehow.
* php/CSharp is not for the "CSharp" target, but for the "Antlr4cs" old, unmaintained code, e.g., fails on "_mode" references. Antlr4cs != CSharp.

kaby76:
* Lexer symbol "Match" hides "Match" in the CSharp runtime.

kaby76:
* There is a difference in how the Antlr4 test plugin works vs the templates used for trgen in file streams.
* The default should always create with [CharStreams.fromFileName with UTF-8](https://github.com/antlr/grammars-v4/blob/02a99eff4482f809626812dba0fcc9a8a4bf49a4/_scripts/templates/Java/Program.java#L49). Right now [it's not](https://github.com/antlr/grammars-v4/blob/02a99eff4482f809626812dba0fcc9a8a4bf49a4/_scripts/templates/Java/Program.java#L47).
* The Antlr4 test plugin uses the new [CharStreams.fromPath](https://github.com/antlr/antlr4test-maven-plugin/blob/ffb6d5bcd7985a032e01c0c2dd3b27ab503b5eec/src/main/java/com/khubla/antlr/antlr4test/ScenarioExecutor.java#L128), [documented here](https://www.antlr.org/api/Java/org/antlr/v4/runtime/CharStreams.html#fromPath(java.nio.file.Path,java.nio.charset.Charset)). But, for case insensitive grammars--which is this grammar (php)--[the plugin uses an custom class](https://github.com/antlr/antlr4test-maven-plugin/blob/ffb6d5bcd7985a032e01c0c2dd3b27ab503b5eec/src/main/java/com/khubla/antlr/antlr4test/ScenarioExecutor.java#L130). That class [inherits from the old ANTLRFileStream](https://github.com/antlr/antlr4test-maven-plugin/blob/ffb6d5bcd7985a032e01c0c2dd3b27ab503b5eec/src/main/java/com/khubla/antlr/antlr4test/filestream/AntlrCaseInsensitiveFileStream.java#L37). That class [is deprecated](https://www.antlr.org/api/Java/org/antlr/v4/runtime/ANTLRFileStream.html), and differs entirely from what was published as the ["case changing class"](https://github.com/antlr/antlr4/blob/27ed0e9bb29d26ff249ccd800ca78b1df0aebb7e/doc/resources/CaseChangingCharStream.java#L14).
* Replacing the code in the trgen template for Java "fixes" the parse fail.
* Using [ANTLRFileStream](https://github.com/antlr/antlr4test-maven-plugin/blob/ffb6d5bcd7985a032e01c0c2dd3b27ab503b5eec/src/main/java/com/khubla/antlr/antlr4test/filestream/AntlrCaseInsensitiveFileStream.java#L41) instead of [wrapping](https://github.com/antlr/grammars-v4/blob/02a99eff4482f809626812dba0fcc9a8a4bf49a4/_scripts/templates/Java/Program.java#L52) the stream [created here](https://github.com/antlr/grammars-v4/blob/02a99eff4482f809626812dba0fcc9a8a4bf49a4/_scripts/templates/Java/Program.java#L49), ***the code works***.

***Conclusion: CharStreams.fromFileName(file_name, charset) does not work!***

kaby76:
* [In line 4 of php/examples/identifiers.php](https://github.com/antlr/grammars-v4/blob/02a99eff4482f809626812dba0fcc9a8a4bf49a4/php/examples/identifiers.php#L4), the second character should have a code point of 134071 (see [this converter](https://onlineutf8tools.com/convert-utf8-to-code-points)). CharStreams.fromFileName(fn, charset) is reading the code point correctly.
* ANTLRFileStream(fn, "UTF-8") does not read UTF-8 correctly. It returns two erroneous code points (55362 then 57271).
* The fact that the code that uses CharStreams.fromFileName fails to parse the input means that the grammar needs to be updated.
* examples/identifiers.php is valid UTF-8.
* The Antlr4 test plugin must be modified to use CharStreams.fromFileName because ANTLRFileStream does not work. I will file an issue.
* The grammar should be updated so that it accepts "identifiers.php".

kaby76:
The PhpLexer.g4 grammar is incorrect. It does not handle the full range of code points (see [NameString](https://github.com/antlr/grammars-v4/blob/02a99eff4482f809626812dba0fcc9a8a4bf49a4/php/PhpLexer.g4#L352)). But because ANTLRFileStream would break up the single code point--incorrectly--into two 16 bit code points, the grammar would "work".

The grammar should be updated to accept >16-bit code points. If this is done, the parser should work on identifiers.php. But, it will no longer work on [older Antlr4](https://github.com/antlr/antlr4/blob/master/doc/unicode.md#lexers-and-unicode-text) like 4.6.6 Antlr4cs that people for some reason are so fond of.

ANTLRFileStream in the [tester](https://github.com/antlr/antlr4test-maven-plugin/blob/ffb6d5bcd7985a032e01c0c2dd3b27ab503b5eec/src/main/java/com/khubla/antlr/antlr4test/filestream/AntlrCaseInsensitiveFileStream.java#L37) should not be used.

studentmain:
> While php 8.xxx works with identifiers.php, it should not. It's best to remove the test.

It works just because the file is using UTF-8 encoding, all non-ASCII characters are encoded into [\x80-\xff]{2,4}, thus accepted by official impl. I wonder if PHP "standard" said anything about using Unicode.

kaby76:
By the way, the closest "specification" for the language seems to be at:
* https://phplang.org/
which is a collection of Markdown files in:
* https://github.com/php/php-langspec

You can `grep -i -e unicode` of the files.

Interesting SO discussions:
* https://stackoverflow.com/questions/3417180/exotic-names-for-methods-constants-variables-and-fields-bug-or-feature
* https://stackoverflow.com/questions/5358824/unicode-identifiers-function-names-for-non-localization-purposes-advisable

KvanTTT:
After rebasing on the current master it works better but some tests are still failing.

kaby76:
* CSharp target `Grammars that failed:  algol60 php sql/tsql sql/postgresql` --I'll check it out..

kaby76:
* Ops. Trgen fails to parse the algol60 grammar.

KvanTTT:
> Did you change the Antlr tool grammar for your caseInsensitive feature?

Yes, but it's the easy change: https://github.com/parrt/intellij-plugin-v4/commit/c99fb3e7b1c6e9933f0de9b05f59856dace17cf6

kaby76:
> > Did you change the Antlr tool grammar for your caseInsensitive feature?
> 
> Yes, but it's the easy change: [parrt/intellij-plugin-v4@c99fb3e](https://github.com/parrt/intellij-plugin-v4/commit/c99fb3e7b1c6e9933f0de9b05f59856dace17cf6)

GFC. I made the change, but it still doesn't work. So I decided to check using the Antlr4/Intellij plugin version 1.18 since that has the new grammar. I then noticed that grammars-v4/php/ grammar fails on this line [here](https://github.com/KvanTTT/grammars-v4/blob/3dbce9b505adffec798aee83b5d7233a810585e5/php/PhpLexer.g4#L353) at the opening of the charset for the rule! So, although trgen and the plugin use slightly different grammars (pain in the a..), that's why trgen fails to parse php/. Not sure why it's now failing, but I'm looking at the tokens to make sure this gets fixed.

PR #2577, opened 2022-04-14
Support CTE with create view statement in MySQL parser

ani-sha:


sql/mysql/Positive-Technologies/MySqlParser.g4
```diff
@@ -234,14 +234,31 @@ createTrigger
       routineBody
     ;
 
+withClause
+    : WITH RECURSIVE? commonTableExpressions
+    ;
+
+commonTableExpressions
+    : cteName (('(' cteColumnName (',' cteColumnName)* ')')?  AS '(' dmlStatement ')')
```
KvanTTT, line 242: It looks like outer parantheses are excess:

```suggestion
    : cteName ('(' cteColumnName (',' cteColumnName)* ')')?  AS '(' dmlStatement ')'
```

Issue #2590, opened 2022-04-21
Missing GROUPING SETS syntax


grzeng-go:
eg:

```sql
SELECT
    aggregate_function(column_1)
    column_2,
    column_3,
FROM
    table_name
GROUP BY
    GROUPING SETS (
        (column_2, column_3),
        (column_2),
        (column_3),
        ()
);
```

grzeng-go:
### TSqlLexer.g4
**add:**
SETS:                                   'SETS';

------------------------------------------------------------------------------

### TSqlParser.g4
**change:**
query_specification
    : SELECT allOrDistinct=(ALL | DISTINCT)? top=top_clause?
      columns=select_list
      // https://msdn.microsoft.com/en-us/library/ms188029.aspx
      (INTO into=table_name)?
      from=from_clause?
      where=where_clause?
      // https://msdn.microsoft.com/en-us/library/ms177673.aspx
      (GROUP BY ((groupByAll=ALL? groupBys+=group_by_item (',' groupBys+=group_by_item)*) | GROUPING SETS '(' groupSets+=grouping_sets_item (',' groupSets+=grouping_sets_item)* ')'))?
      (HAVING having=search_condition)?
    ;
**add:**
grouping_sets_item
    : '('? groupSetItems+=group_by_item (',' groupSetItems+=group_by_item)* ')'?
    | '(' ')'
    ;

KvanTTT:
It would be very helpful if you suggested a pull request with fix.

PR #2593, opened 2022-04-22
fix(sql/tsql): add support for GROUPING SETS

grzeng-go:
Related Documentation:
 https://docs.microsoft.com/en-us/sql/t-sql/queries/select-group-by-transact-sql?view=sql-server-ver15

#2590

### DIFF
--- a/sql/tsql/TSqlParser.g4
+++ b/sql/tsql/TSqlParser.g4
@@ -3278,7 +3278,7 @@ query_specification
       (FROM from=table_sources)?
       (WHERE where=search_condition)?
       // https://msdn.microsoft.com/en-us/library/ms177673.aspx
-      (GROUP BY groupByAll=ALL? groupBys+=group_by_item (',' groupBys+=group_by_item)*)?
+      (GROUP BY ((groupByAll=ALL? groupBys+=group_by_item (',' groupBys+=group_by_item)*) | GROUPING SETS '(' groupSets+=grouping_sets_item (',' groupSets+=grouping_sets_item)* ')'))?
       (HAVING having=search_condition)?
     ;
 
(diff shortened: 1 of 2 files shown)


sql/tsql/TSqlParser.g4
```diff
@@ -3278,7 +3278,7 @@ query_specification
       (FROM from=table_sources)?
       (WHERE where=search_condition)?
       // https://msdn.microsoft.com/en-us/library/ms177673.aspx
-      (GROUP BY groupByAll=ALL? groupBys+=group_by_item (',' groupBys+=group_by_item)*)?
+      (GROUP BY ((groupByAll=ALL? groupBys+=group_by_item (',' groupBys+=group_by_item)*) | GROUPING SETS '(' groupSets+=grouping_sets_item (',' groupSets+=grouping_sets_item)* ')'))?
```
KvanTTT, line 3281: It seems like there is the excess pair of parentheses:

```suggestion
      (GROUP BY (groupByAll=ALL? groupBys+=group_by_item (',' groupBys+=group_by_item)* | GROUPING SETS '(' groupSets+=grouping_sets_item (',' groupSets+=grouping_sets_item)* ')'))?
```

grzeng-go: cannot be removed, because if removed, GROUPING will be treated as group_by_item

KvanTTT: `|` has lower priority than concatentation.

grzeng-go: eg: SELECT * FROM A GROUP BY GROUPING SETS (())
 if removed, GROUPING will be identified as group_by_item,
I tested it with the antlr4 plugin and it is like this

PR #2602, opened 2022-05-02
Updates to php/ grammar for case insensitive option in grammar

kaby76:
* Add CSharp target for php/ grammar.
* Move "CSharp" to Antlr4cs because it's not Antlr4.Runtime.Standard compatible.
* Add testing of php/ grammar for Java and CSharp.
* Add comments to examples/identifiers.php until Unicode variable names can be resolved.

Prerequisite for https://github.com/antlr/grammars-v4/pull/2576


php/Antlr4cs/PhpLexerBase.cs
```diff
@@ -0,0 +1,205 @@
...
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+using Antlr4.Runtime;
+
+// Replace PhpParser with its fully qualified name below
+//using static PhpLexer;
+
+public abstract class PhpLexerBase : Lexer
```
KvanTTT, line 31: Is it possible to merge it into `php/CSharp/PhpLexerBase.cs`? Use `#if ANTLR_STANDARD` preprocessor directive if code is different. See [AntlrCaseInsensitiveInputStream.cs ](https://github.com/KvanTTT/DAGE/blob/master/AntlrGrammarEditor/AntlrRuntimes/CSharp/AntlrCaseInsensitiveInputStream.cs) for example.

KvanTTT, line 31: Is it possible to merge it into `php/CSharp/PhpLexerBase.cs`? Use `#if ANTLR_STANDARD` preprocessor directive if code is different. See [AntlrCaseInsensitiveInputStream.cs ](https://github.com/KvanTTT/DAGE/blob/master/AntlrGrammarEditor/AntlrRuntimes/CSharp/AntlrCaseInsensitiveInputStream.cs) for example.

php/examples/identifiers.php
```diff
@@ -1,4 +1,10 @@
 <?php
 
-$привет_мир = 42;
-$𠮷 = "𠮷";
+// Comment these out until ASCII vs UTF-8 with large codepoints is
+// resolved.
+// See discussion at https://github.com/antlr/grammars-v4/pull/2576#issuecomment-1107839866
```
KvanTTT, line 5: It probably makes sense to point out the concrete issue as well.

KvanTTT, line 5: It probably makes sense to point out the concrete issue as well.

php/Antlr4cs/PhpLexerBase.cs
kaby76: I'm not sure what to make of the comment `[r]eplace PhpParser with its fully qualified name below using static PhpLexer;`. It makes absolutely no sense. I'm removing it.

I moved the existing "CSharp/" code to Antlr4cs/ to avoid confusion. But it should just be removed because [your PR](https://github.com/antlr/grammars-v4/pull/2576) is going to modify this grammar with the new option, which is not supported in 4.6.6 Antlr4cs.

Antlr4cs should not be supported because the code generator and runtime are years behind the current Antlr4 code and not being maintained. Some grammars are already incompatible because they use Unicode sets. [Your PR](https://github.com/antlr/grammars-v4/pull/2576) introduces `options { caseInsensitive = true; }`. Are you now planning to support Antlr4cs/ for grammars-v4/php/, adding the old grammars to Antlr4cs/???

I thought we already talked about this. You yourself agreed to drop support for Antlr4cs--at least that's what I understood. https://github.com/antlr/grammars-v4/pull/2497#issuecomment-1036224128

Although we don't test Antlr4cs in grammars-v4, trgen supports Antlr4cs. I support it in trgen because I see a lot of questions still for Antlr4cs on StackOverflow.com. When I tell people to move on to Antlr4.Runtime.Standard, I typically hear it's because there's no "integrated build". But [there is](https://github.com/kaby76/Antlr4BuildTasks).

By the way, [GoLexer.g4](https://github.com/antlr/grammars-v4/blob/62103121ce94e80bf0e8a25680fc699fdbe4b008/golang/GoLexer.g4#L218) uses Unicode sets, but [there's an Antlr4cs target](https://github.com/antlr/grammars-v4/tree/master/golang/Antlr4cs). That cannot possibly work since Antlr4cs is stuck at version 4.6.6.

Likewise, [your PR](https://github.com/antlr/grammars-v4/pull/2576) should remove all the Antlr4cs/ directories of the grammars you are modifying, or you will need to move the old, pre `options { caseInsensitive = true; }` grammars to Antlr4cs/ directories.

php/examples/identifiers.php
kaby76: This file was changed so you can get your PR to work. It's only a placeholder. To make that clear, I'm adding the phrase "TODO" in a comment in this file.

There are two issues to resolve:
* The Maven Antlr testing plugin still uses ANTLRFileStream which does not support Unicode UTF-8. It only happens with "case insensitive" streams, but if you're going to rewrite grammars to use the new option, then it's not as important.
* The implementation supports Unicode variable names; the "spec" does not address the issue, and actually says otherwise. We need to decide whether to support Unicode names because people seemed to make use of this "feature". If so, the grammar will have to be changed.

php/Antlr4cs/PhpLexerBase.cs
KvanTTT: Ok, let's drop support of outdated C# runtime. In this case I suggest removing replacing `CSharp` directory with your code for the actual runtime.

kaby76: Analyzing the comment `[r]eplace PhpParser with its fully qualified name below ["]using static PhpLexer;["]` further, the ["using static PhpLexer;"](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-directive#static-modifier) allows one to reference the token types symbols defined in PhpLexer.cs, without fully qualifying them, e.g., "PHPEnd" vs. "PhpLexer.PHPEnd".

I already adjusted the token types to be fully qualified because I rather prefer to know where the symbol is bound rather than use an IDE, cursor over the symbol, and see what the definition is. The comment is not necessary. And, although the Antlr tool generates another identical definition for PHPEnd in PhpParser.cs, the comment is misleading because the commented code is "using static PhpLexer;" not "using static PhpParser;".

The comment is being deleted.

I am also renaming symbol "Match" in the grammars to be "Match_" because it results in the warning "PhpParser.cs(58,19): warning CS0108: 'PhpParser.Match' hides inherited member 'Parser.Match(int)'. Use the new keyword if hiding was intended." It is a symbol conflict for CSharp.

kaby76:
@KvanTTT I removed the Antlr4cs code. Please check this PR over. Let's keep a "TODO" tag in the test file so we can work on that after we see if this gets your PR further.

KvanTTT:
It looks good. Thanks!

Issue #2597, opened 2022-05-01
Using "operator" in the field name results in a syntax error


yososs:
Checking the Dart2 specification, 'operator' can be used as a field name. Probably the same problem will occur with other marked keywords.

Keyword Specifications for Dart2
* https://dart.dev/guides/language/language-tour#keywords

> Avoid using these words as identifiers. However, if necessary, the keywords marked with superscripts can be identifiers:
> * Words with the superscript 1 are contextual keywords, which have meaning only in specific places. They’re valid identifiers everywhere.
> * Words with the superscript 2 are built-in identifiers. These keywords are valid identifiers in most places, but they can’t be used as class or type names, or as import prefixes.
> * Words with the superscript 3 are limited reserved words related to asynchrony support. You can’t use await or yield as an identifier in any function body marked with async, async*, or sync*.
> All other words in the table are reserved words, which can’t be identifiers.
> 

Reproduced code
```
class A{
  bool isPlusOrMinus(Expression expression) {
    if (expression.operator == '+') return true;
    if (expression.operator == '-') return true;
    return false;
  }
}
```

bkiers:
Then probably this:

```
unconditionalAssignableSelector
  : '[' expression ']'
  | '.' identifier
  ;
```

should become:

```
unconditionalAssignableSelector
  : '[' expression ']'
  | '.' identifier
  | '.' 'operator'
  ;
```

yososs:
Thanks for sharing.
I will check the operation tomorrow.

yososs:
I ran the following unit test code.
It works well, but
I found that there are still a few problems.

* The "new" keyword fails to test.

``` Java
public class DartParserTest {

        // see: https://dart.dev/guides/language/language-tour#keywords
	@Test
	public void testKeywords0() {
		String[] keywords0 = { "assert", "break", "case", "catch", "class", "const", "continue", "default", "do", "else",
				"enum", "extends", "false", "final", "finally", "for", "if", "in", "is", 
				"new", 
				"null", "rethrow",
				"return", "super", "switch", "this", "throw", "true", "try", "var", "void", "while", "with" };

		for (String k : keywords0) {
			String content = "class A{\n"
					+ "  bool isPlusOrMinus(Expression expression) {\n"
					+ "    if (expression."+k+" == '+') return true;\n"
					+ "    if (expression."+k+" == '-') return true;\n"
					+ "    return false;\n"
					+ "  }\n"
					+ "}\n";
[...]

PR #2603, opened 2022-05-02
Fix for Issue 2597

kaby76:
* Copy recent updated grammar from https://github.com/dart-lang/sdk/blob/main/tools/spec_parser/Dart.g to here.
* Convert target-specific to target-agnostic format.
* Add test for #2597.


kaby76:
Right. The grammar at https://github.com/dart-lang/sdk/blob/main/tools/spec_parser/Dart.g isn't the grammar for Dart it seems. Not sure WTH it is. Looking around...otherwise I will need to scrape the grammar from "spec" (not really a spec, but just "doc"), which I did in the past.

kaby76:
Closing this PR off until I have the [scraper](https://github.com/kaby76/ScrapeDartSpec) working. It seems I left that off just at the point of scraping the spec, but not actually refactoring it into a usable grammar because [Trash](https://github.com/kaby76/Domemtech.Trash) wasn't far along enough. It is better, but still I have to fix the critical [unfold](https://github.com/kaby76/Domemtech.Trash/tree/main/trunfold) refactoring.

PR #2605, opened 2022-05-03
Possible to include `*` in BLOCK_COMMENT

### LINKED ISSUE #2604: [Rust] Lexical parsing of block comments may be incorrect
If an asterisk is included in a block comment, it is not recognized as a block comment.

This problem can be reproduced with the following source code

``` rust
/*
 * let a = 1*1;
 */ 
fn main() {
    println!("Hello, world!");
}
```

It works correctly if you remove the asterisk as follows

``` rust
/*
 * let a = 1;
 */ 
fn main() {
    println!("Hello, world!");
}
```

This problem occurs when the source code is commented out, so it is expected to occur frequently.

Reference
* https://doc.rust-lang.org/reference/comments.html

bkiers:
Given [the specs](https://doc.rust-lang.org/reference/comments.html):

```
BLOCK_COMMENT :
     /* (~[* !] | ** | BlockCommentOrDoc) (BlockCommentOrDoc | ~*/)* */
   | /**/
   | /***/
```

it says that no closing block-comment can occur: `~*/`. However, inside the ANTLR grammar it was defined as `~[*]`, which is not the same.

Given that the inner block comment is already matched _ungreedily_ (`*?`), it is safe to replace `~[*]` with `.`. I did have to move `/***/` up, otherwise input like `"/***/"` was not matched properly.

Fixes #2604

### DIFF
--- a/rust/RustLexer.g4
+++ b/rust/RustLexer.g4
@@ -99,8 +99,9 @@ LINE_COMMENT: ('//' (~[/!] | '//') ~[\r\n]* | '//') -> channel (HIDDEN);
 
 BLOCK_COMMENT
    :
-   (
-      '/*'
+   ( '/**/'
+   | '/***/'
+   |  '/*'
       (
          ~[*!]
          | '**'
@@ -108,10 +109,8 @@ BLOCK_COMMENT
       )
       (
          BLOCK_COMMENT_OR_DOC
-         | ~[*]
+         | .
       )*? '*/'
-      | '/**/'
-      | '/***/'
    ) -> channel (HIDDEN)
    ;
 
--- a/rust/examples/comment.rs
+++ b/rust/examples/comment.rs
@@ -19,6 +19,8 @@ pub mod outer_module {
     pub mod inner_module {}
 
     pub mod nested_comments {
+        /* let a = 1*1; */
+
         /* In Rust /* we can /* nest comments */ */ */
 
         // All three types of block comments can contain or be nested inside


studentmain:
and other block comment types, I can't figure out how to do it correctly when I was copy from their spec, so I replaced all of them to ~[*] at that moment.

bkiers:
Hmm, can't quite get the `OUTER_BLOCK_DOC` to work properly: tests from `comments.rs` keep breaking. I'll leave this for someone else.

studentmain:
Looks like some action required in order to implement ~*/

bkiers:
> Looks like some action required in order to implement ~*/

I don't think that would be necessary. When I create a small demo grammar:

```
grammar RustComments;

parse
 : .*? EOF
 ;

BLOCK_COMMENT
 : '/*' '*'? '*/'
 | '/*' ( BlockCommentOrDoc | ~[*!] | '**' ) ( BlockCommentOrDoc | . )*? '*/'
 ;

OUTER_BLOCK_DOC
 : '/**' ( ~'*' | BlockCommentOrDoc ) ( BlockCommentOrDoc | . )*? '*/'
 ;

INNER_BLOCK_DOC
 : '/*!' ( BlockCommentOrDoc | . )*? '*/'
 ;

LINE_COMMENT
 : '//' ( ~[/!] | '//' ) ~[\r\n]*
 | '//'
 ;

INNER_LINE_DOC
 : '//!' ~[\r\n]*
 ;

OUTER_LINE_DOC
 : '///' ~[\r\n]*
 ;

fragment BlockCommentOrDoc
 : BLOCK_COMMENT
 | OUTER_BLOCK_DOC
 | INNER_BLOCK_DOC
 ;

OTHER
 : . -> skip
 ;
```

and test it like this:

```java
String source =
    "//! A doc comment that applies to the implicit anonymous module of this crate\n" +
    "//!  - Inner line doc\n" +
    "//!! - Still an inner line doc (but with a bang at the beginning)\n" +
    "/*!  - Inner block doc */\n" +
    "/*!! - Still an inner block doc (but with a bang at the beginning) */\n" +
    "//   - Only a comment\n" +
    "///  - Outer line doc (exactly 3 slashes)\n" +
    "//// - Only a comment\n" +
    "/*   - Only a comment */\n" +
    "/**  - Outer block doc (exactly) 2 asterisks */\n" +
    "/*** - Only a comment */\n" +
    "/* let a = 1*1; */\n" +
    "/*! let a = 1*1; */\n" +
    "/* In Rust /* we can /* nest comments */ */ */" +
    "/*   /* */  /** */  /*! */  */\n" +
    "/*!  /* */  /** */  /*! */  */\n" +
    "/**  /* */  /** */  /*! */  */";

RustCommentsLexer lexer = new RustCommentsLexer(CharStreams.fromString(source));
CommonTokenStream stream = new CommonTokenStream(lexer);

stream.fill();

for (Token t : stream.getTokens()) {
    System.out.printf("%-20s '%s'\n",
        RustCommentsLexer.VOCABULARY.getSymbolicName(t.getType()),
        t.getText().replace("\n", "\\n"));
}
```

all comments are tokenized properly AFAIK. However, when I incorporate the lexer changes in the main grammar repo, the tests fail. It is not clear to me why these tests fail because all comments are embedded in 1 test file, making it rather hard to debug.

Issue #2609, opened 2022-05-06
[antlr/antlr4] Sync code and bring Antlr4 grammar up to date for Antlr 4.10.1


kaby76:
The antlr/antlr4/ grammar is out of date with 4.10.1. The grammar needs to be updated with the lexer rule-level options. In addition, all the various target base classes (Cpp, CSharp, Java, JavaScript, Python3) are in different states--they do not do the same thing and have methods that aren't used, the result of people modifying one target and purposefully ignoring the other targets.

PR #2610, opened 2022-05-06
[antlr4 grammar/CI build] Fix for 2609

kaby76:
This PR is for the antlr/antlr4/ grammar https://github.com/antlr/grammars-v4/issues/2609. There's also a change to the build, but that is only for supporting Cpp split grammars with base class code that needs to be compiled and linked.

* Updated the antlr/antlr4 grammar with the new 4.10.1 syntax for `caseInsensitive`.
* Synced the base class code between targets. The targets that are implemented are Cpp, CSharp, Java, JavaScript, and Python3; I did not implement any new targets.
* Added the antlr/antlr4/ grammar to CI testing to help prevent targets from becoming out of sync.
* Note, forking of the antlr/antlr4/ grammar was required for Cpp and Python3. This is because there really is no such thing as "target agnostic format", but what i call "pseudo target agnostic format". It's a syntax that is identical for most, but not all, targets. We can get "target agnostic format", but only if we introduce a syntax for "self", e.g., [$parser and $lexer handles with a '.' reference](https://github.com/antlr/antlr4/issues/3508). Cpp will also need to generate a `#include` for the base class, which the Antlr4 tool does not do.
* trgen needed to be updated to fix a bug in the reference to the "additional_sources" attribute used in the CMakeLists.txt templated file. 
* There is a difference between this grammar and the one in the Intellij Antlr plugin [in lexing options](https://github.com/antlr/intellij-plugin-v4/blob/1f36fde17f7fa63cb18d7eeb9cb213815ac658fb/src/main/antlr/org/antlr/intellij/plugin/parser/ANTLRv4Lexer.g4#L602), but I haven't resolved that difference. Instead, I just added a "TODO".


teverett:
@kaby76  thanks

Issue #2612, opened 2022-05-06
[CSharp] Parsing source with syntax error causes ClassCastExceptin


yososs:
I found this to be a cast error when entering the wrong C++ syntax into CSharp's parser.
This does not appear to be the correct way to handle syntax errors.

Sources containing syntax error
``` C#
class A{
  void a() {
    int[] nums = {1, 3, 5, 7};
    for (int num : nums){
        b(num);
    }
  }
}
```
Added the following cast check to avoid cast errors in CSharpParserBase.java.

``` C#
public abstract class CSharpParserBase extends Parser
{
    protected CSharpParserBase(TokenStream input)
    {
	super(input);
    }

    protected boolean IsLocalVariableDeclaration()
    {
   // fix
    if (!(this._ctx instanceof CSharpParser.Local_variable_declarationContext)) {
    	return false;
    }

	CSharpParser.Local_variable_declarationContext local_var_decl = (CSharpParser.Local_variable_declarationContext)this._ctx;
	if (local_var_decl == null) return true;
	CSharpParser.Local_variable_typeContext local_variable_type = local_var_decl.local_variable_type();
	if (local_variable_type == null) return true;
	if (local_variable_type.getText().equals("var")) return false;
	return true;
    }
}
```

~~However, it should be a syntax error, but it is not a syntax error.
Few people may consider this use case important.
Maybe it should be managed as a separate issue.~~

kaby76:
The grammars-v4/csharp/ parser gives a syntax error for that input:

```
# run in Bash shell.
git clone https://github.com/antlr/grammars-v4.git
cd grammars-v4/csharp
trgen
cd Generated
make
cat - << HERE > test.in
class A{
  void a() {
    int[] nums = {1, 3, 5, 7};
    for (int num : nums){
        b(num);
    }
  }
}
HERE
dotnet run -file test.in
```
Output:
```
line 4:17 mismatched input ':' expecting ';'
Time: 00:00:00.1457023
Parse failed.
```

Please clarify what you are trying to do.

yososs:
I am intentionally entering source code that results in a syntax error. The expected result is that CSharpParser reports a syntax error.

1. the actual result is that ClassCastException is thrown
2. if I fix this problem by adding a cast check, ~~no syntax error is reported~~

2 was an error in my method of syntax checking and was a pointing error.

To reproduce 1, follow the steps

You need to test on a java target.

1. save the above source in examples/test.cs
2. mvn3 test

```
[...]

PR #2615, opened 2022-05-07
[csharp grammar/Java target] Fix issue 2612

kaby76:
https://github.com/antlr/grammars-v4/issues/2612

* Fix the base class code, which @yososs wrote [here](https://github.com/antlr/grammars-v4/issues/2612).
* Add the test to examples/.
* Add csharp/ grammar testing for Java target (now testing targets CSharp and Java).


teverett:
@kaby76  thanks!

PR #2620, opened 2022-05-10
[CSS3] Adding license file

SergeyRusskih:
Using software without a license involves legal risks to the users and might result in rejecting using the software. This is the request to add the license file in a form of an MIT license file. I didn't manage to find the original author by the login and also the original date of creation. So, I put Tom Everett as the person who ported the code and the date of the portation.


teverett:
@SergeyRusskih  thanks

PR #2623, opened 2022-05-16
Add wren grammar

besok:
add grammar for wren language


kaby76:
So, the PHP target for the wren grammar works *very* slowly. In fact, the parse of the first test in examples/ had to be killed. (I wrote [trwdog](https://github.com/kaby76/Domemtech.Trash/tree/main/trwdog) so an admin doesn't have to manually kill the build.)

The reason it is slow is because the grammar is ambiguous over [atomExpression](https://github.com/besok/grammars-v4/blob/98f4986dff0ae440ab215458c14788b4b7ea1743/wren/WrenParser.g4#L202), alts 1 and 9. An example is `_item` in [this test](https://github.com/besok/grammars-v4/blob/98f4986dff0ae440ab215458c14788b4b7ea1743/wren/examples/binary_tree.wren#L5). It is derived by one of two ways:

    atomExpression => id => IDENTIFIER
    atomExpression => call => callHead => id

You can see other smaller problems using the Intellij Antlr4 plugin on the grammar.

You can either remove the ambiguity or disable the testing of the PHP target by adding it to the ["skip list"](https://github.com/besok/grammars-v4/blob/master/_scripts/skip-php.txt).

kaby76:
Sounds good. At any point, you can simply add `wren` to the php-skip.txt list and it'll bypass the grammar, then come back to it later.

besok:
yep, this is always a backup plan :)

teverett:
@besok  good to pull this PR?

besok:
Sure!

teverett:
@besok  thanks!

PR #2625, opened 2022-05-17
javascript: enable yield in non strict mode

joelmo:


KvanTTT:
Could you also please add a sample that covers the suggested changes?

teverett:
@joelmo thanks

PR #2627, opened 2022-05-19
Create parse tree formatter

joelmo:


kaby76:
Java 11 is the maximum required version for this grammar. I don't know if the plugin recognizes a specification of the required version of Java, but trgen assumes verison 11.

teverett:
@kaby76  thanks for bringing this up.  I've been a little concerned about it for a while.  Java has moved on and even Java 11 is numerous revisions behind the current one.   I agree that we need to update the docs and I also think we need to update our min java version across this repo, the formatter and the tester repos.   I think first step is to update the tester: 

`https://github.com/antlr/antlr4test-maven-plugin
`

Is there a minimum java version we should be targetting?

kaby76:
> @kaby76 thanks for bringing this up. I've been a little concerned about it for a while. Java has moved on and even Java 11 is numerous revisions behind the current one. I agree that we need to update the docs and I also think we need to update our min java version across this repo, the formatter and the tester repos. I think first step is to update the tester:
> 
> `https://github.com/antlr/antlr4test-maven-plugin `
> 
> Is there a minimum java version we should be targeting?

Hi Tom, I'm not sure what the answer is here. I opened [an Issue](https://github.com/antlr/grammars-v4/issues/2631) for tracking the problem.

joelmo:
> Hi! I don't think grammars repository is the correct place for JavaScriptFormatter. It contains grammars and testing utils, but not parse tree helpers. What's the purpose of this formatter in grammars?

Maybe someone needs to translate a ParseTree into source. I use this to build a transpiler and wanted to share this. This only works with this parser and it would be harder to find such tools in another repo. I understand if you don't think this belongs here, you can decide what to do with this request.

teverett:
@parrt Could this be a new repo in the Antlr org?

kaby76:
> Maybe someone needs to translate a ParseTree into source.

This formatter is tied into the generated parser code because it uses the generated token names (JavaScriptParser.Else), rule numbers (JavaScriptParser.RULE_block) parse tree types (JavaScriptParser.IdentifierContext), and visitor methods (`public java.lang.Void visitIfStatement(JavaScriptParser.IfStatementContext ctx)`) in Java code. Given that, it would seem appropriate to place it here. But, the problem with this idea is that people would be adding all sorts of switches and features, not to mention other targets. We should not be coding tools like this because everything would be tightly coupled. To remove the tight coupling, it should all be done at a symbolic level, and then placed in a separate package.

Also, you could write the pretty printer this way, but I would suggest a ML-based PP instead, such as [Parr's Codebuff](https://github.com/antlr/codebuff), which I ported to [C#](https://github.com/kaby76/cs-codebuff). Although Parr's code is hard to read :), it's a great idea. Flag-based PPs are the current coding standard, but ML is likely cleaner and more customizable.

parrt:
I'm almost to the point where I can try to use a neural network to redo my CodeBuff tool, @kaby76. Should work a lot better. Also, this doesn't seem to be a parse tree formatter. this appears to be a JavaScript formatter correct? The title is misleading.

I'm not sure ANTLR org is the right place for specific code formatters... I suppose we could create a repository for a collection of formatters, but would they all have a similar better functionality or would each one be a separate isolated tool?

joelmo:
You are right, the title is incorrect.

To loosen coupling a bit this can be put in a separate folder (_tools or _formatters). If these are tested it should be fine to break them when patching parsers and the tool/formatter maintainer would maybe fix them later.

PR #2633, opened 2022-05-21
Add Lucene grammar

bkiers:
Extracted from the original JavaCC grammar: https://github.com/apache/lucene/blob/main/lucene/queryparser/src/java/org/apache/lucene/queryparser/flexible/standard/parser/StandardSyntaxParser.jj

Examples scraped from: https://lucene.apache.org/core/2_9_4/queryparsersyntax.html


teverett:
@bkiers  thanks!

PR #2639, opened 2022-06-01
[Python grammar, Python3 target] Issue 2638

kaby76:
This is a fix for the grammars-v4/python/python/ grammar for the Python3 target.
* All files in grammars-v4/python/python/Python are moved to .../Python3. This is because the directory name indicates the target source files that need to be overladed on top of the rest of the source. There is no way this can be run in Python2.
* Update the GitHub Actions workflow so that it uses Python 3.10. This grammar requires the latest Python3, and the build doesn't offer a distinct version of Python per grammar. 
* General changes to get the code to work:
    * The comments in the two base class files are wrong. Don't include anything--it's all generated for you by the Antlr tool.
    * [Line 183 of PythonLexerBase.py](https://github.com/antlr/grammars-v4/blob/2228b048e67b7cf709ce3885b16d22e2549d9f7d/python/python/Python/PythonLexerBase.py#L183) cannot be parsed with Python 3.9.
    * Comments on [circular dependency](https://github.com/antlr/grammars-v4/blob/2228b048e67b7cf709ce3885b16d22e2549d9f7d/python/python/Python/PythonLexerBase.py#L32) aren't a problem I found, at least not in Python 3.10. Remove this code and replace "_tokens[...]" with token references.
* The grammar was forked because Python requires a "self" object reference in the generated parser and lexer code for actions. ***I really hate forking the grammar, but until Antlr has a universal "self" object reference that gets converted to "this" (Java, C#) or "self" (Python3) or "this->" (C++), we have to fork grammars somewhere because there is no such thing as a "target agnostic format", only a "pseudo target agnostic format".***

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -102,7 +102,7 @@ jobs:
       if: ${{ matrix.language == 'Python3' }}
       uses: actions/setup-python@v3
       with:
-        python-version: '3.8'
+        python-version: '3.10'
     - name: Do not trust Python env
       if: ${{ matrix.language == 'Python3' }}
       run: |
--- a/_scripts/skip-python3.txt
+++ b/_scripts/skip-python3.txt
@@ -65,7 +65,6 @@ pl0
 ply
 powerbuilder
 protobuf3
-python/python
 python/python2
 python/python3
 python/python3-py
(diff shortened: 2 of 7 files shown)


teverett:
@kaby76 thanks

PR #2644, opened 2022-06-07
fix arrayCreation is parsed into functionCall

qiaodandedidi:
The php `array()` keyword is recognized as a function, it should be an `arrayCreation`.

### What version of Antlr are you using?

Version 4.9.3，but i think almost all versions have the same problem

###  Reproduce

```php
<?php
        $arr = array(1, 2, 3);

```

Actual result
image

Expected result
image

### DIFF
--- a/php/PhpParser.g4
+++ b/php/PhpParser.g4
@@ -449,6 +449,7 @@ expression
 
     | Print expression                                          #PrintExpression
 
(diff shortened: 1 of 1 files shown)


qiaodandedidi:
> Thank you for the fix and welcome to ANTLR grammars repository!

My pleasure

PR #2650, opened 2022-06-10
Support PLSQL MAXTRANS and LOB names using quoted identifiers

Naros:


sql/plsql/PlSqlParser.g4
```diff
@@ -2688,7 +2689,7 @@ lob_segname
     ;
 
 lob_item
-    : regular_id
+    : (regular_id | quoted_string)
```
KvanTTT, line 2692: Parentheses are excess, first-level alternatives should be on new lines:

```suggestion
    : regular_id
    | quoted_string
```

PR #2651, opened 2022-06-11
promql: support vector match operations

facetoe:
Add support for vector match operations to the grammar - https://prometheus.io/docs/prometheus/latest/querying/operators/#one-to-one-vector-matches


teverett:
@facetoe thanks

PR #2655, opened 2022-06-14
[PromQL] support case insensitive keywords

facetoe:
The official PromQL parser accepts keywords in any case, so do the same
for the PromQL ANTLR grammar.

### DIFF
--- a/promql/PromQLLexer.g4
+++ b/promql/PromQLLexer.g4
@@ -28,6 +28,10 @@
 
 lexer grammar PromQLLexer;
 
+// All keywords in PromQL are case insensitive, it is just function,
+// label and metric names that are not.
+options { caseInsensitive=true; }
+
 NUMBER: [0-9]+ ('.' [0-9]+)?;
 
 STRING
@@ -91,7 +95,7 @@ AGGREGATION_OPERATOR
     | 'quantile'
     ;
 
-FUNCTION
+FUNCTION options { caseInsensitive=false; }
     : 'abs'
     | 'absent'
     | 'absent_over_time'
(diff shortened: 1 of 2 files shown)


KvanTTT (changes requested):
Please use [`caseInsensitive`](https://github.com/antlr/antlr4/blob/master/doc/options.md#caseinsensitive) option instead of manual tokens transformation. Moreover not all tokens were transformed to the case-insensitive format.

facetoe:
> Please use [caseInsensitive](https://github.com/antlr/antlr4/blob/master/doc/options.md#caseinsensitive) option instead of manual tokens transformation

Ahh, nice, I thought there must have been such a thing but didn't find this 👍

> Moreover not all tokens were transformed to the case-insensitive format.

It is only the list here that are accepted as lower case - https://github.com/prometheus/prometheus/blob/main/promql/parser/lex.go#L95, it is done here - https://github.com/prometheus/prometheus/blob/main/promql/parser/lex.go#L789. Functions and friends are still case sensitive.

KvanTTT:
> Ahh, nice, I thought there must have been such a thing but didn't find this 👍

Yes, it appeared not long time ago.

> It is only the list here that are accepted as lower case -

In this case you can use rule-targeted `caseInsensitive` option: https://github.com/antlr/antlr4/blob/master/doc/lexer-rules.md#caseinsensitive

promql/PromQLLexer.g4
```diff
@@ -160,7 +164,7 @@ TIME_RANGE
 
 DURATION: [0-9]+ ('s' | 'm' | 'h' | 'd' | 'w' | 'y');
 
-METRIC_NAME: [a-zA-Z_:] [a-zA-Z0-9_:]*;
-LABEL_NAME:  [a-zA-Z_] [a-zA-Z0-9_]*;
+METRIC_NAME options { caseInsensitive=false; }: [a-zA-Z_:] [a-zA-Z0-9_:]*;
+LABEL_NAME options { caseInsensitive=false; }:  [a-zA-Z_] [a-zA-Z0-9_]*;
```
KvanTTT, line 169: It can be rewritten without using of option, ANTLR handles ranges as well:

```suggestion
METRIC_NAME: [a-z_:] [a-z0-9_:]*;
LABEL_NAME:  [a-z_] [a-z0-9_]*;
```

facetoe: Nice, much cleaner.

KvanTTT (approved):

PR #2658, opened 2022-06-15
[PromQL] accept functions with no parameters

facetoe:
Thes are also valid functions, for example "time()".


KvanTTT:
Could you please add a sample that covers the case?

Issue #2405, opened 2021-12-01
Malformed start rules in grammars


kaby76:
I was reading on StackOverflow [this question on identifying a start rule](https://stackoverflow.com/questions/70171495/antlr4-python-binding-for-java8-missing-start-rule) for a grammar. The [answer](https://stackoverflow.com/questions/70171495/antlr4-python-binding-for-java8-missing-start-rule/70171977#70171977), from @bkiers , was to look for the rule that contains an EOF.

So, I decided to see what grammars in this repo contain a rule with an EOF using this script:

    for i in `find . -name 'pom.xml' | sed 's#/pom.xml##' | sort -u`
    do
        pushd $i > /dev/null 2>&1
        shopt -s nullglob
        compgen -W *.g4 > /dev/null 2>&1
        case $? in
            0)
                echo ""
                echo *.g4
                trparse *.g4 | trxgrep ' //parserRuleSpec[*//TOKEN_REF/text()="EOF"]/RULE_REF' | trtext
                ;;
            1)
                echo ""
                echo *.g4
                trparse *.g4 | trxgrep ' //parserRuleSpec[*//TOKEN_REF/text()="EOF"]/RULE_REF' | trtext
                ;;
            2)  ;;
        esac
        popd > /dev/null 2>&1
    done

(Note, I had to fix a bug in Trash to get this to work.)

Surprisingly, there are many grammars that do not contain a start rule with EOF, e.g., acme, agc, alef, asm/*, asn, bcpl, calculator, dot, fasta, http, json, pascal, turing, xpath (not inclusive).

Augmenting the start rule with the EOF symbol at the end is important because it forces the parser to consume the entire file. See https://github.com/antlr/antlr4/issues/118. To illustrate the problem, modify the `calculator` grammar with a rule "file: expression;", and try running "parser.file();" on input "1 2". It succeeds, although it really should fail.

Alternatively, I could modify the templates for testing to check the state of the parse tree input against the lexer token stream to see whether the input is fully consumed by the parser.

kaby76:
Following up on [what we learned in the Scala grammar](https://github.com/antlr/grammars-v4/issues/2668), I wrote a new script to go over every grammar and find grammars without an EOF terminated start rule. The script then generates a parser, modifies the start rule of the grammar to terminate with an EOF, then redid the tests. This now forces the parser to consume the entire input in each test.

## Results
### Grammars that are missing EOF terminated start rules
```
[...]

PR #2675, opened 2022-06-22
Missing EOF in start rules

kaby76:
This PR is a fix for #2405 , where the start rules in the grammars don't end in an EOF. This is necessary for the input files to be consumed completely.


teverett:
@kaby76 thanks very much for this.

Issue #2682, opened 2022-06-24
[sparql] Symbol conflict "string"


kaby76:
As reported [here](https://github.com/antlr/antlr4/issues/3758), there's a symbol conflict in the Go target with "string". Looking over the Antlr4 tool code, [it's in the "reserved" word list](https://github.com/antlr/antlr4/blob/ee62e23da9669c20509984e0013664a851565e8f/tool/src/org/antlr/v4/codegen/target/GoTarget.java#L33), but it's not being escaped for whatever reason. In lieu of fixing the tool, which I can't do efficiently (not my code), I'm changing [the grammar from "string" to "string_"](https://github.com/antlr/grammars-v4/blob/7a17c2d437e9a6d2d7cf3c9193be41f1270657bc/sparql/Sparql.g4#L330) and moving on. We already have tons of these "_"-renamed symbols, and will rename back in due time.

PR #2683, opened 2022-06-24
Fix for #2682, sparql for Go.

kaby76:
Renamed "string" to "string_" and added the Go target.


kaby76:
@KvanTTT This PR is here for tactical reasons: sparql should work, and mostly does. But, the Antlr4 tool 4.10.1 doesn't seem to escape "string" even though it is listed as a keyword. So the tool needs to be fixed, too.

PR #2687, opened 2022-06-27
Fixed runtime variable declaration

SergeyRusskih:
This is the fix for the runtime variable declaration bug.

**Changes**:

- Removed "-" from the identifier and explicitly defined "minus minus" syntax
- Added the google truth as a dependency for unit tests

### DIFF
--- a/scss/pom.xml
+++ b/scss/pom.xml
@@ -9,14 +9,6 @@
 		<artifactId>grammarsv4</artifactId>
 		<version>1.0-SNAPSHOT</version>
 	</parent>
-    <dependencies>
-        <dependency>
-            <groupId>com.google.truth</groupId>
-            <artifactId>truth</artifactId>
-            <version>1.1.3</version>
-            <scope>test</scope>
-        </dependency>
-    </dependencies>
     <build>
 		<plugins>
 			<plugin>
--- a/scss/pom.xml
+++ b/scss/pom.xml
@@ -9,6 +9,14 @@
 		<artifactId>grammarsv4</artifactId>
 		<version>1.0-SNAPSHOT</version>
 	</parent>
+    <dependencies>
+        <dependency>
+            <groupId>com.google.truth</groupId>
+            <artifactId>truth</artifactId>
+            <version>1.1.3</version>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
     <build>
 		<plugins>
 			<plugin>
--- a/scss/pom.xml
+++ b/scss/pom.xml
@@ -9,7 +9,15 @@
 		<artifactId>grammarsv4</artifactId>
 		<version>1.0-SNAPSHOT</version>
 	</parent>
-	<build>
+    <dependencies>
+        <dependency>
+            <groupId>com.google.truth</groupId>
+            <artifactId>truth</artifactId>
+            <version>1.1.3</version>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+    <build>
 		<plugins>
 			<plugin>
 				<groupId>org.antlr</groupId>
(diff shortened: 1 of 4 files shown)


KvanTTT:
Something wrong with tests, please fix them: https://github.com/antlr/grammars-v4/runs/7077953267?check_suite_focus=true

SergeyRusskih:
On it!

scss/pom.xml
```diff
@@ -9,7 +9,15 @@
...
-	<build>
+    <dependencies>
+        <dependency>
+            <groupId>com.google.truth</groupId>
+            <artifactId>truth</artifactId>
+            <version>1.1.3</version>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+    <build>
```
KvanTTT, line 20: Why is it required?

SergeyRusskih: This is used for the unit tests. So, added this to be able to run them on local.

SergeyRusskih: Is it something that should be removed?

KvanTTT: What tests do you mean? We don't have this dependency for other grammars.

kaby76: There seems to be [non-standard tests](https://github.com/antlr/grammars-v4/tree/master/scss/testsrc/test/java) and a [gradle build file](https://github.com/antlr/grammars-v4/blob/master/scss/build.gradle) for the grammar. Is this run in the CI build? The non-standard tests look at the CST (don't know why .tree files didn't suffice), and sets the parser prediction mode https://github.com/antlr/grammars-v4/blob/1ec0abf0736a7624e3ca3fc62f50da1bf72ad435/scss/testsrc/test/java/TestBase.java#L56 . But it's also tested the usual way for C#, Java, Javascript.

SergeyRusskih: I'm happy with either leaving it or removing.

kaby76: How is the test code run? Maybe it can be just added to CI.

SergeyRusskih: It only runs on request, so I can run it on local but it is not executed as part of CI which is bad. I'm not sure if we could add it to CI because it's not a standard approach and I'd stick to the universal one. But it does provide some value when developing on local - for me it's more convenient to run the unit tests to verify particular parsing results than just run the existing tests that parse the examples (so I could verify the correctness, not only that we didn't have any errors).

kaby76: Tests in grammars-v4 run via Github Actions, which executes a [command at a Bash shell](https://github.com/antlr/grammars-v4/blob/1ec0abf0736a7624e3ca3fc62f50da1bf72ad435/.github/workflows/main.yml#L28). Can the tests be run from the command line (`mvnw clean test` or `java -jar junit-platform-console-standalone-1.8.2.jar .....`)? Does the prediction mode need to be changed? Otherwise, it makes more sense to move the tests over to files in the examples/ directory and keep with a standardized grammar with tests, no?

SergeyRusskih: - It can be run from the command line but I don't think we should (explanation below).
- No, don't think so. There is no good reason to do it.
- Yes, it does make sense and I'm working on it. I took over this grammar and the tests were already in place, I just made them work for the sake of caching any regression while I'm doing minor fixes (I need fixes now). In parallel, I'm working on refining the grammar - make it follow CSS3 standard, adding support of new features, adding testing examples.

I think we could do the following:
- Add the dependency as proposed (preferred) - so, I use them to catch regression while working on a better solution
- Do nothing (preferred) - remove added dependency, not a big deal. I'm working on proper testing anyway.
- Add the dependency + run them in CI (discouraged) - I think we need to stick to one testing approach
- Remove tests and add the proper examples instead (discouraged) - I'm reluctant to do it because I'll do double work in addition to grammar refactoring.

kaby76: I agree. It may be useful to integrate in some way later.

PR #2689, opened 2022-06-29
sqlite: add support for returning clause

thokra:
This PR adds support for [RETURNING](https://www.sqlite.org/syntax/returning-clause.html) in SQLite. 

When testing the example file, with test queries copied from the official SQLite source, it failed on `ON CONFLICT(a) DO UPDATE SET b=100`. Therefore, this PR also includes a change to  `upsert_clause` by changing `EQ` to `ASSIGN`.
Let me know if you want this as a separate PR.

I have no previous experience with ANTLR, so please let me know if there are any changes required (feel free to just modify the PR).


KvanTTT:
Thanks for the contribution and welcome to grammars-v4 community!

PR #2690, opened 2022-06-30
Remove RECORD from identifier

### LINKED ISSUE #2685: Java: record in classes are identified as methods
Consider the code:
public class A {
    record R1(int x) {

    }
}
It seems that R1 is identified as method instead of record

To fix it I modified memberDeclaration in JavaParser.g4 so recordDeclaration will be before methodDeclaration.

bkiers:
Removed the `RECORD` token from the `identifier` parser rule. Now the input:

```java
public class Record {
  record R1(int x) {
  }
}
```

will _not_ be parsed as:

Screenshot 2022-06-30 at 12 54 52

but as:

Screenshot 2022-06-30 at 12 54 32

Fixes #2685

### DIFF
--- a/java/java/JavaParser.g4
+++ b/java/java/JavaParser.g4
@@ -469,7 +469,6 @@ identifier
     | YIELD
     | SEALED
     | PERMITS
-    | RECORD
     | VAR
     ;
 
--- a/java/java/examples/Record.java
+++ b/java/java/examples/Record.java
@@ -1,0 +1,4 @@
+public class Record {
+  record R1(int x) {
+  }
+}


KvanTTT:
There are some errors on CI: https://github.com/antlr/grammars-v4/runs/7129951772?check_suite_focus=true#step:17:45

bkiers:
> There are likely similar errors in `identifier` with other keywords. The JLS states "Identifier: [IdentifierChars](https://docs.oracle.com/javase/specs/jls/se18/html/jls-3.html#jls-IdentifierChars) but not a [Keyword](https://docs.oracle.com/javase/specs/jls/se18/html/jls-3.html#jls-Keyword) or [BooleanLiteral](https://docs.oracle.com/javase/specs/jls/se18/html/jls-3.html#jls-BooleanLiteral) or [NullLiteral](https://docs.oracle.com/javase/specs/jls/se18/html/jls-3.html#jls-NullLiteral)". But, we see for [keywords](https://docs.oracle.com/javase/specs/jls/se18/html/jls-3.html#jls-Keyword) that OPENS and REQUIRES are contextual, just as RECORD.

That could very well be, but I don't know since which Java version certain keywords are being used and when not being allowed as identifiers. I'll leave them out of this PR.

KvanTTT:
I don't think the fix is correct. The following code is valid in Java 17:

```java
public class Java {
    void test() {
        var record = 5;
    }

    public static class Record {
        record R1(int x) {
        }
    }
}
```

Both identifier `record` and `record` class are valid.

KvanTTT (changes requested):
I don't think the fix is correct. The following code is valid in Java 17:

```java
public class Java {
    void test() {
        var record = 5;
    }

    public static class Record {
        record R1(int x) {
        }
    }
}
```

Both identifier `record` and `record` class are valid.

KvanTTT:
It looks like a context-dependent keyword, it should be processed in another way.

PR #2693, opened 2022-07-02
SQL/T-SQL: no longer requires special case treatment

pguyot:
Updated README and Test class.


KvanTTT:
Thanks for the contribution and welcome to grammars-v4 community!

PR #2698, opened 2022-07-03
[TSql] Synchronize keywords

pguyot:
Keywords in TSqlLexer.g4 were not in the same order in TSqlParser.g4 keyword rule.
Some were missing. Some in the list are not keywords that can be used as ids.
Please note that I did leave ADD in keyword grammar rule, because it was added explicitely by https://github.com/antlr/grammars-v4/commit/d9024d8df1c910daf08d2972f7a2289d6ebcc74b
But this keyword cannot be used as an id within current T-SQL grammar as documented by Microsoft.
https://docs.microsoft.com/en-us/sql/t-sql/language-elements/reserved-keywords-transact-sql?view=sql-server-ver16


sql/tsql/TSqlLexer.g4
```diff
@@ -379,7 +376,6 @@ GB:                                    'GB';
 GETDATE:                               'GETDATE';
 GETUTCDATE:                            'GETUTCDATE';
 GLOBAL:                                'GLOBAL';
-GO_BATCH:                              [\r\n]* [\t ]* 'GO'[\t\n\r ];
```
KvanTTT, line 382: Thanks for removing this stuff! It looks suspicious.

sql/tsql/TSqlLexer.g4
```diff
@@ -267,7 +266,6 @@ BACKUP_PRIORITY:                       'BACKUP_PRIORITY';
 BEGIN_DIALOG:                          'BEGIN_DIALOG';
 BIGINT:                                'BIGINT';
 BINARY_KEYWORD:                        'BINARY';
-BINARY_BASE64:                         'BINARY BASE64';
```
KvanTTT, line 270: Also, it looks weird, thanks!

sql/tsql/TSqlLexer.g4
```diff
@@ -893,6 +888,13 @@ XACT_ABORT:                            'XACT_ABORT';
 VARCHAR:                               'VARCHAR';
 NVARCHAR:                              'NVARCHAR';
 
+//Combinations that cannot be used as IDs
+BINARY_BASE64:                         'BINARY BASE64';
```
KvanTTT, line 892: Can it be split into two keywords?

sql/tsql/TSqlLexer.g4
```diff
@@ -893,6 +888,13 @@ XACT_ABORT:                            'XACT_ABORT';
...
+DISK_DRIVE:                            [A-Z][:];
+GO_BATCH:                              [\r\n]* [\t ]* 'GO'[\t\n\r ];
```
KvanTTT, line 894: It looks weird. Probably it's better to check if `GO` starts with new line using semantic predicates. Also, it may affect performance.

sql/tsql/TSqlLexer.g4
```diff
@@ -893,6 +888,13 @@ XACT_ABORT:                            'XACT_ABORT';
...
+DOLLAR_ACTION:                         '$ACTION';
+EVENTDATA:                             'EVENTDATA' '(' ')';
```
KvanTTT, line 896: Also, can it be split into several tokens? `EVENTDATA` and parentheses.

sql/tsql/TSqlLexer.g4
pguyot: I agree. Of course this was there before, I only moved it, but I will check the semantic predicate to clean that up. Thanks!

pguyot: Actually, I did simplify the logic around go and got rid of this token.

KvanTTT:
Thanks! It looks much better!

PR #2702, opened 2022-07-06
Tsql additional Sybase (legacy) support

olebhansen:
I'm parsing through some of our Sybase ASE DDL-files, and patched the grammar to match what we are presumably able to feed into our DB instance. 

I hope I can do additional updates when I get to triggers, functions and the rest of our procedures... Not sure if you prefer 1 PR for each feature, or one feature = one comit, but multiple features in 1 PR (i.e. like I'm doing now). Happy to take input on how to go about this, as it is my first PR here :-)

### DIFF
--- a/sql/tsql/TSqlParser.g4
+++ b/sql/tsql/TSqlParser.g4
@@ -1959,7 +1959,7 @@ create_xml_index
 
 // https://msdn.microsoft.com/en-us/library/ms187926(v=sql.120).aspx
 create_or_alter_procedure
-    : ((CREATE (OR ALTER| OR REPLACE)?) | ALTER) proc=(PROC | PROCEDURE) procName=func_proc_name_schema (';' DECIMAL)?
+    : ((CREATE (OR (ALTER | REPLACE))?) | ALTER) proc=(PROC | PROCEDURE) procName=func_proc_name_schema (';' DECIMAL)?
       ('('? procedure_param (',' procedure_param)* ')'?)?
       (WITH procedure_option (',' procedure_option)*)?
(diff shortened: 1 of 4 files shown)


sql/tsql/TSqlParser.g4
```diff
@@ -1959,7 +1959,7 @@ create_xml_index
 
 // https://msdn.microsoft.com/en-us/library/ms187926(v=sql.120).aspx
 create_or_alter_procedure
-    : ((CREATE (OR ALTER)?) | ALTER) proc=(PROC | PROCEDURE) procName=func_proc_name_schema (';' DECIMAL)?
+    : ((CREATE (OR ALTER| OR REPLACE)?) | ALTER) proc=(PROC | PROCEDURE) procName=func_proc_name_schema (';' DECIMAL)?
```
KvanTTT, line 1962:
```suggestion
    : ((CREATE OR (ALTER | REPLACE)?) | ALTER) proc=(PROC | PROCEDURE) procName=func_proc_name_schema (';' DECIMAL)?
```

olebhansen: I think this will break a regular "create procedure". But "((CREATE (OR ( ALTER| REPLACE))?) | ALTER) proc..." is perhaps even better.

KvanTTT: Yes, sorry, you are right.

olebhansen: I did an update in [b9b2747](https://github.com/antlr/grammars-v4/pull/2702/commits/b9b27477e72d886cbbde677439f98aecab3ce8a0) that should address both concerns :-)

PR #2712, opened 2022-07-13
[JavaScript/Python3 target] Fix for #2703

### LINKED ISSUE #2703: python target, JavaScript lexer
in JavaScriptLexer.py, that I've got after using Java 18 and antlr4.10,1 lib and JavaScriptLexer.g4 which I've got here: https://github.com/antlr/grammars-v4/tree/master/javascript/javascript and then transformed using transformGrammar.py from https://github.com/antlr/grammars-v4/tree/master/javascript/javascript/Python3 :
there were this.IsStrictMode()
instead of self.IsStrictMode()

this issue was in 12 functions.

kaby76:
This change fixes #2703 , adds "preprocessing" to grammars using "python3 transformGrammar.py" and adds the script to the build. Until we have "target-agnostic grammars", we are stuck in hell. But, at least this change will force people to consider what they do when modifying a grammar while not considering all targets.

### DIFF
--- a/_scripts/regtest.sh
+++ b/_scripts/regtest.sh
@@ -201,6 +201,10 @@ part2()
     # 2) Build driver code.
     echo "Building."
     date
+    if [[ -f transformGrammar.py ]]
+    then
+        python3 transformGrammar.py
+    fi
     case "$target" in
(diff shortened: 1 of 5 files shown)


teverett:
@kaby76  thanks

PR #2716, opened 2022-07-15
Support for `TRIM` function has been added.

scrappyCoco:
Support for built in function `TRIM` has been added:
https://docs.microsoft.com/ru-ru/sql/t-sql/functions/trim-transact-sql?view=sql-server-ver16


sql/tsql/TSqlLexer.g4
```diff
@@ -642,6 +642,7 @@ TINYINT:                               'TINYINT';
 TORN_PAGE_DETECTION:                   'TORN_PAGE_DETECTION';
 TRACKING:                              'TRACKING';
 TRANSFORM_NOISE_WORDS:                 'TRANSFORM_NOISE_WORDS';
+TRIM:                                  'TRIM';
```
KvanTTT, line 645: Should it be also added to the [`keyword`](https://github.com/antlr/grammars-v4/blob/master/sql/tsql/TSqlParser.g4#L4211) rule?

scrappyCoco: For most functions in `built_in_functions` names are presented in keywords. For an example: `STRING_AGG` is a keyword, `REPLACE` is also the keyword and so on. As I understand, `TRIM` also should be added  to the keywords list for similarity. Or exists a better solution?

scrappyCoco: Thanks!
Sorry, I was inattentive.

KvanTTT (approved):

PR #2720, opened 2022-07-20
[golang/ grammar, Python3 target] Adding the Python3 target.

### LINKED ISSUE #2719: [golang/ grammar, Python3 target] Port the Go grammar to Python3
This [Stackoverflow question]() pointed out that the Python3 port for the Go grammar doesn't exist. It's an easy port.

kaby76:
This PR fixes #2719 where the Python3 port for the Go grammar is missing. It was mentioned in [this Stackoverflow problem](https://stackoverflow.com/questions/73050657/antlr4-golang-grammar-file). The port was easy, but because there's no "target-agnostic format" for grammars, I rewrote the grammar slightly so the base class method is called with an object reference ("this."). For the Python3 port, run "transformGrammar.py".

Since the port works, I added it to the build.

### DIFF
--- a/_scripts/skip-python3.txt
+++ b/_scripts/skip-python3.txt
@@ -23,7 +23,6 @@ fortran77
 gdscript
 gff3
 gml
-golang
 haskell
 html
 hypertalk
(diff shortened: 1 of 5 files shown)


teverett:
@kaby76 thanks!

PR #2728, opened 2022-07-24
Fortify Lua grammar to fix related parse issues

matthargett:
The twin of https://github.com/pmd/pmd/pull/4061

Found parse issues while trying to use CPD with a Lua codebase. Took the opportunity to widen the regression test (added in file comments) and group together several small fixes that now allow hundreds of KLOC of open source Lua to lex/parse without issues.


teverett:
@matthargett thanks!

Issue #2730, opened 2022-07-27
Python Runtime incompatibilities with PHP Grammar 


dkazantzas:
There are multiple occurences of actions and predicates such as the following:
https://github.com/antlr/grammars-v4/blob/0c9a11ea096dd182862dac03f25766cd5546ff95/php/PhpLexer.g4#L47
https://github.com/antlr/grammars-v4/blob/0c9a11ea096dd182862dac03f25766cd5546ff95/php/PhpLexer.g4#L63

When the Python files are generated, they result in cases such as the following:
```
def Shebang_sempred(self, localctx:RuleContext, predIndex:int):                                                                                                                     
    if predIndex == 0:                                                                                                                                                          
        return  this.IsNewLineOrStart(-2)   
```

However, as the `self` keyword is used instead of `this`, the result is Python throwing a name error.

Not sure what the correct solution to this issue would be. At the very least, a warning should be added to the `README.md`, so that people that wish to use the Python runtimes don't have this issue.

kaby76:
This PHP grammar is a mess. The solution is to follow exactly what I did here: https://github.com/antlr/grammars-v4/commit/4794c0c5371ce6b0711a306256f8f7706394ec6d , adding transformGrammar.py, a new Readme.md, etc. In this PR, I added instructions to run "python3 transformGrammar.py". That will modify the grammar for Python3. There is no universal "object reference call to method" across all targets. The only viable solution is to standardize on "this." object reference calls, and modify them to produce a grammar that works with the target. Could you add this code please?

dkazantzas:
Good idea. I've added the file. This will save help users automate the whole process

kaby76:
@dkazantzas Thank you so much! I'll peek at the PR to make sure it's working.

PR #2731, opened 2022-07-27
Updated PHP Grammar Documentation about Python Runtimes

dkazantzas:
Added a section in the `README.md` of the PHP Grammar so that users can address the issue that is reported in #2730


dkazantzas:
I've followed the suggestion by @kaby76 in the discussion of #2730 and added a script to transform the Grammar

kaby76:
Please also remove [the "php" line from the skip-python3.txt file](https://github.com/dkazantzas/grammars-v4/blob/4d1a0dc07d8c05e14df681cd10a7034bc74650a2/_scripts/skip-python3.txt#L63). The file tells the build script to "skip" doing the build for the "php/" grammar. When the line is removed, the build will attempt to run regression tests on the grammar.

dkazantzas:
Ok done.

kaby76:
Almost there.

* The directory php/Python/ should be renamed php/Python3/. There really isn't a "Python" target, but either "Python2" or "Python3". The build script checks for the directory in the source, then copies that code for that target in with the rest of the code (where the .g4's reside). After renaming, it fails with several other errors.
* There are two syntax errors in PhpLexer.py. To correct, update transformGrammar.py with the a few lines to change "true" to "True". Context is added because there is a grammar option with "true" that shouldn't be changed.
```
$ diff dkazantzas/php/Python/transformGrammar.py dkazantzas2/php/Python/
30,33d29
<         if '_scriptTag = true;' in x:
<             x = x.replace('_scriptTag = true;', '_scriptTag = True;')
<         if '_styleTag = true;' in x:
<             x = x.replace('_styleTag = true;', '_styleTag = True;')
```
* The PhpParserBase.py file looks like it always contained errors, which wouldn't be detected because the compiler doesn't flag it until executed unless you run the test suite in examples/.
```
$ diff dkazantzas/php/Python/PhpLexerBase.py  dkazantzas2/php/Python/
121c121
<             self._channel = self.SKIP
---
>             self.channel(SkipChannel)
```

dkazantzas:
Made those changes too. I really appreciate your help in this.

teverett:
@kaby76  thanks.  Let's fix the Dart issue before it gets pulled.

kaby76:
> @kaby76 thanks. Let's fix the Dart issue before it gets pulled.

@teverett 

I created a Github Issue for this https://github.com/antlr/grammars-v4/issues/2743

As I explained there, we've been getting a "Connection closed before full header was received" on Window and Macos. I think it's because the network connection for the Github Actions servers or (more likely) the Dart servers are boxes of unreliable junk. It could be both.

To work around this problem I will submit a separate PR to retry the "dart pub get" until the return code is zero or some could like 10x's.

Could you invoke the action for this PR again in order to test the non-determinism of the machines?

teverett:
@kaby76  yup!

kaby76:
@teverett  Thanks. It happened again. I'll check it out locally on my machines to see what I can do.

kaby76:
@teverett I created a PR to wrap the "dart pub get" in a loop to repeat on error. It doesn't hurt the build, may help in these dart connection drops, but who knows. Once merged, maybe this PR can be rebased and checked?

teverett:
@github rebase

kaby76:
@dkazantzas Please fast forward your repo so it can be rebuilt.

dkazantzas:
@kaby76 Done

kaby76:
@dkazantzas That looks good. @teverett Build? Let's see if this works.

kaby76:
@dkazantzas Please fast forward your repo so it can be rebuilt with the latest fixes for Go builds. Thanks.

dkazantzas:
@kaby76 I've synced my repo, it should now be ok

kaby76:
@teverett This is ready to rebuild. Could you set a build going? I think we should no longer have Dart and Go build errors, but let's just see. Thanks.

teverett:
@kaby76  done, thanks for the reminder.

kaby76:
@teverett Works perfectly.

teverett:
@dkazantzas  @kaby76  thanks

Issue #2734, opened 2022-08-01
[python/python grammar | all targets] Base class creates illegal token LINE_BREAK and WS!!


kaby76:
I have a Python3 file with this code that I want to rewrite using Piggy:
```
# https://stackoverflow.com/q/73190865/4779853

a = 1
result = 'even' if a % 2 == 0 else 'odd'
print(result)

a = 2
result = 'even' if a % 2 == 0 else 'odd'
print(result)

```
The exact contents of the file in hex and ascii using "od" is:
```
$ od -t c -t x1 condition.py
0000000   #       h   t   t   p   s   :   /   /   s   t   a   c   k   o
         23  20  68  74  74  70  73  3a  2f  2f  73  74  61  63  6b  6f
0000020   v   e   r   f   l   o   w   .   c   o   m   /   q   /   7   3
         76  65  72  66  6c  6f  77  2e  63  6f  6d  2f  71  2f  37  33
0000040   1   9   0   8   6   5   /   4   7   7   9   8   5   3  \r  \n
         31  39  30  38  36  35  2f  34  37  37  39  38  35  33  0d  0a
0000060  \r  \n   a       =       1  \r  \n   r   e   s   u   l   t
         0d  0a  61  20  3d  20  31  0d  0a  72  65  73  75  6c  74  20
0000100   =       '   e   v   e   n   '       i   f       a       %
         3d  20  27  65  76  65  6e  27  20  69  66  20  61  20  25  20
0000120   2       =   =       0       e   l   s   e       '   o   d   d
         32  20  3d  3d  20  30  20  65  6c  73  65  20  27  6f  64  64
0000140   '  \r  \n   p   r   i   n   t   (   r   e   s   u   l   t   )
         27  0d  0a  70  72  69  6e  74  28  72  65  73  75  6c  74  29
0000160  \r  \n  \r  \n   a       =       2  \r  \n   r   e   s   u   l
         0d  0a  0d  0a  61  20  3d  20  32  0d  0a  72  65  73  75  6c
0000200   t       =       '   e   v   e   n   '       i   f       a
         74  20  3d  20  27  65  76  65  6e  27  20  69  66  20  61  20
0000220   %       2       =   =       0       e   l   s   e       '   o
         25  20  32  20  3d  3d  20  30  20  65  6c  73  65  20  27  6f
0000240   d   d   '  \r  \n   p   r   i   n   t   (   r   e   s   u   l
[...]

PR #2739, opened 2022-08-03
[python/python grammar, all targets] Fixes for #2734, and add Cpp target for python/python/ grammar!!

kaby76:
* Fixed token creation issues that occur across all targets. The main problem here was that the stop index for a token was "off by one". So, [you now see "-1" added to the stop index in the char stream](https://github.com/antlr/grammars-v4/pull/2739/files#diff-8d5542a2ad6a6fffc037278d648a42f793f531175257db4dccaf1524b7117e6f), wherever CommonToken are created.
* While I was here, I decided to add the Cpp target for the grammar! To get this to work, I changed the semantic predicate method calls in the grammar with an explicit "this." call. I then added a transformGrammar.py script to modify the grammar for the Cpp target so "this." is changed to "this->" for C++. One other rather annoying fact is that Antlr does not generate a '#include' for base classes. I added comments in the grammars where transformGrammar.py needs to add `@header` code. Note, the use of `std::unique_ptr<Token>` for tokens in the Antlr API was a royal pain to get used to, and really gets in the way of creating a buffer of tokens. I'm not sure why the API uses `std::unique_ptr<Token>` instead of `std::shared_ptr<Token>`. I had to create copies of CommonToken objects and add `std::move()` in many places. In any case, the port works wonderfully!
* Fixed bugs in calling transformGrammar.sh from _scripts/regtest.sh, where it wasn't really being called. Instead, I pushed the "python3 transformGrammar.py" to all the per-target makefiles.
* The rest of the python/* grammars are left as is, but now that we have a formula to get all targets working, these should be cleaned up at some point!!


kaby76:
Converting back to draft because I'm going to add the Python3 target.

kaby76:
Never mind. It's already there!

PR ready for review.

teverett:
@kaby76  thanks!

PR #2760, opened 2022-08-16
[rexx] Parse hexstrings and binarystrings correctly per ANSI standard.

### LINKED ISSUE #2741: [rexx] Hexstrings and binarystrings aren't parsed correctly
Rexx has support for string literals in hexadecimal and binary.  Hexstrings are, per ISO standard section 6.2.2.38, strings of traditional hex digits without regard to case, with optional blanks between pairs of digits (counting from the right), with an upper or lower case `X` appended after the closing quote (_e.g._, `"1234"x`, `'12 34'X`,).  Similarly, binarystrings are, per section 6.2.2.41, strings of zeroes and ones with optional blanks between quadtuples of digits, with an upper or lower case `B` appended after the closing quote (_e.g._, `"10100001"B`, `'1010 0001'b`.

The rexx grammar collapses those into the `STRING` token type, and accepts invalid values (_e.g._,  `"1234"B`, `'12 3'x`).  The grammar should represent hex and binary strings separately from "normal" strings, and should not accept invalid forms.

I am working on a change to implement correct parsing for hex and binary strings.  It will also do so in pure ANTLR, unlike the current implementation, which embeds Java code in a lexer action.

RossPatterson:
Parse hexstrings and binarystrings correctly per ANSI standard.

Fixes #2741.

* Identify hexstrings and binarystrings in the parse tree by their own parser rules.
* Do not accept hexstrings containing invalid hexadecimal characters.
* Do not accept binarystrings containing invalid binary characters.
* Accept blanks between pairs of hexadecimal digits, counting from the right (ANSI sec. 6.2.2.38 _et seq._).
* Accept blanks between 4-tuples of binary digits, counting from the right (ANSI sec. 6.2.2.41 _et seq._).

In addition, replace the existing Java lexer action that identifies the trailing radix ("b", "B", "x", or "X")
with pure-ANTLR lexer rules, to remove one of the only barriers to a target-agnostic grammar.

This change adds a new `test_bitstring.rex` set of test cases.  There are "good" and "bad" cases in this set, obvious by their names (_e.g._, `binary_good1` and `hex_bad2`).  The "good" cases should parse to either `binary_string` or `hex_string`.  The "bad" cases have different parsing interpretations, because they're invalid binary- or hex-strings, but they're valid as some other expression.  There are also a few cases that are obviously not binary- or hex-strings (_e.g._, `str_good3`), and that should never parse as such (_e.g._, `"hi mom!"x`).  `mvn test` will compare the parse of this file to a reference file (`test_bitstring.rexx.tree`), which has the expected parsing results.

Passes `mvn test`, including the new test cases.


RossPatterson:
Apparently there are linend issues in the test-case-verification stuff.  I'll sort that out and push an update, probably tomorrow.

RossPatterson:
I'll reorganize and resubmit.

RossPatterson:
Replaced by PR #2769.

PR #2761, opened 2022-08-16
[Hive grammer] Fix the priority problem of 'BETWEEN AND' and 'AND' 

jiefei30:
Fix the priority problem of 'BETWEEN AND' expression and 'AND' logical operator.
For example:
```sql
SELECT a FROM t1 WHERE a BETWEEN 1*2 AND 10 AND b IN (SELECT b FROM t2) AND c > 1
```
the old parse tree:

It can be clearly seen that the priorities of 'BETWEEN AND' expression and 'AND' logical operator are wrong, which may lead to errors in analysis

the new parse tree:


teverett:
@jiefei30 thanks!

PR #2762, opened 2022-08-16
Fix for issue #2749 -- "go get" connection link errors

kaby76:
* Changes to _scripts/regtest.sh so we don't uninstall/reinstall Trash tools. It's wasted bandwidth.
* Add a loop around "go get" to repeat 5 times or fail.
* In response to https://github.com/antlr/grammars-v4/pull/2731 build issues.


teverett:
@kaby76  thanks!

PR #2769, opened 2022-08-17
[rexx] Parse hexstrings and binarystrings correctly per ANSI standard.

RossPatterson:
Parse hexstrings and binarystrings correctly per ANSI standard.

Fixes https://github.com/antlr/grammars-v4/issues/2741.

* Identify hexstrings and binarystrings in the parse tree by their own parser rules.
* Do not accept hexstrings containing invalid hexadecimal characters.
* Do not accept binarystrings containing invalid binary characters.
* Accept blanks between pairs of hexadecimal digits, counting from the right (ANSI sec. 6.2.2.38 _et seq._).
* Accept blanks between 4-tuples of binary digits, counting from the right (ANSI sec. 6.2.2.41 _et seq._).

In addition, replace the existing Java lexer action that identifies the trailing radix ("b", "B", "x", or "X")
with pure-ANTLR lexer rules, to remove one of the only barriers to a target-agnostic grammar.

This change adds a new `test_bitstring.rex` set of test cases. There are "good" and "bad" cases in this set, obvious by their names (_e.g._, `binary_good1` and `hex_bad2`). The "good" cases should parse to either `binary_string` or `hex_string`. The "bad" cases have different parsing interpretations, because they're invalid binary- or hex-strings, but they're valid as some other expression. There are also a few cases that are obviously not binary- or hex-strings (_e.g._, `str_good3`), and that should never parse as such (_e.g._, `"hi mom!"x`). `mvn test` will compare the parse of this file to a reference file (`test_bitstring.rexx.tree`), which has the expected parsing results.

Tests that need to have their parse trees verified should be named `whatever` placed in the `examples` directory, along with a corresponding `whatever.tree` reference file created by `grun Rexx file_ -tree whatever > whatever.tree`.  All files in the `examples` directory will have their linends normalized to LF before parsing and verification, since Git checks out files on Windows machines with CRLF linends.

**NOTE:** If you create a reference file on Windows, it will contain `(delim \r\n)` sequences.  Replace all of these with `(delim \n)` before adding the file to the `examples` directory.

Passes `mvn test`, including the new test cases.


teverett:
@RossPatterson  thanks!

PR #2770, opened 2022-08-18
[rexx] Add Python3 target support

### LINKED ISSUE #2747: [rexx] Add Python target support
The Rexx parser contains a small amount of Java code in a couple of semating predicates.  There does not appear to be a way to accomplish the parse without the predicates.  I would like to build a Python target, as I'm more comfortable working in Python than in Java.

I am working on a change to export the Java code to a parser base class, and to create a similar Python base class.  I would prefer to make the grammar target-agnosftic, but it does not seem possible, so I'll reluctantly settle for multi-target.

@kaby76 - If your bag of tricks has something that might help me go fully agnostic, I'd love to hear about it.

RossPatterson:
Fixes #2747.

### DIFF
--- a/rexx/Java/RexxParserBase.java
+++ b/rexx/Java/RexxParserBase.java
@@ -1,0 +1,20 @@
+import org.antlr.v4.runtime.*;
+
+public abstract class RexxParserBase extends Parser
+{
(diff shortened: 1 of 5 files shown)


teverett:
@RossPatterson thanks!

PR #2773, opened 2022-08-18
Support PLSQL JSON-based CHECK constraints

Naros:


teverett:
@KvanTTT  good to pull this?

Issue #2757, opened 2022-08-15
[sql/postgresql] Parser prints parse errors for "case.sql" (and others), but passes testers!!


kaby76:
This is pretty weird. [sql/postgresql](https://github.com/antlr/grammars-v4/tree/master/sql/postgresql) produces errors for [case.sql](https://github.com/antlr/grammars-v4/blob/master/sql/postgresql/examples/case.sql), but it is not flagged in the test results. Note, it would skip a bad parse if an ".error" file exists, but in this case, no case.sql.errors file exists. The parser flags a parse error for the Maven tester and regtest.sh. So, something is wrong in the Antlr runtime. It should not be ignoring a parse error.

kaby76:
* It seems it's only in Java: `cd sql/postgresql; mvn clean test; bash ../../_scripts Java`. There isn't any errors for C#. Very strange.

kaby76:
There are likely two problems here.

First, for the Java target, I'm getting "mismatched input" parse errors. A second problem is that the "syntax error" count associated with the Parser class does not get incremented. In other words, "mismatched input" parser errors are not counted as parse errors correctly even when they are reported. That seems contradictory.

* The grammar is [here](https://github.com/antlr/grammars-v4/blob/9f1d0ba046aedd88fb4cbc96355790e7b0b341ad/sql/postgresql). The input is [here](https://github.com/antlr/grammars-v4/blob/9f1d0ba046aedd88fb4cbc96355790e7b0b341ad/sql/postgresql/examples/case.sql).
* The driver for Java is similar to that for CSharp, produced by Trash's trgen. I replaced the .jar reference for the runtime in Java with a copy of the current Java runtime 4.10.1.

* DefaultErrorStrategy is used in both Java and CSharp.

* The Java code to open the file is `CharStreams.fromFileName(file_name)`. In CSharp, the file open code is `CharStreams.fromPath(file_name)`.

* The output from both parsers is here: [out-java.txt](https://github.com/antlr/antlr4/files/9363605/out-java.txt)  [out-csharp.txt](https://github.com/antlr/antlr4/files/9363609/out-csharp.txt)

* The tokens recognized for Java and CSharp are identical.

* In Java, for sql/postgresql, input case.sql, there is an error that is raised during the parse.

```
[...]

PR #2774, opened 2022-08-19
[Postgresql, Java and CSharp] Fix for #2757

kaby76:
* Replaced queuing of error messages--which didn't work--with delegated error messages to the main parser error listeners.
* Cleaned up parser base code by using "abstract".
* Fixed error in Java code TrimQuotes().


teverett:
@kaby76 thanks

PR #2775, opened 2022-08-20
[PromQL] support group modifiers with no labels

gfrolov:
PromQL supports group modifiers (`group_left`, `group_right`) with or without label lists.

Example:

`method_code:http_errors:rate5m / ignoring(code) group_left method:http_requests:rate5m`

Reference -- https://prometheus.io/docs/prometheus/latest/querying/operators/#group-modifiers


teverett:
@gfrolov thanks!

PR #2780, opened 2022-08-23
TSql: fix grammar for hints

pguyot:
Table aliases were parsed as hints. Maintained some compatibility
with sybase legacy hints grammar (as we have an example), however
they may be parsed as table aliases now (which SQL Server does)


sql/tsql/TSqlParser.g4
```diff
@@ -3716,8 +3715,8 @@ table_source_item_joined
     ;
 
 table_source_item
-    : table_name_with_hint        as_table_alias?
-    | full_table_name             as_table_alias?
+    : full_table_name             deprecated_table_hint as_table_alias // this is currently allowed
+    | full_table_name             as_table_alias? ( with_table_hints? | deprecated_table_hint? | sybase_legacy_hints? )
```
KvanTTT, line 3719: It can be simplified:

```suggestion
    | full_table_name             as_table_alias? (with_table_hints | deprecated_table_hint | sybase_legacy_hints)?
```

Issue #2768, opened 2022-08-17
Error messages are not the same between Windows and other systems


pguyot:
Referring to this build:

https://github.com/antlr/grammars-v4/runs/7878016714?check_suite_focus=true

The test consists in parsing this statement with sql/tsql grammar:

```
select fillfactor from (select 1 as fillfactor) t;
```

This shouldn't work as fillfactor is a reserved keyword. On every platform, it fails. However, the second error message is different on Windows and on other platforms, as if it was not deterministic.

On Windows, it is:
```
line 1:36 mismatched input 'fillfactor' expecting {'ADD', 'PRECISION', 'ABORT', 'ABSO ...
```

On every other platform, it is:
```
line 1:36 no viable alternative at input 'selectfillfactor'
```

This was introduced by the commit in the PR.

PR #2781, opened 2022-08-23
TSql: Fix grammar for xml functions

pguyot:
XML functions are called on columns which may be specified with a table name.


pguyot:
(will rebase once #2780 is merged)

pguyot:
CI didn't pass. Should we revert it until test is fixed?

KvanTTT:
Yes. Sorry, I thought the fail doesn't relate to your changes.

pguyot:
This is bug #2768 : on Windows and on Linux, we have different trees. I will try to isolate a small case to debug that.

pguyot:
Fix is in #2784

PR #2782, opened 2022-08-23
added python-3.8.12 and removed python3-without-actions

RobEin:
Please remove the old python3-without-actions folder.


kaby76:
* You'll need to delete lines like this: https://github.com/RobEin/grammars-v4/blob/0e64067842e7d5e3a4391f2a35660f6b948511e3/_scripts/skip-cpp.txt#L167 . The "skip" files tell the build not to work on these grammars. But, there's a bug where the build fails if it can't find the skipped directory. You'll have to delete these lines that mention the "-without-actions" grammar for all targets.
* This grammar doesn't work with Java 11, but does with v18. We don't have "per-grammar" requirements, but we should.

RobEin:
I think you mean "this grammar doesn't work with Java" 18...
Thanks.
As soon as possible, I will improve these.

PR #2785, opened 2022-08-28
upgraded to October 2021 specs

innomon:
For changes in specs see:
https://github.com/graphql/graphql-spec/blob/main/changelogs/October2021.md#specgraphqlmd 
 and
https://spec.graphql.org/October2021/


teverett:
@innomon thanks!

PR #2790, opened 2022-08-31
Closes #2789

### LINKED ISSUE #2789: VBA Grammar - line continuation for method / property calls broken.
Line continuation for method or property chaining only parses correctly when the 'dot' is on the new line. The below are all valid VBA syntax but only the first `With` block parses without error.

```vb
With A _
    .B
End With

With A. _
    B
End With

With A _
    . _
    B
End With
```

SSlinky:
Fix to enable all styles of method chaining when continued over multiple lines.


teverett:
@SSlinky thanks!

PR #2802, opened 2022-09-07
Updates for 4.11.1

kaby76:


kaby76:
Swift/swift3 crashing in get(), [here](https://github.com/antlr/grammars-v4/blob/a167cc13f2b61ff5d730020d526a56e12df55d02/swift/swift3/src/main/java/SwiftSupport.java#L217-#L221). It should not be calling get(), but LA() or LT(). [This code, too, is wrong](https://github.com/antlr/grammars-v4/blob/a167cc13f2b61ff5d730020d526a56e12df55d02/swift/swift3/src/main/java/SwiftSupport.java#L215)!

kaby76:
asm8080 and asmZ80 do not pass now under Antlr 4.11.1 for Go. Unclear why except that the code does not complete in 5 minutes, and trwdog kills the process (correctly doing so because otherwise the build would hang).

KvanTTT:
Probably we should address @jimidle

teverett:
@kaby76  @KvanTTT  ready to merge this?

kaby76:
It's ready to merge.

teverett:
@kaby76  thanks!

PR #2807, opened 2022-09-09
Fix problems with asm8080 and asmZ80 grammars

kaby76:
This PR fixes a few problems with the Go target, which were noticed with https://github.com/antlr/grammars-v4/pull/2802

* Fixed asm8080 and asmZ80 grammars. COMMENT was marked "skip", but this would cause massive slowdowns in Go target.
* Fixed Haskell grammar. Update the StackQueue.cs data structure, which I copied from [Antlr4BuildTasks](https://github.com/kaby76/Antlr4BuildTasks/blob/master/Antlr4BuildTasks/Util/StackQueue.cs). The previous version had bugs, and was overly complicated. 
* Fixed template for Go. This referenced the old 4.10.1 runtime for Go. The 4.11.1 Go runtime is under the subdirectory "v4".

### DIFF
--- a/_scripts/skip-cpp.txt
+++ b/_scripts/skip-cpp.txt
@@ -10,10 +10,8 @@ apex
 apt
 arithmetic
 asm/asm6502
-asm/asm8080
 asm/asm8086
 asm/asmMASM
-asm/asmZ80
 asm/masm
 asm/pdp7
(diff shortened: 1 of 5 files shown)


teverett:
@kaby76  thanks!

PR #2814, opened 2022-09-11
Add Python3 target of python/python3 grammar to build

kaby76:
PR https://github.com/antlr/grammars-v4/pull/2811 was a good addition. The only thing missing was that the grammar wasn't integrated into the build. This is important because it helps verify that Antlr4 works across targets. This PR changes the code slightly so that it can be tested (renames Python to Python3, the name of the port target; adds transformGrammar.py so "this" gets changed to "self"; and a readme.md to explain how to use the port). The parser works fine, albeit it is kind of slow. I think this may be because of ambiguities in the grammar. This should be investigated at some point. Fixes https://github…com/antlr/grammars-v4/issues/2813

### DIFF
--- a/_scripts/skip-python3.txt
+++ b/_scripts/skip-python3.txt
@@ -65,7 +65,6 @@ ply
 powerbuilder
 protobuf3
 python/python2
-python/python3
 python/python3-py
 python/python3-ts
 python/python3-without-actions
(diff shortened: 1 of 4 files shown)


teverett:
@kaby76 thanks

PR #2825, opened 2022-09-16
[rexx] Set EOLs in test files in a way that works for both Maven and trgen.

### LINKED ISSUE #2821: [rexx] Generated parser isn't being tested in CI builds
In [ANTLR discussion 3321](https://github.com/antlr/antlr4/discussions/3321#discussioncomment-3642335), @kaby76  wrote:

> I also renamed the [temp-examples](https://github.com/antlr/grammars-v4/blob/553522db0536ca80e0535351a19bdd8cbcc14e77/rexx/pom.xml#L49) in the pom (which tells the tools how to run tests on the grammar) to "examples/" so there is some testing. Currently, the generated parser for the grammar is not tested in the CI build!

 I made explicit changes to make those tests function correctly in the ANTLR build system in commits 88bbc3af298b4cc5abd588fc30412c12e98ebd70, and b9b9e0c4023f9f1fff66a47804b53caed7bec2e9, and they appeared to work.  Obviously there's more work to do, so I need to do it.

RossPatterson:
As suggested by @kaby76 in #2821: "_If the .tree files contain EOL tokens because the rexx grammar [places them on the default channel](https://github.com/antlr/grammars-v4/blob/553522db0536ca80e0535351a19bdd8cbcc14e77/rexx/RexxLexer.g4#L141), why don't we just set git attributes for these files on check out?_"

Fixes #2821.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -91,6 +91,9 @@ access_log text
 unicode/graphemes/examples/*.txt eol=lf
 # T-SQL tests with trees can fail with multiline strings
 sql/tsql/examples/*.sql eol=lf
+# Rexx places EOL on default channel, so the .tree's contain EOL.
+# Make sure EOL's are consistent regardless of OS.
+rexx/examples/* eol=lf
 
 # Declare files that will always have CRLF line endings on checkout.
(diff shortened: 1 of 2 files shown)


teverett:
@RossPatterson thanks!

PR #2828, opened 2022-09-16
sql/sqlite: add labels and separate rules for assignments in UPDATE statement

a8m:
It makes it simpler to work with the parsed tree.


KvanTTT:
Could you please add a test sample to the examples dir?

teverett:
@a8m are you planning to add an example for this?

PR #2832, opened 2022-09-20
[Hive] Sync #2761's changes to hive v2 about fix the priority problem

jiefei30:
The priority problem of Hive v3 was modified in [#2761](https://github.com/antlr/grammars-v4/pull/2761), but I forgot to synchronize to v2, the pr only synchronizes the same modification


teverett:
@jiefei30  thanks!

PR #2833, opened 2022-09-20
[MySql] Add support for seting stats_sample_pages=default in alter table statements

nirolevy:
This is supported in MySql since version 5.6
https://github.com/mysql/mysql-workbench/blob/6.3/library/parsers/grammars/MySQLSimpleParser.g#L2702


KvanTTT:
Thanks for the contribution and welcome to grammars-v4 repository!

PR #2839, opened 2022-09-22
🐞 fix(JavaScript): Add optional chaining to member index expression

caoccao:
There is a bug in the `MemberIndexExpression` when optional chaining is used. E.g.

```js
a?.['b']; // Antlr complains '[' is not expected at position 4.
a.map(b => { return b?.['c']; }); // Antlr complains '(' is not expected at position 6.
```

The fix is as following.

1. Add `QuestionMarkDot:                '?.';`.
2. Update MemberIndexExpression to `| singleExpression '?.'? '[' expressionSequence ']'                     # MemberIndexExpression`.

I have tried the fix in my personal project and it works well.

### DIFF
--- a/javascript/javascript/JavaScriptLexer.g4
+++ b/javascript/javascript/JavaScriptLexer.g4
@@ -50,6 +50,7 @@ SemiColon:                      ';';
 Comma:                          ',';
 Assign:                         '=';
 QuestionMark:                   '?';
+QuestionMarkDot:                '?.';
 Colon:                          ':';
 Ellipsis:                       '...';
 Dot:                            '.';
--- a/javascript/javascript/JavaScriptParser.g4
+++ b/javascript/javascript/JavaScriptParser.g4
@@ -311,6 +311,7 @@ expressionSequence
 singleExpression
     : anonymousFunction                                                     # FunctionExpression
     | Class identifier? classTail                                           # ClassExpression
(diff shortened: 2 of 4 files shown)


kaby76:
This is great to see, but I don't think it's complete. See https://262.ecma-international.org/11.0/#prod-OptionalChain for the syntax. For example, `a?.b` should be recognized as an optional chain expression. And, there are no tests for this added in examples/. @KvanTTT what do you think?

caoccao:
@kaby76 Thank you for the quick response. Here are my thoughts for your reference.

- There is no dedicated optional chain expression in the JavaScript parser file, so I followed the current convention by patching the `MemberIndexExpression` directly. If there's a need for `OptionalChainExpression`, I wonder if that would be a separate refactoring work.
- I'm not aware of the test stuff as that's the first time I made changes to this project. Let me see if I could add some test cases and amend this PR.

KvanTTT:
Agree with @kaby76

caoccao:
It seems the current JavaScript parser is not ready for optional chain because the test breaks at `var d = e?.f?.g;`. I'll try to fix it.

caoccao:
**Update**

- Added `OptionalChainExpression`
- Added a few test cases
- CI passed

teverett:
@caoccao thanks!

PR #2841, opened 2022-09-22
[Swift] Support willSet / didSet accessors without var init

JanWittler:
Currently, the Swift v5 grammar fails to correctly parse `willSet` / `didSet` on a variable without an initializer:
```Swift
var x: Int = 0 { didSet { print("works!") }
var y: Int?    { didSet { print("fails!) }
```
This happens, since the grammar interprets the outer block already as a getter and then the inner block as a closure. This was fixed by reordering the rules.

### DIFF
--- a/swift/swift5/Swift5Parser.g4
+++ b/swift/swift5/Swift5Parser.g4
@@ -314,9 +314,9 @@ variable_declaration:
 		variable_name (
 			initializer willSet_didSet_block
 			| type_annotation (
-				getter_setter_block // contains code_block
+			    initializer? willSet_didSet_block
+				| getter_setter_block // contains code_block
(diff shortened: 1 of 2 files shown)


KvanTTT:
Could you also please add a sample that covers suggested changed to the examples directory?

JanWittler:
I adapted the examples to cover more cases.
However, the described case was already present in the examples, it was just parsed incorrectly (as described above). Exemplary, see the code excerpt below:

https://github.com/antlr/grammars-v4/blob/8ba1b0cac7efbbae395523c1f2cd0b4977f41918/swift/swift5/examples/Objects%20and%20Classes/Contents.swift#L110-L114

KvanTTT:

PR #2857, opened 2022-10-03
refactored esolang’s into a directory

teverett:


teverett:
@KOLANICH  thoughts on this?

KOLANICH:
Well, IDK. In my PR a different set of langs is being moved and into a different subdir (I have googled a bit about each grammar for a lang I have never heard about before). Feel free to take over my whole PR and modify it the way you feel right (I feel like that before the consensus about how the hierarchy should look like it is a waste of time and manpower to create these pom files required by Maven), but keep in mind: several people have expressed the opinion that a flat pile of dirs in the root + an automated index in GitHub Pages is better than hierarchy and it seems that this opinion dominates. But not for my workflow. When I work with source code, I usually navigate git repos using file manager rather than explore the code in websites.

KOLANICH:
Also I think that it may make sense to squash all the commits into one

PR #2868, opened 2022-10-08
Add indexing of grammars

kaby76:
This PR adds a workflow to generate and check-in a file, grammars.json, into the repo with "git push" by @teverett . The workflow must be separated from the "CI" workflow because it is not to be invoked with pull requests--only "git push" by Tom.

The [workflow](https://github.com/antlr/grammars-v4/pull/2868/files#diff-36415f9303b4f431a1209c1e933eedab56a3a09a2083114ed965cf66c039a13e) goes like this:
* "Install Python", which is required by @parrt 's [mkindex.py](https://github.com/antlr/grammars-v4/blob/master/_scripts/mkindex.py).
* "Do not trust Python env", which just tests the version installed for Python.
* "Run actions/checkout v3", which checks out the code.
* "Create local changes", which runs mkindex.py to generate grammars.json.
* "Push changes", which pushes the changes to the grammars-v4 repo, if any.
* "Post" steps.

Right now, the only thing that is generated is the grammars.json file. The mkindex.py script was updated to include example files, but as @RossPatterson mentions, it will need refinement.

Some care was taken to not update anything if a "git diff" yielded nothing new, and not check-in anything. It might be unnecessary.

Virtually all this was tested verbatim in a repo I created under my userid. So it should work.


RossPatterson:
The list of examples in the index includes `.errors` and `.tree` files.  You should probably filter them out.

kaby76:
> The list of examples in the index includes `.errors` and `.tree` files. You should probably filter them out.

@RossPatterson Yes, it should. I'm still worried that the build doesn't get into an infinite loop with this addition.

kaby76:
Not quite ready...

kaby76:
@teverett Ready for review. If you merge, you should see a second workflow executed that does the index generate. I didn't add yet any html generation. The primary purpose was to make sure this thing will work so Antlr4-lab can see the repo and load grammars and example inputs.

teverett:
@kaby76  thanks!

_scripts/mkindex.py
```diff
@@ -115,7 +115,7 @@ def index_grammars(root : str) -> Sequence[dict]:
                 'lexer' : lexer,
                 'parser' : parser,
                 'start' : get_single_value("entryPoint", pom),
-                'example' : examplesFiles[0]
```
parrt, line 118: I thought I saw an example where there were multiple examples? And I think it might always come back with a list anyway.

kaby76: Right, it should be a list of files, not just one file. Otherwise why offer a selection for input files?

But, something didn't work because there's no grammars.json file generated.  Checking my repo copy to see where the workflow doesn't work.

parrt: Haha oh right. Duh. Right.

RossPatterson:
I'm seeing occasional executions of this, and commiting of the results, on my fork of this repo (_e.g._, https://github.com/RossPatterson/grammars-v4/commit/d71bb8663e0f9283ace55c4a218cb6e1d2fc8b4e).  Is there some way to prevent that?  I'm having to discard those commits before I can sync from `antlr/grammars-v4`

kaby76:
> I'm seeing occasional executions of this, and commiting of the results, on my fork of this repo (_e.g._, [RossPatterson@d71bb86](https://github.com/RossPatterson/grammars-v4/commit/d71bb8663e0f9283ace55c4a218cb6e1d2fc8b4e)). Is there some way to prevent that? I'm having to discard those commits before I can sync from `antlr/grammars-v4`

I have a PR for fix this, which involves grammars.json. The indexing file is used by lab.antlr.org. I'm waiting for this to be merged. https://github.com/antlr/grammars-v4/pull/2920 
@teverett  The PR generates grammar.json as a Github Artifact instead of a checkin. Github offers fine grain personal access tokens, which only came available in the last month.

parrt:
> Is there some way to prevent that? I'm having to discard those commits before I can sync from antlr/grammars-v4

Ah. ok, now I think I see the issue.  Hmm....grrrr... Isn't there an easy way to simply turn it off for any repository other than the original?

KvanTTT:
Commits with generated code are definitely useless and excess. grammars.json should be artifact.

kaby76:
> Commits with generated code are definitely useless and excess. grammars.json should be artifact.

It's going to be a Github Artifact--if this PR https://github.com/antlr/grammars-v4/pull/2920 gets merged. I am waiting. Once merged, someone can needs to add a fine-grain personal access token for specifically to grammars-v4 with read-only access. Fine-grained PATs only came about in the last month--probably because people have been complaining for years that it is absolutely brain-dead to offer artifacts, but not offer a way to access them except through the website. Github is finally getting its act together.

KvanTTT:
> It's going to be a Github Artifact--if this PR https://github.com/antlr/grammars-v4/pull/2920 gets merged. I am waiting.

Done! Thanks.

parrt:
Do we need to protect the artifact? I.e., can we avoid that complexity and just generate an open grammars.json artifact?

kaby76:
> Do we need to protect the artifact? I.e., can we avoid that complexity and just generate an open grammars.json artifact?

We could request webpages through JS, parse the page looking for links, then find the link for the artifact. This seems a round-about way of getting the .zip file. Instead, Github offers an API that requires a PAT.

With fine-grain PATs, the artifact will be read-only because we will only grant read-only access to the files in grammars-v4, nothing else. The only hard part is that we need to make two requests to the repo, parse the JSON results, in order to determine the name of the artifact to request. See [my comment on the three downloads](https://github.com/antlr/grammars-v4/pull/2920#issuecomment-1312770914). (Note, the last "curl" would skip the user id parameter and just give the PAT.)

https://github.blog/2022-10-18-introducing-fine-grained-personal-access-tokens-for-github/#:~:text=Fine%2Dgrained%20personal%20access%20tokens%20offer%20enhanced%20security%20to%20developers,your%20data%20of%20compromised%20tokens.&text=Stolen%20and%20compromised%20credentials%20are,data%20breaches%20across%20the%20industry.

parrt:
Won't the artifact always be in the same spot and therefore easy to get?

parrt:
oh wow. so it's something weird like 453691604 in the url. ugh. That's kinda annoying and a complicated way. Personally I'd like to find a different way but I'll defer to you guys.   Given that I load the json exactly once and the repo can get out of sync with the antlr lab, maybe I'll just run the script manually upon server start to avoid all of this.

kaby76:
> Won't the artifact always be in the same spot and therefore easy to get?

Yeah, that's what I thought. But, the artifact is per-github-action. I really wish there was a "get the latest successful action" URL, but I'm okay with Github implemented the API this far. (It'll take years before they offer a constant URL to the latest artifact.) The only constant URL would be a checked-in copy of grammars.json.

PR #2869, opened 2022-10-09
Fixed incorrect records parsing including testcase

### LINKED ISSUE #2847: Incorrect parsing of records in Java
Hi
I know this has been up in topics before, but it does not seem resolved and none of the posts suggests how to approach the problem, so I bring it up again to get some clarity. (https://github.com/antlr/grammars-v4/pull/2690, https://github.com/antlr/grammars-v4/issues/2685) 

The problem is that it seems records gets incorrectly mistaken for methods in simple records such as 
```
public class IncorrectAntlrParsing {
   public record BindingsSnapshot(int a) { }
}
```

image

As discussed in the previous posts records only became a keyword in version 14 of Java, and could potentially be a method as the parser now parses it (to a method), but this is no longer the case in version 14+. (Or have I misunderstood something?)

The obvious fix to the problem seems to be to move record to the top of the memberDeclaration in JavaParser.g4:
```
memberDeclaration
    : recordDeclaration //Java17
    | genericMethodDeclaration
    | fieldDeclaration
    | constructorDeclaration
    | genericConstructorDeclaration
    | interfaceDeclaration
    | annotationTypeDeclaration
    | classDeclaration
    | enumDeclaration
    | methodDeclaration
    ;
```

So my question is: 
- Is this a correct fix, or do you foresee any other complications with it?  (It has been suggested before and the simple records like these are in multiple places in the Java Core module)
- If this is not a correct fix, any other suggestions on how to approach the problem?

Edit: Uploaded wrong image first (with the fixed version)

DagSonntag:
Closes #2847 with empty records being incorrectly parsed. Included simple test-case with correct parsing.


teverett:
@DagSonntag thanks

PR #2886, opened 2022-10-14
Add `PRIMARY` as keyword to MySQL grammar to fix index hints

joschi:
Index hints using the primary index (`USE INDEX(PRIMARY)`) are currently failing because `PRIMARY` is missing from the list of keywords in the ANTLR-based MySQL parser.

According to https://dev.mysql.com/doc/refman/5.7/en/keywords.html#keywords-5-7-detailed-P, `PRIMARY` is a reserved keyword.

Index hints: https://dev.mysql.com/doc/refman/5.7/en/index-hints.html


joschi:
Refs https://github.com/debezium/debezium/pull/3957#issuecomment-1278494795

KvanTTT:

Issue #2887, opened 2022-10-14
mkindex.py skips combined grammars


kaby76:
* [This line](https://github.com/antlr/grammars-v4/blob/9f66ac21bfb628426c1bd21378704d7692cfdaf5/_scripts/mkindex.py#L95) does not work. It filters out combined grammars; it should not.

PR #2888, opened 2022-10-14
Fix for #2887

kaby76:
The mkindex.py script filters out combined grammars. It shouldn't. This change adds combined grammars to the grammars.json index so antlr4-lab can work with those grammars.


teverett:
@kaby76  thanks!

PR #2890, opened 2022-10-18
Add support for all (System)Verilog compiler directives

msagca:
- Added new lexer modes to properly parse one-line compiler directives.
- Restricted allowed character range for tokens to ASCII space as per the spec.


teverett:
@msagca thanks!

PR #2891, opened 2022-10-19
snowflake db grammar

mlorek:


kaby76:
The Php target fails because there is a copy_options symbol and also a COPY_OPTIONS symbol. Rename one because PHP is case insensitive.

teverett:
@mlorek  thanks!

PR #2892, opened 2022-10-19
Add PICT grammar

bkiers:
Add the [PICT](https://github.com/Microsoft/pict/blob/main/doc/pict.md) grammar + examples

### DIFF
--- a/pict/PICT.g4
+++ b/pict/PICT.g4
@@ -1,7 +1,184 @@
 grammar PICT;
 
+options { caseInsensitive = true; }
+
+/// https://github.com/Microsoft/pict/blob/main/doc/pict.md
 input
- : OTHER* EOF
+ : parameter_definitions sub_models constraints EOF
+ ;
+
+parameter_definitions
+ : parameter_definition*
+ ;
+
+parameter_definition
+ : parameter COLON parameter_values
+ ;
+
+parameter
+ : VALUE+
+ ;
+
+parameter_values
+ : parameter_value ( COMMA parameter_value )*
+ ;
+
+parameter_value
+ : parameter_value OPAR parameter_value CPAR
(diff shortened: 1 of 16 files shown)


xjtxjt:
`@bkiers` Sir,I am xiangjt from Stack Overflow [https://stackoverflow.com/questions/74109888/how-can-i-tell-antlr-to-use-only-the-branch-i-specify-not-other-branches?noredirect=1#comment130878402_74109888](url) 
there seems a little bug in this grammar,it's right in the alternative:
```
parameter_definition
 : parameter COLON parameter_values
 ;
```
when I input:
```
    Type:           Primary, Logical, Single, Span, Stripe, Mirror, RAID-5
    Size:           10, 100, 500, 1000, 5000, 10000, 40000
    Format method:  quick, slow
    File system:    FAT, FAT32, NTFS
    Cluster size:   512, 1024, 2048, 4096, 8192, 16384, 32768, 65536
    Compression:    on, off

    IF [File system] <> "NTFS" OR
     ( [File system] =  "NTFS" AND [Cluster size] > 4096 ) 
    THEN [Compression] = "Off";

    IF NOT ( [File system] = "NTFS" OR 
           ( [File system] = "NTFS" AND NOT [Cluster size] <= 4096 )) 
    THEN [Compression] = "Off"
    ELSE [Compression] = "on";

```
and the syntax tree show:
antlr4_parse_tree

the bug is right here:
image
`File system` should be a `parameter`,but ` File` is matched as a value of `Format method`
image
`Cluster size` should be a `parameter`,but ` Cluster`  is matched as a value of `Flie System`

bkiers:
@xjtxjt feel free to fix it yourself. I am more than happy to review your fix.

PR #2894, opened 2022-10-23
apache derby grammar

mlorek:


teverett:
@KvanTTT  good to pull this?

teverett:
@mlorek thanks!

sql/derby/DerbyLexer.g4
```diff
@@ -0,0 +1,435 @@
...
+LINE_COMMENT:           '--' ~[\r\n]*                 -> channel(HIDDEN);
+
+// TODO: ID can be not only Latin.
+DOUBLE_QUOTE_ID:        '"' ~'"'+ '"';
+SINGLE_QUOTE:           '\'';
+
+ID:                     [A-Z_] ( [A-Z0-9_] )*;
+
+
+STRING_LITERAL       options { caseInsensitive=false; }     : '\'' (~'\'' | '\'\'')* '\'';
```
KvanTTT, line 371: Using this option is not incorrect but it looks redundant since there is no using of case sensitive literals in token definition.

PR #2899, opened 2022-10-26
derby grammar PR comments

mlorek:


sql/derby/DerbyLexer.g4
```diff
@@ -353,9 +353,6 @@ XMLSERIALIZE:                                          'XMLSERIALIZE';
 YEAR:                                                  'YEAR';
 
 
-DUMMY: 'DUMMY';
```
KvanTTT, line 356: So, finally it was incorrect token since you removed it at all?

mlorek: > So, finally it was incorrect token since you removed it at all?

it was WIP token

PR #2900, opened 2022-10-26
Add extend option

DmitryMaletin:
Extend definition added back as it was omitted in the current grammar.


teverett:
@DmitryMaletin thanks!

Issue #2908, opened 2022-11-01
Antlr4 Java grammar Java 17 mods not handling yield in switch expression.


gmtschlr:
I got an error parsing the version java.util.Calendar in the Java 17.0.3 source base, using antlr4 (4.10.1), which has a switch expression at about line 1491.  

The "yield" statement in the second case ("iso8601") was parsed as a local variable declaration instead of a yield statement.

The problem appears to be in the syntax for blockStatement, which has localVariableDeclaration before statement.

With some trepidation, I switched them, and the yield statement is now recognized, without any other errors noted in parsing several dozen files.  I remain concerned whether this fix is the proper one.

Relevant grammar file (as modified) and Java test case in attached zip file.
[JavaParser.zip](https://github.com/antlr/grammars-v4/files/9911817/JavaParser.zip)

DagSonntag:
I think this issue can be closed with the merged branch.

PR #2910, opened 2022-11-02
Fix for Java parser incorrectly parsing Yield BlockStatements 

DagSonntag:
Updated the JavaParser.g4 to correctly handle yield expressions in Blockstatements in switch-expressions and added corresponding unit tests. The issue itself (and solution) is explained in #2908 and hence no additional issue was opened for this.
Closing #2908


DagSonntag:
Did this pull request pass the tests? I got some strange error on the reindexing that I don't understand?
image

DagSonntag:
Will try to do a new pull request shortly due to the failed tests. It did Probably have to do with that the branch on my cloned repo was behind during one commit

PR #2911, opened 2022-11-02
Fixed Java incorrect YieldParsing (attempt 2)

DagSonntag:
Updated the JavaParser.g4 to correctly handle yield expressions in Blockstatements in switch-expressions and added corresponding unit tests. The issue itself (and solution) is explained in #2908  and hence no additional issue was opened for this.
Closing #2908 

Previous pull request #2910 failed due to some failed head on the branch. This might have been due to that my local clone was not up to date when the commit was performed. This is now solved and hence this new attempt on a pull request.


teverett:
@DagSonntag thanks!

Issue #2909, opened 2022-11-02
javascript cpp port crashes when parsing a js file


nj00001:
this is my js file content, and I know there is something wrong with the syntax of the following code.
```"use strict";
//------------------------------------------------------------------------------
// Expression Bodies
// More expressive closure syntax.
// http://es6-features.org/#ExpressionBodies
//------------------------------------------------------------------------------

odds  = evens.map(v => v + 1)
pairs = evens.map(v => ({ even: v, odd: v + 1 }))
nums  = evens.map((v, i) => v + i)

//------------------------------------------------------------------------------
// Statement Bodies
// More expressive closure syntax.
// http://es6-features.org/#StatementBodies
//------------------------------------------------------------------------------

nums.forEach(v => {
   if (v % 5 === 0)
       fives.push(v)
})

//------------------------------------------------------------------------------
// Lexical this
// More intuitive handling of current object context.
// http://es6-features.org/#Lexicalthis
//------------------------------------------------------------------------------

this.nums.forEach((v) => {
    if (v % 5 === 0)
        this.fives.push(v)
})
```

this is my antlr4 code, its crashed in `parser.program()`
```
        ANTLRInputStream input(targetString);
        JavaScriptLexer lexer(&input);
        CommonTokenStream commonTokens(&lexer);

        JavaScriptParser parser(&commonTokens);

        tree::ParseTree *tree = parser.program();
```

kaby76:
What grammar are you using? Neither [javascript/javascript](https://github.com/antlr/grammars-v4/tree/master/javascript/javascript) nor [javascript/jsx](https://github.com/antlr/grammars-v4/tree/master/javascript/jsx) even compile for Cpp. The testing of both of these grammars is [turned off](https://github.com/antlr/grammars-v4/blob/c4775c1c07ef227055682c65cece28add2233782/_scripts/skip-cpp.txt#L83-L84).

lizhenhua20:
> What grammar are you using? Neither [javascript/javascript](https://github.com/antlr/grammars-v4/tree/master/javascript/javascript) nor [javascript/jsx](https://github.com/antlr/grammars-v4/tree/master/javascript/jsx) even compile for Cpp. The testing of both of these grammars is [turned off](https://github.com/antlr/grammars-v4/blob/c4775c1c07ef227055682c65cece28add2233782/_scripts/skip-cpp.txt#L83-L84).

[...]

PR #2912, opened 2022-11-05
Fixes for #2909 "javascript cpp port crashes when parsing a js file"

kaby76:
This is a partial fix for https://github.com/antlr/grammars-v4/issues/2909. There is a change required in the Cpp runtime, which will be made in a PR in https://github.com/antlr/antlr4. A new release will be needed before adding the target back to CI testing.

* Add in changes for automated testing (transformGrammar.py).
* Fix incorrect code in Cpp port of the lexer base class.
* Femove Antlr4cs since it uses Unicode litreals, which are not supported in the old Antlr4 fork.


teverett:
@kaby76  thanks!

Issue #2919, opened 2022-11-10
Ignoring merge conflicts caused by grammars.json


msagca:
[grammars.json](https://github.com/antlr/grammars-v4/blob/master/grammars.json) file creates unnecessary merge conflicts when a PR that modified this file gets merged before some other open PRs. It should be okay to ignore merge conflicts caused by this file since the index is updated anyway in case of a change. If you're okay with this, the method described [here](https://gist.github.com/tmaybe/4c9d94712711229cd506) seems to work.

kaby76:
Yes, it's a generated file, but not a "build" product per se, although it is generated on each PR. It's used in lab.antlr.org right now. Eventually, I will be using this file to produce a top-level page for a Github Pages website that contains sortable tables, searching, links, classifications, targets available, etc., for grammars.  Please make a PR to fix the conflict. It seems like the conflict ignore procedure should work.

https://github.com/antlr/grammars-v4/issues/2866
https://github.com/antlr/grammars-v4/issues/2905
https://github.com/antlr/grammars-v4/pull/2830
https://github.com/antlr/grammars-v4/discussions/2872

KvanTTT:
Honestly, I don't like the idea of automatic commits and I still think lab.antlr.org should process repository itself instead of using generated file.

kaby76:
> Honestly, I don't like the idea of automatic commits and I still think lab.antlr.org should process [the] repository itself instead of using [a] generated file.

I'll change the generate of grammars.json around to make it a Github Action artifact. That will work for a while.

But, Antlr4test-maven-plugin, antlr4-maven-plugin, trgen, and the build scripts contain the logic to understand the structure of the grammars-v4 repo. I don't think it's a good idea to add another tool that duplicates the logic. The pom.xml do not encode everything that is needed, e.g., what targets can be run.

PR #2920, opened 2022-11-12
Generate grammars.json, but leave as artifact.

kaby76:
https://github.com/antlr/grammars-v4/issues/2919
https://github.com/antlr/antlr4-lab/issues/52

For now, process the grammars in the repo to create a simple list of objects that describe each grammar, but leave as an artifact. This will be a multi-step checkin for this and antlr4-lab repos.


kaby76:
It seems to access the artifact, multiple accesses will be required:
* curl https://api.github.com/repos/kaby76/temp-with-actions/actions/artifacts => From the returned json output, find the latest build, then get the artifact ID, e.g., 434781932
* curl https://api.github.com/repos/kaby76/temp-with-actions/actions/artifacts/434781932 => get the link to the artifact from the returned json, 
* curl -H 'Accept: application/vnd.github+json' -H 'Authorization: token kaby76 OAUTH-TOKEN' -L https://api.github.com/kaby76/temp-with-actions/artifacts/434781932/zip => This step requires a Github userid and authorization token. The output is a zip file with the grammars.json file.

The rate-limiting step that is required here is that we must have an authorization token to access the grammars-v4 repo artifacts. I don't know how to create a public one.

KvanTTT:
Thanks, I like idea of grammars.json as artifact much more than artificial commit.

kaby76:
I don't have a way currently to create a public token for github.com that has simply read-only rights. It looks like the best way to move forward is to check in this code, then try to set up an authorized app.

kaby76:
Fine-grain personal access tokens with read only access to repos are in beta in Github.com. But, yes, it does work! 

```
curl -H 'Accept: application/vnd.github+json' -H 'Authorization: token <fine-grain-PAT>' -L https://api.github.com/repos/kaby76/temp-with-actions/actions/artifacts/453691604/zip --output
foobar.zip
```

@teverett  Please merge. This will turn off the checkin, and just run the indexer leaving an artifact that we can then "curl" in lab.antlr.org clients.

parrt:
So people are complaining that we are generating grammars.json in every dir of grammars-v4 or just in their forks?  Can they ignore an extra file?

Issue #2922, opened 2022-11-16
Java/java does not handle generic localVariableDeclaration anymore


DagSonntag:
With a recent fix to handle yield properly (by me) #2911 I noticed I accidentally broke other functionality such as a local variable declarations with generic types i.e. 
`ArrayList<Integer> myList = new ArrayList<>();`

By moving statements before LocalVariableDeclarations blockStatement this now meant that the left side of the assignments got resolved as expressions ((ArrayList < Integer) > myList). 

Hence the "fix" has to be reverted, and the best way to fix #2908 that I came up with (without breaking other functionality) was to separate identifiers from valid typeIdentifiers (see pull-request added shortly). This solves the issue and is "valid" since the yield keyword is not valid in type declarations as of Java 14. 

However, it breaks existing code using the grammar since it changes the classOrInterfaceType structure. Moreover, if it should be fully used there might be more places where it is appropriate to use the new typeIdentifier type. 

So, I am happy to take feedback and thoughts on other solutions or if you think I should do something else as well. I have added and updated examples to avoid breaking this functionality again.

kaby76:
Parsing `ArrayList<Integer> myList = new ArrayList<>();` is broken in the current grammar because it thinks `ArrayList<Integer> myList` is an expression. But, it's not because of semantics.

The problem is that expressions are not parsed correctly with relational operations. [It says in the spec](https://docs.oracle.com/javase/specs/jls/se19/html/jls-15.html#jls-15.20):

> The relational operators are syntactically left-associative (they group left-to-right).
>
> However, this fact is not useful. For example, a<b<c parses as (a<b)<c, which is always a compile-time error, because the type of a<b is always boolean and < is not an operator on boolean values.
>
> The type of a relational expression is always boolean.

It might be possible to remedy the situation with rudimentary parse tree checks of the form `a < b < c` using parser-time visitors.

DagSonntag:
> Parsing `ArrayList<Integer> myList = new ArrayList<>();` is broken in the current grammar because it thinks `ArrayList<Integer> myList` is an expression. But, it's not because of semantics.

Hi
Hmm, that is true but I don't think the left association is what causes the issue this time. It is more the order in which the rules are tested, i.e. whether to test a block statement as a LocalVariableDeclaration or as an statement expression first. If one test the latter first then the resolution gets wrong as described.

DagSonntag:
Ah, yes you are right, sorry, I misunderstood.
[...]

PR #2923, opened 2022-11-16
Handling of yield keyword without breaking LocalVariableDeclarations

DagSonntag:
Discussion in issue #2922

Added new code and new and updated examples to fix the issue discussed


teverett:
@DagSonntag  thanks!

Issue #2924, opened 2022-11-21
Java/Java does identifies "var" as as ClassOrInterfaceType in LocalVariableDeclaration 


DagSonntag:
Currently if one uses the `var` keyword (introduced in Java 10) in localvariabledeclarations this is not identified as a keyword but as a `ClassOrInterfaceType` (the more generic case). This is due to the order in the definition in the parser .g4 file: 

```g4
localVariableDeclaration
    : variableModifier* (typeType variableDeclarators | VAR identifier '=' expression)
```

In this case the second or-option will never be true. By changing the order this is properly evaluated and VAR takes precedence:

```g4
localVariableDeclaration
    : variableModifier* (VAR identifier '=' expression | typeType variableDeclarators)
```

I will shortly submit a pull request with the change and a test-case. If anyone has a case where this does not work, please let me know!

PR #2925, opened 2022-11-21
VAR handling in LocalVariableDeclaration

DagSonntag:
The pull request for the issue discussed in issue #2924. I.e. that VAR is never detected as a keyword in current implementation of localvariabledeclaration.

Included examples with correct parsing


java/java/JavaParser.g4
```diff
@@ -254,7 +254,7 @@ arrayInitializer
     ;
 
 classOrInterfaceType
-    : typeIdentifier typeArguments? ('.' identifier typeArguments?)*
+    : (identifier typeArguments? '.')* typeIdentifier typeArguments?
```
kaby76, line 257: Looks good. Annotations? https://docs.oracle.com/javase/specs/jls/se19/html/jls-4.html#jls-ClassOrInterfaceType

teverett:
@DagSonntag  thanks!

java/java/JavaParser.g4
DagSonntag: Hmm, this was closed before I could respond, but this is covered in the typearguments no?
Anyway, let's check it up when it comes up next time...

```
typeArguments
    : '<' typeArgument (',' typeArgument)* '>'
    ;

typeArgument
    : typeType
    | annotation* '?' ((EXTENDS | SUPER) typeType)?
    ;

typeType
    : annotation* (classOrInterfaceType | primitiveType) (annotation* '[' ']')*
    ;
```

raghav-deepsource:
@DagSonntag , could you tell me how you generated the .tree file?

kaby76:
> could you tell me how you generated the .tree file?

@raghav-deepsource I never liked how this is implemented in the tester. These files should be retained so that they can be renamed for remastering the test. For now, create an empty .tree file for each test input file you want to record, then run `mvn test`. In the output, edit the empty .tree files and add in the tree that mvn prints out. The .tree files should be one line, and no newline chars at the end of the file.

raghav-deepsource:
hmm, yup, that worked. I was very confused when I saw that diff syntax the antlr test gave me. It would be better to directly print the parse result, so that can be more easily compared with the existing file (if the changed data were pasted into the original file, provided the file was checked into git already, there would be a nice git diff representation of the changes) instead of directly writing a diff out.

PR #2928, opened 2022-11-25
add support of drop cluster

jiangtao69039:


sql/plsql/PlSqlParser.g4
```diff
@@ -2218,6 +2219,10 @@ alter_cluster
         ';'
     ;
 
+drop_cluster
+    : DROP CLUSTER cluster_name (INCLUDING TABLES (CASCADE CONSTRAINTS)?)?
```
KvanTTT, line 2223: Please simplify:

```suggestion
    : DROP CLUSTER cluster_name (INCLUDING TABLES CASCADE CONSTRAINTS?)?
```

Issue #2929, opened 2022-11-25
Bring Python3 grammar up to date with current Spec grammar


kaby76:
This concerns requests for updating the Python grammar to the latest standard grammar, which is a CPython grammar at https://docs.python.org/3/reference/grammar.html.

* Missing match_stmt. https://stackoverflow.com/q/74563209/4779853. A description is [here](https://learnpython.com/blog/python-match-case-statement/).
* https://github.com/antlr/grammars-v4/issues/2462

Note, the grammar is significantly different from the Spec. This should be minimized, with comments to explain why rules were refactors, and to show the actual refactoring transformations. Without this information, we cannot determine whether the refactoring was correct or not.

RobEin:
Maybe you could try this:
[https://github.com/RobEin/ANTLR4-Python-grammar-by-PEG](https://github.com/RobEin/ANTLR4-Python-grammar-by-PEG)
It has some bugs, but it basically works.
Unfortunately, I don't have time to develop it further at the moment.

kaby76:
I do not know why, but replies to emails are not getting posted in the Issue thread. I have noticed this several times before, on different repos. Github has a serious, major bug, which can negatively affect all development.

Please do not reply via email to an issue thread. Go to the webpage, and manually enter it on the page.

https://github.com/community/community/discussions/3287
https://www.google.com/search?q=email+replies+are+not+posted+in+issue+thread+github&rlz=1C1NDCM_enUS742US742&oq=email+replies+are+not+posted+in+issue+thread+github&aqs=chrome..69i57j69i64.13320j0j4&sourceid=chrome&ie=UTF-8

Specifically, this was written by @RobEin 

> Maybe you could try this:
> https://github.com/RobEin/ANTLR4-Python-grammar-by-PEG
> It has some bugs, but it basically works.
> Unfortunately, I don't have time to develop it further at the moment.

For sure, I will be looking over your grammar to examine the rewrites you made to get a functional grammar. But, from here, I am working out automated steps to scrape the grammar directly from the text at https://github.com/python/cpython/tree/3.11/Grammar

I have an Antlr4 grammar that reads Pegen grammars [here](https://github.com/kaby76/Domemtech.Trash/tree/main/src/supported_grammars/pegen). Using that and Trash, it's easy to get the grammar in Antlr4 syntax minus changes for semantic predicates.

KvanTTT:
> Please do not reply via email to an issue thread. Go to the webpage, and manually enter it on the page.

[...]

PR #2930, opened 2022-11-25
Issue 2929

kaby76:
Preliminary changes to bring python/python3 grammar up to date with 3.11 and Spec.

https://github.com/antlr/grammars-v4/issues/2929

### Renamings
* Renamed `simple_stmt` to `simple_stmts` to be the plural of `simple_stmt`, corresponding to the Spec grammar.
* Renamed `small_stmt` to `simple_stmt`, corresponding to the Spec grammar.
* Renamed `suite` to `block`, corresponding to the Spec grammar.
### Reordering
* Reordered keywords in lexer grammar so you can find them. It is ridiculous finding and adding a new keyword without some basic sanity. Note, this is possible because Antlr lexer rules match (1) maximum-lengthed strings, (2) match the first lexer rule if multiple lexer rules match the string. Fortunately, all keyword rules were already grouped together prior to this PR.
### Added rules
* Added `match_stmt` to `compound_stmt` as in the Spec.
* Added rules associated with `match_stmt`.
* Added '_' to `atom` in order to back support for-statement. __I'm not sure if the underscore should be a different token, but I'm tending toward this in order to support patterns in match statements.__
* Added rule `name : NAME | .....;` rule in order to allow certain keywords as identifiers. Renamed applied occurrences of `NAME` to `name`.
* Removed mutual-left recursion in `attr` and `name_or_attr` present in Spec grammar, via unfolding of `name_or_attr` into RHS of `attr`.
### Other grammar changes
* Removed useless parentheses.
## Code changes
* Added reset()/Reset() to Java/CSharp base class lexer code.
* Added parser base classes for semantic predicates. Those predicates currently are stubs returing true.
* Untabified target code because tabs expand to a variable number of spaces per editor.
## Git changes
* Added "Grammars*" to .gitignore for new trgen target directories, in which trgen can generate drivers for all targets for side-by-side comparison without moving Generated/ code around.
## Test changes
* match.py
* match2.py
* Added the JavaScript testing of python/python3.
* Added test-parsing.sh to test parser against entire cpython source code. 156 failed to parse out of 1897 total, which is about 8% failed. The total time for an unloaded Ryzen 7 2700, 16GB machine to parse with C# was 25.75m.
## Ignored
* Note, no attempt was made to update python/python. That is a "unified" grammar for Python2 and Python3. I really don't understand how one can deal with all the similarly named rules and conflicts in merging rules for similar syntax between Python2 and Python3. In my opinion, this is extremely difficult to do correctly, which is why I will not touch that grammar.
* I did not attempt to update the target-specific Python3 grammars, like [python3-cpp](https://github.com/antlr/grammars-v4/tree/master/python/python3-cpp), [python3-py](https://github.com/antlr/grammars-v4/tree/master/python/python3-py), [python3-ts](https://github.com/antlr/grammars-v4/tree/master/python/python3-ts). Those are old and should just be removed. Whoever wrote them should determine whether to delete or not.
## To do
* Semantic predicates will need to be added, but there are already many in the python/python3 grammar.
* The generated parser is slow. This is specifically due to the rules for expressions. I will fix this and the expressions in the grammar to use optimized Antlr rules in another PR after this.
* I am still working on Dart, Go, and PHP ports. Those will be in another PR after this.


hasii2011:
attaching as per your [instructions](https://stackoverflow.com/questions/74563209/python-3-10-antlr-parser?noredirect=1#comment131636403_74563209)
[Archive.zip](https://github.com/antlr/grammars-v4/files/10097393/Archive.zip)

kaby76:
My new grammar parses these files. I had to pitch much of the old grammar and use the Spec grammar. It will be in the next PR after this is merged.

hasii2011:
I'll keep my eyes open for the merge

teverett:
@kaby76 thanks!

hasii2011:
Ok,  this update works great with Python 3.10 code;  For me other than a minor file edit to Python3LexerBase to account for my package structure, it was essentially a drop-in replacement once I generated the new code.  Thanks for this work

PR #2934, opened 2022-11-29
[Thrift] fix bug and add bad case for  literal in thrift

alingse:
i found this in my project  https://github.com/thrift-labs/thrift-parser/pull/8/files
the origin `Thrift.g4` is not good enough.

here is some failed test case: https://github.com/antlr/grammars-v4/actions/runs/3575542658/jobs/6012214625

the thrift source code was here

https://github.com/apache/thrift/blob/master/compiler/cpp/src/thrift/thriftl.ll#L285-L332

it check a start with `'` or `"` and when it match `\`, the next char must be `n|t|r` or the start char.
and finally it must be end with the start quote.

translate this logic to regex was this

```

LITERAL
    : ( ( '"'  (('\\' ('r'|'n'|'t'|'"'|'\''|'\\') ) | ~('\\' | '"' ))* '"')
        |
        ( '\'' (('\\' ('r'|'n'|'t'|'"'|'\''|'\\') ) | ~('\\' | '\''))* '\'')
      )
    ;
```

it was really some complex.


msagca:
I think this looks better:
```antlr
LITERAL
    : '"' ( ESC_SEQ | ~[\\"] )* '"'
    | '\'' ( ESC_SEQ | ~[\\'] )* '\''
    ;
fragment ESC_SEQ : '\\' [rnt"'\\] ;
```

alingse:
@msagca done. thanks, i haven't know fragment before. it's really clear than the old one.

alingse:
cc @KvanTTT

PR #2935, opened 2022-11-30
sql/sqlite: formatting sql2.sql

haochengxia:


teverett:
@haochengxia  thanks!

PR #2945, opened 2022-12-06
[MySql] DBZ-5904 Suppport INSERT INTO statements with dots in column names

nirolevy:


sql/mysql/Positive-Technologies/MySqlParser.g4
```diff
@@ -2043,7 +2043,7 @@ roleName
 
 fullColumnName
     : uid (dottedId dottedId? )?
-    | . dottedId dottedId?
+    | .? dottedId dottedId?
```
KvanTTT, line 2046: Not sure it's correct. Why `fullColumnName` accepts any token at the beginning? Can it be rewritten without using `.`?

nirolevy: The original reason is explained here https://github.com/antlr/grammars-v4/pull/1885

Will removing the optional from the any token make it more agreeable? i.e.
```antlr
fullColumnName
    : uid (dottedId dottedId? )?        #basicCulomnName
    | dottedId (dottedId dottedId? )?   #culomnNameAnchoredToThisTable
    | . dottedId dottedId?              #culomnNameAithArbitraryTableAlias
    ;
```

Or do you prefer I do as @jpechane originally suggested and change it to `keywordsCanBeId`?
```antlr
fullColumnName
    : (uid | dottedId) (dottedId dottedId? )?
    | keywordsCanBeId dottedId dottedId?
    ;
```
(This breaks tests, specifically ``    SELECT trigger.num FROM test `trigger`;    ``, which can of course be easily fixed by adding `TRIGGER` to `keywordsCanBeId`, but I'm afraid it'll break other aliases - in current keywords missing from `keywordsCanBeId` or any future additions...)

PR #2947, opened 2022-12-07
CPP14Parser Grammar File Updated with Else-If

AnirudhPotturi:
The CPP14Parser Grammar File had no rules written for an else-if selection statement.


msagca:
The rule is correct. `statement` following `else` can be another `if` statement, hence an `else if` statement.

teverett:
@AnirudhPotturi  @KvanTTT  @msagca  should we close this PR?

msagca:
> @AnirudhPotturi @KvanTTT @msagca should we close this PR?

I think we can close this.

PR #2955, opened 2022-12-09
Add support of create rollback segment

jiangtao69039:


sql/plsql/PlSqlParser.g4
```diff
@@ -246,6 +247,10 @@ drop_rollback_segment
     : DROP ROLLBACK SEGMENT rollback_segment_name
     ;
 
+create_rollback_segment
+    : CREATE PUBLIC? ROLLBACK SEGMENT rollback_segment_name ((TABLESPACE tablespace) | storage_clause)*
```
KvanTTT, line 251: Excess parentheses:

```suggestion
    : CREATE PUBLIC? ROLLBACK SEGMENT rollback_segment_name (TABLESPACE tablespace | storage_clause)*
```

KvanTTT (changes requested):

PR #2962, opened 2022-12-13
Protobuf3 grammar: support optional field

DmitryMaletin:
Add support for optional field as documented here: https://developers.google.com/protocol-buffers/docs/proto3#specifying_field_rules

### DIFF
--- a/protobuf3/Protobuf3.g4
+++ b/protobuf3/Protobuf3.g4
@@ -60,7 +60,7 @@ fieldLabel
   ;
 
 field
-  : ( fieldLabel )? type_ fieldName EQ fieldNumber ( LB fieldOptions RB )? SEMI
+  : fieldLabel? type_ fieldName EQ fieldNumber ( LB fieldOptions RB )? SEMI
   ;
 
 fieldOptions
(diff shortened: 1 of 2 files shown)


protobuf3/Protobuf3.g4
```diff
@@ -55,9 +55,12 @@ optionName
   ;
 
 // Normal Field
+fieldLabel
+  : OPTIONAL | REPEATED
+  ;
 
 field
-  : ( REPEATED )? type_ fieldName EQ fieldNumber ( LB fieldOptions RB )? SEMI
+  : ( fieldLabel )? type_ fieldName EQ fieldNumber ( LB fieldOptions RB )? SEMI
```
KvanTTT, line 63: Please remove excess parentheses:

```suggestion
  : fieldLabel? type_ fieldName EQ fieldNumber ( LB fieldOptions RB )? SEMI
```

KvanTTT:
Could you please add an example to `examples` subdirectory that covers suggested changes?

DmitryMaletin:
> Could you please add an example to `examples` subdirectory that covers suggested changes?

Would it be enough  or is it better to modify/add  to examples/conformance.proto? 
Looks like singular fields there named as optional

KvanTTT:
I think it's ok, thanks.

Issue #1228, opened 2018-09-10
AQL (ArangoDB Query Language) grammar


ziflex:
I've noticed that there is no AQL grammar.

https://en.wikipedia.org/wiki/AQL_(ArangoDB_Query_Language)

PR #2963, opened 2022-12-15
#1228 - ArangoDb - grammar and tests

mlorek:


aql/ArangoDbParser.g4
```diff
@@ -0,0 +1,220 @@
...
+    | with_collection_list?
+        FOR v=id_ (',' e=id_ (',' p=id_)? )?
+        IN (numeric_literal ('..' numeric_literal)? )?
+        (OUTBOUND | INBOUND | ANY) expr
+        (GRAPH expr | collection_list) // collections support direction: ANY
+        (PRUNE expr)?
+        filter*
+        options_?
+        return_expr?
+    ;
```
KvanTTT, line 48: I suggest factoring:

```suggestion
data_access_query
    : with_collection_list? ( for_op* return_expr
                            | FOR v=id_ (',' e=id_ (',' p=id_)? )?
                             IN (numeric_literal ('..' numeric_literal)? )?
                             (OUTBOUND | INBOUND | ANY) expr
                             (GRAPH expr | collection_list) // collections support direction: ANY
                             (PRUNE expr)?
                             filter*
                             options_?
                             return_expr? )
    ;
```

aql/ArangoDbParser.g4
```diff
@@ -0,0 +1,220 @@
...
+    | with_collection_list? for_op*
+        REPLACE (expr WITH)? object_literal IN collection options_?
+         return_expr?
+    | with_collection_list? for_op*
+        (REMOVE expr IN collection options_?)+
+        return_expr?
+    | with_collection_list? for_op*
+        UPSERT expr INSERT expr (UPDATE | REPLACE) expr IN collection options_?
+        return_expr?
+    ;
```
KvanTTT, line 75:
```suggestion
data_modification_query
    : with_collection_list? for_op*
        ( INSERT object_literal in_into collection options_?  
        | UPDATE (expr WITH)? object_literal IN collection options_?
        | REPLACE (expr WITH)? object_literal IN collection options_?
        | (REMOVE expr IN collection options_?)+
        | UPSERT expr INSERT expr (UPDATE | REPLACE) expr IN collection options_?
        )
      return_expr?
    ;
```

aql/ArangoDbParser.g4
```diff
@@ -0,0 +1,220 @@
...
+    : id_ | BIND_PARAMETER_COLL
+    ;
+
+collect
+    : COLLECT expr (INTO expr)? options_?
+    | COLLECT expr (INTO id_ KEEP id_)? options_?
+    | COLLECT expr WITH COUNT INTO id_ options_?
+    | COLLECT expr? aggregate_assign (INTO id_)? options_?
+    | COLLECT WITH COUNT INTO id_ options_?
+    ;
```
KvanTTT, line 133:
```suggestion
collect
    : COLLECT ( expr (INTO expr)?
              | expr (INTO id_ KEEP id_)?
              | expr WITH COUNT INTO id_
              | expr? aggregate_assign (INTO id_)?
              | WITH COUNT INTO id_
              )
      options_?
    ;
```

aql/ArangoDbParser.g4
```diff
@@ -0,0 +1,220 @@
...
+
+window
+    : WINDOW object_literal aggregate_assign
+    | WINDOW expr WITH object_literal aggregate_assign
+    ;
```
KvanTTT, line 138:
```suggestion
window
    : WINDOW (expr WITH)? object_literal aggregate_assign
    ;
```

aql/ArangoDbParser.g4
```diff
@@ -0,0 +1,220 @@
...
+    | expr RANGE expr
+    | expr NOT? (IN | LIKE) expr
+    | ('!' | NOT)+ expr
+    | expr ('&&' | AND) expr
+    | expr ('||' | OR) expr
+    ;
+
+literal
+    : numeric_literal
+    | TRUE | FALSE
```
KvanTTT, line 179:
```suggestion
    | TRUE
    | FALSE
```

aql/ArangoDbParser.g4
```diff
@@ -0,0 +1,220 @@
...
+let_list
+    : (LET expr_list)+
+    ;
+
+id_
+    : ID
+    ;
+
+new_old
+    : NEW | OLD
```
KvanTTT, line 153: Please fix formatting (first-level alternatives on new lines):

```suggestion
    : NEW
    | OLD
```

aql/ArangoDbParser.g4
```diff
@@ -0,0 +1,220 @@
...
+with_collection_list
+    : WITH collection_list
+    ;
+
+collection_list
+    : collection (',' collection)*
+    ;
+
+collection
+    : id_ | BIND_PARAMETER_COLL
```
KvanTTT, line 124:
```suggestion
    : id_
    | BIND_PARAMETER_COLL
```

aql/ArangoDbParser.g4
```diff
@@ -0,0 +1,220 @@
...
+    | new_old
+    ;
+
+array_literal
+    : '[' expr_list? ']'
+    ;
+
+object_literal
+    : '{' pair_list? '}'
+    | '{' id_ (',' id_)* '}' // accepted in RETURN clause
```
KvanTTT, line 195:
```suggestion
    : '{' (pair_list | id_ (',' id_)*)? '}' // accepted in RETURN clause
```

KvanTTT (changes requested):
Thank you for the new grammar but please factor common parts and fix formatting.

mlorek:
php build fails with
image

KvanTTT:
@kaby76 could you please take a look at PHP build? Does it make sense to restart the build?

kaby76:
The PHP target for this grammar uses more than the available memory in PHP 8.1.12. If it doesn't work for that version of PHP, then it won't work for 8.0.x, which is being used in the CI build. In this case, php doesn't terminate. Fortunately, we have the trwdog app making sure this doesn't kill the build so much that intervention is required.

But, the target doesn't work well with PHP. I suggest you just add "aql" to the [skip list for PHP](https://github.com/antlr/grammars-v4/blob/master/_scripts/skip-php.txt). Doesn't matter where, but in alphabetic order is preferred.

Now, as far as what the parser fails on is "for.aql". This is a 2-line query. The grammar is ambiguous for expression, requiring a very, very large read ahead to determine. I think the grammar has some problems. But, this grammar should not fail like this for PHP. An issue should be created in antlr/antlr4 to determine why it's so bad.

kaby76:
OK, the problem with this grammar is the choice between `expr : ... expr (',' expr)+ ...` and `expr : ... expr '[' (expr_list | '*') ']' ...` in which we have `expr_list : expr (',' expr)*;`. It cannot decide where the expr ends and where it begins in the expr_list. Suggest the alt be `expr '[' (expr | '*') ']'`, or something like that.

mlorek:
> Yes, this looks good. Does it work with "mvn clean test"? If so, then this should work for PHP--and speed up all the other targets as well.

it did work locally

kaby76:
Oh well--still doesn't work for PHP. After the build settles down, add aql to the skip-php.txt file to get past this problem for now.

kaby76:
I added a "-d memory_limit=4G" option to the call to php on the command line, and although it runs longer, it still fails with a timeout. All tests parse fine except for.aql, and quickly. So, there is something quite wrong with the PHP target. This needs to be raised in github.com/antlr/antlr4/issues. Working with 4.11.1 and PHP 8.1.12

KvanTTT (approved):

kaby76:
The PR was merged, but the build was still broken. skip-php.txt should contain aql. I'll submit a PR to fix this, and an issue in antlr/antlr4 for the problem in parsing for.aql.

kaby76:
The out of memory condition for aql/for.aql has been observed in another grammar. https://github.com/antlr/antlr-php-runtime/issues/12#issuecomment-1356316535 Tracking it down. (We could change the grammar, but then the bug will still be there.)

PR #2966, opened 2022-12-15
Give the various `expr`s in `select_core` labels.

markww:
In this antlr grammar, `select_core` is a complex grammatical element containing many subpieces. I think it was written this way to mirror the structure of the documentation of `select_core` in [the official SQLite docs](https://www.sqlite.org/syntax/select-core.html). This would be somewhat tricky for a SQLite visitor to read, as the role of the various `expr`s could be ambiguous.

This change adds labels to those `expr`s to allow visitors to quickly see what role each of the `expr`s is playing, reducing the possibility for confusion.


teverett:
@KvanTTT  good to merge this?

KvanTTT:
Yes, good for me, thanks!

PR #2968, opened 2022-12-16
Fix build for aql grammar/PHP target.

kaby76:
* Added grammar aql to the PHP skip list so that the target is not tested on this grammar. All other targets tested.


teverett:
@kaby76  thanks!

PR #2977, opened 2022-12-26
Partial fix for #3462, agc grammar

kaby76:
This PR changes several symbols in the grammar to allow PHP to work.

PHP tests work although several take 2m. The ATN trace and PT output is the same with C#.

```
#
err=0
SAVEIFS=$IFS
IFS=$(echo -en "\n\b")
for g in `find examples -type f | grep -v '.errors$' | grep -v '.tree$'`
do
  file="$g"
  x1="${g##*.}"
  if [ "$x1" != "errors" ]
  then
    echo "$file"
    pushd Generated-PHP
    php -d memory_limit=1G Test.php -tree -trace "../$file" > o.txt
    dos2unix o.txt
    popd
    pushd Generated-CSharp
    ./bin/Debug/net6.0/Test.exe -tree -trace "../$file" > o.txt
    dos2unix o.txt
    popd
    diff Generated-PHP/o.txt Generated-CSharp/o.txt
  fi
done
```

### DIFF
--- a/agc/agc.g4
+++ b/agc/agc.g4
@@ -87,31 +87,31 @@ argument
    ;
 
 ws
-   : WS
+   : WS_
    ;
 
 eol
-   : WS? EOL
+   : WS_? EOL_
    ;
 
 comment
-   : COMMENT
+   : COMMENT_
    ;
 
 label
-   : LABEL
+   : LABEL_
    | register_
    | standard_opcode
    | pseudo_opcode
    ;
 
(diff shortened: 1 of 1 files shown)


kaby76:
Let's call the partial list here. I need to work on the grouped parsing and PHP runtime working.

kaby76:
Needing to refork this repo, and squash commits. Will open a new PR.

PR #2980, opened 2022-12-29
Allow escapes in template strings

### LINKED ISSUE #2978: Typescript template string grammar can't process escaped backtick
Hi, I'm new to antlr and trying to make a typescript-like language.
using lexer and parser from `https://github.com/antlr/grammars-v4/tree/master/javascript/typescript`, failed to parse the escaped `` ` ``
```javascript
// parse failed by antlr defined grammar
// native output: a = '`'
var a = ` \` `;

// a = '${'
var a = `\${`

```
besides, is it costly to define every char in the template string as a node? can we define a rule to merge the char into string fragments?
in the native typescript grammar, the above statement works well.

bkiers:
Fixes #2978

### DIFF
--- a/javascript/typescript/TypeScriptLexer.g4
+++ b/javascript/typescript/TypeScriptLexer.g4
@@ -195,9 +195,10 @@ UnexpectedCharacter:            . -> channel(ERROR);
 
 mode TEMPLATE;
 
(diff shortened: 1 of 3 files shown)


KvanTTT:
Thanks!

PR #2984, opened 2022-12-30
Fixes for Snowflake SQL grammar

mattmcd:
A few fixes for the Snowflake SQL grammar:
* table_type rule - local or global qualifier is optional
* added create_table_as_select rule
* fixed typo in DEFAULT_DDL_COLLATION_ usage in session_parameter rule
* fixes to at_before rule: AT -> AT_KEYWORD, ARROW -> ASSOC 
* added IFF builtin function
* added builtin functions, data types, and keywords to id_ rule


mattmcd:
Looks like the PHP test build step is failing due to timeout on testing select   https://github.com/antlr/grammars-v4/actions/runs/3806751652/jobs/6489098075#step:8:5917

mattmcd:
It looks like the timeout of PHP build was caused by the test for datatypes as column names i.e. in `select.sql`

    select user, date, iff, current_timestamp, cluster, type, region, function from edge;

This test is currently commented out so I'm not sure what is the best approach here: the original version of the grammar would error if it tried to parse a table with data types as column names, which is incorrect as these tables
work fine in Snowflake.  My updated `id_` rule fixes this but times out on the PHP target.

teverett:
@KvanTTT good to merge this?

mattmcd:
@teverett @KvanTTT are there any changes you think are needed to this PR?

KvanTTT:
Thanks for the contribution!

Issue #2988, opened 2023-01-03
Change individual to grouped parsing


kaby76:
Testing in the V4 repo parse one input file per test program, which I call ***individual parsing***. But, we can save a good deal of time by parsing multiple input files per test program, called ***grouped parsing***. Let's see what this looks like in the output from a build:
```
Testing java/java9

~/issues/issue-2988/before/java/java9/Generated ~/issues/issue-2988/before
bash test.sh
../examples/AllInOne8.java
Time: 00:00:05.9106526
Parse succeeded.
../examples/helloworld.java
Time: 00:00:00.5270543
Parse succeeded.
../examples/IdentifierTest.java
Time: 00:00:00.6478677
Parse succeeded.
../examples/Instanceof.java
Time: 00:00:00.2960225
Parse succeeded.
../examples/ManyStringsConcat.java
Time: 00:00:00.1571820
Parse succeeded.
../examples/module-info.java
Time: 00:00:00.0614978
Parse succeeded.
../examples/TryWithResourceDemo.java
Time: 00:00:00.7459264
Parse succeeded.
../examples/Unicode.java
Time: 00:00:00.5391444
Parse succeeded.
Duration: 0 hours 0 minutes 12 seconds
```
With _grouped parsing_, the run time for inputs after the first are much shorter:
```
Testing java/java9

~/issues/issue-2988/grammars-v4/java/java9/Generated-CSharp ~/issues/issue-2988/grammars-v4
bash test.sh
CSharp 0 ../examples/AllInOne8.java success 6.1742676
CSharp 1 ../examples/helloworld.java success 0.0255556
CSharp 2 ../examples/IdentifierTest.java success 0.1810686
CSharp 3 ../examples/Instanceof.java success 0.0866523
CSharp 4 ../examples/ManyStringsConcat.java success 0.0374591
CSharp 5 ../examples/module-info.java success 0.0031916
CSharp 6 ../examples/TryWithResourceDemo.java success 0.1530053
CSharp 7 ../examples/Unicode.java success 0.044574
Total Time: 6.987732
```
Grouped parsing will help for Java, CSharp, and probably a few other targets. But, it may not fix targets like PHP, which show no speed-up with warm-up. See https://github.com/antlr/antlr-php-runtime/issues/36.

To implement grouped parsing, the newest version of Trash trgen will need to be employed, as well as the templates in this repo updated.

There is still a significant amount of time required to generate and compile the drivers, so the builds will still be slow.

PR #2989, opened 2023-01-04
Fix for Issue 2988: Change individual to grouped parsing

kaby76:
## Statement of Problem
This PR is to address https://github.com/antlr/grammars-v4/issues/2988.

Tests in the V4 repo parse one input file per program invocation, which I call ***individual parsing***. But, we can save a good deal of time by parsing multiple input files per program invocation, called ***grouped parsing***. Let's see what this looks like in the output from a build.
* For `Individual parsing`, one test is processed by one program call:
```
./bin/Debug/net6.0/Test.exe ../examples/AllInOne8.java
Time: 00:00:05.9106526
Parse succeeded.
./bin/Debug/net6.0/Test.exe ../examples/helloworld.java
Time: 00:00:00.5270543
Parse succeeded.
./bin/Debug/net6.0/Test.exe ../examples/IdentifierTest.java
Time: 00:00:00.6478677
Parse succeeded.
...
```
* For `Grouped parsing`, all tests are processed by one program call:
```
./bin/Debug/net6.0/Test.exe ../examples/AllInOne8.java ../examples/helloworld.java ../examples/IdentifierTest.java ...
CSharp 0 ../examples/AllInOne8.java success 6.1742676
CSharp 1 ../examples/helloworld.java success 0.0255556
CSharp 2 ../examples/IdentifierTest.java success 0.1810686
...
```
With _grouped parsing_, the time for parsing each successive input *should* be shorter relative to the time the input took with _individual parsing_. Internally, Antlr keeps a cache of [prediction context](https://github.com/antlr/antlr4/blob/master/runtime/CSharp/src/Atn/PredictionContext.cs) between each parse for faster parsing. Grouped parsing should help for Java, CSharp, and probably a few other targets. But, it may not help with all targets, especially PHP, which shows no speed-up with warm-up, likely a bug. See https://github.com/antlr/antlr-php-runtime/issues/36.

## Requirements, Design, and Implementation
To implement grouped parsing, the newest version of Trash trgen will need to be employed, as well as the templates in this repo updated. This PR changes the CI build tests, and some of the `.errors` and `.tree` files (more on this explained below).

After a detailed reading and testing of the current _scripts/test.ps1 and templates, I came to the conclusion that there was a significant disconnect in the requirements, design, and implementation between the [Bash](https://github.com/antlr/grammars-v4/blob/56dcd3d9251506d7514a68e86a4c192735c48f59/_scripts/regtest.sh) and [Powershell](https://github.com/antlr/grammars-v4/blob/56dcd3d9251506d7514a68e86a4c192735c48f59/_scripts/test.ps1) test scripts. Therefore, we need to state some requirements for these scripts and the generated scripts for trgen.
#### Requirements
* The basic requirement for the Bash or Powershell test scripts is to provide a simple, self-contained method to build and test the code in a Bash or Powershell environment.
* The scripts should take minimal parameters so that people can focus on updating changes to a grammar rather than wasting time figuring out how to test the grammar across eight targets. (Note, to be implemented for #2883.)
* trgen should generate scripts [test.sh](https://github.com/kaby76/Domemtech.Trash/blob/8f7b1174e40f6ac66500e49b1a854eeb6bd104c8/src/trgen/templates/Java/test.sh) and [test.ps1](https://github.com/kaby76/Domemtech.Trash/blob/8f7b1174e40f6ac66500e49b1a854eeb6bd104c8/src/trgen/templates/Java/test.ps1). The Bash and Powershell scripts should mirror each other step-by-step. The scripts should also mirror each other across targets (e.g., [Java](https://github.com/kaby76/Domemtech.Trash/blob/8f7b1174e40f6ac66500e49b1a854eeb6bd104c8/src/trgen/templates/Java/test.sh) and [Dart](https://github.com/kaby76/Domemtech.Trash/blob/8f7b1174e40f6ac66500e49b1a854eeb6bd104c8/src/trgen/templates/Dart/test.sh).
* There should be scripts for [building](https://github.com/kaby76/Domemtech.Trash/blob/8f7b1174e40f6ac66500e49b1a854eeb6bd104c8/src/trgen/templates/Java/build.ps1), [testing](https://github.com/kaby76/Domemtech.Trash/blob/8f7b1174e40f6ac66500e49b1a854eeb6bd104c8/src/trgen/templates/Java/test.sh), and [cleaning up](https://github.com/kaby76/Domemtech.Trash/blob/8f7b1174e40f6ac66500e49b1a854eeb6bd104c8/src/trgen/templates/Java/clean.sh), both Bash and Powershell.
* The top-level scripts [regtest.sh](https://github.com/antlr/grammars-v4/blob/fd9daa7b0cff4c64bdb527133e6dfa217d15310b/_scripts/regtest.sh) and test.ps1 (replacing [tester.ps1](https://github.com/antlr/grammars-v4/blob/fd9daa7b0cff4c64bdb527133e6dfa217d15310b/_scripts/test.ps1)) should be incremental (taking into account git status), and runable from any directory (including a grammar, or a collection of grammars, like asm/, antlr, or sql/). (**NB: This requirement will be addressed in https://github.com/antlr/grammars-v4/issues/2883.**)
* With grouped parsing, the parser driver program must generate `.errors` and `.tree` files; it cannot be redirected because output for all inputs would then be grouped together. NB: Careful attention is placed on capturing all parser output. Any output to stderr that is not for trees and standard parsing errors is considered an error in the execution of the parser. For example, this can happen especially with dynamically typed and parsed languages like JavaScript, Python3, and PHP in which the base class is not correct, and the interpreter throws an error.
* [git](https://git-scm.com/) will be used to implement diffing of the `.errors` and `.tree` files.
#### Discussion of requirements
It's hard to remember how to run the compiler and program of the parser program for a particular target. It's even worse trying to remember the syntax and runtimes for the targets. That's the whole point of trgen.

Unfortunately, [tester.psm1](https://github.com/antlr/grammars-v4/blob/c59afdbe12c110f709a8e3bc7969a20abe874413/_scripts/templates/CSharp/tester.psm1) is not a stand-alone Powershell script because the extension is ".psm1", not ".ps1".

Further, the script assumed the test input was always [in a directory called "examples/"](https://github.com/antlr/grammars-v4/blob/c59afdbe12c110f709a8e3bc7969a20abe874413/_scripts/test.ps1#L59). [trgen](https://github.com/kaby76/Domemtech.Trash/tree/main/src/trgen) reads the pom.xml to get the location of the input tests.

Since grouped parsing can't use Bash/Powershell capture of the parse tree--because there would be multiple trees captured for multiple input files--the driver program must place the output in a specific output file. But, then, how does the script check diffs of these parse tree files??

In order to implement parse tree file diffs, I use `git diff` to find those files that are changed. It's laborious to write code to do diffs in Bash and Powershell. In fact, the Powershell "diff" strips out newlines of a multiline output file contained in a string. Let's make the code simpler by just using "git". If there's no git, or the grammar has been removed from the rest of the cloned repo, then the tests should just default to returning any errors found, even if they are expexted.

The `.errors` files seem to have extra newlines at the end. Extremely annoying. I don't know if they are hand-edited or the drivers that created them include extra printf's for the hell of it. In any case, I remastered a number of these files with this PR in order for the tests to pass.

***From this point forward, test.sh or test.ps1 should be used to remaster all `.tree` and `.errors` files, as pondered [in this comment](https://github.com/antlr/grammars-v4/pull/2925#issuecomment-1368787212).***
## Results
Although not scientific, I compared one testing of the V4 grammars with _individual parsing_ vs _grouped parsing_. On an unloaded Windows 11 machine (Ryzen 7 2700, 8-cores, DDR4 16GB, SanDisk SDSSDH3 1TB, Antlr4.11.1, NET SDK 7.0.101), _individual parsing_ of the CSharp target completed in `1h 44m`, _grouped parsing_ completed in `1h 11m`. That is significantly faster.

## Conclusions
This is an important PR. Grouped parsing helps a lot in the parse time for the build. But, overall, the main bottleneck in the build is the compilation of the generated code. It really goes back to https://github.com/antlr/grammars-v4/issues/2883. The "incremental" testing isn't really working because, somehow, the information on what specifically to test is ignored. In addition, we shouldn't be testing combined grammars because they don't have target-specific action code.


kaby76:
For a C# target test, the combined time for parsing all of the test-suite took (started at `Wed Jan  4 15:38:22 UTC 2023` and ended at `Wed Jan  4 15:46:36 UTC 2023`). That's 8m 14s.

| Phase | Time |
|---|---|
| Generate | 14s |
| Compile | 1h 4m |
| Run | 8m 14s |

For comparison.
| Phase | Time |
|---|---|
| Generate | 13m 58s |
| Compile | 1h 17m 38s |
| Run | 34m 22s |

If you consider the slowdown linear between these two runs, then the parsing should be 1.2x times slower if the old way (`(480+14)*1.2 = 592.2s = 9m 52s. But it was much slower the old way, 34m 22s.

kaby76:
In order to implement parse tree file diffs, I use `git diff` to find those files that are changed. The `.errors` files seem to have extra newlines at the end. Extremely annoying. Also, this helps to remaster all .tree and .errors files, as posed [here](https://github.com/antlr/grammars-v4/pull/2925#issuecomment-1368787212).

kaby76:
The next set of changes are for testing CSharp in a Powershell environment. After a detailed reading and testing of the current _scripts/test.ps1 and templates, it's clear there was a significant disconnect of the requirements for building. ***The point of makefile and test.sh is provided a simple, self-contained method to build and test the code in an environment, which completely hides how to call the target-specific compiler and runner.***

[tester.psm1](https://github.com/antlr/grammars-v4/blob/c59afdbe12c110f709a8e3bc7969a20abe874413/_scripts/templates/CSharp/tester.psm1) is clearly not a stand-alone script because the extension is ".psm1", not ".ps1". In order to call the program, I had to say "Import-Module ./tester.psm1", then call "Test-Case(....)", feeding in parameters I have no clue what to do. That is very opaque compared to just typing "./tester.ps1" in the generated driver code directory, which is the requirement.

Further, the whole point of [trgen](https://github.com/kaby76/Domemtech.Trash/tree/main/src/trgen) is to let this tool generate a driver without you know jack about the pom.xml file. Unfortunately, test.ps1 [hardwired "examples/"](https://github.com/antlr/grammars-v4/blob/c59afdbe12c110f709a8e3bc7969a20abe874413/_scripts/test.ps1#L59) as the test inputs. Why this code is in test.ps1 instead of tester.psm1 is beyond me. 

So, the tester.psm1 was completely reworked to follow the requirements and style of makefile and test.sh, but for a Powershell environment.

kaby76:
* Grouped parsing requires more time per program because more files are being tested per program run. Right now, the default is 300s.
* sql/mysql-Python3 combination times out on the second input file. But, combing through the old Action logs, I find out that [sql/mysql](https://github.com/antlr/grammars-v4/actions/runs/3630269742/jobs/6123427602#step:17:7233) was never actually being tested on input files! JFC.

kaby76:
@teverett @KvanTTT After working on this PR for two weeks, I think I finally have this PR ready for review. The new Bash and Powershell scripts should resemble each other more closely and across all Antlr targets. That should make it easier to understand and maintain.

After this, I plan to fix https://github.com/antlr/grammars-v4/issues/2883

kaby76:
Needing to refork this repo, and squash commits. Will open a new PR.

PR #2992, opened 2023-01-05
fix a bug

k88936:
the "compilationUnit" was put behind so when using antlr preview, it will load primaryExpression instead (at least in idea) ,so i just changed the order

### DIFF
--- a/c/C.g4
+++ b/c/C.g4
@@ -29,6 +29,13 @@
 /** C 2011 grammar built from the C11 Spec */
 grammar C;
 
+compilationUnit
+    :   translationUnit? EOF
+    ;
+
+translationUnit
+    :   externalDeclaration+
+    ;
 
 primaryExpression
     :   Identifier
@@ -484,13 +491,7 @@ jumpStatement
     ';'
     ;
 
-compilationUnit
-    :   translationUnit? EOF
-    ;
 
-translationUnit
-    :   externalDeclaration+
-    ;
 
 externalDeclaration
     :   functionDefinition


kaby76:
> the "compilationUnit" was put behind so when using antlr preview, it will load primaryExpression instead (at least in idea) ,so i just changed the order

The order of parser rules does not change how the parser works. Can you please explain the bug fully?

k88936:
yes it is ,but in PLUGINS in idea the preview function seemingly to read the first one as Start Rule, so before that it use "primaryExpression"and preview not function well.....and I noticed other .g4 here (java,cpp) all start with somethingUnit

kaby76:
Antlr does not have syntax in the grammar itself to declare a start symbol, so we are forced to come up with conventions to declare the "start rule". https://github.com/antlr/antlr4/discussions/3763

This repo explicitly declares the start rule(s) in the pom.xml file in order to be able to test the grammar. That is what should be used. Note, in some pom.xml, there are multiple entry points declared.

Another convention would be to explicitly declare a rule that has an EOF at the end of the rule. Not only does this help to declare the start rule, but it's also necessary in order to force the parser to consume all the input. That is why I modified hundreds of grammars in this repo to force the parser to read the input to completion. https://github.com/antlr/grammars-v4/pull/2675

A third convention would be to declare the start rule as the first rule in the grammar. But, some grammars have multiple "start rules". And, the order of the grammar rules was chosen because it was alphabetic (for fast searching), or in the order that was declared in an official document that declares the grammar for the language (e.g., the Java Language Spec). For C and C++, I have automated scrapers that read the specs for C and C++ online and convert into Antlr grammars.

There are scores of grammars that do not follow the convention of "first rule is the start rule":
```
./ada/ada2005/Ada2005Parser.g4
./ada/ada2012/AdaParser.g4
./ada/ada83/Ada83Parser.g4
./ada/ada95/Ada95Parser.g4
./algol60/algol60.g4
./antlr/antlr4/LexBasic.g4
./bcpl/bcpl.g4
./bcpl/origbcpl.g4
./c/C.g4
./caql/CaQLParser.g4
./caql/Metrics.g4
./clif/CLIF.g4
./cql3/CqlParser.g4
./csharp/CSharpPreprocessorParser.g4
./dart2/Dart2Parser.g4
./datalog/datalog.g4
./fortran77/Fortran77Parser.g4
./glsl/GLSLPreParser.g4
./html/HTMLParser.g4
./iso8601/iso8601.g4
./java/java8/Java8Parser.g4
./java/java9/Java9Parser.g4
./javascript/typescript/TypeScriptParser.g4
./kotlin/kotlin/UnicodeClasses.g4
./kotlin/kotlin-formal/UnicodeClasses.g4
./modula2pim4/m2pim4.g4
./mumps/mumps.g4
./oberon/oberon.g4
./oncrpc/oncrpcv2.g4
./oncrpc/xdr.g4
./powerquery/PowerQueryParser.g4
./properties/PropertiesParser.g4
./python/python2/Python2.g4
./python/python2-js/Python2.g4
./python/python3/Python3Parser.g4
./python/python3-py/Python3.g4
./python/python3-ts/Python3.g4
./python/python3alt/AltPython3.g4
./scala/Scala.g4
./sql/hive/v2/FromClauseParser.g4
./sql/hive/v2/IdentifiersParser.g4
./sql/hive/v2/SelectClauseParser.g4
./sql/hive/v3/FromClauseParser.g4
./sql/hive/v3/IdentifiersParser.g4
./sql/hive/v3/ResourcePlanParser.g4
./sql/hive/v3/SelectClauseParser.g4
./stringtemplate/LexBasic.g4
./stringtemplate/LexUnicode.g4
./tcpheader/tcp.g4
./unicode/graphemes/Graphemes.g4
./verilog/systemverilog/SystemVerilogPreParser.g4
./verilog/verilog/VerilogPreParser.g4
./vhdl/vhdl.g4
./wat/WatParser.g4
```

PR #2997, opened 2023-01-06
Factor the `VALUES` clause into its own grammatical element.

markww:
Prior to this commit, the `VALUES` clause was written in a fashion directly mirroring [its description in the SQLite
documentation](https://www.sqlite.org/syntax/select-core.html). The problem with this approach is that `VALUES` does not have a flat structure; the parens indicate the start and end of rows in the table produced by the `VALUES` clause. It is easier for visitors to handle this structure if the `VALUES` clause is factored to make each of the rows be parsed in their own parser context.

In addition, this allows factoring the code shared between the `INSERT` and `SELECT` statements so that they can share the same specification of their `VALUES` clause.


KvanTTT:
Please cover your changes with some examples (put them into `sql\sqlite\examples` directory).

markww:
I've added some examples showing the top-level `VALUES` case. The `VALUES` clause in an `INSERT` is already covered in `examples/returning.sql`.

PR #3001, opened 2023-01-10
feat: add column aliases, scoped table ref for CREATE MATERIALIZED VIEW in sql/plsql

poWer4aiX:
Option to specify a column alias list and scoped tables reference constraints in the CREATE MATERIALIZED VIEW statement for sql/plsql (was commented as TODO)

### DIFF
--- a/sql/plsql/PlSqlParser.g4
+++ b/sql/plsql/PlSqlParser.g4
@@ -1800,7 +1800,7 @@ create_zonemap_as_subquery
 create_materialized_view
     : CREATE MATERIALIZED VIEW tableview_name
       (OF type_name )?
-//scoped_table_ref and column alias goes here  TODO
+        ( '(' (scoped_table_ref_constraint | mv_column_alias) (',' (scoped_table_ref_constraint | mv_column_alias))* ')' )?
         ( ON PREBUILT TABLE ( (WITH | WITHOUT) REDUCED PRECISION)?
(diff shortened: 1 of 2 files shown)


sql/plsql/PlSqlParser.g4
```diff
@@ -1814,6 +1814,23 @@ create_materialized_view
...
+        (schema_name '.')? scope_table_name_or_c_alias
+    ;
+
+ref_column_or_attribute
+    : identifier
+    ;
+
+scope_table_name_or_c_alias
+    : identifier
+    ;
```
KvanTTT, line 1828: Do we have strong need of introducing new rules that only have `identifier` inside? Could `identifier` be inlined in the respective use-sites?

```suggestion
scoped_table_ref_constraint
    : SCOPE FOR '(' identifier ')' IS (schema_name '.')? identifier
    ;
```

Also, [rule element labels](https://github.com/antlr/antlr4/blob/master/doc/parser-rules.md#rule-element-labels) can be used if necessary:

```sql
scoped_table_ref_constraint
    : SCOPE FOR '(' ref_column_or_attribute=identifier ')' IS (schema_name '.')? scope_table_name_or_c_alias=identifier
    ;
```

poWer4aiX: Thanks for pointing out the labeling feature, where I was not aware of. It simplifies the the resulting parser implementation while keeping the information within the grammar.

PR #3002, opened 2023-01-10
fortran77: fix gotoStatement

BBreiden:
The gotoStatement should accept "GO TO" and "GOTO".

### DIFF
--- a/fortran77/Fortran77Lexer.g4
+++ b/fortran77/Fortran77Lexer.g4
@@ -535,7 +535,7 @@ CTRLREC
 
 
 TO
-   : 'TO'
+   : 'TO' | 'to'
    ;
 
 
--- a/fortran77/Fortran77Parser.g4
+++ b/fortran77/Fortran77Parser.g4
@@ -338,7 +338,7 @@ dataImpliedDoListWhat
    ;
 
 gotoStatement
-   : ((GO | GOTO) to) (unconditionalGoto | computedGoto | assignedGoto)
+   : (GO TO| GOTO) (unconditionalGoto | computedGoto | assignedGoto)
    ;
 
 unconditionalGoto
@@ -881,8 +881,4 @@ logicalConstant
 identifier
    : NAME
    | REAL
-   ;
-
-to
-   : NAME
    ;
--- a/fortran77/examples/goto-example.txt
+++ b/fortran77/examples/goto-example.txt
@@ -6,9 +6,11 @@ c
 
       go to 20
 
-10    write ( *, '(a)' ) '  Hello, world!'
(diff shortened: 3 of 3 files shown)


KvanTTT:
Could you please add an example (to `fortran77/examples`) that covers suggested changes?

BBreiden:
I have added an example. I hope this is ok. Otherwise just let me know!

KvanTTT:
CI says there are errors on Fortran: https://github.com/antlr/grammars-v4/actions/runs/3890697712/jobs/6644788312#step:17:52

BBreiden:
I will try to reproduce and try to fix it.

kaby76:
The old tester.psm1/test.ps1 scripts doesn't print out the error messages for [the new test input](https://github.com/antlr/grammars-v4/actions/runs/3890697712/jobs/6644788312#step:17:48). That's will be fixed with my rewrite of the test scripts. https://github.com/antlr/grammars-v4/pull/2989
```
Java 0 ../examples/example1.txt success 0.03
Java 1 ../examples/example2.txt success 0.018
line 7:9 missing 'TO' at 'to'
line 7:12 extraneous input '20' expecting EOL
line 9:6 no viable alternative at input '10write'
line 12:6 no viable alternative at input '20goto'
Java 2 ../examples/goto-example.txt fail 0.03
Total Time: 0.229
```

BBreiden:
The TO token was only defined as uppercase but the example uses lower case. I fixed the token file. Can we please rerun CI?

kaby76:
> The TO token was only defined as uppercase but the example uses lower case. I fixed the token file. Can we please rerun CI?

Is it Fortran77 if lowercase is used? Fortran77 uses only uppercase for the language but can process lowercase data. I recall wasting a lot of money on buying punch cards on this issue.

BBreiden:
Well, you're more expert than me, if you remember punch cards... :-)
But you are correct, I think. Looking at the Oracle documentation, they only use uppercase. However, the current version of the grammar allows lower case. The TO was an exception of not being defined in both (upper and lower) cases. The example2.txt also uses lower case. And GNU F77 didn't complain, when I tried to compile my example... ;-)

kaby76:
If other keywords are accepted in lowercase in the grammar, make it consistent.  But the compiler is wrong and I do recall the CDC compiler barfing on l.c. code. It makes me wonder what else is wrong with this grammar.

BBreiden:
I think, it is far from perfect at the moment. E.g. labels are only allowed on CONTINUE statements but not on other executable statements. Also in an ELSE IF statement you cannot use comments (I opened another PR to fix this).

PR #3003, opened 2023-01-10
fortran77: elseIfStatement allows comments

BBreiden:
elseIfStatement should also allow comments mixed with statements like the thenStatement or elseStatement.

### DIFF
--- a/fortran77/Fortran77Parser.g4
+++ b/fortran77/Fortran77Parser.g4
@@ -31,13 +31,8 @@ parser grammar Fortran77Parser;
 options
    { tokenVocab = Fortran77Lexer; }
 
-// multi-line comments?
-commentStatement
-    : COMMENT+
-    ;
-    
 program
-   : commentStatement* (executableUnit commentStatement*)+ EOL* EOF
+   : executableUnit+ EOL* EOF
    ;
 
 executableUnit
@@ -113,7 +108,7 @@ statement
    ;
 
 subprogramBody
-   : commentStatement* (wholeStatement commentStatement*)+ endStatement
+   : wholeStatement+ endStatement
    ;
 
 wholeStatement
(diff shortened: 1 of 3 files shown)


fortran77/Fortran77Parser.g4
```diff
@@ -382,7 +382,7 @@ firstIfBlock
    ;
 
 elseIfStatement
-   : (ELSEIF | (ELSE IF)) LPAREN logicalExpression RPAREN THEN EOL? wholeStatement+
+   : (ELSEIF | (ELSE IF)) LPAREN logicalExpression RPAREN THEN EOL? commentStatement* (wholeStatement commentStatement*)+
```
KvanTTT, line 385: Does it make sense to put comments to the `HIDDEN` channel (`-> channel(HIDDEN)`)? 

https://github.com/antlr/grammars-v4/blob/38857b99fbb6effa129e2d49182d3814013f8e46/fortran77/Fortran77Lexer.g4#L652-L654

In this case there won't be need of using `commentStatement` on parser level.

BBreiden:
I have added an example for the if statement with an else-if statement.

fortran77/Fortran77Parser.g4
KvanTTT: Sorry, but I don't think it's a good fix of the current problem. In the current implementation `commentStatement` is placed to certain parse rules but is not placed to others. That's why potential bugs still can be encountered. I suggest using standard technics for ignoring comments:

* Add `-> channel(HIDDEN)` to [`COMMENT`](https://github.com/antlr/grammars-v4/blob/master/fortran77/Fortran77Lexer.g4#L652) lexer rule. In this case lexer just puts tokens to another channel that is invisible for parser. `HIDDEN` channel is reserved in ANTLR and it's not neccessary to introduce a new one. Read more about lexer commands in the [ANTLR doc](https://github.com/antlr/antlr4/blob/master/doc/lexer-rules.md#channel).
* Remove `commentStatement` and its using from the parser.

KvanTTT (changes requested):
Please make `COMMENT` ignorable.

fortran77/Fortran77Parser.g4
BBreiden: OK. I will try to update the grammar according to your suggestions. Thanks for pointing it out to me. 👍

BBreiden:
I redirected the comments to the `HIDDEN` channel and removed the `commentStatement`. In my setup the examples are still parsed without errors.

PR #3004, opened 2023-01-10
fortran77: continue should only allow single label

BBreiden:


BBreiden:
I have added an example. Hope this is ok!

KvanTTT:
Unfortunately, there are errors on CI: https://github.com/antlr/grammars-v4/actions/runs/3891113700/jobs/6644828450#step:17:44

BBreiden:
I will try to reproduce this error and try to fix it.

BBreiden:
I was able to reproduce the problem locally and updated the example. Can you pls rerun CI?

KvanTTT:
Thanks for the contribution and welcome to the ANTLR grammars repository!

PR #3005, opened 2023-01-10
[C#] Avoid member accesses being parsed as `type_` when followed by `unary_expressions`.

MEPalma:
Fixes `C#` parser to prevent member accesses being parsed as `type_` sub-productions when followed by `unary_expressions`. This is achieved by ensuring that `expression` productions have precedence over `type_` productions.

For example in the following derivation:
```csharp
public class ClassName {
    Constructor MethodName(Object a, String b, int c) {
        return new Constructor(field, a.b, a.b + c.d, e.f.g + h);
    }
}
```
sub-derivation `a.b + c.d` is now parsed as:
```
...argument...additive_expression(
    multiplicative_expression...(...a, member_access('.', ...b)),
    '+',
    multiplicative_expression...(...a, ...member_access('.', ...b)),
 )
```

instead of:
```
...argument(
    _type...(a, '.', b),
    ...expression...(
        ...'+', 
        ...primary_expression(
            ...a,
            ...member_access('.', ...b)
        )
    )
```

A similar argument can be made for sub-production `a.f.g + h`, which is now parsed as an `additive_epression` of instead of a `_type(e.f.g), unary_expression('+', '+ h')`.

### DIFF
--- a/csharp/CSharpParser.g4
+++ b/csharp/CSharpParser.g4
@@ -16,7 +16,7 @@ compilation_unit
 
 //B.2.1 Basic concepts
 
-namespace_or_type_name 
+namespace_or_type_name
 	: (identifier type_argument_list? | qualified_alias_member) ('.' identifier type_argument_list?)*
 	;
 
@@ -40,18 +40,18 @@ tuple_element
     : type_ identifier?
     ;
 
-simple_type 
+simple_type
 	: numeric_type
 	| BOOL
 	;
 
-numeric_type 
+numeric_type
 	: integral_type
 	| floating_point_type
 	| DECIMAL
 	;
 
-integral_type 
+integral_type
 	: SBYTE
 	| BYTE
 	| SHORT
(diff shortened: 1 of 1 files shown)


teverett:
@MEPalma thanks!

ningpp:
thank you!
this pull request solve my problem when parsing code in below
`avgSum += Math.Sqrt(Math.Pow(x[i] - x[j], 2) + Math.Pow(y[i] - y[j], 2));`

KvanTTT:
@MEPalma or @ningpp could you please send an another PR with an example (to [csharp/examples](https://github.com/antlr/grammars-v4/tree/master/csharp/examples)) that cover changes from the current PR?

MEPalma:
@KvanTTT I have added an example in this PR https://github.com/antlr/grammars-v4/pull/3022, which also includes @ningpp's instance. Thank you!

PR #3006, opened 2023-01-10
[PlSql]feat: handle sqlplus TIMING command

poWer4aiX:


sql/plsql/PlSqlLexer.g4
```diff
@@ -1973,6 +1973,7 @@ TIMEZONE_MINUTE:              'TIMEZONE_MINUTE';
 TIMEZONE_OFFSET:              'TIMEZONE_OFFSET';
 TIMEZONE_REGION:              'TIMEZONE_REGION';
 TIME_ZONE:                    'TIME_ZONE';
+TIMING:                       'TIMING';
```
KvanTTT, line 1976: I guess it should be also placed to [`non_reserved_keywords_in_12c`](https://github.com/antlr/grammars-v4/blob/master/sql/plsql/PlSqlParser.g4#L4981) or [`non_reserved_keywords_pre12c`](https://github.com/antlr/grammars-v4/blob/master/sql/plsql/PlSqlParser.g4#L5295).

KvanTTT (changes requested):

sql/plsql/PlSqlLexer.g4
poWer4aiX: I had the expectation that only such keywords should be added, which were not reserve in 12c / in any release before 12c. As the SqlPlus command `TIMING` exists before 12c (e.g. in [9.2](https://docs.oracle.com/cd/B10501_01/server.920/a90842/ch13.htm#1014053)) it should not be added here, or am I'm wrong?

sql/plsql/PlSqlParser.g4
```diff
@@ -4389,6 +4390,10 @@ set_command
     : SET regular_id (CHAR_STRING | ON | OFF | /*EXACT_NUM_LIT*/numeric | regular_id)
     ;
 
+timing_command
+    : TIMING (START (timing_text=id_expression)* | SHOW | STOP)?
```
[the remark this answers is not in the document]
poWer4aiX: fixed, thanks

sql/plsql/PlSqlLexer.g4
poWer4aiX: I got it now. Without adding `TIMING` to the list of non_reserved_keywords_in_* (part of regular_id) the parser will reject using `TIMING` as an regular_id (e.g. in a `SELECT` statement) as an unexpected identifier, which should not be the case.

Added now to `non_reserved_keywords_pre12c` and extended the test.

KvanTTT (approved):
Thanks a lot for the contribution and fixes!

Issue #3018, opened 2023-01-18
TypeScript Cpp Port LexerBase file missing header file


ppppqp:
**Summary:** TypeScript Cpp Port `TypeScriptLexerBase.cpp` misses the header file `TypeScriptLexer.h`.

**Bug:** Building the target throws the error "use of undeclared identifier 'TypeScriptLexer'" in `TypeScriptLexerBase.cpp`. On closer examination, the cpp file does not include `TypeScriptLexer.h` but [uses](https://github.com/antlr/grammars-v4/blob/56dcd3d9251506d7514a68e86a4c192735c48f59/javascript/typescript/Cpp/TypeScriptLexerBase.cpp#L60) the symbol defined in it. 

The fix is tested locally and submitted as a pull request.

PR #3019, opened 2023-01-18
Fix: TypeScript Cpp port missing header file

ppppqp:
Fixes issue[#3018](https://github.com/antlr/grammars-v4/issues/3018)


teverett:
@ppppqp  thanks!

PR #3022, opened 2023-01-18
[csharp] Example for changes introduced in PR3005

MEPalma:
Adds short example for the changes introduced in https://github.com/antlr/grammars-v4/pull/3005.


teverett:
@MEPalma thanks!

Issue #3027, opened 2023-01-25
TSQL: double quoted strings are not accepted


cceelen:
The current TSQL lexer rule for strings does not allow double quoted strings:
`STRING options { caseInsensitive=false; } : 'N'? '\'' (~'\'' | '\'\'')* '\'' ;`
[Microsoft uses only single quoted strings in their grammar description.](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/execute-transact-sql?view=sql-server-ver16)
This means that the following is rejected:
`CREATE PROCEDURE bla_bla AS
BEGIN
EXEC ("DELETE " + @param + 'more' + @param);
END`
Rejecting this, means that input accepted by a number of other DB engines gets rejected by the generated parser.

cceelen:
I was trying to replace it with this rule:
`STRING options { caseInsensitive=false; } : 'N'? ('\'' (~'\'' | '\'\'')* '\''  | '"' (~'"' | '\\"' )* '"' );`
But this does not seem to work either.

kaby76:
> Rejecting this, means that input accepted by a number of other DB engines gets rejected by the generated parser.

Is this Microsoft TSQL or a product from another company? The MariaDB grammar was forked from the MySQL grammar because MariaDB is not MySQL, and there were no semantic predicates for selecting the MariaDB extensions to the MySQL grammar, nor even commented as an extension. https://github.com/antlr/grammars-v4/pull/2940

cceelen:
Hi @kaby76 
i was using the t-sql grammars and not the specific one for maria DB. 

https://github.com/antlr/grammars-v4/tree/master/sql/tsql

The SQL I want to parse is from Sybase. So it's good to know that we can just go in and override some of the rules without the need to fork or rewrite them for a different engine.

I was trying to solve it in this way:
`STRING options { caseInsensitive=false; } : 'N'? '\'' (~'\'' | '\'\'')* '\''  ;
DOUBLE_QUOTED_STRING options { caseInsensitive=false; } : '"' (~'"' | '\\"' )* '"' ;

execute_var_string
    : LOCAL_ID (OUTPUT | OUT)? ('+' LOCAL_ID ('+' execute_var_string)?)?
    | STRING ('+' LOCAL_ID ('+' execute_var_string)?)?
    | DOUBLE_QUOTED_STRING ('+' LOCAL_ID ('+' execute_var_string)?)?
    ;
`
This kind of worked when using that in the IDE plugin. 
[...]

PR #3031, opened 2023-01-27
#3027 creating a variant of the TSQL to support Sybase

cceelen:
(!) not ready for merge. The test is still failing. (!)


KvanTTT (changes requested):
Thanks for the contribution but I don't think duplication of `TSqlLexer` and `TSqlParser` is a good idea. Is it possible to get rid of duplication?

cceelen:
Hi @KvanTTT 

I'd love to avoid the duplication. Reading up online i was not finding ways to import other grammars from other directories. If there is a way to do this cleanly, then please let me know.
I also had to disable the tokenVocab added in TSqlParser to avoid warnings as a variant was already imported by SybaseParser.g4. If I leave the warning, then generating code from the grammar requires more explicit steps.

Both issues might be worthwhile changes on ANTLR itself.

cceelen:
Hi @KvanTTT,
I ran out of ideas on how to avoid the copy of the tsql grammar files. 
Two issues: 
- Imports are only working within the same directory.
- Even with imports i had to disable the import of the TSqlLexer in the TSqlParser grammer.

I see only two possible solutions:
- accept duplication, or
- merge my changes into the existing TSQL grammar, meaning that code that might be rejected by SQL Server gets accepted by the grammar.

My first thought was to actually merge it into the TSQL grammar, but given the wide consequences I'm not sure every user of that grammar would appreciate the consequences. 

If you see an alternative, then please let me know.

Issue #3033, opened 2023-01-28
PostgreSQL - tableelement rule precedence issue


paulirwin:
ANTLR 4.11.1. I encountered a likely precedence issue with the PostgreSQL `tableelement` rule. For reference, here is the rule in this repo today:

```antlr
tableelement
   : columnDef
   | tablelikeclause
   | tableconstraint
   ;
```

When trying to parse the Sakila sample database schema, this parses the `CONSTRAINT` as a `columnDef` rule with column name `CONSTRAINT`:

```sql
CREATE TABLE payment_p2007_01 (CONSTRAINT payment_p2007_01_payment_date_check CHECK (((payment_date >= '2007-01-01 00:00:00'::timestamp without time zone) AND (payment_date < '2007-02-01 00:00:00'::timestamp without time zone)))
)
INHERITS (payment);
```

I have fixed this locally by swapping the order of the rules. Since `tableconstraint` has the lexer token `CONSTRAINT` as its first child, this correctly parses the CONSTRAINT as a `tableconstraint` instead of a `columnDef`:

```antlr
tableelement
   : tableconstraint
   | tablelikeclause
   | columnDef
   ;
```

Creating this issue for discussion and to make sure that my understanding of the grammar is correct. I'll submit a PR for this change as well.

PR #3034, opened 2023-01-28
[PostgreSQL] Swap order of columnDef and tableconstraint in tableelement rule #3033

paulirwin:
See issue #3033

### DIFF
--- a/sql/postgresql/PostgreSQLParser.g4
+++ b/sql/postgresql/PostgreSQLParser.g4
@@ -672,9 +672,9 @@ typedtableelementlist
    ;
 
 tableelement
-   : columnDef
+   : tableconstraint
    | tablelikeclause
-   | tableconstraint
+   | columnDef
    ;
 
 typedtableelement


teverett:
@KvanTTT  good to merge?

paulirwin:
I've done some more testing with this change and it has been working well. Since `tableconstraint` begins with the lexer rule `CONSTRAINT` and `tablelikeclause` begins with the lexer rule `LIKE`, they are not going to accidentally match on anything that should be a `columnDef` (since `CONSTRAINT` and `LIKE` are not valid column names), allowing other valid column names to pass through to the `columnDef` rule as expected.

KvanTTT:
> I've done some more testing with this change and it has been working well. 

Is it possible to add some testing examples to the `examples` directory?

paulirwin:
@KvanTTT I went to go add an example, and confirmed there already is an example that is parsed incorrectly in `create_table.sql`: 

```sql
CREATE TABLE parted (
	a text,
	b int NOT NULL DEFAULT 0,
	CONSTRAINT check_a CHECK (length(a) > 0)
) PARTITION BY LIST (a);
```

`check_a` gets parsed as a `columnDef` rule (as a column named `CONSTRAINT` of type `check_a`) instead of `tableconstraint`. This is likely a problem with other CONSTRAINT table elements in the example.

Let me know if this is insufficient for demonstrating this bug. Thanks!

KvanTTT:
Ok, it's enough. Thanks for the contribution and welcome to ANTLR grammars repository!

PR #3038, opened 2023-01-31
[SemVer] adding grammar for the semantic version spec

myarichuk:
This is a grammar for [Semantic Version](https://semver.org/) spec with a small addition (that doesn't violate the spec but allows for a more straightforward interpretation of the parse results).
I've added tokens that would recognize "build tags" such as "alpha3" or "beta.4" -> this should allow developers easier implementation of SemVer comparison.


myarichuk:
Note: the work on the parser is not yet finished so it is likely I will submit another PR for the commit with both C# "interpreter" that uses the grammar and relevant unit tests, which can also serve as examples of the semver spec and it's edge-cases.  Also, since I am a noob with ANTLR, I'd love to see any feedback :)
  
* WIP of a project that uses the grammar can be seen [here](https://github.com/myarichuk/ReleaseTools/tree/master/src/Parser.ConventionalCommit)
* Unit test project (also WIP) can be seen [here](https://github.com/myarichuk/ReleaseTools/blob/master/tests/ParserTests/SemanticVersion.cs)

kaby76:
> Note: the work on the parser is not yet finished so it is likely I will submit another PR for the commit with both C# "interpreter" that uses the grammar (example of intended usage of the grammar) and relevant unit tests, which can serve as examples of the semver spec and it's edge-cases. Also, since I am a noob with ANTLR, I'd love to see any feedback :)
> 
> * WIP of a project that uses the grammar can be seen [here](https://github.com/myarichuk/ReleaseTools/tree/master/src/Parser.ConventionalCommit)
> * Unit test project (also WIP) can be seen [here](https://github.com/myarichuk/ReleaseTools/blob/master/tests/ParserTests/SemanticVersion.cs)

Do not include driver code for your grammar. The grammar is tested by generating a driver from template with the Github Actions build. https://github.com/antlr/grammars-v4/tree/master/_scripts/templates

Include a pom.xml in the directory "grammars-v4/semver/". Make a copy from something like grammars-v4/abb/pom.xml. 

Include a line in the "root pom.xml" a reference to your new grammar, somewhere around [here](https://github.com/antlr/grammars-v4/blob/87bcdac9d4487e73ac6566d07e44161915c2bb85/pom.xml#L264) (alphabetic sorting).

Include a directory of test files in "grammars-v4/semver/examples/". All files in that are considered source except ".errors" and ".tree" files.

myarichuk:
Awesome, thanks for so many detailed suggestions! Plan to follow the suggestions as soon as I have enough free time.

semver/SemanticVersion.g4
```diff
@@ -0,0 +1,98 @@
...
+    ;
+
+versionCore: major DOT minor DOT patch;
+
+semver:
+    versionCore                               #SemverWithCore
+  | versionCore DASH preRelease               #SemverWithPreRelease
+  | versionCore PLUS build                    #SemverWithBuild
+  | versionCore DASH preRelease PLUS build    #SemverWithPreReleaseAndBuild
+  ;
```
KvanTTT, line 98: Honeslty, I'm not sure such splitting on alternaitves makes big sense. I suggest simplifying:

```suggestion
semver
  : versionCore (DASH preRelease (PLUS build)? | PLUS build)?
  ;
```

KvanTTT (changes requested):
Welcome to the grammars repository and thanks for the contribution.

But some issues should be fixed.

semver/SemanticVersion.g4
```diff
@@ -0,0 +1,98 @@
...
+fragment Q : [qQ];
+fragment R : [rR];
+fragment S : [sS];
+fragment T : [tT];
+fragment U : [uU];
+fragment V : [vV];
+fragment W : [wW];
+fragment X : [xX];
+fragment Y : [yY];
+fragment Z : [zZ];
```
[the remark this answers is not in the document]
kaby76: If caseInsensitive is used, it'll no longer work with Antlr4cs. I wasn't sure if it was required.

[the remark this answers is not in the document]
myarichuk: Ideally, I would like the grammar to be as compatible as possible. Wasn't aware of "caseInsensitive" feature, I will definitely use it for my own stuff :)
@KvanTTT - thanks for the detailed review and suggestions. They make perfect sense, I will address them.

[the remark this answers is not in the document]
KvanTTT: > it'll no longer work with Antlr4cs.

I don't consider it's useful to support an outdated runtime.

myarichuk:
Addressed the review suggestions, thanks again for your detailed feedback!

* Added a "local" ``pom.xml`` in the grammar folder
* Split the combined grammar into lexer/parser files
* Used the "caseIncensitive" flag for the lexer.
* Removed unnecessary parentheses from lexer token definitions
* Removed the alternative splitting for the ``semver`` rule - it makes more sense to combine them, yeah.
* Updated the "global" ``pom.xml`` to include the grammar name. I assume the name there should have the same name mentioned in the ``pom.xml`` in the grammar folder, right?
* Added some examples in the ``examples`` folder - one example per text file. Judging by what I see in other folder that should work, right?

I wonder if it makes sense to include a version with case-insensitive fragments for people who use the **Antlr4cs**. Wdyt?

myarichuk:
note: missed ``artifactId`` from pom.xml - looks like it made the build fail. I changed ``artifactId`` so it would be unique.

semver/SemanticVersion.g4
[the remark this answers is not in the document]
KvanTTT: Also, we use `caseInsensitive` option for other grammars in this repository.

semver/README.md
```diff
@@ -0,0 +1,58 @@
...
+fragment PREVIEW: P | P R E | P R E V I E W;
+fragment DEV: D E V | D E V E L | D E V E L O P M E N T;
+fragment MILESTONE: M T | M I L E S T O N E;
+fragment DAILY: D A I L Y;
+fragment NIGHTLY: N I G H T L Y;
+fragment BUILD: B L D | B U I L D;
+fragment TEST: T E S T;
+fragment EXPERIMENTAL: E X P E R I M E N T A L;
+
+```
```
KvanTTT, line 58: Sorry, but I don't consider this info useful, we don't have such info in other case insensitive grammars (check [`sql`](https://github.com/antlr/grammars-v4/tree/master/sql) for example). It's well-known practice to use case insensitive fragments. Also, I think there is a small number of outdated Antlr4cs runtime.

```suggestion
```

semver/SemanticVersionParser.g4
```diff
@@ -0,0 +1,27 @@
...
+options { tokenVocab = SemanticVersionLexer; }
+
+patch: value = NUMBER;
+minor: value = NUMBER;
+major: value = NUMBER;
+
+build:
+      value = NUMBER                                             #BuildNumber
+    | DASH                                                       #BuildDash
+    | tag = TAG (DOT | DASH)? version = (NUMBER | IDENTIFIER)?   #BuildTagged
```
KvanTTT, line 12: If `tag` is parser rule, then the following is actual:

```suggestion
    | tag (DOT | DASH)? version = (NUMBER | IDENTIFIER)?   #BuildTagged
```

semver/SemanticVersionParser.g4
```diff
@@ -0,0 +1,27 @@
...
+    | value = IDENTIFIER                                         #BuildIdentifier
+    | left = build (DOT | DASH)? right = build                   #BuildIdentifierExtended
+    ;
+
+preRelease:
+      value = NUMBER                                            #PreReleaseNumber
+    | DASH                                                      #PreReleaseDash
+    | tag = TAG (DOT | DASH)? version = (NUMBER | IDENTIFIER)?  #PreReleaseTagged
```
KvanTTT, line 20:
```suggestion
    | tag (DOT | DASH)? version = (NUMBER | IDENTIFIER)?  #PreReleaseTagged
```

semver/SemanticVersionParser.g4
```diff
@@ -0,0 +1,27 @@
...
+    | value = IDENTIFIER                                        #PreReleaseIdentifier
+    | left = preRelease (DOT | DASH)? right = preRelease        #PreReleaseIdentifierExtended
+    ;
+
+versionCore: major DOT minor DOT patch;
```
KvanTTT, line 25: The following definition is shorter:

```suggestion
versionCore: major=NUMBER DOT minor=NUMBER DOT patch=NUMBER;
```

semver/SemanticVersionParser.g4
```diff
@@ -0,0 +1,27 @@
+parser grammar SemanticVersionParser;
+
+options { tokenVocab = SemanticVersionLexer; }
+
+patch: value = NUMBER;
+minor: value = NUMBER;
+major: value = NUMBER;
```
KvanTTT, line 7: No need in these rules if use [rule element labels](https://github.com/antlr/antlr4/blob/master/doc/parser-rules.md#rule-element-labels) (see below).

```suggestion
```

KvanTTT (changes requested):
Thanks, but some issues are still presented.

semver/README.md
myarichuk: Makes sense, I will remove this section in the readme

myarichuk:
Hey, did some fixes. The test show pass now (``grammarName`` in the ``antlr4test`` config was wrong so the plugin couldn't find the generated grammar classes)
Also addressed the comments.

Thanks for the detailed comments!

teverett:
@KvanTTT  good to merge this?

KvanTTT (approved):
Thanks again for the new grammar and its fixes!

myarichuk:
Thanks for the detailed feedback, I really appreciate it!

PR #3040, opened 2023-01-31
[Typescript grammar] Add `Boolean` and `Number` as keywords

ariel-levy:
Expressions such as

```ts
mapEnumToSchema(Joi.boolean())
```

or

```ts
const codesByType = Joi.object()
  .keys({
    type: Joi.string().required(),
    limit: Joi.number().optional(),
    skip: Joi.number().optional(),
  })
  .required();
```

are not parsed properly (`boolean` is caught by the lexer and is not expected by the grammer)

### DIFF
--- a/javascript/typescript/TypeScriptParser.g4
+++ b/javascript/typescript/TypeScriptParser.g4
@@ -835,6 +835,8 @@ keyword
     | Require
     | TypeAlias
     | String
+    | Boolean
+    | Number
     ;
 
 getter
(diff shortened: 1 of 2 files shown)


KvanTTT:
@ariel-levy welcome to antlr grammars repository. Could you please cover your changes with a test example? Just add a new file or extend an exiting one in the [`javascript/typescript/examples`](https://github.com/antlr/grammars-v4/tree/master/javascript/javascript/examples) directory.

ariel-levy:
@KvanTTT thank you :) 
added, hope it's in the right place

KvanTTT:
Thanks for the contribution and welcome to the ANTLR grammars repository!

PR #3041, opened 2023-02-01
[Tsql] Fix grammar for revert

jiangtao69039:
According to the official documentation, there should be no parentheses.

Syntax from the documentation:

```sql
REVERT  
    [ WITH COOKIE = @varbinary_variable ]
```


pguyot:
Could you please fix the missing endline at the end of new file revert.sql?

jiangtao69039:
ok i fixed it.

PR #3047, opened 2023-02-03
add LSHIFT, RSHIFT, and URSHIFT

Instein98:
Hi! I created this pull request because when I am using the Lexer generated by Java8Lexer.g4, shift operations are not correctly tokenized in my opinion (for example, '>>' is tokenized as two '>'). But I might be wrong because it seems you do this deliberately. I'd like to hear your opinions anyway.

### DIFF
--- a/java/java8/Java8Lexer.g4
+++ b/java/java8/Java8Lexer.g4
@@ -426,6 +426,9 @@ BITAND : '&';
 BITOR : '|';
 CARET : '^';
 MOD : '%';
+LSHIFT : '<<';
+RSHIFT : '>>';
+URSHIFT : '>>>';
 ARROW : '->';
 COLONCOLON : '::';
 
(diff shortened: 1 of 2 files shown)


kaby76:
This is a well-known problem, and you need to research why this has been tried and why it doesn't work. I highly recommend that you test your change against the entire source for the OpenSDK using the before and after of your changes to the grammar, looking for regressions.

If you tokenize this way, you will need to do one of two things: (a) add semantic predicates that somehow differentiate expressions like "a << b" versus generic declarations like "Set<Set<int>>"; or (b), you will need to use lexer modes. AdaptivePredict() calls the lexer, but does not pass any state information of the parser. And know that some people tokenize the entire input even before calling the parser.

KvanTTT:
Thanks for the contribution, but the change is incorrect unfortunately. Thanks @kaby76 for the detail explanation.

PR #3052, opened 2023-02-06
Account for nested braces in template strings

### LINKED ISSUE #3051: Typescript grammar can not handle CloseBrace '}' properly in template expression
e.g. `` `${[1, 2, 3].map(x => { return x * 2;}).join("")}` ``
in the template string expression, there is a block. The grammar didn't handle the '}' correctly. 

When processing the `${` token, the analyzer pushes DEFAULT MODE, and the mode stack would be: 

`[DEFAULT, TEMPLATE, DEFAULT]`

 but in default mode, we can not recognize if the `}` is closing the block or the template expression. 

according to the grammar definition, I think the closeBrace is recognized as the templateCloseBrace
```
TemplateCloseBrace:             {this.IsInTemplateString()}? '}' -> popMode;

```

bkiers:
Fixes #3051

### DIFF
--- a/javascript/typescript/CSharp/TypeScriptLexerBase.cs
+++ b/javascript/typescript/CSharp/TypeScriptLexerBase.cs
@@ -31,7 +31,7 @@ public abstract class TypeScriptLexerBase : Lexer
     private bool _useStrictCurrent;
 
     /// <summary>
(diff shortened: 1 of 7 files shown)


teverett:
@bkiers  thanks!

PR #3053, opened 2023-02-06
[TSql] Add many dbcc commands and remove default dbcc rule

pguyot:


sql/tsql/TSqlLexer.g4
```diff
@@ -1142,6 +1176,7 @@ STRING options { caseInsensitive=false; } : 'N'? '\'' (~'\'' | '\'\'')* '\'';
 BINARY:              '0' 'X' HEX_DIGIT*;
 FLOAT:               DEC_DOT_DEC;
 REAL:                (DECIMAL | DEC_DOT_DEC) ('E' [+-]? DEC_DIGIT+);
+ZERO:                '0';
```
KvanTTT, line 1179: This rule looks unreachable since`DECIMAL` lexer rule always wins. It's declared above and accepts digit sequences. To fix, you can place `ZERO` lexer rule before `DECIMAL`, declare a new parser rule:

```g4
decimal
    : DECIMAL
    | ZERO
    ;
```

And use `decimal` rule instead of `DECIMAL` in all use sites.

KvanTTT (changes requested):

sql/tsql/TSqlLexer.g4
pguyot: True. To avoid over complexity, I used DECIMAL in the rules (so they'll accept any number) and I added tests for these weird statements.

PR #2994, opened 2023-01-06
Extended TSql Parser and Lexer

ealight:
- Added all the math functions to built-ins. ([Source](https://learn.microsoft.com/en-us/sql/t-sql/functions/mathematical-functions-transact-sql?view=sql-server-ver16))
- Extended date built-ins functions up-to-date. ([Source](https://learn.microsoft.com/en-us/sql/t-sql/functions/date-and-time-data-types-and-functions-transact-sql?view=sql-server-ver16))
- Added procedure's temporary tables recognizing. Now temporary tables which start with '#' are going to temp_table_name/full_temp_table_name contexts. ([Source](https://www.red-gate.com/simple-talk/databases/sql-server/t-sql-programming-sql-server/temporary-tables-in-sql-server/))
- Added database engine stored procedures handling. For now it's only 'sp_executesql' but there is a field to easy work with. ([Source](https://learn.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/database-engine-stored-procedures-transact-sql?view=sql-server-ver16)).
- Added old fashion join - NON ANSI Join handling. They're going to own context. ([Source](https://sqlenlight.com/support/help/sa0006/))
- Added dateparts for date built-in functions to match documentation. (dateparts_9 - [DATE_BUCKET](https://learn.microsoft.com/en-us/sql/t-sql/functions/date-bucket-transact-sql?view=sql-server-ver16), dateparts_12 - [DATE_ADD](https://learn.microsoft.com/en-us/sql/t-sql/functions/dateadd-transact-sql?view=sql-server-ver16), [DATE_DIFF](https://learn.microsoft.com/en-us/sql/t-sql/functions/datediff-transact-sql?view=sql-server-ver16), [DATEDIFF_BIG](https://learn.microsoft.com/en-us/sql/t-sql/functions/datediff-big-transact-sql?view=sql-server-ver16), dateparts_15 - [DATENAME](https://learn.microsoft.com/en-us/sql/t-sql/functions/datename-transact-sql?view=sql-server-ver16), [DATEPART](https://learn.microsoft.com/en-us/sql/t-sql/functions/datepart-transact-sql?view=sql-server-ver16), dateparts_datetrunc - [DATETRUNC](https://learn.microsoft.com/en-us/sql/t-sql/functions/datetrunc-transact-sql?view=sql-server-ver16))

PR #3055, opened 2023-02-07
[TSql] Fix ambiguity of temp tables (partial revert of #2994)

pguyot:
The new `temp_*table_name` rules accept the same tokens as the old rules renamed `persist_*table_name` as `id_` can be a `TEMP_ID`. This creates an ambiguity that slows down parsing.

Besides, from a semantic standpoint, temporary tables can be quoted (`SQUARE_BRACKET_ID`) and therefore some temporary table names would only be captured by `persist_*table_name` rules.

So this PR reverts this bit of PR #2994.

/cc @ealight


ealight:
lgtm

PR #3059, opened 2023-02-08
[TSql] Add support for square bracketed xml methods

pguyot:


sql/tsql/TSqlParser.g4
```diff
@@ -4499,31 +4499,31 @@ value_method
     ;
 
 value_call
-    :  VALUE '(' xquery=STRING ',' sqltype=STRING ')'
+    :  (VALUE | VALUE_SQUARE_BRACKET) '(' xquery=STRING ',' sqltype=STRING ')'
```
KvanTTT, line 4502: Can't we use the following code without introducing a new token?

```suggestion
    :  (VALUE | '[' VALUE  ']') '(' xquery=STRING ',' sqltype=STRING ')'
```

sql/tsql/TSqlLexer.g4
```diff
@@ -401,6 +401,7 @@ ESTIMATEONLY:                          'ESTIMATEONLY';
 EXCLUSIVE:                             'EXCLUSIVE';
 EXECUTABLE:                            'EXECUTABLE';
 EXIST:                                 'EXIST';
+EXIST_SQUARE_BRACKET:                  '[EXIST]';
```
KvanTTT, line 404: Why is it not possible to use `'[' EXIST ']'` instead of introducing the new token? BTW, the current token doesn't support whitespaces between `[` and `EXIST`. Is it correct?

pguyot: Square brackets are like quotes. Microsoft SSMS doesn't support spaces in this context.
Should we add a test to make sure the grammar doesn't accept it either? Can we do this without new tokens?

image

We globally have a problem with brackets, for example with `data_type` rule:

https://github.com/antlr/grammars-v4/blob/bd3d04274567cf4fa86a37fdf99e67fa2724f953/sql/tsql/TSqlParser.g4#L4905-L4912

We added SQUARE_BRACKET_ID here but it really should be `[varchar]`, `[nvarchar]`, `[binary]`, `[varbinary]`.

pguyot: Trying further, square bracketed names do not seem to be possible for built-in functions.

image

The variants introduced in this PR are not for every uses of the non-quoted keywords.

```
KILL QUERY NOTIFICATION SUBSCRIPTION ALL
```
is accepted by SQL Server 2022, but not:
```
KILL [QUERY] NOTIFICATION SUBSCRIPTION ALL
```

pguyot: Actually few can exist with squared brackets. Built-in functions such as ˋrank()` don't, but xml methods do. I can create a generic rule where any bracketed text would be accepted, instead of the four xml methods we currently accept, like we do with data types followed by `(max)`.

KvanTTT: Ok, let's postpone it.

PR #3061, opened 2023-02-08
Fixes for #2883

kaby76:
## Statement of Problem
This is a fix for https://github.com/antlr/grammars-v4/issues/2883

Testing in the V4 repo should be incremental, meaning that if one grammar changes, the build will only test the one grammar. But, that is not what currently happens. Usually, a single change will test all grammars because the main culprit, [regtest](https://github.com/antlr/grammars-v4/blob/6239468762b9ce5f7926e42b32a0ecfaebeeb525/_scripts/regtest.sh), does not take into consideration the files that changed. Therefore, despite [test.ps1](https://github.com/antlr/grammars-v4/blob/6239468762b9ce5f7926e42b32a0ecfaebeeb525/_scripts/test.ps1) testing incrementally, the rest of the runs slow the entire build down.

In addition, the build output is very long. This can be a problem when trying to find the problem with the build.

This PR addresses these two problems, as well as several other minor issues.
* The scripts for testing should be named the same, i.e., _scripts/test.ps1 and _scripts/test.sh.
* The location of the antlr jar should be passed into the scripts, not read via environmental variables, and the version set only in the top-level workflow file, .github/workflows/main.xml.
* The "skip lists" and the root pom.xml are a common choke point for new grammars, which then cause a full rebuild. The targets valid for a grammar should be placed in the directory with the grammar.
* Plain ol' "git diff" in the examples directory is insufficient because it reports too many files that diff, including input files that you haven't yet checked in the repo.
* Grammar abb defines a symbol called "module", but that conflicts with the new TypeScript target.

## Implementation
The "skip list" implementation has now been replaced with desc.xml files, which are contained in the directory with the grammar. Each desc.xml file contains a list of targets that work for the grammar. The pom.xml files are still valid in order to support the Maven tester. The tests scripts are now called "test.ps1" and "test.sh".

Test.sh now redirects output to a file, and only outputs the file in case of error. To determine the error, just search for "failed" in the Github workflow output. Both of the test scripts can now be called within the directory containing the grammar. By default for the Github workflow, both test scripts perform testing of the grammars that changed.

## Results and Expectations
By default, the Github build should be incremental, meaning that it will test only the grammars that have changed, for both Bash and Powershell scripts. We will very rarely see full builds. Most builds should now take minutes.


teverett:
This is great.   We could extend the XML to include grammar name, etc. (not this PR, a  later one)

@KvanTTT  good to merge?

kaby76:
@teverett Yes, this PR is ready to go. After merging, I will monitor for any problems and fix. Yes, we should figure out things we'd like to add to the desc.xml file. I was thinking about the usual (start symbol, test directories, etc), but maybe documentation (author, links to source, etc).

teverett:
@kaby76  thanks!

teverett:
Merged.  the one question I have is, "is there a way to do a full build, for example when the antlr version changes?"

teverett:
also, is it time to remove the pom.xml files?

PR #3066, opened 2023-02-09
Apex 1

kzherebilo:
APEX. I suggest we update the **expression** rule to account for safe navigation operator ("?.")

### DIFF
--- a/apex/apex.g4
+++ b/apex/apex.g4
@@ -521,7 +521,7 @@ expression
     |   expression '.' Identifier
     |   expression '.' THIS
     |   expression '.' NEW
-    |   expression ('.' | '?.')
+    |   expression '.'
         (   DB_INSERT
         |   DB_UPSERT
         |   DB_UPDATE
--- a/apex/apex.g4
+++ b/apex/apex.g4
@@ -518,7 +518,7 @@ expression
     :   primary
     |   expression '.' GET '(' expressionList? ')'
     |   expression '.' SET '(' expressionList? ')'
-    |   expression ('.' | '?.') Identifier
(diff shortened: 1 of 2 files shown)


KvanTTT (commented):
Welcome to ANTLR grammars repository and thanks for the contribution! But could you please add a sample that covers the suggested changes to the [examples](https://github.com/antlr/grammars-v4/tree/master/apex/examples)) directory?

kzherebilo:
Hello @KvanTTT! Thank you for your reply! That's the very first time I contribute to any public project. Please let me know if I've missed anything else)
I have just pushed an example named 'ExampleSafeDereference.cls'.

apex/apex.g4
```diff
@@ -518,7 +518,7 @@ expression
     :   primary
     |   expression '.' GET '(' expressionList? ')'
     |   expression '.' SET '(' expressionList? ')'
-    |   expression '.' Identifier
+    |   expression ('.' | '?.') Identifier
```
KvanTTT, line 521: I afraid it doesn't work since `?.` token is absent in the APEX grammar. You should introduce either the `?.` token:

```g4
QUESTION_DOT: '?.';
```

or use two tokens instead: `'?' '.'`:

```suggestion
    |   expression '?'? '.' Identifier
```


I don't know what option is more correct in terms of APEX specification.

KvanTTT (changes requested):

kzherebilo:
You're right! I believe the latter option you suggested is much simpler! Can we proceed with that one? Thank you!

KvanTTT:
Thanks again!

Issue #3009, opened 2023-01-12
TSQL grammar parser has tokens not present in lexer


fedorusov:
They are:
- IPV4_ADDR
- IPV6_ADDR
- FILESTREAM_ON

kaby76:
This will need to be fixed. https://github.com/antlr/grammars-v4/issues/2998 https://github.com/antlr/grammars-v4/issues/2990 https://github.com/antlr/grammars-v4/issues/2972#issuecomment-1362331961

PR #3072, opened 2023-02-13
#3009 - ipv[46]_addr, filestream_on tokens

mlorek:
added missing lexer tokens


KvanTTT (changes requested):
Could you please separate tokens reordering and new tokens adding? Use different commits or different pull requests.

Also, shouldn't the new tokens be added to the [keyword](https://github.com/antlr/grammars-v4/blob/master/sql/tsql/TSqlParser.g4#L5009) parse rules? It looks like they can be identifiers.

mlorek:
> Please fix failing tests.

trying to, but they work locally

PR #3074, opened 2023-02-13
add base classes for TypeScript

zwqStella:
Add base classes of TypeScript version.

### DIFF
--- a/python/python3/TypeScript/Python3LexerBase.ts
+++ b/python/python3/TypeScript/Python3LexerBase.ts
@@ -1,0 +1,128 @@
+import { CommonToken } from "antlr4ts/CommonToken";
+import { Lexer } from "antlr4ts/Lexer";
+import { Python3Parser } from './Python3Parser';
+
+export default abstract class Python3LexerBase extends Lexer {
+    tokens: any[];
+    indents: any[];
+    opened: number;
+
+    constructor(input) {
+        super(input);
+        this.tokens = [];
+        this.indents = [];
+        this.opened = 0;
+    }
+
+    reset() {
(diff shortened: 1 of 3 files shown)


kaby76:
This is great! Thanks for doing this.

Antlr doesn't support TypeScript yet, but it's on the verge of supporting it in a few weeks. You can play around with the latest. https://www.npmjs.com/package/antlr4. Just set your package.json to reference 4.12.0-beta.9. I've tested it with abb/ and started adding [templates for TypeScript to trgen](). When Antlr 4.12 is released in a few weeks, I'll update the CI builds to Antlr 4.12 and include testing of the TypeScript target for grammars-v4.

In fact, you might want to play around with the new target right now, and report any things you see missing [here](https://github.com/antlr/antlr4/issues/4110).

kaby76:
Looks like the official Antlr TypeScript target isn't exporting all the required classes and methods needed in this port. I'm working with Eric to iron these problems out before 4.12.0 is released.

zwqStella:
Accidentally closed this PR and the approvals are reset... Are we going to merge this for now, or hold it until we could do some tests with the official Antlr TypeScript supports?

kaby76:
> Accidentally closed this PR and the approvals are reset... Are we going to merge this for now, or hold it until we could do some tests with the official Antlr TypeScript supports?

If it were up to me, I would merge your changes in so it would not be lost, and keep the Issue still open, since the code is incomplete with respect to the new target.

kaby76:
Required changes in antlr4 4.12.0 release for Typescript port to work. https://github.com/antlr/antlr4/pull/4119

teverett:
@zwqStella thanks!

PR #3075, opened 2023-02-13
Support all privileges of MariaDB 10.6 in MariaDB grammar

joschi:
Some privileges supported by MariaDB are missing from the grammar.

This change set adds the missing tokens and tests them against a `GRANT` query with all privileges supported by MariaDB 10.6.x: https://mariadb.com/docs/server/ref/es10.6/privileges/


KvanTTT (changes requested):
We already have separated grammar for MariaDB: https://github.com/antlr/grammars-v4/tree/master/sql/mariadb

@uurl could you check if it's correct to add MariaDB changes to the current grammar? Or they should be added to the separated grammar.

joschi:
> We already have separated grammar for MariaDB: https://github.com/antlr/grammars-v4/tree/master/sql/mariadb

@KvanTTT Yes, I saw that and updated this MR. 😄

KvanTTT (approved):

PR #3076, opened 2023-02-13
[MySQL] Add support for VALUES Statement

joschi:
https://dev.mysql.com/doc/refman/8.0/en/values.html


joschi:
Refs https://github.com/antlr/grammars-v4/pull/3077

Issue #3043, opened 2023-02-02
sql/plsql not parsing apparently valid query string


kaby76:
* https://stackoverflow.com/questions/75292088/is-there-an-example-to-build-a-antlr4-grammar-parser-for-oracle-pl-sql-with-the/75295213#comment132891227_75292088

```sql
CREATE TABLE "X"."Y" ( "OID" NUMBER(19,0) NOT NULL ENABLE, "STANDARDIZED_HOME_PHONE" GENERATED ALWAYS AS (115) VIRTUAL , ) ;
CREATE TABLE "A"."B" ( "FIELD1" NUMBER(19,0) NOT NULL ENABLE, "FIELD2" NUMBER(19,0) NOT NULL ENABLE, PRIMARY KEY ("FIELD1") ENABLE ) ORGANIZATION INDEX NOCOMPRESS ; 
```

The only issue I noticed is that the author says "If anyone knows which SQL grammar to use , it would be greatly appreciates. Otherwise, I might have to edit or preprocess the file. – Timothy Hilgenberg". It's either plsql or it's not. How could there be any question of what server you are using?

PR #3079, opened 2023-02-14
[Plsql] Fix column_definition

jiangtao69039:
Add visible_clause:

image


mlorek:
this is fixed as part of PR #3043

PR #3084, opened 2023-02-16
Move ':' to COLON

nocarend:
It makes sense when you want to separate the lexer and parser rules.

### DIFF
--- a/pascal/pascal.g4
+++ b/pascal/pascal.g4
@@ -401,7 +401,7 @@ actualParameter
    ;
 
 parameterwidth
-   : ':' expression
+   : COLON expression
    ;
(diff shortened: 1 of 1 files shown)


kaby76:
This is a good change because the grammar was not consistently using lexer symbols for string literals in parser rules. Note, Antlr does handle split grammars with string literals in the parser grammar if and only if there is a lexer rule for the string literal. So, if the grammar when split works (`trparse pascal.g4 | trsplit | trsponge -o split; cd split; trgen -s program; cd Generated-CSharp; dotnet build`).

KvanTTT (approved):

PR #3093, opened 2023-02-17
Update OCL.g4

kevinlano:
Corrected additiveExpression grammar so that x - y + z is correctly parsed.

### DIFF
--- a/ocl/OCL.g4
+++ b/ocl/OCL.g4
@@ -5,7 +5,7 @@
 * Arrow operators ->op are used consistently for any OCL 
 * operator, not just collection operators. 
 * 
-* Copyright (c) 2003--2022 Kevin Lano
+* Copyright (c) 2003--2023 Kevin Lano
 * This program and the accompanying materials are made available under the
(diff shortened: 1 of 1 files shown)


KvanTTT (changes requested):
Please add an example that covers the suggested changes to the [ocl/examples](https://github.com/antlr/grammars-v4/tree/master/ocl/examples) directory.

ocl/OCL.g4
```diff
@@ -177,14 +177,15 @@ equalityExpression
     ; 
 
 additiveExpression
-    : factorExpression ('+' | '-' | '..' | '|->') 
-                              additiveExpression 
+    : additiveExpression '+' additiveExpression 
+    | additiveExpression '-' factorExpression
+    | factorExpression ('..' | '|->') factorExpression
```
KvanTTT, line 182: I suppose it should be the following:

```suggestion
    | additiveExpression ('..' | '|->') additiveExpression                         
```

KvanTTT (changes requested):

Issue #3028, opened 2023-01-25
Verilog grammar status


quantrpeter:
Hi Mr @msagca
     How is the status of verilog grammar, is it stable now?
thanks
Peter

msagca:
Hi @quantrpeter 

I think the grammar is in a good state. Of course it can be further optimized and improved but it works fine enough. As a user of it what do you think?

quantrpeter:
I was using the old version to build our tool, I remember last year there is a big change, grammar changed a lot, and listener stub are all changed. Now I want to use it to build a logic synthesizer, that's why I ask if it is stable. thanks for creating it Mr Msagca.

quantrpeter:
I just download a very big verilog, over 10000 line and test your grammar, it successful parsed it. thanks for your great effort, i can start building the logic synthesizer now.

quantrpeter:
when do we need to use VerilogPreParser?

msagca:
> when do we need to use VerilogPreParser?

It parses preprocessor directives. You need to handle them first before beginning the actual parse. If you're sure that the input text does not contain any directives that could impact the resulting parse tree then you can skip this step.

quantrpeter:
thanks, if i put any text to parse for VerilogPreParser, it doesn't give me any error, may I know why please?

```
		VerilogLexer lexer = new VerilogLexer(CharStreams.fromString("hello this is fake"));
		CommonTokenStream tokenStream = new CommonTokenStream(lexer);
		VerilogPreParser parser = new VerilogPreParser(tokenStream);
		parser.source_text();
```

msagca:
@quantrpeter

Tokens belonging to preprocessor directives are on the [DIRECTIVES](https://github.com/antlr/grammars-v4/blob/master/verilog/verilog/VerilogLexer.g4#L26) channel. You have to pass the correct channel ID (3) to [CommonTokenStream](https://www.antlr.org/api/Java/org/antlr/v4/runtime/CommonTokenStream.html).

quantrpeter:
```
String str = IOUtils.toString(TestCompile.class.getClassLoader().getResourceAsStream("verilogexample/tb.v"), "utf-8");
VerilogLexer lexer = new VerilogLexer(CharStreams.fromString(str));
CommonTokenStream tokenStream = new CommonTokenStream(lexer, VerilogLexer.DIRECTIVES);
tokenStream.fill();
VerilogPreParser parser = new VerilogPreParser(tokenStream);
[...]

PR #3097, opened 2023-02-18
[GLSL,(System)Verilog] fix comments/spaces inside directives

msagca:
Adds limitations to where in compiler directives comments and spaces can exist.
Existing examples already cover and comply with the changes.
#3028


teverett:
@msagca thanks!

PR #3098, opened 2023-02-18
snowflake: fix parsing of `COMMENT` in column declaration

michael-the1:
Fix parsing of `COMMENT` in the column declaration of a `CREATE TABLE` statement. A `COMMENT` in a column declaration is done _without_ the equals (`=`) sign (see [documentation](https://docs.snowflake.com/en/sql-reference/sql/create-table)).

While there, replace `COMMENT EQ string` with the already existing `comment_clause`.

### DIFF
--- a/sql/snowflake/SnowflakeParser.g4
+++ b/sql/snowflake/SnowflakeParser.g4
@@ -2216,7 +2216,7 @@ out_of_line_constraint
 full_col_decl
     : col_decl
         collate?
-        comment_clause?
+        (COMMENT string)?
         default_value?
         not_null?
         with_masking_policy?
--- a/sql/snowflake/SnowflakeParser.g4
+++ b/sql/snowflake/SnowflakeParser.g4
@@ -1409,7 +1409,7 @@ create_account
             EDITION EQ ( STANDARD | ENTERPRISE | BUSINESS_CRITICAL )
           ( REGION_GROUP EQ region_group_id )?
           ( REGION EQ snowflake_region_id )?
-          ( comment_clause )?
+          comment_clause?
     ;
 
 create_api_integration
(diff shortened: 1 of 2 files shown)


sql/snowflake/SnowflakeParser.g4
```diff
@@ -1448,7 +1448,7 @@ create_object_clone
     ;
 
 create_connection
-    : CREATE CONNECTION if_not_exists? id_ ( ( COMMENT EQ string )? | (AS REPLICA OF id_ DOT id_ DOT id_ ( COMMENT EQ string )?) )
+    : CREATE CONNECTION if_not_exists? id_ ( ( comment_clause )? | (AS REPLICA OF id_ DOT id_ DOT id_ ( comment_clause )?) )
```
KvanTTT, line 1451:
```suggestion
    : CREATE CONNECTION if_not_exists? id_ ( comment_clause? | (AS REPLICA OF id_ DOT id_ DOT id_ comment_clause?) )
```

sql/snowflake/SnowflakeParser.g4
```diff
@@ -2261,7 +2261,7 @@ create_table_as_select
     ;
 
 create_tag
-    : CREATE or_replace? TAG if_not_exists? id_ ( COMMENT EQ string )?
+    : CREATE or_replace? TAG if_not_exists? id_ ( comment_clause )?
```
KvanTTT, line 2264:
```suggestion
    : CREATE or_replace? TAG if_not_exists? id_ comment_clause?
```

KvanTTT (changes requested):

But please remove redundant parentheses.

michael-the1:
> Thanks for the contribution and welcome to ANTLR grammars repository!
> 
> But please remove redundant parentheses.

Thanks! I find-replaced them without looking, should've known better 😂

KvanTTT (approved):

MasterKuat:
Hi @michael-the1,
As options could appear in any order full_col_decl grammar should be something like :
```
full_col_decl
    : col_decl
        (collate|(COMMENT string)|default_value|not_null|with_masking_policy|with_tags|inline_constraint)*
    ;

```
By the way, not sure of syntax as I'm a noob with antlr but I test it and seems good.
Many thanks for your work 😃

michael-the1:
@MasterKuat Yes, I also ran into that and came to the same solution. I will probably open a PR for it next week, or you can totally do it! I'm also pretty new at ANTLR, but I found contributing like this quite easy. Just add a few test statements and it should be good to go.

PR #3101, opened 2023-02-19
Update expressionExamples.txt

kevinlano:
Added new example as requested


teverett:
@kevinlano  thanks!

kaby76:
Multiple problems with the build for this.
* Problem in _scripts/test.sh. I fixed this and making a PR for the change.
* Grammar ocl. The pom.xml does not have the proper entry point!
* ocl/desc.xml has no targets that are tested!! This is because the script I wrote looks at the grammars, tries all targets, and adds them to "desc.xml". Unfortunately, the grammar is messed up, so the initial version is completely empty.

PR #3107, opened 2023-02-20
[Snowflake]Fix parsing "AT" stream

MasterKuat:
AT lexer symbol in at_before1 was not defined.
Also, I reword BEFORE to BEFORE_KEYWORD in order to be consitent with the original AT_KEYWORD

I also use LR_BRACKET and RR_BRACKET symbol. 
What is best pratice on this subjet ? Lexer rule or direct caracter ?


bkiers:
`AT_KEYWORD` was named like that since there was a keyword (`'AT'`) and there is the `'@'` char. There is only 1 `'BEFORE'` so there is no real need to append `_KEYWORD` to this token. Also, by renaming it, you're breaking code that relies on the grammar containing a `BEFORE` token.

Personally, I'd just do this in this PR:

```
at_before1
    : AT_KEYWORD | BEFORE
    ;
```

MasterKuat:
@bkiers Revert to BEFORE done.
Thank you for your input

teverett:
@MasterKuat  there are merge conflicts.

MasterKuat:
Hi @teverett Fixed :)

PR #3108, opened 2023-02-20
[Snowflake] Enable and improve `COMMENT ON` for all object types

michael-the1:
`COMMENT ON` can be applied to all object types, see [documentation](https://docs.snowflake.com/en/sql-reference/sql/comment).

`COMMENT ON COLUMN` needs a full column name, at least `table.column` and can accept up to `database.schema.table.column`.


teverett:
@michael-the1  thanks!

sql/snowflake/SnowflakeParser.g4
```diff
@@ -226,12 +226,8 @@ copy_into_location
...
 
 comment
-    : COMMENT if_exists? ON comment_object_type object_name IS string
-    | COMMENT if_exists? ON COLUMN column_name IS string
-    ;
-
-comment_object_type
-    : TABLE
+    : COMMENT if_exists? ON object_type_name object_name IS string
+    | COMMENT if_exists? ON COLUMN full_column_name IS string
```
KvanTTT, line 230: It can be simplified:

```suggestion
    : COMMENT if_exists? ON (object_type_name object_name | COLUMN full_column_name) IS string
```

PR #3109, opened 2023-02-20
[Snowflake] Fix column declaration

michael-the1:
Fix column declaration with the following changes:
1. Tag declarations can be done with a full object name
2. `collate`, `inline_constraint`, `default_value`, and `not_null` can appear in any order of each other
3. `COMMENT` always goes last. This is not according to [the documentation](https://docs.snowflake.com/en/sql-reference/sql/create-table), but the documentation is wrong here :)

While there, change `EMPTY` to `EMPTY_` to fix a warning.


kaby76:
This looks okay. The build fails are from a bad package.json file that I wrote. One issue I see is this https://github.com/antlr/grammars-v4/actions/runs/4223816295/jobs/7333993472#step:18:56

EMPTY should be renamed to EMPTY_ in the snowflake grammar because warnings will be treated as errors at some point soon.

teverett:
@kaby76  does @michael-the1  need to close the PR and re-open, or is just restarting the builds sufficient?

kaby76:
It looks like he is using the old versions of the test scripts likely because the branch was from the old version of master. Rebase? Or close this and make a new branch with the latest master code.

michael-the1:
Or just a merge commit :)

michael-the1:
cc: @MasterKuat I think you were interested in this as well

MasterKuat:
Hi @michael-the1,
Yeah. Thanks for the ping but I already subscribe to it yesterday 😄 
By the way I do a PR also yesterday : #3107 
I think for the time being we should sync our work.

michael-the1:
> Hi @michael-the1, Yeah. Thanks for the ping but I already subscribe to it yesterday 😄 By the way I do a PR also yesterday : #3107 I think for the time being we should sync our work.

I don't think there will be any conflicts so we should be good.

teverett:
@michael-the1  thanks!

PR #3112, opened 2023-02-20
Oclfix

teverett:


ocl/desc.xml
```diff
@@ -1,3 +1,4 @@
-<desc>
+<?xml version="1.0" encoding="UTF-8" ?>
+<desc xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../_scripts/desc.xsd">
    <targets>Java</targets>
 </desc>
```
KvanTTT, line 4: Again, there are strange changes that doesn't relate to PR title.

Issue #3123, opened 2023-02-21
Test.sh is somehow testing a grammar more than once, but it should not


kaby76:
A recent build had a grammar listed twice, and got tested twice. For the snowflake grammar and the CSharp target, that could save 4 minutes build time, and make the build half as long. https://github.com/antlr/grammars-v4/actions/runs/4231715348/jobs/7350576397#step:17:31

The Bash test.sh script needs to perform a "sort -u" on the list of grammars to test prior to testing.

PR #3124, opened 2023-02-21
Fix for #3123: sort and unique the list of grammars to test

kaby76:
This PR fixes a minor problem with the builds using the Bash script [test.sh](https://github.com/antlr/grammars-v4/blob/8b00eaf4b4eab17bfe5493e96260ee89b6f05efa/_scripts/test.sh), where the array of strings [grammars](https://github.com/antlr/grammars-v4/blob/8b00eaf4b4eab17bfe5493e96260ee89b6f05efa/_scripts/test.sh#L88) contained the same grammar twice. https://github.com/antlr/grammars-v4/issues/3123. This should not happen because it is a waste of resources, and slows the build down because the grammar is tested more than once. The Powershell code [already checks for this](https://github.com/antlr/grammars-v4/blob/8b00eaf4b4eab17bfe5493e96260ee89b6f05efa/_scripts/test.ps1#L181).

### DIFF
--- a/_scripts/test.sh
+++ b/_scripts/test.sh
@@ -229,10 +229,10 @@ then
             do
                 if [ -f `pwd`/desc.xml ]
                 then
-                break
+                    break
                 elif [ `pwd` == "$prefix" ]
                 then
-                break
+                    break
                 fi
                 cd ..
             done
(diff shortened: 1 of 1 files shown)


kaby76:
For this PR, the Github Actions worked perfectly. It noted that there were zero grammars to test, which is true because I only changed _scripts/test.sh. The build for the 30 jobs (12 of test.sh, 18 of test.ps1) on 30 machines was only 4 minutes. This is exactly how the builds should work: test only grammars that change, using the updated test scripts. However, if the templates change, then there should be some tests. The only question is how much. But, this is such a fringe case, I may not solve it.

teverett:
@kaby76  thanks!

PR #3129, opened 2023-02-23
with sql parser

caicancai:
Common Table Expression (CTE) was introduced in MySQL 5.7 WITH the WITH statement to support the syntax of CTE. CTE is a general syntax used to improve the readability, maintainability, and reuse of SQL statements. It can define one or more expressions using simple syntax and use them to build more complex expressions into the query, making the query easier to understand and debug. In this sense, Ctes are important for data integration in MySQL 5.7.

### DIFF
--- a/sql/mysql/Positive-Technologies/MySqlParser.g4
+++ b/sql/mysql/Positive-Technologies/MySqlParser.g4
@@ -68,7 +68,7 @@ dmlStatement
     : selectStatement | insertStatement | updateStatement
     | deleteStatement | replaceStatement | callStatement
     | loadDataStatement | loadXmlStatement | doStatement
-    | handlerStatement | valuesStatement
+    | handlerStatement | valuesStatement | withStatement
     ;
 
 transactionStatement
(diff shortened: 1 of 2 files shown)


caicancai:
I in the g4 file condition with keyword, support with keyword parsing
```
withStatement
  : 'WITH' commonTableExpressions (',' commonTableExpressions)*
  ;
```

caicancai:
update MySqlParser.g4

KvanTTT (changes requested):

But please add a sample to [sql/mysql/Positive-Technologies/examples](https://github.com/antlr/grammars-v4/tree/master/sql/mysql/Positive-Technologies/examples) that covers suggested changes.

caicancai:
The original MySqlParser.g4 does not support the parsing of cte sql, but through commonTableExpressions extension, the MySqlParser.g4 supports the parsing of cte sql

caicancai:
@KvanTTT

KvanTTT (approved):

PR #3131, opened 2023-02-23
Add bencoding grammar

bkiers:
Specification: https://wiki.theory.org/BitTorrentSpecification#Bencoding


kaby76:
@bkiers You will need to create a desc.xml file to specify what targets will work with the grammar ([see another grammar for example](https://github.com/antlr/grammars-v4/blob/e1aea27925fe1d33cf4b24b871b8f156b3215c6a/abb/desc.xml)). By default, it's not a grammar if there's no desc.xml file in the same directory as the pom.xml. Although it is currently require to also have a pom.xml--to specify the start symbol, the grammar files, and test directories--the Maven tester is no longer used because it tests too much. The Maven tester does not perform a git diff to limit the set of grammars to test.

bkiers:
Ah, that must be new. How is the start rule now discovered @kaby76?

kaby76:
> Ah, that must be new. How is the start rule now discovered @kaby76? Or does that still come from the pom file?

Yes, I just added the desc.xml files in the last week because the "skip lists" were "too centralized" in the _scripts/ directory. Although the grammar files are parsed, I still use the pom.xml to as final authority of the start symbol, grammar files, grammar name, and examples to parse. At some point, we might fold the pom.xml into the desc.xml and/or use only information from the grammar and directory. But, I do think the Maven tester has value.

teverett:
@bkiers  thanks!

PR #3132, opened 2023-02-23
[Snowflake]Procedure with JAVASCRIPT language

MasterKuat:
Create procedure could be also with Javascript language.
Procedure could be also SECURE.
Reference : https://docs.snowflake.com/en/sql-reference/sql/create-procedure

JAVA / SCALA / PYTHON could be also correct language but syntax has more options.

### DIFF
--- a/sql/snowflake/SnowflakeLexer.g4
+++ b/sql/snowflake/SnowflakeLexer.g4
@@ -413,6 +413,7 @@ IS:                                                    'IS';
 ISNULL:                                                'ISNULL';
 ISNUMERIC:                                             'ISNUMERIC';
 ISOLATION:                                             'ISOLATION';
+JAVASCRIPT:                                            'JAVASCRIPT';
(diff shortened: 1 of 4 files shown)


sql/snowflake/SnowflakeLexer.g4
```diff
@@ -413,6 +413,7 @@ IS:                                                    'IS';
 ISNULL:                                                'ISNULL';
 ISNUMERIC:                                             'ISNUMERIC';
 ISOLATION:                                             'ISOLATION';
+JAVASCRIPT:                                            'JAVASCRIPT';
```
KvanTTT, line 416: I guess it should be added to [non_reserved_words](https://github.com/antlr/grammars-v4/blob/master/sql/snowflake/SnowflakeParser.g4#L3267) since it can be identifier.

KvanTTT (approved):

MasterKuat:
Thank you @KvanTTT 😃

Issue #3121, opened 2023-02-20
MacOS github testing is rate limiting factor


kaby76:
I'm looking over the builds, and it looks like Ubuntu and Windows builds of the CI workflow complete fairly quickly. But there are fiver or so actions that are queued for a MacOS server to complete the rest of the workflow. I think it's fair to conclude that Google has a limited number of the MacOS servers,  probably because Apple machines are ridiculously expensive.

I propose the build skip all MacOS testing.

teverett:
Well, I agree it is slower.  However some people (like me) use OS X.

kaby76:
It looks like the number of concurrent jobs for Mac has a column just for itself. https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits

So, yes, it seems that Mac servers are fewer in number. I'm wondering whether we could skip either Powershell or Bash testing, but have at least one. That way we still get the grammars tested for all targets on the Mac, but not two times. All testing is done basically twice because trgen generates roughly identical scripts for a target (it's almost the same line for line, except one in Powershell, the other in Bash).

PR #3136, opened 2023-02-24
plsql - create controlfile and alter db update

mlorek:


kaby76:
@mlorek The JavaScript test on MacOS is failing because it's timing out after 5 minutes of parsing. The parse is done using "group parsing" but I don't think JavaScript gives a good speed up when done this way. (For other targets, like CSharp and Java, the speed up is at least 4x.) The build script doesn't print that there was a time out (bug), which doesn't help. The grammar itself has contained performance issues for some time, which is the root cause of the problem, but harder to clean out than fixing the build scripts or restructuring the tests.

There are several tests that are very long running. These should go into a "long-running" directory for now until I add a way to go back to individual parsing into the build. If I parse individually for JavaScript, each test will get 5 minutes runtime rather than the aggregate of parsing all input files. These would be examples/aggregate01.sql examples/analytic_query.sql examples/cast_multiset07.sql examples/datetime02.sql examples/order_by07.sql examples/query_factoring07.sql

mlorek:
> @kaby76 what is the location of 'long-running' directory?

kaby76:
> > @kaby76 what is the location of 'long-running' directory?

Just create the directory and move those SQL files and any associated .errors and .tree files to the new directory. Then, check in, push, and let's see if that takes care of the problem for now.

I'm thinking about adding a "group" or "individual" setting in the desc.xml file. I'll fix this and the time-out-no-error-message problem after I get Antlr 4.12 merged in.

sql/plsql/PlSqlParser.g4
```diff
@@ -1808,11 +1843,10 @@ datafile_tempfile_spec
     : (CHAR_STRING)? (SIZE size_clause)? REUSE? autoextend_clause?
```
KvanTTT, line 1843: Redundant parentheses:

```suggestion
    : CHAR_STRING? (SIZE size_clause)? REUSE? autoextend_clause?
```

kaby76:
@mlorek You might also want to not spawn off too many PRs at once until I get a fix for https://github.com/antlr/grammars-v4/issues/3138. The Mac servers for Github are showing to be the rate-limiting factor. https://github.com/antlr/grammars-v4/issues/3121. It looks like Github/Microsoft doesn't want to buy too many of those machines, and they don't seem particularly speedy. I will add some code to find out what machine type the server is.

mlorek:
@kaby76 how many is too many?

kaby76:
> @kaby76 how many is too many?

Don't know! I think we're competing right now with someone else out there. None of the actions that you or I spawned for this repo are getting any time. The server is working on something else. Nothing we can do except just wait this out.

kaby76:
"Re-run triggered 1 hour ago". Seems like a bug if a workflow is waiting there and Github doesn't want to assign a server to the task except when manually restarting a queue. Not impressed with Github.

KvanTTT (approved):

PR #3139, opened 2023-02-24
[Snowflake]Improvement create File Format

MasterKuat:
Improvement for CREATE FILE FORMAT : 
* type specification is optional (default to csv in snowflake)
* TYPE could be quoted or unquoted
* RECORD_DELIMITER and FIELD_DELIMITER  could be a string (20 char max but I didn't find a way to acheive this)
* NONE value could be quoted also


MasterKuat:
`
CHAR_LITERAL:       '\'' (~['\\\r\n] | EscapeSequence) '\'';
` 
overlap
`
STRING       options { caseInsensitive=false; }     : '\'' (~'\'' | '\'\'')* '\'';
`
so grammar `| FIELD_OPTIONALLY_ENCLOSED_BY EQ (character | NONE | NONE_Q)` never match.

I try to play with pushMode and other things but I do not find any solution 😢 
I include in this PR some examples. There are all valid in Snowflake.

Help will be greatly appreciate 🙏

KvanTTT:
> STRING       options { caseInsensitive=false; }     : '\'' (~'\'' | '\'\'')* '\'';

BTW, using option is useless because string literal doesn't contain case sensitive chars inside definition.

sql/snowflake/examples/create_fileformat.sql
```diff
@@ -0,0 +1,8 @@
...
+--CREATE FILE FORMAT FMT_CSV_ANSI TYPE = 'CSV' FIELD_DELIMITER = ';' 	SKIP_HEADER = 1 TRIM_SPACE = TRUE FIELD_OPTIONALLY_ENCLOSED_BY = NONE NULL_IF = ('') ERROR_ON_COLUMN_COUNT_MISMATCH = FALSE 	ESCAPE = NONE 	ESCAPE_UNENCLOSED_FIELD = NONE 	ENCODING = WINDOWS1252;
+CREATE FILE FORMAT FMT_XML TYPE = 'XML'  COMPRESSION = 'AUTO'   PRESERVE_SPACE = FALSE  STRIP_OUTER_ELEMENT = TRUE   DISABLE_SNOWFLAKE_DATA = FALSE      DISABLE_AUTO_CONVERT = FALSE     IGNORE_UTF8_ERRORS = FALSE;
+CREATE FILE FORMAT FMT_CSV_UTF8_PIPE FIELD_DELIMITER = '|' SKIP_HEADER = 1 ESCAPE_UNENCLOSED_FIELD = 'NONE' TRIM_SPACE = TRUE NULL_IF = ('') ERROR_ON_COLUMN_COUNT_MISMATCH = FALSE ;
+CREATE FILE FORMAT FMT_CSV_ANSI_BS FIELD_DELIMITER = '\b' SKIP_HEADER = 1 ESCAPE_UNENCLOSED_FIELD = 'NONE' TRIM_SPACE = TRUE NULL_IF = ('') ERROR_ON_COLUMN_COUNT_MISMATCH = FALSE ENCODING = 'WINDOWS1252' ;
+-- Below will fail but don't know how to fix it. Seems CHAR_LITERAL is superseeded by STRING
+CREATE FILE FORMAT FMT_CSVZIP SKIP_HEADER = 1 TRIM_SPACE = TRUE FIELD_OPTIONALLY_ENCLOSED_BY = '\"' NULL_IF = ('') COMPRESSION = BZ2 ERROR_ON_COLUMN_COUNT_MISMATCH = FALSE REPLACE_INVALID_CHARACTERS = TRUE ENCODING = 'utf8' ;
```
MasterKuat, line 8: Hi @kaby76, I'm surprised, checks is all green but
 Line 8 should fail with an : line 8/column 95 mismatched input ''\"'' expecting {'NONE', ''NONE'', CHAR_LITERAL}

Does tests do not try to parse examples files below examples folder ?
(May be I misunderstood)

kaby76: It's supposed to check the directory and all subdirectories. The Powershell script shows all the input files, but the Bash shell does not (I left the Powershell script that way so we see some slightly different output). Let me check to see what is wrong.

kaby76: The test create_fileformat.sql passed and test is correct. I changed the tester to just stop saying "parser produced an error, but it was expected, so the test passes...". The scripts generally only output when there's an error in the test. In this case, the parser produces an error, which is captured exactly to create_fileformat.sql.errors. The script then performs a "git diff" of the .errors and .tree files, and only raises an error when there's a change in those files. This is different from the previous test scripts, where it didn't really check if there was an actual diff in the parse error message.

MasterKuat:
<ins>For history</ins>: 
At time of this PR, Snowflake documentation for [Create file format](https://docs.snowflake.com/en/sql-reference/sql/create-file-format) state that 
`TYPE = CSV | JSON | AVRO | ORC | PARQUET | XML` 
is mandatory but in fact is not.
Snowflake's support raised an internal request (SNOW-749662) to get this corrected.

sql/snowflake/examples/create_fileformat.sql
MasterKuat: Hi @kaby76, 
Ok I had a close look and I think I got the trick.
In fact there is **no** create_fileformat.sql.errors (any .errors file in fact) in the Snowflake exemple  subfolder.
Does it means that we need to add an empty *.errors to ensure that all statement parser successfully ?

kaby76: You could create an empty .errors file, but then the old Maven program would probably flag that as an error. So, I'd recommend not checking in any empty .errors files. The script checks for non-zero .errors files that are new (not checked in using `git ls-files`), and flags that as errors.

MasterKuat: Yeah, I'm sorry to insist, but you wrote ealier that script produce a non-zero "create_fileformat.sql.errors", this one is new. So test for this PR should fail.
Which is obviously not the case. 
So there is a bug in tests 😃

kaby76: Yes, there's a bug there. I'll fix that. But, you need to check in any and all non-empty .errors files since that is how to say to the test scripts that you expect an error.

MasterKuat: Crystal clear. Thank you @kaby76 !
And sorry for the bug.

teverett:
@MasterKuat  thanks!

PR #3144, opened 2023-02-26
[Snowflake]Select parsing fix

MasterKuat:
- OVER clause need an expr before.
- over_clause need at least 1 of partition_by or order by
- in order by, it's each expression which could ordered diffrently, not only the latest one

Statements in select.sql execute in Snowflake UI and grammar.


teverett:
@MasterKuat thanks!

Issue #3145, opened 2023-02-27
_scripts/mkindex.py is not indexing files under javadoc/examples/javadoc


kaby76:
It looks like the [grammars.json file](https://github.com/antlr/grammars-v4/blob/86f1c64ea79c27f10ca1500bd010a60bee684c53/grammars.json) is not correct because [BlockTagsExample.java](https://github.com/antlr/grammars-v4/blob/86f1c64ea79c27f10ca1500bd010a60bee684c53/javadoc/examples/javadoc/BlockTagsExample.java) is not listed. I think [the script does not recursively check for files](https://github.com/antlr/grammars-v4/blob/86f1c64ea79c27f10ca1500bd010a60bee684c53/_scripts/mkindex.py#L118) because [os.listdir](https://docs.python.org/3/library/os.html#os.listdir) does not list recursively files. And, it should not place a directory in the grammars.json file because then lab.antlr.org will need to check and recurse down the directory structure. The problem exists for other gramars that have subdirectories under examples/, like kotlin/kotlin-formal/examples/fuzzer.

Issue #2866, opened 2022-10-07
Add grammars-v4 dir indexer


parrt:
Hi. (I screwed up and push to master despite having a git branch command in my history. Weird.) Related to supporting lab.antlr.og and https://github.com/antlr/antlr4-lab/issues/11...

@teverett @kaby76 @KvanTTT please take a look at _scripts/mkindex.py, which generates a json list from a grammars-v4 path name:

```
$ python mkindex.py .. | jq
[
  {
    "name": "regex",
    "lexer": "https://raw.githubusercontent.com/antlr/grammars-v4/master/xsd-regex/regexLexer.g4",
    "parser": "https://raw.githubusercontent.com/antlr/grammars-v4/master/xsd-regex/regexParser.g4",
    "start": "root",
    "example": "example-chargroup-sub3.txt"
  },
  {
    "name": "abb",
    "lexer": "https://raw.githubusercontent.com/antlr/grammars-v4/master/abb/abbLexer.g4",
    "parser": "https://raw.githubusercontent.com/antlr/grammars-v4/master/abb/abbParser.g4",
    "start": "module",
    "example": "robdata.sys"
  },
  {
    "name": "DGS",
    "lexer": "https://raw.githubusercontent.com/antlr/grammars-v4/master/graphstream-dgs/DGSLexer.g4",
    "parser": "https://raw.githubusercontent.com/antlr/grammars-v4/master/graphstream-dgs/DGSParser.g4",
    "start": "dgs",
    "example": "removeAttribute.dgs"
  },
  {
    "name": "SwiftFin",
    "lexer": "https://raw.githubusercontent.com/antlr/grammars-v4/master/swift-fin/SwiftFinLexer.g4",
    "parser": "https://raw.githubusercontent.com/antlr/grammars-v4/master/swift-fin/SwiftFinParser.g4",
    "start": "messages",
    "example": "test1.txt"
  },
  {
    "name": "Lucene",
    "lexer": "https://raw.githubusercontent.com/antlr/grammars-v4/master/lucene/LuceneLexer.g4",
    "parser": "https://raw.githubusercontent.com/antlr/grammars-v4/master/lucene/LuceneParser.g4",
    "start": "topLevelQuery",
    "example": "boolean-3.txt"
  },
  {
    "name": "Cql",
    "lexer": "https://raw.githubusercontent.com/antlr/grammars-v4/master/cql3/CqlLexer.g4",
[...]

PR #3157, opened 2023-02-28
Fix for 3145: get mkindex.py to list out test case files in subdirectories

kaby76:
Fix for https://github.com/antlr/grammars-v4/issues/3145

## _scripts/mkindex.py script changes:
* Changed os.libdir() to Path(), so example files in subdirectories are listed.
* Added checks for ".errors" and ".tree" files, so they are not listed in lab.antlr.org as an input file to parse. Python 3.10 is now required.
* Added code to replace backslashes in paths to forward slashes.
* Added code to format json file, so we can read the results.

## Grammars.json file:
* A number of grammars were not listed, start symbols wrong, and otherwise new test file examples added over the last few months to the repo. I updated the grammars.json file to support the current version of the "Antlr lab" running at lab.antlr.org.

## To do:
* https://github.com/antlr/grammars-v4/issues/2866 should be closed. @parrt 
* We still need to do https://github.com/antlr/antlr4-lab/issues/52. To do that, we'll need a fine-grained personal access token for this repo. It should have only read access.


kaby76:
Path names should be relative to the examples directory. It's currently relative to the root directory. It should not.

parrt:
> https://github.com/antlr/grammars-v4/issues/2866 should be closed. 

Done.

Regarding the access token... I guess that's something I need to feed to my server when I start it up? Are you sure there's no way I can just download via regular http? This is friction to my world and I'm already barely hanging on in terms of my workload.

PR #3160, opened 2023-02-28
[Typescript grammer] re-arrange class decorations and declaration grammer entries

ariel-levy:
This should fix 3 main issues: 
 - a class with which had both `export` in front of it *and* decorators would break the grammer
 - generally, `decoratorsList` appears directly in each allowed context except for class declarations, it's not very convenient to get a class's decorators that way + it's more elegant that way IMO
 - there was a redundancy between `ClassExpression` and `classDeclaration`, on top of that if a class would be parsed as `ClassExpression` it would miss the decorators etc

### DIFF
--- a/javascript/typescript/TypeScriptParser.g4
+++ b/javascript/typescript/TypeScriptParser.g4
@@ -338,7 +338,6 @@ statement
     | exportStatement
     | emptyStatement_
     | abstractDeclaration //ADDED
-    | decoratorList
     | classDeclaration
     | interfaceDeclaration //ADDED
     | namespaceDeclaration //ADDED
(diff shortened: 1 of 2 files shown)


KvanTTT (changes requested):

kaby76:
The branch you modified is from January 30. The test script code and Github Action workflows have changed significantly, so that is why the build failed. You need to update your fork and branch.

ariel-levy:
@kaby76 I got a notification about some CI workflows failing, but seems unrelated to the naked eye 
image

is there something to do about it?

KvanTTT:
It looks like everything is ok now.

KvanTTT (approved):

Issue #2264, opened 2021-07-21
java 8 grammar fails to parse string with unicode


mlorek:
grammar fails to parse string  "\"" when \ character is represented using unicode notation:

```java
class Foo {

	String s = "\u005c"";
    
}

studentmain:
I remember java replace all \uxxxx before parse, so `cl\u0061ss A{}` is valid too

studentmain:
https://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.2

A custom input stream or a preprocessor is needed.

mlorek:
how about this?

fragment EscapeSequence
    : '\\' [btnfr"'\\]
    | '\\' ([0-3]? [0-7])? [0-7]
    | '\\' 'u'+ HexDigit HexDigit HexDigit HexDigit
    | '\\u005c' [btnfr"'\\]
    ;

studentmain:
> how about this?
> 
> fragment EscapeSequence
> : '\' [btnfr"'\]
> | '\' ([0-3]? [0-7])? [0-7]
> | '\' 'u'+ HexDigit HexDigit HexDigit HexDigit
> | '\u005c' [btnfr"'\]
> ;

Problem is not only "string with unicode", problem is everything with unicode

PR #3163, opened 2023-03-01
java - fix for #2264

mlorek:


java/java/JavaLexer.g4
```diff
@@ -211,8 +211,13 @@ fragment ExponentPart
...
+fragment EscapeChar
+    : '\\'
+    | '\\u005c'
+    ;
+
 fragment EscapeSequence
-    : '\\' [btnfr"'\\]
+    : EscapeChar [btnfr"'\\]
     | '\\' ([0-3]? [0-7])? [0-7]
     | '\\' 'u'+ HexDigit HexDigit HexDigit HexDigit
```
KvanTTT, line 222: What's the meaning of introducing the new fragment `EscapeChar`? I see only one use-site. Could it be simplified in the following way?

```suggestion
    : '\\' ( 'u005c'? [btnfr"'\\]
           |  ([0-3]? [0-7])? [0-7]
           |  'u'+ HexDigit HexDigit HexDigit HexDigit
           )
```

kaby76:
The TypeScript target tests aren't working. I'm working on why the scripts aren't working, but it seems it has nothing to do with this change. The parse is hanging on the input, but when I run without `-x` and instead use `examples/*.java`, the grouped parse takes only 9s. In the meanwhile, just remove the TypeScript target to get through the tests.

kaby76:
The issue which the tests failed is because test.sh, which is generated by trgen for testing, does not call ts-node correctly. It's an easy fix, which will result in a cleaner script. I'll have later on today after a full test of all grammars on Windows and Ubuntu. https://github.com/antlr/grammars-v4/issues/3173

kaby76:
> The failed build is because of some bad code in test.sh. It does not call ts-node correctly. It's an easy fix, which will result in a cleaner script. I'll have later on today after a full test of all grammars on Windows and Ubuntu. #3173

The _scripts/templates/TypeScript/test.{sh,ps1} scripts are fixed. You can update to them after that PR is merge.

kaby76 (approved):
Looks fine. In the next round of changes to trgen, the pom.xml will no longer be used. Instead, all information will be grabbed from the desc.xml (which may contain information that the pom.xml had, e.g., different test case scenarios), or just extract the information directly from the grammar files (e.g., grammar name, start symbol) and tests (e.g., examples/).

KvanTTT (approved):

kaby76:
The java/java9 grammar looks like it's having problems.

First, there's only three targets tested (Cpp, CSharp, Dart). That was set up as the initial version by one of my scripts that goes through every target and tests it, noting whether the grammar works for the target or not. (I didn't want to set up the desc.xml file by hand, so that's why I wrote the script to do this in an automated way.) The grammar should always work for Java, which it does. At the time I ran the script, it seems the tester was broken. Java should be added to the desc.xml file.

But, then, I compared the testing of java/java9 between `master` vs your changes in this PR. You added tests "*/examples/Escapes.java". That works fine for java/java/, but not for java/java9, for Cpp and Dart. So, the java/java9,{Dart,Cpp} ports don't work with your change, but java/java9,{Java,CSharp} port do.

mlorek:
> The java/java9 grammar looks like it's having problems.
> 
> First, there's only three targets tested (Cpp, CSharp, Dart). That was set up as the initial version by one of my scripts that goes through every target and tests it, noting whether the grammar works for the target or not. (I didn't want to set up the desc.xml file by hand, so that's why I wrote the script to do this in an automated way.) The grammar should always work for Java, which it does. At the time I ran the script, it seems the tester was broken. Java should be added to the desc.xml file.
> 
> But, then, I compared the testing of java/java9 between `master` vs your changes in this PR. You added tests "*/examples/Escapes.java". That works fine for the java/java9,{Java,CSharp} ports. But, the java/java9,{Dart,Cpp} ports don't work with your change.

interesting as it's a valid Java code in Escapes.java

kaby76:
> interesting as it's a valid Java code in Escapes.java

Grammar java/java/ doesn't have base class code and it works there. It's likely the base class code has issues because it works for two ports but not for the other two. I probably wrote those codes. Sorry! :)

kaby76 (approved):

teverett:
@mlorek thanks!

PR #3164, opened 2023-03-01
TS grammer - Add module as reserverd and fix 'propertyMemberBase' order

ariel-levy:
- expressions of the sort of `module.export...` broke the parsing
- `'public' modifier must precede 'async' modifier.ts(1029)`

### DIFF
--- a/javascript/typescript/TypeScriptParser.g4
+++ b/javascript/typescript/TypeScriptParser.g4
@@ -504,7 +504,7 @@ functionDeclaration
 
 //Ovveride ECMA
 classDeclaration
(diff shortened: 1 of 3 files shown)


teverett:
@KvanTTT good to merge?

KvanTTT (approved):

KvanTTT:
Yes, it's ok.

ariel-levy:
@KvanTTT generally - are multiple small PRs spammy for you? if it's preferable I can accumulate them to 1 PR
(I'm working on something using this grammer and every now and then I get a parsing error that hints a grammer error)

KvanTTT:
Hi! Not spammy at all. Multiple requests are easier for review than a big one.

PR #3180, opened 2023-03-02
Add Desktop Entry Grammar

msagca:
https://www.freedesktop.org/wiki/Specifications/desktop-entry-spec/


msagca:
@kaby76 I've already added one. Thanks.

teverett:
@msagca  thanks!

Issue #3183, opened 2023-03-03
[PlSql] Cannot parsing create table statement


jiangtao69039:
The following statement has parsing error:

```sql
CREATE TABLE T_ORDER 
(   ORDER_ID NUMBER(10) GENERATED BY DEFAULT AS IDENTITY START WITH 1 MAXVALUE 100,
    STATUS VARCHAR2(10), 
    COLUMN1 VARCHAR2(10)
);
```

jiangtao69039:
> According to [the Oracle SQL Language Reference, Version 21](https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/CREATE-TABLE.html#GUID-F9CE0CC3-13AE-4744-A43C-EAC7A71AAAB6__GUID-3F840F33-7A29-4910-9E43-2F1DAE4FFE76), shouldn't identity_options be nested in parentheses?

According to the documentation, it is true that parentheses should be included. But in fact, I can execute successfully without parentheses.
![image](https://user-images.githubusercontent.com/20572131/222748674-f6df4ad8-cdd4-4983-93ff-e8efff47389d.png)

mlorek:
> According to [the Oracle SQL Language Reference, Version 21](https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/CREATE-TABLE.html#GUID-F9CE0CC3-13AE-4744-A43C-EAC7A71AAAB6__GUID-3F840F33-7A29-4910-9E43-2F1DAE4FFE76), shouldn't identity_options be nested in parentheses?

oracle doc is full of inconsistencies

kaby76:
> > According to [the Oracle SQL Language Reference, Version 21](https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/CREATE-TABLE.html#GUID-F9CE0CC3-13AE-4744-A43C-EAC7A71AAAB6__GUID-3F840F33-7A29-4910-9E43-2F1DAE4FFE76), shouldn't identity_options be nested in parentheses?
> 
> oracle doc is full of inconsistencies

We don't test queries here on a live Oracle DB, so we are in a pickle because we won't know whether the Antlr grammar or the Oracle doc is correct. So, for this and future PRs, shouldn't we add a comment to the grammar with a bug report link to Oracle or a statement saying the doc is wrong against version such-and-such of Oracle? Otherwise, someone might remove the syntax a year or two from now because it doesn't fit the doc.

PR #3194, opened 2023-03-07
[PlSql] Fix optional parentheses of identity_options (#3183)

jiangtao69039:
#3183

### DIFF
--- a/sql/plsql/PlSqlParser.g4
+++ b/sql/plsql/PlSqlParser.g4
@@ -4982,7 +4982,11 @@ column_collation_name
     ;
 
 identity_clause
-    : GENERATED (ALWAYS | BY DEFAULT (ON NULL_)?)? AS IDENTITY ('(' identity_options* ')')?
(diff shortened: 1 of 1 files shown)


sql/plsql/PlSqlParser.g4
```diff
@@ -4982,7 +4982,11 @@ column_collation_name
     ;
 
 identity_clause
-    : GENERATED (ALWAYS | BY DEFAULT (ON NULL_)?)? AS IDENTITY ('(' identity_options* ')')?
+    : GENERATED (ALWAYS | BY DEFAULT (ON NULL_)?)? AS IDENTITY identity_options_parentheses?
+    ;
+
+identity_options_parentheses
+    : '('? identity_options+ ')'?
```
KvanTTT, line 4989: I guess the following code is more correct:

```suggestion
    : identity_options+
    | '(' identity_options+ ')'
```

Because it allows single opening parenthesis that looks weird.

KvanTTT (changes requested):
Also, please add an example that covers suggested changes.

KvanTTT (approved):

Issue #3138, opened 2023-02-24
Long-running examples require individual parsing, not group parsing


kaby76:
Plsql has several examples that take 20+ seconds for JavaScript on MacOS. Right now, we do "group parsing," which is supposed to speed up parsing of inputs after the first file. But, I don't think that grouped parsing helps with the JavaScript target, and certainly not with PHP. We should have a way to specify which type of testing to do in the desc file. Additionally, we need a way to do multiple groups of testing, like there is for pom.xml, i.e., "individual parsing for long-running/ tests; group parsing for examples/ tests", and be able to subdivide that with the target and OS, skipping or not.

kaby76:
This issue requires a large number of changes to [trgen](https://github.com/kaby76/Domemtech.Trash/tree/main/src/trgen), which is a completely terrifying mess. I am fixing, and it will look but better, more organized and logical.

kaby76:
* .errors files are different. This really needs to be fixed. But, the changes here will work past it. https://github.com/antlr/antlr4/issues/3700

PR #3195, opened 2023-03-07
[build] Fix for Issue 3138

kaby76:
## Statement of Problem
This is a fix for https://github.com/antlr/grammars-v4/issues/3138: upgrade the build to allow for group or individual parsing, to allow multiple tests per grammar (i.e., "scenarios" in the Maven Antlr test plugin), and to fix a couple of bugs in the build.

## Implementation
This PR has the following changes:

* Use the enhanced trgen version 0.20.6 and test scripts to allow for different test scenarios. This version of trgen uses an enhanced "desc.xml" file, which can contain new element types: `<test>`, `<inputs>`, `<grammar-name>`, `<grammar-files>`, `<targets>`, `<parsing-type>`, and `<entry-point>`. XML is case sensitive, so `<Targets>` is not recognized as `<targets>`. **NB: Make sure to spell the elements correctly, and use valid values. The trgen tool does not check for misspelled elements, so write carefully!** Since I am not a fan of the Camel, Pascal, flat, or upper case naming conventions for elements, I chose Lisp case for all the two-word elements. These elements can be used outside a `<test>` element. The only requirement is to have a `<targets>` within a `<desc>` element still.
* Enhance trgen to use only the desc.xml for testing. The Maven Antlr tester is still valid, albeit only for Java.
* Enhance trgen to derive the grammar name and start rule from the .g4 files instead of from the desc.xml (or pom.xml). If there are multiple grammar files in a grammar that results in an ambiguous start symbol and grammar, the desc.xml can be used to tell the build how to proceed. The start rule is an Antlr parser rule with EOF. For a few grammars, alterations were made to have trgen find the start rule.
* Change the sql/tsql grammar to remove the "symbol conflict" for the TypeScript target to work.
* Add an "Info" step to the build so that we may see what hardware and operating system is being used for testing.
* Removed the kludge in the Antlr4cs templates for the hardwired namespace "Test". Although that target still is not tested here, most grammars work with it.
* Add "set -e" to the build.sh templates so that when there is a build failure, the result is passed immediately to the calling script.
* Add "git config --global pager.diff false" in order for git to stop using a pager, which interferes with testing in an automated way.
* Fix error reporting of .errors files. "$?" is reset to zero in Bash if statements.
* Fix the TypeScript template test.sh to use "tsc -p tsconfig.json --pretty". The TypeScript compiler outputs certain error messages to the console. Using the "--pretty" option forces the compiler to output all error messages to stderr so it can be captured and displayed. https://github.com/antlr/grammars-v4/issues/3204
* Changed the cql3 grammar to use EOF on the start rule, rather than indirectly via symbol "eof".
* Changed the csharp grammar desc.xml to specify the name of the grammar. The csharp grammar contains two grammars, one for preprocessing and the other post preprocessing. Without the grammar-name specified, trgen picks erroneously the preprocessor grammar to use for testing.
* Add testing of two testing scenarios, one for the start rule "specification", and the other for "expressions". Without the new scenario, expressionExamples.txt would fail since the input cannot be parsed from "specification".
* Changed sql/tsql to rename "expression_list" to "expression_list_" so that the grammar can be used in TypeScript. Note, the generated TypeScript parser is very slow, so I copied a few of the examples to a new directory and run the TypeScript target on that smaller test suite.
* Update systemverilog and verilog grammars desc.xml with unambiguously specifying what grammar to use.

## To do
* https://github.com/antlr/grammars-v4/issues/3212


kaby76:
This PR is ready.

As usual, I will make sure to monitor the builds that others do for any problems introduced.

kaby76:
Changing back to draft so I can do one more full rebuild just to make sure all is really working. May take a few days.

kaby76:
All set wit this PR. Works on my machine. Will keep an eye out for problems of course.

teverett:
@kaby76  thanks!

PR #3196, opened 2023-03-07
[Tsql] Sybase - additional trigger syntax

olebhansen:
* Add support for trigger-syntax "create or replace" - observed working in ASE 16

* Add support for trigger syntax "if update()" to enforce referential integrity - See https://infocenter.sybase.com/help/index.jsp?topic=/com.sybase.infocenter.dc32300.1600/doc/html/san1390612503040.html

### DIFF
--- a/sql/tsql/TSqlParser.g4
+++ b/sql/tsql/TSqlParser.g4
@@ -44,6 +44,13 @@ batch_level_statement
     | create_or_alter_trigger
     | create_view
     ;
+
+dml_trigger_sql_clauses
+    : dml_trigger_cfl_statement SEMI?
+    | sql_clauses SEMI?
+    | SEMI
+    ;
+
 sql_clauses
     : dml_clause SEMI?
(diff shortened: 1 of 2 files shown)


olebhansen:
I'm not sure the "if update()" is valid in MSSql, or if the two tsql-dialects are diverging enough to warrant something Sybase-specific. I also tried to limit the changes/impact to usage from within a trigger-statement, which resulted in some slightly more complications/duplications in the parser.

kaby76:
I suggest gating the different syntax by using a semantic predicate at the beginning of an alt.

olebhansen:
Thanks, I was not aware of that. I'm not too versed in the grammar-syntax, so good to learn more :-) I'll close this PR and resubmit the "create or replace" on it's own (which I think is fine as-is), and a re-worked suggestion for the "if update()" support when I have had a stab at the suggested approach using semantic predicate.

PR #3198, opened 2023-03-08
[Typescript] Allow type hint for 'lastFormalParameterArg'

ariel-levy:
```
const post = (...args: any[]) => {
};
```
broke the parser until this fix


kaby76:
Snowflake is failing in your build because it wasn't working and the build scripts were broken (sorry, my error). Since then, Snowflake and the scripts have been updated. Update to the latest and your changes should work.

KvanTTT (approved):

PR #3201, opened 2023-03-09
[PromQL] support preserving comments

facetoe:
Send the comments and whitespace to separate channels. This is useful for code formatters etc that want to have access to these.


teverett:
@facetoe  thanks!

PR #3205, opened 2023-03-09
[Antlr4] Rule labels are not supported in lexer rules

renatahodovan:
The patch updates the grammar accordingly.


KvanTTT (approved):

kaby76:
We need a test case for this change since this grammar is used in several tools. Please add one. Thanks.

teverett:
@renatahodovan  thanks!

PR #3213, opened 2023-03-11
[ASM] Use caseInsensitive=true for ASM grammars

KvanTTT:


kaby76 (approved):
Excellent. Much, much cleaner. We should probably add something to the desc.xml to indicate a minimum version of Antlr 4.10 to make sure folks understand this cannot be done with an older version of the tool. Something like:
```
   <targets>Antlr4cs;CSharp;Cpp;Dart;Go;Java;JavaScript;TypeScript;PHP;Python3</targets>
   <version>^4.10</version>
</desc>
```
Maybe a JavaScript npm package.json version string? What do you think?

KvanTTT:
> We should probably add something to the desc.xml to indicate a minimum version of Antlr 4.10 to make sure folks understand this cannot be done with an older version of the tool.

It sounds reasonable.

> Maybe a JavaScript npm package.json version string? What do you think?

I suggest creating an issue for that and discussing there.

teverett:
@KvanTTT  all good to merge?

KvanTTT:
Yes, I think so.

@kaby76 , I've created an issue: https://github.com/antlr/grammars-v4/issues/3214

PR #3222, opened 2023-03-12
ocl: Corrected priorities of numeric operators

kevinlano:


kaby76:
Rule `expressions` is required for testing.

This grammar defines two entry points: `specification` and `expressions`, the later I added. This is because someone added the test file [expressionExamples.txt](https://github.com/antlr/grammars-v4/blob/master/ocl/examples-expr/expressionExamples.txt), which never actually parsed from `specification`. The new test scripts allow for multiple test cases to be tested, one for [specification](https://github.com/antlr/grammars-v4/blob/07d0791b2303af81f640515a7806955430cad3f7/ocl/desc.xml#L4) and the other for [expressions](https://github.com/antlr/grammars-v4/blob/07d0791b2303af81f640515a7806955430cad3f7/ocl/desc.xml#L9). The .errors file for that test was never checked in, and the scripts had a bug in not checking whether there were expecting errors in the parse. The test scripts now check whether the parse had errors and if the output is different from before.

You will need to uncomment `expressions`.

kaby76:
@kevinlano The grammar works fine (parses the inputs) across all targets. The desc.xml can be updated to reflect that: `<targets>Java</targets>` => `<targets>Antlr4cs;CSharp;Cpp;Dart;Go;Java;JavaScript;TypeScript;PHP;Python3</targets>`.

ocl/OCL.g4
```diff
@@ -19,9 +19,9 @@ specification
   : 'package' ID '{' classifier* '}' EOF
   ;
 
-expressions
-  : expression (';' expression)* ';'? EOF
-  ;
+// expressions
+//   : expression (';' expression)* ';'? EOF
+//   ;
```
KvanTTT, line 24: Please remove not used code.

ocl/OCL.g4
```diff
@@ -199,7 +198,9 @@ factorExpression
 // ->subrange is used for ->substring and ->subSequence
 
 factor2Expression
-  : factor2Expression '->size()' 
+  : '-' factor2Expression 
+  | '+' factor2Expression
```
KvanTTT, line 202: Do `'-'` and `'+'` have the same priority? I guess yes and it can be simplified:

```suggestion
  : ('-' | '+') factor2Expression 
```

KvanTTT (changes requested):

ocl/OCL.g4
kaby76: Either the `expressions` rule be uncommented so the tests pass, or the rule, the test examples-expr/expressionExamples.txt, and the corresponding test in desc.xml all removed. It has to be one way other the other.

kevinlano: Yes, uncomment the "expressions" rule.

ocl/OCL.g4
kevinlano: Yes, that would be ok to change like this.

teverett:
@kevinlano  thanks

Issue #3017, opened 2023-01-16
php grammar: Java target crashes on tests when printing trees


kaby76:
The [php]() grammar crashes when trying to print out trees (`trgen -t Java; cd Generated-Java; make; make test`).
```
Exception in thread "main" java.lang.NullPointerException
        at org.antlr.v4.runtime.misc.Utils.escapeWhitespace(Utils.java:63)
        at org.antlr.v4.runtime.tree.Trees.toStringTree(Trees.java:49)
        at org.antlr.v4.runtime.tree.Trees.toStringTree(Trees.java:58)
        at org.antlr.v4.runtime.tree.Trees.toStringTree(Trees.java:58)
        at org.antlr.v4.runtime.tree.Trees.toStringTree(Trees.java:58)
        at org.antlr.v4.runtime.tree.Trees.toStringTree(Trees.java:58)
        at org.antlr.v4.runtime.tree.Trees.toStringTree(Trees.java:58)
        at org.antlr.v4.runtime.tree.Trees.toStringTree(Trees.java:42)
        at org.antlr.v4.runtime.RuleContext.toStringTree(RuleContext.java:189)
        at Test.DoParse(Test.java:206)
        at Test.ParseFilename(Test.java:135)
        at Test.main(Test.java:107)
```
The code crashes because a [token is created with no text](https://github.com/antlr/grammars-v4/blob/56dcd3d9251506d7514a68e86a4c192735c48f59/php/Java/PhpLexerBase.java#L75), whereas in other targets, text is assigned ([Python3](https://github.com/antlr/grammars-v4/blob/56dcd3d9251506d7514a68e86a4c192735c48f59/php/Python3/PhpLexerBase.py#L64), [JavaScript](https://github.com/antlr/grammars-v4/blob/56dcd3d9251506d7514a68e86a4c192735c48f59/php/JavaScript/PhpLexerBase.js#L64)). Why the Antlr API allows one to create a CommonToken that is basically inconsistent (null text field, unknow values for start and stop, etc.) is dubious. That, or the toStringTree() API should NOT crash on null pointers!

[...]

PR #3259, opened 2023-03-16
[php] Fix for #3017: crash in Java target

kaby76:
The base class for the lexer created a semi-colon token but gave no text to the symbol. Changed it to just mutate the token type to a semicolon, as done in the other targets. https://github.com/antlr/grammars-v4/issues/3017


teverett:
@kaby76 thanks!

Issue #3266, opened 2023-03-16
Fail to parse Try statements containing qualified names  [Java/Java]


DagSonntag:
Hi
I noticed that the Java/Java grammar fails to parse Try statements containing qualified names (i.e. identifiers separated with dots) like:
An example is in java.base/sun/security/ssl/SSLSocketImpl.java: 779

```java
try (conContext.inputRecord) {
                    appInput.deplete();
                }
```

Looking into the handbook one can see that this is due to that a resource (the conContext.inputRecord) in this case, can be a qualifiedname, not just an identifier as detailed in the grammar now.

I will do a merge request to fix this shortly.

bkiers:
Looking at the [java](https://github.com/antlr/grammars-v4/tree/master/java/java) grammar:

```
statement
    : ...
    | TRY resourceSpecification block catchClause* finallyBlock?
    | ...
    ;

resourceSpecification
    : '(' resources ';'? ')'
    ;

resources
    : resource (';' resource)*
    ;

resource
    : ...
    | qualifiedName
    ;

qualifiedName
    : identifier ('.' identifier)*
    ;
```

it all seems OK. So I tried both the [java](https://github.com/antlr/grammars-v4/tree/master/java/java) and [java9](https://github.com/antlr/grammars-v4/tree/master/java/java9) grammars with the input:

```java
class Foo {
    public void mu() {
        try (conContext.inputRecord) {
            // ...
        } catch (Exception e) {
            // ...
        }
    }
}
```

and both parse the input fine.

I tested like this:

```java
String source = "class Foo {\n" +
        "    public void mu() {\n" +
        "        try (conContext.inputRecord) {\n" +
        "            // ...\n" +
        "        } catch (Exception e) {\n" +
        "            // ...\n" +
        "        }\n" +
        "    }\n" +
        "}";

Java9Lexer lexer = new Java9Lexer(CharStreams.fromString(source));
Java9Parser parser = new Java9Parser(new CommonTokenStream(lexer));
ParseTree root = parser.compilationUnit();

System.out.println(root.toStringTree(parser));
```

Note that I did not try the Java 8 grammar since the [try-with-resrouce](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) was introduced in Java 9.

PR #3267, opened 2023-03-16
Fix for resource to be qualified names

DagSonntag:
Fix for issue #3266
Including examples


DagSonntag:
Ooops, my mistake. Accidentally also put in the fix for #3268 on the same merge request. Let me know if I should try to split it into two separate or if  one is ok for both. Hopefully "both" fixes will be ok though (they are quite straightforward)

teverett:
@DagSonntag  thanks!

PR #3271, opened 2023-03-17
TOML - key token fix

DanielBeckwith:
This fixes situations where a key is a valid number, which would cause the respective number token to be produced, subsequently causing parsing to incorrectly fail when the TOML document was valid.

How I implemented the lexer modes feels messy, so feel free to suggest improvements.


teverett:
@DanielBeckwith  thanks!

Issue #3260, opened 2023-03-16
Scala.g4 fails to parse public spark code 


talwgx:
Attempting to parse AST of Spark code: https://github.com/apache/spark/blob/master/resource-managers/mesos/src/main/scala/org/apache/spark/scheduler/cluster/mesos/MesosCoarseGrainedSchedulerBackend.scala using Scala.g4 (latest ver on master) results in errors below (multi-line comments were removed during pre-processing, also double check with: https://tools.knowledgewalls.com/remove-comments-from-any-program-online)

742:10 mismatched input 'def' expecting {'implicit', 'lazy', 'case', 'override', 'abstract', 'final', 'sealed', 'private', 'protected', 'class', 'object', 'trait'}
757:11 mismatched input 'def' expecting {'implicit', 'lazy', 'case', 'override', 'abstract', 'final', 'sealed', 'private', 'protected', 'class', 'object', 'trait'}
761:11 mismatched input 'def' expecting {'implicit', 'lazy', 'case', 'override', 'abstract', 'final', 'sealed', 'private', 'protected', 'class', 'object', 'trait'}
766:11 mismatched input 'def' expecting {'implicit', 'lazy', 'case', 'override', 'abstract', 'final', 'sealed', 'private', 'protected', 'class', 'object', 'trait'}
772:11 mismatched input 'def' expecting {'implicit', 'lazy', 'case', 'override', 'abstract', 'final', 'sealed', 'private', 'protected', 'class', 'object', 'trait'}
785:11 mismatched input 'def' expecting {'implicit', 'lazy', 'case', 'override', 'abstract', 'final', 'sealed', 'private', 'protected', 'class', 'object', 'trait'}
801:21 mismatched input 'def' expecting {'implicit', 'lazy', 'case', 'override', 'abstract', 'final', 'sealed', 'private', 'protected', 'class', 'object', 'trait'}
805:10 mismatched input 'def' expecting {'implicit', 'lazy', 'case', 'override', 'abstract', 'final', 'sealed', 'private', 'protected', 'class', 'object', 'trait'}

pic is from a locally run antlr-lab (the online ver aborts bc of the grammar warn)

![antlr](https://user-images.githubusercontent.com/73443663/225481988-a237c66f-3136-445c-a5ee-66622803153a.jpg)

Any help would be much appreciated!

@Marti2203

kaby76:
[...]

PR #3277, opened 2023-03-17
[scala] fix for 3260 (fails to parse public spark code)

kaby76:
* Fix for https://github.com/antlr/grammars-v4/issues/3260
* Preliminary changes adapted from https://github.com/antlr/grammars-v4/pull/2228
* Initial commit is just to get the ball rolling.

## To do
* Refactor @Marti2203's token stream code into lexer base class.
* The plan is to test with as many of the 4k .scala files at https://github.com/apache/spark
* This grammar continues with changes to the current grammar. But, Scala has an official grammar, which should be scraped by automation.


kaby76:
* For the last week, I needed to work out a LSP server for Antlr4 grammars with VSCode, but especially VS2022. I couldn't debug this grammar without jumping around the grammar. The base class code still needs work at NextToken().

kaby76:
After thinking about this for a long time, I don't think this code can be hammered into submission. This solution is terribly ineligant, and extremely fragile, which is why it breaks all over the 4500+ Spark code.

The rules stated in the Spec--obviously derived after writing the parser implementation, which I wouldn't expect anything less by the "compiler experts" nowadays--require parser context. Interpreting these rules as straight-forward parser-state-independent lookaround code in the lexer is incorrect. These rules have to be converted into rules that require much larger lookaround in order to duplicate what the parser does. To convert these to a straight parse-state-independent lexer would require new heuristics to be discovered, completely from scratch.

Instead, I'm going to try something completely different. Since the parser is ridiculously slow anyways, and since any further changes probably won't make a difference in speed, my plan is to refactor the grammar as follows:
* Replace `NL` occurrences with new non-terminals that derive the empty string, and place all `NL` on the hidden channel.
* In the lexer, shunt all NL to hidden.
* On exit of the start symbol, add a step to just go through the tree and token stream, and validate `NL` occur properly relative to the tree.
* I'll probably still discover things in the Spark code that should parse, but don't because the grammar in the Spec is wrong. Correct the grammar because the Spec would, of course, be wrong, because the folks who wrote the section on the [rules for newlines](https://www.scala-lang.org/files/archive/spec/2.13/01-lexical-syntax.html#newline-characters) got it wrong.

kaby76:
Refactoring the grammar as described above, it seems to be parsing much better. There was a problem with the [classParams](https://github.com/antlr/grammars-v4/blob/c176029ba00e5fe23cc83a8340f5a5cc26c901dd/scala/ScalaParser.g4#L430) rule.

Unfortunately, the grammar results in a slower parser (for the moment). For example, on `spark/common/sketch/src/test/scala/org/apache/spark/util/sketch/CountMinSketchSuite.scala`, the parse is 30s in individual parse mode, and 25s in group parse mode. The fact that group parsing does not result in a good speed-up is interesting in itself. It likely means that there is some deep-order problems in the Antlr4 AdaptivePredict() code. I will look into this in the future.

For now, the plan is to continue the test of the Spark code until I'm satisfied that it works better. Then, I need to go through the grammar with perf testing to find out the most egregious mal-desigsn in the grammar, and try to correct these problems to improve performance.

kaby76:
* 1.25 hours of group parsing of the Spark code, only ~250 files. Horrible.
* I did an initial performance analysis, and the interaction of [stableId](https://github.com/antlr/grammars-v4/blob/2c8bfc06f0926a0196ffc78fa9c2b3386f9aa066/scala/ScalaParser.g4#L67-L70) and [importExpr](https://github.com/antlr/grammars-v4/blob/2c8bfc06f0926a0196ffc78fa9c2b3386f9aa066/scala/ScalaParser.g4#L498-L500) is particularly bad. This is a classic ***anti-pattern*** I've observed in many grammars-v4 grammars (e.g., verilog), where the parser tries to determine the end of and the rest of the handle of `importExpr` after the reference of `stableId`. I haven't debugged as to why AdaptivePredict is so bad here, but it feels like it computes the transitions at least n^2 times the length of the stableId string.
* The two productions are basically from the Spec grammar.
* The fix is to unfold stableId into importExpr and refactor to remove the ambiguity.

PR #3281, opened 2023-03-20
CUDA PTX ISA 1.0

mlorek:


teverett:
@mlorek thanks!

kaby76:
I did a project to emulate a CUDA machine. https://code.google.com/archive/p/cuda-waste/source/default/source. There's a grammar there too, I think for ISA 3.something, and a huge test suite.

Issue #3284, opened 2023-03-21
[bibtex] Standardize grammar name on split grammar, add EOF for start rule, fix targets


kaby76:
This grammar isn't in correct format for testing. It requires several changes:
* The name of a grammar in split format should follow the standard convention, which is the parser grammar name should end with "Parser", and the lexer grammar name should end with "Lexer". The main problem is that it won't be handled correctly by trgen.
* A start rule that does not end in EOF can quit prematurely on error and say the input it just fine! It is best to always end a start rule with EOF to force the parser to consume all input. It also further identifies to trgen what the start rule is.
* The target list does not include Cpp nor CSharp. But those targets work. The CSharp target is used by the Trash toolkit, so I highly recommend adding the target.

PR #3285, opened 2023-03-21
[bibtex] Fix for #3284 -- standardize the grammar to work in build.

kaby76:
Fix for https://github.com/antlr/grammars-v4/issues/3284
* Rename parser grammar BibTex.g4 to BibTexParser.g4.
* Add EOF to the start rule.
* Add Cpp and CSharp targets.


teverett:
@kaby76 thanks!

Issue #3287, opened 2023-03-21
[mysql] Maven testing pom.xml for mysql messed up


kaby76:
`mvn test` in grammars-v4/sql/mysql/Positive-Technologies` results in errors. This is because Maven requires not only downward links to file system sub-directories and files, but it requires parent chaining as well, which is really quite annoying. It seems you cannot skip subdirectories in that chain.

For example, [grammars-v4/sql/pom.xml points to mysql/Positive-Technologies/pom.xml](https://github.com/antlr/grammars-v4/blob/556e9305a95d7fc487e1794318d455c65c5e6346/sql/pom.xml#L21). But, [grammars-v4/sql/mysql/Positive-Technologies/pom.xml](https://github.com/antlr/grammars-v4/blob/556e9305a95d7fc487e1794318d455c65c5e6346/sql/mysql/Positive-Technologies/pom.xml#L9) points back to "sqlparent", but Maven does have a notion of a "virtual file system" here. Maven cannot find "sqlparent" in 'grammars-v4/sql/mysql", so the testing fails immediately.

There is still a problem in checking with Maven because [this](https://github.com/kaby76/grammars-v4/blob/556e9305a95d7fc487e1794318d455c65c5e6346/pom.xml#L13) line does not use a test plugin that uses a 4.12 runtime. https://github.com/antlr/antlr4test-maven-plugin

If the pom.xml are still there, some minimal support should be done.

PR #3289, opened 2023-03-21
[mysql] Fix pom.xml's for mysql grammar with Maven

kaby76:
Fix for https://github.com/antlr/grammars-v4/issues/3287
```
cd grammars-v4/sql/mysql
mvn test
```
This now works.


teverett:
@kaby76 thanks!

PR #3295, opened 2023-03-22
[SNOWFLAKE] Add missing function

MasterKuat:
predicate could be a select item
add trim familly function
add like, ilike and rlike predicate


sql/snowflake/SnowflakeParser.g4
```diff
@@ -3889,7 +3900,9 @@ predicate
     | expr comparison_operator (ALL | SOME | ANY) '(' subquery ')'
     | expr NOT* BETWEEN expr AND expr
     | expr NOT* IN '(' (subquery | expr_list) ')'
```
hellozrh, line 3902: the keyword NOT should only be used one or zero times. So :
```g4
    | expr NOT? BETWEEN expr AND expr
    | expr NOT? IN '(' (subquery | expr_list) ')'
```

sql/snowflake/SnowflakeParser.g4
```diff
@@ -3641,6 +3651,7 @@ as_alias
 
 expression_elem
     : expr as_alias?
+    | predicate as_alias?
```
hellozrh, line 3654:
```g4
    : (expr | predicate) as_alias?
```

MasterKuat: Fixed, thanks

sql/snowflake/SnowflakeParser.g4
MasterKuat: Wasn't mine but sure fixed.

sql/snowflake/SnowflakeParser.g4
```diff
@@ -3331,6 +3335,7 @@ binary_or_ternary_builtin_function
     : CHARINDEX
     | REPLACE
     | substring=( SUBSTRING | SUBSTR )
+    | ( LIKE | ILIKE )
```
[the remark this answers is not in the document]
MasterKuat: Fixed

PR #3297, opened 2023-03-23
[MySQL] Allow using of alias in single delete statement

mfvitale:


KvanTTT (changes requested):
Welcome to the ANTLR grammars repository!

Thanks for the contribution but please add an example to corresponding file that covers suggested changes to the [sql/mysql/Positive-Technologies/examples](https://github.com/antlr/grammars-v4/tree/master/sql/mysql/Positive-Technologies/examples) directory.

PR #3301, opened 2023-03-27
Fix tableview name

jiangtao69039:
example:
```sql
SELECT * FROM scott.emp@hq.acme.com;
```

**PERIOD link_name** could be more than one


hellozrh (commented):
LGTM，^_^

KvanTTT (approved):

PR #3302, opened 2023-03-27
[TSQL] Allow "raw" for ID

### LINKED ISSUE #3300: [TSQL] Cannot parse "raw" as ID
**What happened:**
I want to be able to create objects called "raw".
However, as there is an own lexer rule `RAW` that appears before the generic `ID` the token is not recognized as such.

**Expected:** I want to parse for example,
- `CREATE DATABASE raw`
- `CREATE SCHEMA raw`
- `CREATE TABLE raw`
- `SELECT * FROM raw`
- ...

(Using brackets, a la `[raw]`, works)

**Actual:** Parsing fails 

---

I can confirm that moving lexer rule `RAW` after `ID` solves it.
 
❓ However, given the complexity of the grammar, I am uncertain whether this is the best fix here? 
As most likely the ordering issue applies to other ID names besides "raw" as well...

---

image

svengiegerich:
**Motivation**: Fix #3300 

**Changes**: As [suggested](https://github.com/antlr/grammars-v4/issues/3300#issuecomment-1484053327) by @bkiers, extend the parser rule `id_` by the lexer rule `RAW`.


hellozrh (commented):
Nice work, this SGTM.

kaby76:
> @KvanTTT I'm not too familiar with the testing regime here, but it seems to fail for cases unrelated to the change (& that are supposed to fail)? E.g. [this error](https://github.com/antlr/grammars-v4/pull/3302/checks#step:18:41) relates to [that line](https://github.com/antlr/grammars-v4/blob/07bf5e052b5bf61f679c0e82a3c6abcdd3ae071b/sql/tsql/examples/constants.sql#L13), or what am I missing?

The reason it is failing is because the parsing error messages--which are expected--have changed because you changed the grammar. You need to remaster the .errors and .tree fils if the changes are expected.

Test works as follows:
* Parsing is accomplished using a driver generated from templates. The templates are in [_scripts/templates](https://github.com/antlr/grammars-v4/tree/master/_scripts/templates). The "test.sh" or "test.ps1" scripts do the actual testing.
* Parsing is first done for all input. The drivers create .errors and .tree files. The .errors files are non-empty files that contain the parse errors for a particular file. E.g., [constants.sql.errors](https://github.com/antlr/grammars-v4/blob/fe8ee8e03ffc4af9270e430a17817d25480b72f5/sql/tsql/examples/constants.sql.errors). Error messages from the parse change when you change the grammar. The .tree files are non-empty files that contain the parse tree as outputted by [toStringTree()](https://javadoc.io/doc/org.antlr/antlr4-runtime/latest/org/antlr/v4/runtime/tree/ParseTree.html#toStringTree(org.antlr.v4.runtime.Parser)).
* If there are any diffs in the .errors or .tree files, the test fails.

To remaster the .errors and .tree files, the easiest is to use [trgen](https://github.com/kaby76/Domemtech.Trash/tree/main/src/trgen) to generate a driver, build it, and run it. Once completed, check in those files that have changed. Otherwise you can hand-edit the files, but that's error prone and may take you a while.

KvanTTT (approved):

KvanTTT:
Welcome to ANTLR 4 grammars repository!

PR #3303, opened 2023-03-27
Fix wrong symbol in diSubrange rule

hernanponcedeleon:
I think there is an error in the symbol of the `diSubrange` rule (I was not able to parse several llvm files containing debug information in the form of metadata). 

According to my understanding from the [llvm documentation](https://llvm.org/doxygen/classllvm_1_1Metadata.html), the `diSubrange` rule should use `DISubrange` instead of `DICompileUnit`.

### DIFF
--- a/llvm-ir/examples/disubrange.ll
+++ b/llvm-ir/examples/disubrange.ll
@@ -1,0 +1,1 @@
+!0 = !DISubrange(count: 3)


hernanponcedeleon:
Thanks @KvanTTT for the feedback!

Would a simple example (along the lines of [this one](https://github.com/antlr/grammars-v4/blob/master/llvm-ir/examples/diexpression.ll)) be sufficient? 

A file containing
```
!0 = !DISubrange(count: 3)
```
shows that the current version of the grammar finds no viable alternative at input `!DISubrange`.

KvanTTT:
Yes, it's enough.

KvanTTT (approved):

PR #3304, opened 2023-03-28
postgresql moudle:  remove the unnecessary code.

hellozrh:


sql/postgresql/Java/PostgreSQLParserBase.java
```diff
@@ -42,7 +42,6 @@ public void ParseRoutineBody(PostgreSQLParser.Createfunc_opt_listContext _localc
         }
         if (func_as != null) {
             String txt = GetRoutineBodyString(func_as.func_as().sconst(0));
-            int line = func_as.func_as().sconst(0).start.getLine();
```
KvanTTT, line 45: Could you please also remove this line from other base files? For instance, I see the similar in [PostgreSQLParserBase.cs](https://github.com/antlr/grammars-v4/blob/master/sql/postgresql/CSharp/PostgreSQLParserBase.cs#L38)

PR #3008, opened 2023-01-11
[PostgreSQL] Built-in functions, merge statement and Non Ansi Join

ealight:
- Added all the mathematics built-in functions. ([Source](https://www.postgresql.org/docs/15/functions-math.html))
- Added all the string built-in functions. ([Source](https://www.postgresql.org/docs/15/functions-string.html))
- Added all the date built-in functions. ([Source](https://www.postgresql.org/docs/15/functions-datetime.html))
- Added all the date formatting built-in functions. ([Source](https://www.postgresql.org/docs/15/functions-formatting.html))
- Added MERGE statement handler([Source](https://www.postgresql.org/docs/current/sql-merge.html))
- Added old fashion Non Ansi Join handler ([Source](https://sqlenlight.com/support/help/sa0006/))

PR #3308, opened 2023-03-30
[PostgreSql] Fix 'LEFT/RIGHT' joinType would be recognized as table alias

jiefei30:
Because the "**builtin_function_name**" rule was added in #3008, which means that "**colid**" rule can match the "LEFT/RIGHT" keywords, which leads to "LEFT/RIGHT" being recognized as table alias when user use the "LEFT/RIGHT" joinType and the master table without alias.

The old parse tree:

The new parse tree:

### DIFF
--- a/sql/postgresql/PostgreSQLParser.g4
+++ b/sql/postgresql/PostgreSQLParser.g4
@@ -3164,8 +3164,12 @@ alias_clause
    ;
 
 opt_alias_clause
-   : AS? table_alias
+   : table_alias_clause
    |
+   ;
+
+table_alias_clause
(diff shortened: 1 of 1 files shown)


KvanTTT (approved):

KvanTTT (changes requested):
Could you please add an sample that covers suggested changes?

jiefei30:
> Could you please add an sample that covers suggested changes?

This modification does not add new syntax, but adjusts the matching logic of the LEFT/RIGHT keywords. Corresponding examples are available in grammars-v4/sql/postgresql/examples/join.sql

PR #3318, opened 2023-04-05
[Snowflake]handle boolean operation

MasterKuat:
In select statement we could have boolean operation witch result in boolean field.


teverett:
@MasterKuat thanks!

PR #3324, opened 2023-04-07
Change order of expression for Java/Java grammar to parse correctly

### LINKED ISSUE #3323: Java/Java order of expression has been updated wrong over time
"New" expressions in the Java/Java grammar (such as lambdaExpressions, switchexpressions or Java8 methodreferences) have been added in the wrong place over time, causing incorrect parsing of for example lambda-expressions in ternary operators. 

```
import java.util.regex.Pattern.CharPredicate;
public class LambdaInTernaryOperator {
    public static void main(String[] args){
        Charpredicate a = func("a", true);
    }
    public static CharPredicate func(String name, boolean caseIns){
        return caseIns ? c -> Character.isUpperCase(c) ||
                                Character.isLowerCase(c) ||
                                Character.isTitleCase(c)
                        : Character::isUpperCase;;

    }
    
}
```

Proposed solution: Change the order for expressions to be resolved.
 
Adding merge request to resolve the issue shortly.

DagSonntag:
Closes #3323

See issue for discussion on the problem.


teverett:
@DagSonntag thanks!

PR #3325, opened 2023-04-08
C grammar, standard C90

andr1972:
C Grammar , old C90 standard, tested on 216 includes standard library , mainly from /usr/include, some from /usr/include/x86_64-linux-gnu/bits. Only these headers, which not caused error when GCC was called with "-std=c90 -fsyntax-only -E" to file /dev/shm/file.c, next parsed with Antr4 and C++ runtime. 
C99, C11, C17 can be created incrementally, and could be tested on big C repositories.

### DIFF
--- a/swift/swift5/CSharp/SwiftSupport.cs
+++ b/swift/swift5/CSharp/SwiftSupport.cs
@@ -1,0 +1,443 @@
+using System;
+using System.Collections;
+using System.IO;
+using Antlr4.Runtime;
+using Antlr4.Runtime.Misc;
+
+public abstract class SwiftSupport : Parser
+{
+
+    protected ITokenStream _input;
+
+    protected SwiftSupport(ITokenStream input, TextWriter output, TextWriter errorOutput)
+        : base(input, output, errorOutput)
+    {
+        _input = TokenStream;
+    }
+    /* TODO
+    There is one caveat to the rules above. If the ! or ? predefined operator
+     has no whitespace on the left, it is treated as a postfix operator,
+     regardless of whether it has whitespace on the right. To use the ? as
+     the optional-chaining operator, it must not have whitespace on the left.
+      To use it in the ternary conditional (? :) operator, it must have
+      whitespace around both sides.
+    */
+
+    /*
+operator-head : /  =  -  +  !  *  %  <  >  &  |  ^  ~  ?
+  | [\u00A1-\u00A7]
+  | [\u00A9\u00AB]
+  | [\u00AC\u00AE]
+  | [\u00B0-\u00B1\u00B6\u00BB\u00BF\u00D7\u00F7]
+  | [\u2016-\u2017\u2020-\u2027]
+  | [\u2030-\u203E]
+  | [\u2041-\u2053]
+  | [\u2055-\u205E]
+  | [\u2190-\u23FF]
+  | [\u2500-\u2775]
+  | [\u2794-\u2BFF]
+  | [\u2E00-\u2E7F]
+  | [\u3001-\u3003]
+  | [\u3008-\u3030]
+  ;
+ */
+    public static BitArray operatorHead = new BitArray(0x10000);
+    public static BitArray operatorCharacter;
+
+    public static BitArray leftWS = new BitArray(255);
+    public static BitArray rightWS = new BitArray(255);
+
+    static SwiftSupport()
+    {
+        // operator-head → /  =­  -­  +­  !­  *­  %­  <­  >­  &­  |­  ^­  ~­  ?­
+        operatorHead.Set('/', true);
+        operatorHead.Set('=', true);
+        operatorHead.Set('-', true);
+        operatorHead.Set('+', true);
+        operatorHead.Set('!', true);
+        operatorHead.Set('*', true);
+        operatorHead.Set('%', true);
+        operatorHead.Set('<', true);
+        operatorHead.Set('>', true);
+        operatorHead.Set('&', true);
+        operatorHead.Set('|', true);
+        operatorHead.Set('^', true);
+        operatorHead.Set('~', true);
+        operatorHead.Set('?', true);
+
+        // operator-head → U+00A1–U+00A7
+        _setBitRange(ref operatorHead, 0x00A1, 0x00A7);
+
+        // operator-head → U+00A9 or U+00AB
+        operatorHead.Set((0x00A9), true);
+        operatorHead.Set((0x00AB), true);
+
+        // operator-head → U+00AC or U+00AE
+        operatorHead.Set((0x00AC), true);
+        operatorHead.Set((0x00AE), true);
+
+        // operator-head → U+00A9 or U+00AB
+        operatorHead.Set((0x00A9), true);
+        operatorHead.Set((0x00AB), true);
+
+        // operator-head → U+00B0–U+00B1, U+00B6, U+00BB, U+00BF, U+00D7, or U+00F7
+
+        operatorHead.Set(0x00B0, true);
+        operatorHead.Set(0x00B1, true);
+        operatorHead.Set((0x00B6), true);
+        operatorHead.Set((0x00BB), true);
+        operatorHead.Set((0x00BF), true);
+        operatorHead.Set((0x00D7), true);
+        operatorHead.Set((0x00F7), true);
+
+        // operator-head → U+2016–U+2017 or U+2020–U+2027
+        operatorHead.Set(0x2016, true);
+        operatorHead.Set(0x2017, true);
+        _setBitRange(ref operatorHead, 0x2020, 0x2027);
+
+        // operator-head → U+2030–U+203E
+        _setBitRange(ref operatorHead, 0x2030, 0x203E);
+
+        // operator-head → U+2041–U+2053
+        _setBitRange(ref operatorHead, 0x2041, 0x2053);
+
+        // operator-head → U+2055–U+205E
+        _setBitRange(ref operatorHead, 0x2055, 0x205E);
+
+        // operator-head → U+2190–U+23FF
+        _setBitRange(ref operatorHead, 0x2190, 0x23FF);
+
+        // operator-head → U+2500–U+2775
+        _setBitRange(ref operatorHead, 0x2500, 0x2775);
+
+        // operator-head → U+2794–U+2BFF
+        _setBitRange(ref operatorHead, 0x2794, 0x2BFF);
+
+        // operator-head → U+2E00–U+2E7F
+        _setBitRange(ref operatorHead, 0x2E00, 0x2E7F);
+
+        // operator-head → U+3001–U+3003
+        _setBitRange(ref operatorHead, 0x3001, 0x3003);
+
+        // operator-head → U+3008–U+3030
+        _setBitRange(ref operatorHead, 0x3008, 0x3030);
+
+        // operator-character → operator-head­
+        operatorCharacter = (BitArray)operatorHead.Clone();
+
+        // operator-character → U+0300–U+036F
+        _setBitRange(ref operatorCharacter, 0x0300, 0x036F);
+
+        // operator-character → U+1DC0–U+1DFF
+        _setBitRange(ref operatorCharacter, 0x1DC0, 0x1DFF);
+
+        // operator-character → U+20D0–U+20FF
+        _setBitRange(ref operatorCharacter, 0x20D0, 0x20FF);
+
+        // operator-character → U+FE00–U+FE0F
+        _setBitRange(ref operatorCharacter, 0xFE00, 0xFE0F);
+
+        // operator-character → U+FE20–U+FE2F
+        _setBitRange(ref operatorCharacter, 0xFE20, 0xFE2F);
+
+        // operator-character → U+E0100–U+E01EF
+        // Java works with 16-bit unicode chars. However, it can work for targets in other languages, e.g. in Swift
+        // operatorCharacter.set(0xE0100,0xE01EF+1);
+
+        leftWS.Set(Swift5Parser.WS, true);
+        leftWS.Set(Swift5Parser.LPAREN, true);
+        leftWS.Set(Swift5Parser.LBRACK, true);
+        leftWS.Set(Swift5Parser.LCURLY, true);
+        leftWS.Set(Swift5Parser.COMMA, true);
+        leftWS.Set(Swift5Parser.COLON, true);
+        leftWS.Set(Swift5Parser.SEMI, true);
+        leftWS.Set(Swift5Parser.Interpolation_multi_line, true);
+        leftWS.Set(Swift5Parser.Interpolation_single_line, true);
+
+        rightWS.Set(Swift5Parser.WS, true);
+        rightWS.Set(Swift5Parser.RPAREN, true);
+        rightWS.Set(Swift5Parser.RBRACK, true);
+        rightWS.Set(Swift5Parser.RCURLY, true);
+        rightWS.Set(Swift5Parser.COMMA, true);
+        rightWS.Set(Swift5Parser.COLON, true);
+        rightWS.Set(Swift5Parser.SEMI, true);
+        rightWS.Set(Swift5Parser.Line_comment, true);
+        rightWS.Set(Swift5Parser.Block_comment, true);
+    }
+
+    private static void _setBitRange(ref BitArray bitArray, int firstIndex, int lastIndex)
+    {
+        for (var i = firstIndex; i < lastIndex; i++)
+        {
+            bitArray.Set(i, true);
+        }
+    }
+
+    private static bool _IsCharacterFromSet(IToken token, BitArray bitArray)
+    {
+        if (token.Type == Eof)
+        {
+            return false;
+        }
+
+        String text = token.Text;
+        int codePoint = text[0];
+
+        //  Determine number of characters needed to represent the codePoint character
+        if ((codePoint >= 0x10000 ? 2 : 1) != text.Length)
+        {
+            // not a single character
+            return false;
+        }
+
+        return operatorCharacter.Get(codePoint);
+    }
+
+    public bool isOperatorHead(IToken token)
+    {
+        return _IsCharacterFromSet(token, operatorHead);
+    }
+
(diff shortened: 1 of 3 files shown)


borneq:
FuncCallwithVarArgs.c :
"int a2(int param1, param2)" pass but it is Clang quirk and parser don't distinguish between parameter
without type and named type without paramter.

FunctionReturningPointer.c 
My compilers don't compile 
int MyStruct *
f3 (
   int             param1,
   char            param2
  );
where both int and MyStruct are types

pr403.c:
Joining lines at \ is the first preprocessor job, 
\ has the highest priority
GCC allows even:
#include <stdio.\
h>

borneq:
although I made FunctionReturningPointer.c.errors 
I have
```
========
Parsing tests of c for Java failed.
Difference in output.
diff --git a/c/examples/FunctionReturningPointer.c.errors b/c/examples/FunctionReturningPointer.c.errors
index b2bae35a..5634b8f9 100644
--- a/c/examples/FunctionReturningPointer.c.errors
+++ b/c/examples/FunctionReturningPointer.c.errors
@@ -1 +1 @@
-line 16:13 no viable alternative at input 'intMyStruct*'
\ No newline at end of file
+line 16:13 no viable alternative at input 'intMyStruct*'
Test failed.
========
Error: Process completed with exit code 1.
```
https://github.com/parstools/grammars-v4/actions/runs/4657790061/jobs/8242732094#step:18:45

kaby76:
> diff --git a/c/examples/FunctionReturningPointer.c.errors b/c/examples/FunctionReturningPointer.c.errors
> index b2bae35a..5634b8f9 100644
> --- a/c/examples/FunctionReturningPointer.c.errors
> +++ b/c/examples/FunctionReturningPointer.c.errors
> @@ -1 +1 @@
> -line 16:13 no viable alternative at input 'intMyStruct*'
> \ No newline at end of file
> +line 16:13 no viable alternative at input 'intMyStruct*'
> Test failed.
> ========
> Error: Process completed with exit code 1.
> ```
> 
> https://github.com/parstools/grammars-v4/actions/runs/4657790061/jobs/8242732094#step:18:45

Every single error message from the parser must end with a newline. You should use trgen to remaster your test files. I don't know why you left the bogus input in the file though.

kaby76:
NB!!

You placed an entirely new grammar for C in the same directory as another directory. They are not the same and it confuses everyone--and the build tool--when you have the two in one directory. The build tool can handle multiple grammars, but you have to program for the two.

The build tool opens every grammar in the directory and parses it. It then looks for the grammar name and the start rule. There are two different grammar names and two start rules.

Please create the following directory structure.

c/
c/c/
c/c90/

Place the existing files before your changes in c/c/. Place your new grammar in c/c90/. Add a desc.xml to c/c90/. Copy the c/c/examples/ to c/c90/examples.

kaby76 (changes requested):
* `git mv c/desc.xml c/C/desc.xml`. A copy of desc.xml is placed in the same directory as the .g4. (There should not be a desc.xml without a .g4; every directory that contains a .g4 needs a desc.xml.)
* `cp c/C/desc.xml grammars-v4/c/C90/desc.xml`.
* In `c/C/pom.xml`, change `<packageName>c</packageName>` to `<packageName></packageName>`. (`mvn clean test` was failing.)
* In `c/C90/pom.xml`, change `<packageName>c90</packageName>` to `<packageName></packageName>`. (`mvn clean test` was failing.)
* For the new C90 grammar, can we add the tests you used to validate the grammar? We may need to place those in another directory, like "c/C90/validation-large/" (or whatever you think). We can then validate the grammar against this for Java and CSharp since the other targets may be too slow.

borneq:
How large  can be test grammar? place preprocessed file or not preprocessed
with shell script calling preprocess?
wt., 11 kwi 2023 o 11:15 Ken Domino ***@***.***> napisał(a):

kaby76:
This looks good @borneq

I tested out the grammar with all the targets, and it looks like Cpp, CSharp, Dart, Java all parse fast. The Go target works fast too, but only on the most recent [dev fixes](https://github.com/antlr/antlr4/tree/dev), which we're not testing yet. But, of the targets mentioned, all take 3-6 s. Dart is the fastest at 3s, CSharp and Java slowest of these 5 targets. (My machine a standard AMD Ryzen 7 2700, 16G DDR4 1311 MHz, SSD, B450 motherboard. Single test from command-line, i.e., not a statistical analysis.)

@borneq Let's add in two test scenarios, one for short tests that all targets can pass, and the longer test scenario for the "fast" targets. Please update the c/C90/desc.xml file as follows:
```
   <targets>CSharp;Cpp;Dart;Go;Java;JavaScript;PHP;Python3;TypeScript</targets>
   <test>
      <name>long</name>
      <targets>Cpp;CSharp;Dart;Java</targets>
   </test>
   <test>
      <name>short</name>
      <targets>CSharp;Cpp;Dart;Go;Java;JavaScript;PHP;Python3;TypeScript</targets>
   </test>      
</desc>
```

kaby76 (approved):
@teverett This PR is looks good.

c/C90/C90.g4
```diff
@@ -0,0 +1,672 @@
...
+    | enumDeclaration
+    ;
+
+typeSpecifier
+    :   (storageVarSpecifier | typeQualifier)* typeName typeModifier* typeQualifier*
+    ;
+
+type
+    : typeName
+    | typeQualifier typeName
```
KvanTTT, line 101: Can it be simplified in the following way?

```suggestion
    : typeQualifier? typeName
```

c/C90/C90.g4
```diff
@@ -0,0 +1,672 @@
...
+    : '*'
+    ;
+
+fixedParameterList
+    : parameter (',' parameter)*
+    ;
+
+parameterList
+    : fixedParameterList
+    | fixedParameterList ',' '...'
```
KvanTTT, line 135:
```suggestion
    : fixedParameterList
    | fixedParameterList (',' '...')?
```

c/C90/C90.g4
```diff
@@ -0,0 +1,672 @@
...
+fragment
+SimpleEscapeSequence
+    :   '\\' ['"?abfnrtv\\]
+    ;
+
+fragment
+OctalEscapeSequence
+    :   '\\' OctalDigit
+    |   '\\' OctalDigit OctalDigit
+    |   '\\' OctalDigit OctalDigit OctalDigit
```
KvanTTT, line 641:
```suggestion
    :   '\\' OctalDigit (OctalDigit OctalDigit?)?
```

c/C90/C90.g4
```diff
@@ -0,0 +1,672 @@
...
+    ;
+
+fragment
+HexadecimalEscapeSequence
+    :   '\\x' HexadecimalDigit
+    |   '\\x' HexadecimalDigit HexadecimalDigit
```
KvanTTT, line 647:
```suggestion
    :   '\\x' HexadecimalDigit HexadecimalDigit?
```

c/C90/C90.g4
```diff
@@ -0,0 +1,672 @@
...
+
+type
+    : typeName
+    | typeQualifier typeName
+    | typeDeclaration
+    ;
+
+typeName
+    : 'int'
+    | ('unsigned'|'signed')? 'long' 'int'?
```
KvanTTT, line 107: I suggest extracting `'unsigned'|'signed'` to the separated rule since it's used in many places:

```g4
unsignedOrSigned
    : 'unsigned'
    | 'signed'
    ;
```

```suggestion
    | unsignedOrSigned? 'long' 'int'?
```

c/C90/C90.g4
```diff
@@ -0,0 +1,672 @@
...
+
+fixedParameterList
+    : parameter (',' parameter)*
+    ;
+
+parameterList
+    : fixedParameterList
+    | fixedParameterList ',' '...'
+    | 'void'
+    | /*empty*/
```
KvanTTT, line 137: Please don't use empty alternatives (it may degrade performance and cause generator warnings). Instead, use `?` operator in use sites.

c/C90/C90.g4
```diff
@@ -0,0 +1,672 @@
...
+    ;
+
+fixedParameterOrTypeList
+    : parameterOrType (',' parameterOrType)*
+    ;
+
+parameterOrTypeList
+    : fixedParameterOrTypeList
+    | fixedParameterOrTypeList ',' '...'
```
KvanTTT, line 146:
```suggestion
    : fixedParameterOrTypeList (',' '...')?
```

c/C90/C90.g4
```diff
@@ -0,0 +1,672 @@
...
+    | typeModifier variableName
+    | Identifier
+    ;
+
+variablePlace
+    : '(' variablePlace ')'
+    | typeModifier variablePlace
+    | typeQualifier variablePlace
+    | Identifier
+    |
```
KvanTTT, line 198: Also, get rid of empty alternative.

c/C90/C90.g4
```diff
@@ -0,0 +1,672 @@
...
+    ;
+
+visualExtension
+    : '__cdecl'
+    ;
+
+gccDeclaratorExtension1  //__attribute__ ((visibility ("default"))) or ((__always_inline__))
+    : '__attribute__' '(' '(' Identifier ')' ')'
+    | '__attribute__' '(' '(' Identifier '(' integerConstant ')' ')' ')'
+    | '__attribute__' '(' '(' Identifier '(' StringLiteral ')' ')' ')'
```
KvanTTT, line 422: Please simlify:

```suggestion
    : '__attribute__' '(' '(' Identifier ('(' (StringLiteral | integerConstant) ')')? ')' ')'
```

c/C90/C90.g4
```diff
@@ -0,0 +1,672 @@
...
+    :   gccAttribute (',' gccAttribute)*
+//    |   // empty
+    ;
+
+gccAttribute
+    : Identifier
+    | 'const'
+    | Identifier '(' StringLiteral ')'
+    | Identifier '(' Identifier ( ',' integerConstant )* ')'
+    | Identifier '(' integerConstant ( ',' integerConstant )* ')'
```
KvanTTT, line 444:
```suggestion
    : Identifier ('(' (StringLiteral | (Identifier | integerConstant) ( ',' integerConstant )*) ')')?
    | 'const'
```

c/C90/C90.g4
```diff
@@ -0,0 +1,672 @@
...
+    : 'enum' Identifier? '{' enumList '}'
+    ;
+
+enumList
+    : enumItem (',' enumItem)* ','?
+    ;
+
+enumItem
+    : Identifier
+    | Identifier '=' inclusiveOrExpression
```
KvanTTT, line 457:
```suggestion
    : Identifier ('=' inclusiveOrExpression)?
```

c/C90/C90.g4
```diff
@@ -0,0 +1,672 @@
...
+    :   NonzeroDigit Digit* IntegerSuffix?
+    |   '0' IntegerSuffix?
+    ;
+
+fragment
+IntegerSuffix
+    :   UnsignedSuffix LongSuffix?
+    |   UnsignedSuffix LongLongSuffix
+    |   LongSuffix UnsignedSuffix?
+    |   LongLongSuffix UnsignedSuffix?
```
KvanTTT, line 504:
```suggestion
    :   UnsignedSuffix (LongSuffix | LongLongSuffix)?
    |   (LongSuffix | LongLongSuffix) UnsignedSuffix?
```

c/C90/C90.g4
```diff
@@ -0,0 +1,672 @@
...
+    ;
+
+FloatingConstant
+    :   DecimalFloatingConstant FloatingSuffix?
+    ;
+
+fragment
+DecimalFloatingConstant
+    :   FractionalConstant ExponentPart? FloatingSuffix?
+    |   DigitSequence ExponentPart FloatingSuffix?
```
KvanTTT, line 573:
```suggestion
    :   (FractionalConstant ExponentPart? | DigitSequence ExponentPart) FloatingSuffix?
```

c/C90/C90.g4
```diff
@@ -0,0 +1,672 @@
...
+
+fragment
+DigitSequence
+    :   Digit+
+    ;
+
+fragment
+ExponentPart
+    :   'e' Sign? DigitSequence
+    |   'E' Sign? DigitSequence
```
KvanTTT, line 584:
```suggestion
    :   [eE] Sign? DigitSequence
```

c/C90/C90.g4
```diff
@@ -0,0 +1,672 @@
...
+    :   '"' SChar* '"'
+    ;
+
+fragment
+StringLiteralMulti
+    : '"' SChar* '\\'Newline (SChar* '\\'Newline)* SChar* '"'
+    ;
+
+StringLiteral
+    :   ('L')? (StringLiteralOne|StringLiteralMulti) //L"abc" - elements are ints
```
KvanTTT, line 605:
```suggestion
    :   'L'? (StringLiteralOne | StringLiteralMulti) //L"abc" - elements are ints
```

c/C90/C90.g4
```diff
@@ -0,0 +1,672 @@
...
+fragment
+HexadecimalEscapeSequence
+    :   '\\x' HexadecimalDigit
+    |   '\\x' HexadecimalDigit HexadecimalDigit
+    ;
+
+
+Whitespace
+    :   [ \t]+
+        -> skip
```
KvanTTT, line 653: Please use `channel(HIDDEN)` instead of `skip` (to prevent data loss).

```suggestion
        -> channel(HIDDEN)
```

c/C90/C90.g4
```diff
@@ -0,0 +1,672 @@
...
+    ;
+
+
+Newline
+    :   (   '\r' '\n'?
+        |   '\n'
+        )
+        -> skip
```
KvanTTT, line 661:
```suggestion
        -> channel(HIDDEN)
```

c/C90/C90.g4
```diff
@@ -0,0 +1,672 @@
...
+    ;
+
+BlockComment
+    :   '/*' .*? '*/'
+        -> skip
```
KvanTTT, line 666:
```suggestion
        -> channel(HIDDEN)
```

c/C90/C90.g4
```diff
@@ -0,0 +1,672 @@
...
+    ;
+
+Preprocessor
+    :   '#' ~[\r\n]*
+        -> skip
```
KvanTTT, line 671:
```suggestion
        -> channel(HIDDEN)
```

c/C90/C90.g4
```diff
@@ -0,0 +1,672 @@
...
+    |   HexadecimalConstant //to do integer suffix?
+    ;
+
+sign
+    :   '+' | '-'
+    ;
+
+DecimalConstant
+    :   NonzeroDigit Digit* IntegerSuffix?
+    |   '0' IntegerSuffix?
```
KvanTTT, line 496:
```suggestion
    :   (NonzeroDigit Digit* | '0') IntegerSuffix?
```

KvanTTT (changes requested):
Welcome to ANTLR grammars repository and thanks for the contribution!

But please fix the noted issues or answer why it's impossible.

borneq:
Thanks, I I'll take a look

borneq:
all done, except modifiersWithoutVariable must have empty alternative, ale
is left recursive

borneq:
Last commit [0782230](https://github.com/antlr/grammars-v4/pull/3325/commits/078223015e8e07438ecf9d1f05bb9361bd1adc50) with speedup. Main cause of slowing was K&R style functionDefinitionKandR which generated the same token as functionDefinition, now parametersKandRlist is not optional for K&R rule, is is option for general definition.
Rule parametersKandRlist has two similar alternatives, it is faster than option for parametersKandRlist .
I fix declarator.
Now speedup for tested file from 37 s to 4.2 s.
I am planning rewrite expression chain, to generate more flat trees:
instead of 
A-> Ab
A-> b
will
A->b+
for additions, multiplication and other, for example int a = 0+1+2+3+4+5+6+7+8+9;

borneq:
Syntax has GNU extension, although compilable with -std=c90, not -std=gnu90
my proposal: removing from C90.g4 syntax not covered by both gcc and clang, or use GCC not only with -std=c90 but also with -pedantic-errors

borneq:
It is many GCC extension, is not pure C90. Maybe rename C90 to GCC90 and later add second grammar C90 or IsoC90 with removing extensions. (GCC compile with -pedantic-errors) Will 2x faster if remove K&R style function definitions, but ISO C90 don't parse validation-large/std90.c and other preprocessed GCC includes.
K&R is slow because inside definition are semicolons and parser must have very long lookahead.

borneq:
I will correct typeName:  simple generator od type name generates enormous deep lookahead

borneq:
I change intTypeName to can be any combinations of components
```
intTypeName
    : ('int'| complex | signedUnsigned | longShort)+
    ;
```
floatTypeName is specified
```
floatTypeName
    : 'float'
    | 'double'
    | 'long' 'double'
    | 'double' 'long'
....other
```
If I test file
```
void fun() {
    long double f1;
    long double f2;    
    long double f3;
...etc
```
I obtain unlimited maximal k for lookahead.
prediction is from "long double f1" to "}", although now typeName is unambiguous,
"long" can be start type "long int" or alone "long" type.
But if next is "double" why predict further, especially if next is ';' ?
what next: I will test minimal grammar with only typenames and some other rules to see what is cause it.

borneq:
Syntax distinguish between variable and function declaration, speedup.
K&R style slows down only a bit, negligibly.

borneq:
disambiguation, speedup twice on test file

borneq:
Simple expressions like "n = 0" have deep expression chain, depth 20.
I try change expression chain to grammar expr: expr op expr with alternatives precedence like is Rust Antlr grammar, but it slow down by orders of magnitude.

borneq:
Is possible solve conflict by me and synchronize repositories or I must close this and create new pull request?
By way, creating new pull request will be desired, because my c90.g4 grammar has gnu extensions, should remove this extensions and maybe rename it to to c90gnu.

PR #3330, opened 2023-04-12
[java] fix consecutive semicolons problems 

flyinox:
* fix consecutive semicolons problems in packageDeclaration and importDeclaration
* add example

### DIFF
--- a/java/java/JavaParser.g4
+++ b/java/java/JavaParser.g4
@@ -39,11 +39,11 @@ compilationUnit
     ;
 
 packageDeclaration
-    : annotation* PACKAGE qualifiedName ';'
+    : annotation* PACKAGE qualifiedName ';'+
     ;
 
 importDeclaration
-    : IMPORT STATIC? qualifiedName ('.' '*')? ';'
+    : IMPORT STATIC? qualifiedName ('.' '*')? ';'+
     ;
 
 typeDeclaration
--- a/java/java/JavaParser.g4
+++ b/java/java/JavaParser.g4
@@ -34,7 +34,7 @@ parser grammar JavaParser;
 options { tokenVocab=JavaLexer; }
 
 compilationUnit
-    : packageDeclaration? importDeclaration* typeDeclaration*
+    : packageDeclaration? (importDeclaration | ';')* (typeDeclaration | ';')*
     | moduleDeclaration EOF
     ;
 
@@ -44,13 +44,11 @@ packageDeclaration
 
 importDeclaration
     : IMPORT STATIC? qualifiedName ('.' '*')? ';'
-    | ';'
     ;
 
 typeDeclaration
     : classOrInterfaceModifier*
       (classDeclaration | enumDeclaration | interfaceDeclaration | annotationTypeDeclaration | recordDeclaration)
-    | ';'
     ;
 
 modifier
--- a/java/java/JavaParser.g4
+++ b/java/java/JavaParser.g4
@@ -39,11 +39,12 @@ compilationUnit
     ;
 
 packageDeclaration
-    : annotation* PACKAGE qualifiedName ';'+
+    : annotation* PACKAGE qualifiedName ';'
     ;
 
 importDeclaration
-    : IMPORT STATIC? qualifiedName ('.' '*')? ';'+
+    : IMPORT STATIC? qualifiedName ('.' '*')? ';'
+    | ';'
     ;
 
 typeDeclaration
--- a/java/java/examples/ConsecutiveSemicolons.java
+++ b/java/java/examples/ConsecutiveSemicolons.java
@@ -1,0 +1,9 @@
+// Use case to ensure that consecutive semicolons
+// in package declaration and import declaration work properly
+
+import java.util.ArrayList;;
+package consecutive.semicolon;;
+
+public class ConsecutiveSemi {
+}
+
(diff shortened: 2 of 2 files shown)


java/java/JavaParser.g4
```diff
@@ -39,11 +39,11 @@ compilationUnit
     ;
 
 packageDeclaration
-    : annotation* PACKAGE qualifiedName ';'
+    : annotation* PACKAGE qualifiedName ';'+
```
msagca, line 42: I don't see such thing in the [spec](https://docs.oracle.com/javase/specs/jls/se20/html/jls-19.html#jls-19-7). Consecutive semicolons usually suggest that there are empty statements, not multiple semicolons after a single statement.

KvanTTT (changes requested):
I agree with @msagca It doesn't look a correct fix.

java/java/JavaParser.g4
hellozrh, line 42:
```suggestion
    : (annotation* PACKAGE qualifiedName)? ';'
```

java/java/JavaParser.g4
```diff
@@ -39,11 +39,11 @@ compilationUnit
...
     ;
 
 importDeclaration
-    : IMPORT STATIC? qualifiedName ('.' '*')? ';'
+    : IMPORT STATIC? qualifiedName ('.' '*')? ';'+
```
hellozrh, line 46:
```suggestion
    : (IMPORT STATIC? qualifiedName ('.' '*')? )?';'
```

hellozrh (approved):
Your example in the file ConsecutiveSemi.java has Java grammar errors, the ‘package’ statement is not allowed after  'import' statement.

java/java/JavaParser.g4
flyinox: apologize for not stating the question clearly. The situation is as follows:
According to the "compileUnit" rule: "packageDeclaration? importDeclaration* typeDeclaration*"
If there are consecutive semicolons after "packageDeclaration" or in a non-last "importDeclaration", these trailing semicolons will be parsed as "typeDeclaration". As a result, the "importDeclaration" and "typeDeclaration" after the trailing semicolons will not have a chance to be correctly parsed.

java/java/JavaParser.g4
```diff
@@ -39,11 +39,12 @@ compilationUnit
...
 
 packageDeclaration
-    : annotation* PACKAGE qualifiedName ';'+
+    : annotation* PACKAGE qualifiedName ';'
     ;
 
 importDeclaration
-    : IMPORT STATIC? qualifiedName ('.' '*')? ';'+
+    : IMPORT STATIC? qualifiedName ('.' '*')? ';'
+    | ';'
```
msagca, line 47: This is a better fix even though it makes the rule `compilationUnit` ambiguous (semicolons can be recognized as either `importDeclaration` or `importDeclaration`).

flyinox: Yes, and I have noticed this issue in the description of the [spec](https://docs.oracle.com/javase/specs/jls/se20/html/jls-19.html#jls-19-7) as well

teverett:
@KvanTTT  good to merge this?

java/java/JavaParser.g4
hellozrh: > 

The demo in the pr show the double semicolon after the statement of packageDeclaration and importDeclaration, but all the other statement with double semicolon ended is correct。So I suggest We should fix other situations。

flyinox:
@KvanTTT I have made some adjustments, please help review and see if it is ready to be merged.

java/java/JavaParser.g4
```diff
@@ -44,6 +44,7 @@ packageDeclaration
 
 importDeclaration
     : IMPORT STATIC? qualifiedName ('.' '*')? ';'
+    | ';'
```
KvanTTT, line 47: Please simplify in the following way:

```suggestion
    : (IMPORT STATIC? qualifiedName ('.' '*')?)? ';'
```

KvanTTT (changes requested):

KvanTTT:
Can't we use the following structure?

KvanTTT (commented):
Can't we use the following structure instead of your suggestions?

```g4
compilationUnit
    : packageDeclaration? (importDeclaration | ';')* typeDeclaration*
    | moduleDeclaration EOF
    ;

...

importDeclaration
    : IMPORT STATIC? qualifiedName ('.' '*')? ';'
    ;
...

It looks more correct because `importDeclaration` actually always contains `IMPORT` unlike your variant.

File: java/java/JavaParser.g4 (hunk 4 above)
Comment (msagca) [reply java/java/JavaParser.g4:47]: I don't think this makes the rule easier to read/process.

**Comment by msagca** [comment]
As @KvanTTT suggested, empty statements (`;`) inside the rules `importDeclaration` and `typeDeclaration` can be moved to `compilationUnit`:

```antlr
compilationUnit
: packageDeclaration? (importDeclaration | ';')* (typeDeclaration | ';')*
| moduleDeclaration EOF
;
```

flyinox:
> Yes, I also like it more. @flyinox please fix in the suggested way.

Done, `;` have been removed from both the typeDeclaration and importDeclaration

teverett:
@flyinox  thanks!

PR #3331, opened 2023-04-13
[JSON] Fix exponent number not allowing leading 0s

Marcono1234:
This can be seen on https://www.json.org where after `e | E` the grammer only shows `digits` instead of `digit | onenine digits`; or in [RFC 8259](https://www.rfc-editor.org/rfc/rfc8259#section-6):
```
exp = e [ minus / plus ] 1*DIGIT
```


teverett:
@Marcono1234  thanks!

PR #3339, opened 2023-04-23
grammars fix to sql/trino

BigStupideer:
update sql/trino/TrinoLexer.g4  ELSE_ -> ELSE


teverett:
@BigStupideer  thanks!

PR #3343, opened 2023-04-25
Updated order of Java/Java expression parsing order

### LINKED ISSUE #3342: Casting and :: lambda expressions not parsed correctly in Java/Java grammar
Hi
I am sitting with the following java code and the Java/Java grammar
```
public class LambdaOrder {
    void forEachRemaining(Consumer<? super Double> action) {
            forEachRemaining((DoubleConsumer) action::accept);
    }
}
```

In the existing grammar this parses the Casting and Lambda incorrectly into the following parse-tree:
image

As can be seen there the casting is "applied" on the "action" identifier there, while it should be applied on the whole lambda as seen here:
image

This can be solved by changing the order of the casting and :: rules in the expression grammar in the grammar-file.
 (existing syntax is:
```
 expression
    : primary
       ...
    | '(' annotation* typeType ('&' typeType)* ')' expression
      ...
    | expression '::' typeArguments? identifier
    ...
```
 
While the following syntax solves the issue: 
```
expression
    : primary
       ...
    | expression '::' typeArguments? identifier
    | '(' annotation* typeType ('&' typeType)* ')' expression
     ...
```
)

However, I don't understand why this solves the issue?  Does the parsing not start from the outside in, i.e. for`forEachRemaining((DoubleConsumer) action::accept);` it starts with finding the method (with expressions), then take a single expression which it first identifies as a casting expression (since it is before the :: rule), and then finds the sub-expression to the casting expression to be a :: case?

Or does the syntax start from inside out? First identifying `(DoubleConsumer) action` and then go on to `(DoubleConsumer) action::accept`?

Also, is the solution good, or is the correct way to use <assoc=right> rules here instead.

Thankful for any clarifications! I can make the correct pull request after discussions :)

DagSonntag:
As related to the discussion in #3342 this is an attempt to update the table to solve the issue. 
There is also a discussion whether a more major rework is required, but this solves the (incorrect parsing) issue discussed there at least. 
I also added an example to avoid regression.

Closes #3342


kaby76:
This looks better. I like the html link and added comments for each level.

teverett:
@DagSonntag thanks!

PR #3347, opened 2023-04-26
[Snowflake]handle JS Function and add keyword

MasterKuat:
Add builtin function SPLIT, SPLIT_PART and TO_BOOLEAN
Add support for javascripted FUNCTION and new keyword MEMOIZABLE (for SQL function only)
Add lexer token to valid field name
Add expr as predicate (by the way if expr return a boolean could be in where clause)


teverett:
@MasterKuat  thanks!

Issue #3348, opened 2023-04-26
[verilog] VerilogLexer/VerilogPreParser grammar combo produces empty tree


kaby76:
The VerilogPreParser.g4 grammar uses VerilogLexer.g4. It is not currently tested. I decided to fix the desc.xml file to test this grammar pair and found out that VerilogPreParser does not appear to work on any input. Perhaps it is supposed to read other input files? The pom.xml does not seem to test the grammar pair either, so I think it's been untested for quite a while, even though it was change only two weeks ago. How was it tested??

See https://github.com/antlr/antlr4/issues/4247#issuecomment-1523314617.

msagca:
Hi @kaby76 

VerilogPreParser is supposed to parse the compiler directives and tokens belonging to directives are on a separate channel.

kaby76:
The grammar should be fitted into the testing so it can be tested. This can be done one of two ways:

1) Create a separate directory and input files for the directives. Add two `<test>` elements to the desc.xml file, one to test the VerilogLexer/VerilogParser and other to test the VerilogLexer/VerilogPreParser pair.
2) Add another start symbol to the VerilogParser that executes base class code to perform two parses on input that contains normal Verilog code and also with preprocessor directives. One parse would be for VerilogParser and the other for VerilogPreParser. In the base class code, read the input and construct a normal stream that has the default and special channel, and just pass this this to VerilogParser. Since the preprocessor directives are hidden, the parser should still work. In the second parse, create from the token stream a new stream that filters out all but the specialized channel for VerilogPreParser, and changes it to "default" channel. Pass this to the VerilogPreParser to parse. It's not clear how best to combine the parse trees, or even if this is necessary.

msagca:
@kaby76 

For the first option, is there a way to specify the token channel or the grammar name in `desc.xml`?

kaby76:
A desc.xml that could work is shown below. But, it requires the latest trgen, which I need to reference in the grammars-v4 testers. We just need meaningful "example-pre" input files.
```
<?xml version="1.0" encoding="UTF-8" ?>
<desc xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../_scripts/desc.xsd">
   <targets>CSharp;Dart;Java;JavaScript;Python3;TypeScript</targets>
   <test>
      <grammar-files>VerilogLexer.g4;VerilogParser.g4</grammar-files>
[...]

PR #3353, opened 2023-04-27
[GLSL, (System)Verilog] update desc.xml for testing preprocessor grammars

msagca:
#3348


msagca:
Hi @teverett,

For my previous commit, all except 1 test passed but the [remaining one](https://github.com/antlr/grammars-v4/actions/runs/10929847808/job/30341485407?pr=3353) timed out after 6 hours for no obvious reason. Now, again 23 out of 24 passed quickly but [one](https://github.com/antlr/grammars-v4/actions/runs/10945320334/job/30389300941?pr=3353) is still in progress and looks like it's stuck. The OS is MAC in both cases. It was CSharp target then and now it's Java.

Do you have any opinions on why this is happening?
Thanks.

kaby76:
The MacOS github machines are infamous. Always doing timeouts. Should be fine to mrge.

PR #3356, opened 2023-04-29
Create CGTLParser.g4

kevinlano:
Parser file for CGTL


msagca:
Hi @kevinlano 

I think this PR should contain everything that is necessary to build and test the grammar, i.e. parser/lexer grammars, example files, `desc.xml` and `pom.xml`. Can you please merge #3357 with this.

teverett:
@kevinlano yes, I think merging the PR's makes sense.  It also needs a desc.xml and perhaps an example.  Also, thanks for the contribution!

PR #3368, opened 2023-05-04
[antlr3] Removed useless parentheses.

kaby76:
Removed useless parentheses in antlr3.

### DIFF
--- a/antlr/antlr3/ANTLRv3Parser.g4
+++ b/antlr/antlr3/ANTLRv3Parser.g4
@@ -91,7 +91,7 @@ ruleScopeSpec
    ;
 
 block
-   : LPAREN ((optionsSpec)? COLON)?
+   : LPAREN (optionsSpec? COLON)?
                 alternative rewrite (OR alternative rewrite )*
         RPAREN
    ;
@@ -106,7 +106,7 @@ alternative
    ;
 
 exceptionGroup
-   : (exceptionHandler)+ (finallyClause)?
+   : exceptionHandler+ finallyClause?
    | finallyClause
    ;
 
@@ -152,7 +152,7 @@ notSet
    ;
 
 treeSpec
-   : TREE_BEGIN element (element)+ RPAREN
+   : TREE_BEGIN element element+ RPAREN
    ;
 
 ebnf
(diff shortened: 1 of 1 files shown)


KvanTTT (changes requested):
The grammar actually contains other rules with useless parenteses but unfortunately it's not possible to comment not changed code.

Could you please also remove useless parentheses from these lines?

* https://github.com/antlr/grammars-v4/blob/master/antlr/antlr3/ANTLRv3Parser.g4#L94
* https://github.com/antlr/grammars-v4/blob/master/antlr/antlr3/ANTLRv3Parser.g4#L109
* https://github.com/antlr/grammars-v4/blob/master/antlr/antlr3/ANTLRv3Parser.g4#L155

kaby76:
@KvanTTT  Thanks for noting these. It's been hard to get the patterns for useless parentheses correct. Changes added. And, here is the new script.
```
#!/bin/bash
echo Finding useless parentheses in grammars...
for i in `find . -name desc.xml | grep -v Generated\*`
do
  echo $i
  d=`dirname $i`
  pushd $d > /dev/null 2>&1
  trparse -t ANTLRv4 *.g4 | trxgrep '
	(: Find all blocks... :)
	//block[
		(: except not one of these ... :)
		not(./parent::ebnf/blockSuffix and ./altList/OR) and
		not(./parent::ebnf/blockSuffix and count(./altList/alternative/element) > 1) and
		not(./altList/OR and ../../following-sibling::element) and
		not(./altList/OR and ../../preceding-sibling::element)
	]/ancestor::parserRuleSpec' | trtext

  trparse -t ANTLRv4 *.g4 | trxgrep '
	(: Find all blocks... :)
	//lexerBlock[
		(: except not one of these ... :)
		not(./parent::lexerElement/ebnfSuffix and ./lexerAltList/OR) and
		not(./parent::lexerElement/ebnfSuffix and count(./lexerAltList/lexerAlt/lexerElements/lexerElement) > 1) and
		not(./lexerAltList/OR and ../following-sibling::lexerElement) and
		not(./lexerAltList/OR and ../preceding-sibling::lexerElement)
	]/ancestor::lexerRuleSpec' | trtext
  popd > /dev/null 2>&1
done
```

KvanTTT (approved):

KvanTTT:
Thanks a lot!

PR #3369, opened 2023-05-04
[antlr4] Remove useless parentheses

kaby76:
Remove the useless parentheses in the antlr4 grammar. Update to latest trgen since Cpp target is not working with additional cpp files.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -97,7 +97,7 @@ jobs:
     - name: Install trgen
       shell: bash
       run: |
-        for i in trgen triconv trwdog trxml trxml2; do dotnet tool install -g $i --version 0.20.15; done
+        for i in trgen triconv trwdog trxml trxml2; do dotnet tool install -g $i --version 0.20.17; done
     - name: Test
       shell: pwsh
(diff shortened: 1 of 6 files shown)


.github/workflows/main.yml
```diff
@@ -97,7 +97,7 @@ jobs:
     - name: Install trgen
       shell: bash
       run: |
-        for i in trgen triconv trwdog trxml trxml2; do dotnet tool install -g $i --version 0.20.15; done
+        for i in trgen triconv trwdog trxml trxml2; do dotnet tool install -g $i --version 0.20.17; done
```
KvanTTT, line 100: Version update doesn't relate to PR title. Please extract these changes to another PR.

KvanTTT (changes requested):

kaby76:
@KvanTTT Fixed. The PR now has modified only ANTLRv4Parser.g4 and desc.xml. The desc.xml is to get the build to work with Cpp, making explicit the "top-level" grammars that should be run through the Antlr4 tool.

KvanTTT (approved):

PR #3374, opened 2023-05-05
Support table statement and  Positive-Technologies README modification

caicancai:
Support tablestatement in mysql8.0:https://dev.mysql.com/doc/refman/8.0/en/table.html


caicancai:
@KvanTTT

caicancai:
@KvanTTT

PR #3375, opened 2023-05-05
[apex] Removed useless parentheses.

kaby76:
Removed useless parentheses in modifier, classOrInterfaceModifier, arrayInitializer, elementValueArrayInitializer, annotationTypeBody, apexDbUpsertExpression. I don't understand why modifier was grouped the way it was. Strange.


teverett:
@kaby76 thanks!

PR #3381, opened 2023-05-07
[abb] Remove useless parentheses in lexer grammar

kaby76:
Updated my script to check for useless parentheses in the lexer rules. This PR includes several more, which I missed.


teverett:
@kaby76 thanks!

PR #3382, opened 2023-05-07
[abnf] Remove useless parentheses.

kaby76:
Remove the useless parentheses I missed in the lexer rules.

NB: the expression `( ~ '>' )*` is equivalent to the simplified expression `~'>'*`. This works because the tree generated has `~` higher priority than the `*` operator.


teverett:
@kaby76 thanks!

PR #3384, opened 2023-05-08
[ada2012] Remove additional useless parentheses.

kaby76:
Additional useless parentheses discovered and removed.


teverett:
@kaby76  thanks!

PR #3385, opened 2023-05-08
[acme] Additional useless parentheses removed.

kaby76:
More useless parentheses detected and removed after updating my Trash script.


teverett:
@kaby76  thanks!

Issue #3389, opened 2023-05-08
[build] Build of javascript/javascript/ grammar fails due to bug in trgen.


kaby76:
I was reviewing https://github.com/antlr/grammars-v4/pull/3388 and noticed a bug in trgen that causes the test of this grammar to fail. The name of the javascript/javascript/Java/JavaScriptParser.java file is truncated due to a matching problem in the file name, and peels off too much.

This PR moves to the latest of trgen.

PR #3388, opened 2023-05-08
Remove functionDeclaration from anonymousFunction rule (antlr#3387)

ris58h:
Fixes https://github.com/antlr/grammars-v4/issues/3387

PR #3390, opened 2023-05-08
[build] Fix for #3389

kaby76:
This is a fix for #3389. PR https://github.com/antlr/grammars-v4/pull/3388 is failing because trgen computes the wrong filename for base class files. The fix is to use the latest version of trgen, which fixes the problem.


teverett:
@kaby76  thanks!

PR #3405, opened 2023-05-12
informix-sql:add grammars with support of drop access_method/drop aggregate/drop database .

hellozrh:
informix-sql:add grammars with support of drop access_method/drop aggregate/drop database .


sql/informix-sql/InformixSQLParser.g4
```diff
@@ -72,6 +75,33 @@ dropView
...
+    : identifier
+    ;
+
+aggregateName
+    : identifier
+    ;
+
+accessMethodName
+    : identifier
+    ;
```
KvanTTT, line 103: Do we really need separated rules for actually plain identifiers? Can't we use just `identifier` in `dropAccessMethod`, `dropAggregate`, `dropDatabase`?

hellozrh: Use different names for every statements,   grammar prompts will be easier,  in the future.

KvanTTT: > Use different names for every statements, grammar prompts will be easier, in the future.

Not sure because they are used only in one use-site and there are no other identifiers that could ambiguate.

Also, what about using [rule element labels](https://github.com/antlr/antlr4/blob/master/doc/parser-rules.md#rule-element-labels) instead of full-fledged rules? It looks more reasonable and faster.

```g4
dropAccessMethod
    : DROP ACCESS_METHOD (IF EXISTS)? accessMethodName=identifier RESTRICT
    ;
```

hellozrh: Yeah, using rule element labels is the best way. I'll rewrite them。Thank you, @KvanTTT .

KvanTTT (approved):

PR #3424, opened 2023-05-15
[masm] Removed useless parentheses in the masm grammar.

kaby76:
Removing a few unnecessary parentheses.


teverett:
@kaby76  thanks!

PR #3425, opened 2023-05-15
[b] Remove useless parentheses in the "b" grammar.

kaby76:
Removing several parentheses that are unnecessary.


teverett:
@kaby76 thanks!

PR #3432, opened 2023-05-16
[arithmetic] Removed useless parentheses.

kaby76:
Deleted several block parentheses that are unnecessary.


teverett:
@kaby76  thanks!

PR #3438, opened 2023-05-18
Fix python3 parse error

MR-BH:
fix case like:
```
print(0)
print(1)
```
cause error
Parse error: {
      startLine: 2,
      endLine: 2,
      startCol: 8,
      endCol: 13,
      message: "no viable alternative at input '<EOF>'"
    }

### DIFF
--- a/python/python3/Python3Parser.g4
+++ b/python/python3/Python3Parser.g4
@@ -68,7 +68,7 @@ varargslist: (vfpdef ('=' test)? (',' vfpdef ('=' test)?)* (',' (
 vfpdef: name;
 
 stmt: simple_stmts | compound_stmt;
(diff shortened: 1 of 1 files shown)


kaby76:
@MR-BH The Python3 target of the python/python3 grammar sometimes takes over 300s on a heavily loaded Github server, so the parse was killed. The test will need to be split or changed to individual parsing. For now, remove the target in the desc.xml. https://github.com/antlr/grammars-v4/blob/6e704069e7ea9b1d6a03bb1e4325017cdcfd120b/python/python3/desc.xml#L4

KvanTTT (changes requested):

But please add an example that covers suggested changes to the examples subdirectory.

kaby76:
I restructured the python grammars, and restructured the tests in python3. Please update to the latest source and try again. Thanks.

PR #3442, opened 2023-05-18
[build] Moved Trash scripts to https://github.com/kaby76/g4-checks.

kaby76:
Removed a couple of Trash scripts from this repo and moved them to https://github.com/kaby76/g4-checks for easier maintenance.


teverett:
@kaby76  thanks!

Issue #3447, opened 2023-05-19
[json] json/examples/numbers.json is always being checkout with Windows newlines.


kaby76:
Several weeks ago, [this PR](https://github.com/antlr/grammars-v4/commit/89776995d899393f1960a5c20e124218a697d465) added numbers.json. Unfortunately, if the repo is cloned on Linux, the file json/examples/numbers.json contains Windows newlines (0xd 0xa), while the other file gets expanded correctly with Linux newlines:
```
05/19-08:08:44 ~/aaa
$ git clone https://github.com/antlr/grammars-v4.git
Cloning into 'grammars-v4'...
remote: Enumerating objects: 44344, done.
remote: Counting objects: 100% (405/405), done.
remote: Compressing objects: 100% (279/279), done.
remote: Total 44344 (delta 153), reused 350 (delta 117), pack-reused 43939
Receiving objects: 100% (44344/44344), 43.73 MiB | 21.55 MiB/s, done.
Resolving deltas: 100% (23837/23837), done.
Updating files: 100% (8220/8220), done.
05/19-08:09:56 ~/aaa
$ cd grammars-v4/json
05/19-08:10:04 ~/aaa/grammars-v4/json
$ ls
desc.xml  examples/  JSON.g4  pom.xml
05/19-08:10:05 ~/aaa/grammars-v4/json
$ cd examples/
05/19-08:10:09 ~/aaa/grammars-v4/json/examples
$ od -t x1
example1.json  numbers.json
05/19-08:10:09 ~/aaa/grammars-v4/json/examples
$ od -t x1 numbers.json
0000000 5b 0d 0a 20 20 20 20 30 2c 0d 0a 20 20 20 20 2d
0000020 30 2c 0d 0a 20 20 20 20 31 32 33 34 35 36 37 38
0000040 39 30 2c 0d 0a 20 20 20 20 2d 31 2e 31 32 33 34
0000060 35 36 37 38 39 30 2c 0d 0a 20 20 20 20 2d 31 2e
0000100 32 65 33 2c 0d 0a 20 20 20 20 30 2e 30 2c 0d 0a
0000120 20 20 20 20 31 65 2b 31 2c 0d 0a 20 20 20 20 31
0000140 45 2b 31 2c 0d 0a 20 20 20 20 31 65 2d 32 33 2c
0000160 0d 0a 20 20 20 20 31 65 30 30 30 31 2c 0d 0a 20
[...]

PR #3448, opened 2023-05-19
Fix for #3447

kaby76:
Converted numbers.json to unix, added rule in .gitattributes to checkout with newlines according to OS for json/examples/*.json. I tested the clone/checkout of the branch on Windows and Linux--all works perfect now. https://github.com/antlr/grammars-v4/issues/3447


kaby76:
I don't know exactly why json/examples/numbers.json was messed up. I suspect that the person checked it in a file system across systems, like WSL2, checking in numbers.json with Linux newlines on Windows, or Windows newlines on Linux, or some equally bizarre situation. But, this resulted in every `git clone https://github.com/antlr/grammars-v4.git` having json/examples/numbers.json having always Windows newlines, even on Linux boxes.

I re-checked the branch changes, but git was having difficulties still. To check the changes, I ended up doing `git clone https://github.com/kaby76/grammars-v4.git; cd grammars-v4; git switch --discard-changes issue-3447` to verify that the newlines are correct everywhere, Linux or Windows.

Git always seems to give me problems. I do not trust it, and even less so through any UI anywhere.

If this doesn't work, I will force the .gitattribute to change the json/examples/*.json to always be Linux newlines.

kaby76:
https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings#refreshing-a-repository-after-changing-line-endings

teverett:
@kaby76  thanks!

PR #3455, opened 2023-05-22
TSqlParser.g4 add optional OR ALTER in create_view

AKryukov92:
add optional OR ALTER in create_view grammar


KvanTTT (changes requested):
Please cover this change by an example.

KvanTTT (approved):

PR #3458, opened 2023-05-22
TSqlParser.g4 add rule on_partition_or_filegroup in create_table

AKryukov92:


KvanTTT (changes requested):

But please add an example that covers suggested changes to the [tsql examples](https://github.com/antlr/grammars-v4/tree/master/sql/tsql/examples) directory.

PR #3459, opened 2023-05-22
TSql Add rule and keyword for LABEL in OPTIONS of table

AKryukov92:


sql/tsql/TSqlLexer.g4
```diff
@@ -527,6 +527,7 @@ KEY_PATH:                              'KEY_PATH';
 KEY_SOURCE:                            'KEY_SOURCE';
 KEY_STORE_PROVIDER_NAME:               'KEY_STORE_PROVIDER_NAME';
 KILL:                                  'KILL';
+LABEL:                                 'LABEL';
```
KvanTTT, line 530: Can `LABEL` be identifier? I guess so that why add this token to [keyword](https://github.com/antlr/grammars-v4/blob/master/sql/tsql/TSqlParser.g4#L5076) list.

KvanTTT (changes requested):
Please cover suggested changed by an example.

sql/tsql/TSqlLexer.g4
AKryukov92: According to 
https://learn.microsoft.com/en-us/sql/t-sql/language-elements/reserved-keywords-transact-sql?view=azuresqldb-current
LABEL is a keyword in Azure Synapse Analytics, which is tsql dialect too.
I'll remove this keyword here because those rules are general purpose

AKryukov92: If I add only rule in parser without keyword here, then on parsing
```
select whatever from tbl OPTION (MAXDOP 1, LABEL = 'Large');
```
it will throw errors:
```
mismatched input ',' expecting '.'
mismatched input 'OPTION' expecting '.'
mismatched input 'LABEL' expecting {'CONCAT', 'EXPAND', 'FAST', 'FORCE', 'HASH', 'IGNORE_NONCLUSTERED_COLUMNSTORE_INDEX', 'KEEP', 'KEEPFIXED', 'LOOP', 'MAXDOP', 'MAXRECURSION', 'MERGE', 'OPTIMIZE', 'ORDER', 'PARAMETRIZATION', 'RECOMPILE', 'ROBUST', 'USE', LABEL}
```
Is there any other way to support this syntax?

KvanTTT: I didn't mean adding the rule only to parser. I meant adding both token as you did at the beginning:

```g4
LABEL:                                 'LABEL';
```

and to `keyword` list:

```g4
keyword
    ...
    | LABEL
    ...
```

Issue #3460, opened 2023-05-22
[build] In the middle of 4.13.0 release of Antlr, Go builds won't work.


kaby76:
The builds aren't working because 4.12.0 of Go isn't supported.

PR #3461, opened 2023-05-22
[build] Remove Go target testing for now until 4.13.0 is fully out.

kaby76:
This is a fix for #3460, which is a blocking problem for testing cleanly any PRs. The Go target testing is removed for now until 4.13.0 is fully out.


teverett:
@kaby76  thanks

Issue #3468, opened 2023-05-23
[build] Dependabot upgrades for Antlr crash the build.


kaby76:
When Antlr is released, Github offers dependabot updates to upgrade some of the targets. Unfortunately, these updates can crash the build because of some assumptions:

* The Antlr tool version is a separate update from the runtime version update. When dependabot updates the runtime version, it has no clue that the Antlr tool version also needs to be updated.
* All targets use the same version of the tool.

To streamline updating to the next version of Antlr, the version of the tool should be derived from the version of the runtime, thus allowing dependabot updates to work. 

Dependabot will update [JavaScript](https://github.com/antlr/grammars-v4/pull/3465), [Python3](https://github.com/antlr/grammars-v4/pull/3464), [maven-source-plugin](https://github.com/antlr/grammars-v4/pull/3463), and [the root pom.xml](https://github.com/antlr/grammars-v4/pull/3462) (which really isn't used).

This means each target will be on its own version of Antlr. That's probably best because the Dart and PHP runtimes aren't released promptly. They usually take weeks because I have to ask people to cut a release and push it to public servers.

kaby76:
So, let's analyse each of the targets and see how an Antlr upgrade is handled.

* CSharp. The version of Antlr used is the version selected in the .csproj. https://github.com/antlr/grammars-v4/blob/4d7389ea494054dd952cfc6cd946b61e16de1679/_scripts/templates/CSharp/Test.csproj#L15. There's a bit of code to use a specific version of the tool, https://github.com/antlr/grammars-v4/blob/4d7389ea494054dd952cfc6cd946b61e16de1679/_scripts/templates/CSharp/Test.csproj#L11, but that should be removed. The build is accomplished by Antlr4BuildTasks, which downloads the JRE and the Antlr tool needed based on the version set for the runtime.

kaby76:
* Dart. The version of Antlr used is set [here](https://github.com/antlr/grammars-v4/blob/65556f3d8adcf84900bbb14c1dd4c40b7db9996c/_scripts/templates/Dart/pubspec.yaml#L7). We just need to extract the version out of the .yaml file, and make sure to use antlr4 with that version.

kaby76:
[...]

PR #3469, opened 2023-05-23
[build] Fixes for https://github.com/antlr/grammars-v4/issues/3468

kaby76:
Remove all mention of tool version in workflows/main.yml, because the version is per-target so dependabot updates don't crash the build. Instead install antlr4 tool and start "greppin'" for the version of the tool to use.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -69,6 +69,9 @@ jobs:
       if: ${{ matrix.language == 'Python3' }}
       run: |
         python --version
+    - name: Install Antlr tool
+      run: |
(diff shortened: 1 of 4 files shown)


kaby76:
@teverett This PR changes the update process for a release of Antlr. The good news is that Github Dependabot PR merges should no longer crash the build! This is because each target has an independent version of Antlr that it uses. We are now on version 4.13.0 for almost all targets except PHP and Dart. I have to bug the owners of those packages to make releases.

teverett:
@kaby76  thanks!

PR #3479, opened 2023-05-25
[build] For the JavaScript target, use npm scripts to call the Antlr tool.

kaby76:
This PR is a Javascript hack to run Antlr tool using the NPM package manager, which I learned from ChatGPT. This is important so that the versions of the Antlr tool and JavaScript runtime for Antlr are in synch. It also helps keep Dependabot working smoothly, and letting it work without crashing the build.

The hack is a Node javascript script that reads the package.json file and extracts the version of the Antlr runtime dependency. The program then performs an `execSync()` to call the [antlr4 tool](https://github.com/antlr/antlr4-tools). That tool does the actual downloading of the correct version for the Antlr tool .jar. We want the builds to look as much as possible written by a native JavaScript programmer, using the build tools native to JavaScript. Makefiles, Bash, and Powershell scripts are not a natural fit for JS programmers!

### DIFF
--- a/_scripts/templates/JavaScript/build.sh
+++ b/_scripts/templates/JavaScript/build.sh
@@ -3,9 +3,7 @@ set -e
 
 if [ -f transformGrammar.py ]; then python3 transformGrammar.py ; fi
 
-<tool_grammar_tuples:{x |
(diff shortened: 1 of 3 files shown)


kaby76:
Closing this off for now to clean out the PR queue. I'll revisit this at some point.

PR #3483, opened 2023-05-27
[vhdl] Rewrite grammar with caseInsensitive option

AVReshetov:


KvanTTT:
Hi. Welcome to ANTLR grammars repository and thanks for the contribution!

But I think it's more properly and clearly to use [`caseInsensitive`](https://github.com/antlr/antlr4/blob/master/doc/options.md#caseinsensitive) option here. Could you rewrite the grammar with that option?

As an example, you can take a look at [TSql](https://github.com/antlr/grammars-v4/blob/master/sql/tsql/TSqlLexer.g4) and other SQL grammars.

teverett:
@AVReshetov  thanks!

Issue #3478, opened 2023-05-25
[build] Dotnet tools should be updated with Dependabot!


kaby76:
Instead of the listed tasks in the [main.yml](https://github.com/antlr/grammars-v4/blob/4ee2f74c11722d0b7d6c52cc6a86d7f2564fe11c/.github/workflows/main.yml#L97) file that install Dotnet tools with specific version, we should have Dependabot scan and update these. See https://learn.microsoft.com/en-us/dotnet/core/tools/global-tools#install-a-local-tool and https://github.com/dependabot/dependabot-core/issues/1697. Upgrading should be easy as pie! This is the only way to go.

kaby76:
Dependabot on Github does not use the latest dependabot-core package. And, there isn't a way to say what version of the core to use. Closing for now.

PR #3485, opened 2023-05-28
[build] Fix for #3478. Add Dependabot processing of dotnet tools.

kaby76:
This is a fix for #3478.

Dependabot does not understand Bash scripts. So, version numbers for the Dotnet tools that the CI build use are never updated. Dependabot should update the tools because the are critical, and often, updates are in response to a use case in this repo.

Unfortunately, Dependabot only understands "local" dotnet tools specified in .config/dotnet-tools.json. We can get Dependabot to update the Dotnet tools used, but only if these tools are made "local." Local tools are installed via "dotnet tool restore." With the switch from global tools to local tools, all the build templates in _scripts/templates have to be updated.

This PR changes how the Trash tools are used. First, we needed to create the file .config/dotnet-tools.json. This file sets the versions of the tools to restore. The directory .config/ also contains an uninstall.sh script to clear the cache of local tools, if you want to do so, but normally is not required.

Second, we needed to update the .github/dependabot.yml file. Here is where things got tricky because Dependabot-core, the basic Ruby software that Github Dependabot runs, has a bug, where the software does not recognize and maintain the dependabot.yml file _if the directory that .config/ resides doesn't have a .csproj file!_ Consequently, this PR adds the file `.fake.csproj` to the root of the repo in order to get Dependabot to work.

Third, the dependabot.yml file now contains an entry for the dotnet tools, with the stipulation that only Trash tools are updated on major or minor version updates.

The rest of the files modified in this PR are the templates themselves. All tools now need to be run as a sub-command of `dotnet`.


kaby76:
Dependabot does not run with the latest release of dependabot-core. The changes to support dotnet-tools.json are not available. I have no idea when this will all be updated.

kaby76:
The solution, which I explain [here](https://github.com/dependabot/dependabot-core/issues/7390), is to create a "fake.csproj". I might be able to create a "hidden" one for Bash shells by naming it ".fake.csproj".

teverett:
@kaby76  there seem to be merge conflicts

kaby76:
@teverett I fixed the conflicts and double-checked a full build for the CSharp target. This is good to go.

teverett:
@kaby76  thanks!

PR #3491, opened 2023-05-31
[build] Fix for #3480 -- Add Dependabot processing of CSharp.

### LINKED ISSUE #3480: [build] Dependabot does not update .csproj files in templates area.
The Dependabot program is [specified to run on the CSharp templates directory](https://github.com/antlr/grammars-v4/blob/4ee2f74c11722d0b7d6c52cc6a86d7f2564fe11c/.github/dependabot.yml#L19), but I've never once seen a PR created for modifying the [line for antlr4.runtime.standard in the .csproj file in the templates](https://github.com/antlr/grammars-v4/blob/4ee2f74c11722d0b7d6c52cc6a86d7f2564fe11c/_scripts/templates/CSharp/Test.csproj#L15). We know that [antlr4.runtime.standard](https://www.nuget.org/packages?q=antlr4.runtime.standard) was recently released. 

It's likely that Dependabot doesn't parse the .csproj file in template form because the open angle bracket `<` in `<PackageReference Include="Antlr4.Runtime.Standard" Version ="4.13.0">` needs to be escaped as `\<` for StringTemplate. I may be able to change the delimiter characters in StringTemplate to something that doesn't conflict with the Dependabot parser of .csproj files. [This constructor call](https://github.com/kaby76/Domemtech.Trash/blob/b30d2c05794a4e6245032f03bd888b176322d750/src/trgen/Command.cs#L1594) will need to change in trgen. I will also likely need a way to say "use these delimiter characters" for trgen, rather than hardwire to use some other delimiter characters for .csproj files.

Once I get all these Dependabot updates implemented, the build will be fabulous! Updates to new versions of Antlr will be a breeze.

kaby76:
This fixes #3480, where .csproj files were not recognized and updated by Dependabot. In this implementation, file Other.csproj is added, along with Other.csproj.meta. The Other.csproj file is in MSBuild format, not StringTemplate format, as there is a .meta file that turns off StringTemplate processing. With ST turned off, trgen just copies the file "as is" to the generated driver directory. The Test.csproj file now contains an "Import" statement, where it includes Other.csproj. Because there are two csproj files, the rest of the changes clear up the ambiguity with dotnet builds.

When this is merged, there will be a Dependabot PR created for upgrading 4.12.0 to 4.13.0 of antlr4.runtime.standard. This, too, should be merged. I purposefully downgraded the version in this PR so that there is proof that the implementation works.

NB. As noted [in the comments](https://github.com/antlr/grammars-v4/issues/3480#issuecomment-1570155253), the plan will be to update the templates to .stg files, and replace this hack.

@teverett  All set.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -94,7 +94,7 @@ jobs:
     - name: Install trgen
       shell: bash
       run: |
-        for i in trgen triconv trwdog trxml trxml2; do dotnet tool install -g $i --version 0.20.19; done
+        for i in trgen triconv trwdog trxml trxml2; do dotnet tool install -g $i --version 0.20.23; done
     - name: Test
       shell: pwsh
       run: |
@@ -191,7 +191,7 @@ jobs:
     - name: Install trgen
       shell: bash
       run: |
-        for i in trgen triconv trwdog trxml trxml2; do dotnet tool install -g $i --version 0.20.19; done
+        for i in trgen triconv trwdog trxml trxml2; do dotnet tool install -g $i --version 0.20.23; done
     - name: Test
       shell: bash
       run: |
--- a/_scripts/templates/CSharp/Other.csproj
+++ b/_scripts/templates/CSharp/Other.csproj
@@ -1,0 +1,8 @@
+<Project>
+
+  <ItemGroup>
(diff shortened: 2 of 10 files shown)


teverett:
@kaby76  thanks!

Issue #3477, opened 2023-05-25
[build] Dependabot supports Typescript, it should be added.


kaby76:
The typescript code should be updated by dependabot.

https://github.com/antlr/grammars-v4/blob/4ee2f74c11722d0b7d6c52cc6a86d7f2564fe11c/.github/dependabot.yml

Looks like it is only for the Javascript templates. https://github.com/antlr/grammars-v4/blob/4ee2f74c11722d0b7d6c52cc6a86d7f2564fe11c/.github/dependabot.yml#L40

PR #3492, opened 2023-05-31
[build] Fix for #3477 -- Add Dependabot processing of TypeScript.

kaby76:
Update dependabot.yml to check the TypeScript templates. Note, Dependabot can have more than one entry for package-ecosystem "npm". I tested it here, and it works. Also, importantly, the package.json file does not contain any open angle brackets (i.e., `<`). Therefore, the npm manager has no problem parsing and updating the package.json file.

@teverett This PR is good to go.


teverett:
@kaby76  thanks!

Issue #3245, opened 2023-03-14
I hope to support Informix sql syntax


hellozrh:
I hope to support Informix sql syntax。

bkiers:
Seems to be present already: https://github.com/antlr/grammars-v4/tree/master/informix

hellozrh:
> 
What I need is  Informix database dialect of the SQL language, see as
https://www.ibm.com/docs/en/informix-servers/14.10?topic=programming-guide-sql-syntax

bkiers:
Ah, right, indeed the `informix` grammar I linked to seems to be something else than SQL... :)

KvanTTT:
The grammar exists and I guess we can close the issue.

hellozrh:
> The grammar exists and I guess we can close the issue.

Yes, I agree with you. The next we just need to improve it bit by bit  .

PR #3495, opened 2023-06-01
informix-sql: add support of release savepoint/rename column.

hellozrh:
informix-sql: add support of release savepoint/rename column.


hellozrh:
https://github.com/antlr/grammars-v4/issues/3245

KvanTTT (approved):

PR #3499, opened 2023-06-02
Fix import and export rules

### LINKED ISSUE #3484: [JavaScript] Support string literals in import and export
Input:
```
import { "a-b" as a } from "/modules/my-module.js";
```

Errors:
```
line 1:9 extraneous input '"a-b"' expecting {'}', 'null', BooleanLiteral, 'break', 'do', 'instanceof', 'typeof', 'case', 'else', 'new', 'var', 'catch', 'finally', 'return', 'void', 'continue', 'for', 'switch', 'while', 'debugger', 'function', 'this', 'with', 'default', 'if', 'throw', 'delete', 'in', 'try', 'as', 'from', 'class', 'enum', 'extends', 'super', 'const', 'export', 'import', 'async', 'await', 'yield', 'implements', StrictLet, NonStrictLet, 'private', 'public', 'interface', 'package', 'protected', 'static', Identifier}
line 1:18 no viable alternative at input 'as a'
line 1:18 no viable alternative at input 'a'
line 1:20 mismatched input '}' expecting {<EOF>, RegularExpressionLiteral, '[', '(', '{', ';', ',', '=', '?', '?.', '.', '++', '--', '+', '-', '~', '!', '*', '/', '%', '**', '??', '>>', '<<', '>>>', '<', '>', '<=', '>=', '==', '!=', '===', '!==', '&', '^', '|', '&&', '||', '*=', '/=', '%=', '+=', '-=', '<<=', '>>=', '>>>=', '&=', '^=', '|=', '**=', 'null', BooleanLiteral, DecimalLiteral, HexIntegerLiteral, OctalIntegerLiteral, OctalIntegerLiteral2, BinaryIntegerLiteral, BigHexIntegerLiteral, BigOctalIntegerLiteral, BigBinaryIntegerLiteral, BigDecimalIntegerLiteral, 'break', 'do', 'instanceof', 'typeof', 'new', 'var', 'return', 'void', 'continue', 'for', 'switch', 'while', 'debugger', 'function', 'this', 'with', 'if', 'throw', 'delete', 'in', 'try', 'as', 'class', 'super', 'const', 'export', 'import', 'async', 'await', 'yield', StrictLet, NonStrictLet, Identifier, StringLiteral, BackTick}
```

From [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#named_import):
> A module may also export a member as a string literal which is not a valid identifier, in which case you must alias it in order to use it in the current module.

Specification: https://tc39.es/ecma262/multipage/ecmascript-language-scripts-and-modules.html#prod-ModuleExportName

Update: there is a similar issue in export statement https://github.com/antlr/grammars-v4/issues/3484#issuecomment-1569719460

bkiers:
Fixes #3484

### DIFF
--- a/javascript/javascript/examples/ImportAndExport.js
+++ b/javascript/javascript/examples/ImportAndExport.js
@@ -1,0 +1,8 @@
+// https://github.com/antlr/grammars-v4/issues/3484
+
+import { "a-b" as a } from "/modules/my-module.js";
+import { "a-b" as yield } from "/modules/my-module.js";
+import { "a-b" as await } from "/modules/my-module.js";
+
+const a = 1;
+export { a as "a-b" };
--- a/javascript/javascript/JavaScriptParser.g4
+++ b/javascript/javascript/JavaScriptParser.g4
@@ -81,12 +81,28 @@ importStatement
     ;
 
 importFromBlock
-    : importDefault? (importNamespace | moduleItems) importFrom eos
+    : importDefault? (importNamespace | importModuleItems) importFrom eos
     | StringLiteral eos
     ;
 
-moduleItems
-    : '{' (aliasName ',')* (aliasName ','?)? '}'
+importModuleItems
+    : '{' (importAliasName ',')* (importAliasName ','?)? '}'
+    ;
+
+importAliasName
+    : moduleExportName (As importedBinding)?
+    ;
+
+moduleExportName
+    : identifierName
+    | StringLiteral
+    ;
+
+// yield and await are permitted as BindingIdentifier in the grammar
+importedBinding
+    : Identifier
+    | Yield
+    | Await
     ;
 
 importDefault
@@ -112,7 +128,15 @@ exportStatement
 
 exportFromBlock
     : importNamespace importFrom eos
-    | moduleItems importFrom? eos
+    | exportModuleItems importFrom? eos
+    ;
+
+exportModuleItems
+    : '{' (exportAliasName ',')* (exportAliasName ','?)? '}'
+    ;
+
+exportAliasName
+    : moduleExportName (As moduleExportName)?
     ;
 
 declaration


bkiers:
I don't know who was responsible for the (much!) faster builds, but the next time I see them, they'll get a beer (or some other beverage) and a bear hug from me! 🍻 🐻

javascript/javascript/examples/ImportAndExport.js
```diff
@@ -0,0 +1,8 @@
+// https://github.com/antlr/grammars-v4/issues/3484
```
ris58h, line 1: We could use MDN examples here:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#syntax
https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export#syntax
Also I would separate them into two files `import.js` and `export.js`.

bkiers: Yes, _we_ coulld! 😉

kaby76:
> I don't know who was responsible for the (much!) faster builds, but the next time I see them, they'll get a beer (or some other beverage) and a bear hug from me! 🍻 🐻

Thanks. I've been fixing the builds. Yes, they are much faster now, i.e., less painful wait times.

I will be adding a few things in the next few weeks.
* Fix Dependabot updates so I don't need to make manual changes and babysit the build when a new Antlr is released.
* Add test coverage, which will help verify that the parse examples actually test the changes to a grammar for a PR. There will be an artifact where one can view a colorized version of the exact grammar(s) text showing "hotspots."
* Useless parentheses notices.
* Useless symbol notices.

I hope to add many more syntactic and semantic checks.

KvanTTT (changes requested):
There is a problem with Python target. Please fix it before merging.

kaby76:
The parser base class wasn't fully tested. The problematic code is here: https://github.com/antlr/grammars-v4/blob/a755a2ab8bb6a1ddd23ec6e0796c8ad7602dba0f/javascript/javascript/Python3/JavaScriptParserBase.py#L102

Instead, the code should use the `in` operator.
```
def string_contains(string, substring):
    return substring in string

...

return ((tokenType == JavaScriptParser.MultiLineComment and (string_contains(text, "\r") or string_contains(text, "\n"))) or
                (tokenType == JavaScriptParser.LineTerminator))
```
or something like that.

bkiers:
Thanks @kaby76, that worked:

```py
return (tokenType == JavaScriptParser.MultiLineComment and (("\r" in text) or ("\n" in text))) or \
    (tokenType == JavaScriptParser.LineTerminator)
```

The build is green again 👍

KvanTTT (approved):

Issue #3399, opened 2023-05-09
Antlr4 Grammer for Fortran 90/95


AkhilAkkapelli:
Can someone add a Antlr4 grammar for Fortran 90/95 or any above version of Fortran to the repository as I am seeking it?

kaby76:
As noted in the [antlr-discussion thread](https://groups.google.com/g/antlr-discussion/c/CSjpelzgRX0), there are two grammars written, but neither have been submitted by the authors to the repo. The best would be to just scrape the grammar (by automated means) from the [Spec](https://www.iso.org/standard/17366.html), but that costs a lot of money.
* https://github.com/OpenFortranProject/open-fortran-parser. There's no grammar labeled with "90", and no clear cut segregation of the different grammars to know how this is supposed to be organized.
* https://github.com/cyrillemidingoyi/PyCropML/blob/master/src/pycropml/transpiler/antlr_py/grammars/Fortran90Parser.g4; https://github.com/cyrillemidingoyi/PyCropML/blob/master/src/pycropml/transpiler/antlr_py/grammars/Fortran90Lexer.g4.

AkhilAkkapelli:
@kaby76  Thank you for taking the time to respond to my email and offer some suggestions for finding the Lexer and Parser grammar for Fortran 90/95 using Antlr4.

I appreciate the links you have provided to the two grammars written. I will take a closer look at these and see if they meet the requirements of my project and see if they are helpful for my project.

Best regards
Akhil

kaby76:
Please reopen. This is a "to do" item. Fortran version 90 and all the many others need to be added.

PR #3501, opened 2023-06-04
[fortran90] Added fortran/fortran90 grammar.

kaby76:
* Fix for #3399 (added the fortran/fortran90/ grammar).
* Moved Fortran77 grammar to the subdirectory fortran/fortran77.
* All targets except TypeScript work, which is the only target that I don't include in the desc.xml.
* The TypeScript target of the grammar does not work because Antlr 4.13.0 does not generate a correct import statement for including the base class! Otherwise, if you hand-edit the generated Fortran90Lexer.ts code with the correct syntax, the parser works.
* The pom.xml for both fortran77 and fortran90 work.
* The tests are divided into "slow" and "fast" targets. All the interpreted language targets (PHP, JavaScript, Python3) are "slow". (This is a good reason not to use these terrible languages.)


cyrillemidingoyi:
Great! Thank you @kaby76

teverett:
@kaby76  thanks!

Issue #3509, opened 2023-06-05
[fortran90] Fix TypeScript target code.


kaby76:
As noted [here](https://github.com/antlr/grammars-v4/pull/3501#issue-1740454713), I couldn't get the TypeScript code to compile. Since then, I have a fix that changes the base class declaration. It now works great.

PR #3510, opened 2023-06-05
[fortran90] Fix for #3509--TypeScript target.

kaby76:
This PR is a fix for #3509.

This was an easy fix, thanks to the comment from @ericvergnaud [here](https://github.com/antlr/antlr4/issues/4302#issuecomment-1576442166). It's just a change to the declaration for the base class. With it, we now have a grammar that is split and superClass'ed for the lexer working for 8 targets, the first of its kind.


kaby76:
The antlr4 tool is not robust. https://github.com/antlr/antlr4-tools/issues/13

teverett:
@kaby76 thanks!

PR #3511, opened 2023-06-05
[build] Fix for #3365--Add grammar coverage tool.

### LINKED ISSUE #3365: [build] Add grammar coverage tool.
I have a new that I've been working on the last few days called [trcover](https://github.com/kaby76/Domemtech.Trash/tree/main/src/trcover). It analyzes grammars and input files, the determines the "hotspots" in the grammar, finally outputting a HTML-marked up version of the grammar for quick visualization of just what we are testing. I think I can output some other stats like percent of the grammar tested, which might be a good thing to add to the build and to help people write tests for there grammars.

I just did a test of the [java8](https://github.com/antlr/grammars-v4/tree/master/java/java8) grammar, and the visualization shows huge amounts of the grammar not be tested at all! The following zip contains an html file, which can be opened in a browser. Right now, I mark up symbols that are parsed at least one time with the same color. (There may be of course errors in my code!!)

[cover.zip](https://github.com/antlr/grammars-v4/files/11392268/cover.zip)

kaby76:
This PR fixes #3365.

In this PR, the tool [trcover](https://github.com/kaby76/Domemtech.Trash/tree/main/src/trcover) is added, which will produce a nice report of what parser rules are tested. By look at the report, one can find rules that are not tested, then craft tests to complete testing of the entire grammar.

This tool reads the grammar and parses the usual test input. For a grammar, each rule is converted into a minimized NFA. After a parse tree is produced from parsing, the analysis code examines what rules were used during the parse. The tool's output is an annotated version of the grammar file that describes what rules and symbols were parsed. The output is saved as an artifact of the build. Simply navigate to the Action that executed for your PR, and look for "Artifacts" at the top of the page.

To view coverage analysis, click on the "coverage" link to download "coverage.zip". ([Here is an example](https://github.com/antlr/grammars-v4/actions/runs/5317014546#artifacts).) Open the zip file and navigate to the "cover.html" file for the grammars you modified. Open the "cover.html" file in a browser to see the annotated grammar. Symbols with a red background indicate symbols that were tested at least once. If the symbol was tested more often, a darker shade of red is used. At the end of the cover.html file is a percentage of the parser rules hit by all tests.


teverett:
@kaby76  thanks!

PR #3513, opened 2023-06-06
informix-sql:add support of rename index.

hellozrh:
informix-sql:add support of rename index.

### DIFF
--- a/sql/informix-sql/InformixSQLParser.g4
+++ b/sql/informix-sql/InformixSQLParser.g4
@@ -54,9 +54,9 @@ unitStatement
     | databaseStmt
     | releaseSavepoint
     | renameColumn
+    | renameConstraint
     | renameDatabase
     | renameIndex
-    | renameConstraint
     ) SCOL
     ;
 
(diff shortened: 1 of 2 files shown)


kaby76:
Go build failed due to "antlr4" not being robust. https://github.com/antlr/antlr4-tools/issues/13. This is caused because the version of Antlr for the Go target is always the tip, so the "antlr4" script tries to find out what is the latest version of Antlr4. I will try redoing the Go target templates so that it works against 4.13.0 explicitly, which will avoid all this unrobust code in the Python3 script.

Otherwise, the build passed on your changes.

hellozrh:
> 

I see, and this problem also appeared last time, it has been solved after I resubmited something.

hellozrh:
Oh, it failed again.  That's it , wait for you to specify the version.

KvanTTT (approved):

kaby76:
@teverett I think this is good to merge.

KvanTTT (approved):

kaby76:
@KvanTTT Thanks!

PR #3515, opened 2023-06-06
[build] Fix for #3514 -- "antlr4" tool is not robust.

### LINKED ISSUE #3514: [build] Go target (and likely Cpp target) not robust because "antlr4" tool doesn't work reliably.
As I pointed out [here](https://github.com/antlr/antlr4-tools/issues/13), the "antlr4" tool is not reliable if one tries to just use the "latest" version of Antlr, i.e., don't use the `-v` option to say what version you want. It looks like [the query](https://github.com/antlr/antlr4-tools/blob/787e7a500e56d17566086b5d5c04eeea84a387e8/antlr4_tool_runner.py#L26) times out and then the code crashes. This is catastrophic for testing. And, programming around this by detecting this particular crash, as opposed to a legitimate problem with the grammar you are trying to run the tool on, would not be very nice looking scripting code. Already these Bash scripts look like a mess.

We'll need to set the version explicitly for the Go (and likely Cpp) target.

kaby76:
This PR fixes #3514. It should address build errors for Go for [PR#3517](https://github.com/antlr/grammars-v4/pull/3517) [build errors](https://github.com/antlr/grammars-v4/actions/runs/5187609698/jobs/9350179996#step:20:54) and [PR#3613](https://github.com/antlr/grammars-v4/pull/3513) [build errors](https://github.com/antlr/grammars-v4/actions/runs/5186830904/jobs/9348436952#step:21:65) and likely other places.

This change brings back the explicit version tags for Antlr to Cpp and Go targets, avoiding the problem with the "antlr4" Python script. I didn't want to do this because now someone will need to manually change those version strings when Antlr is again released. Unfortunately, there is no Dependabot for CMake files. There may be a way to update the Go target with Dependabot, but the [instructions for using the Go target](https://github.com/antlr/antlr4/blob/master/doc/go-target.md#2-get-the-go-antlr-runtime) explicitly say `go get github.com/antlr4-go/antlr`. Dependabot can't modify `go get github.com/antlr4-go/antlr/v4@v4.13.0` in a Bash file.

### DIFF
--- a/_scripts/templates/Cpp/cmake/FindANTLR.cmake
+++ b/_scripts/templates/Cpp/cmake/FindANTLR.cmake
@@ -1,7 +1,7 @@
 find_package(Java QUIET COMPONENTS Runtime)
 
   execute_process(
-      COMMAND antlr4<if(os_win)>.exe<else><endif>
+      COMMAND antlr4<if(os_win)>.exe<else><endif> -v ${ANTLR4_TAG}
       OUTPUT_VARIABLE ANTLR_COMMAND_OUTPUT
       ERROR_VARIABLE ANTLR_COMMAND_ERROR
       RESULT_VARIABLE ANTLR_COMMAND_RESULT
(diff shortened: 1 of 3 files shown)


teverett:
@kaby76  thanks!

PR #3519, opened 2023-06-07
Reset JavaScriptLexerBase (fixes #3518).

### LINKED ISSUE #3518: JavaScriptLexerBase doesn't reset properly.
`JavaScriptLexerBase` has state but doesn't override `reset` method to reset it's state.

ris58h:
I'm not an expert in languages other than Java so please check other versions.


teverett:
@ris58h  thanks!

Issue #3521, opened 2023-06-07
[javascript] Go target doesn't work for 4.13.0.


kaby76:
The [javascript](https://github.com/antlr/grammars-v4/tree/master/javascript/javascript) grammar has Go-target code but it doesn't compile because all the import paths have changed for 4.13.0. So, the desc.xml (as well as all grammars that have Go-target code) needs to be qualified to only `4.13.0` or greater.

PR #3523, opened 2023-06-08
[javascript] Fix for #3521--Add Go target for the javascript grammar.

kaby76:
This PR fixes https://github.com/antlr/grammars-v4/issues/3521.

The changes in this PR are for the Antlr 4.13.0 release, the [javascript grammar](https://github.com/antlr/grammars-v4/tree/8d58426f64b98637d55f6e49a51544bac5106b5a/javascript/javascript), and the Go target. Consequently, the desc.xml is changed to add testing of the Go target and to specify that this grammar must use 4.13.0 of Antlr or better. This is because the `import` path for the Antlr runtime has changed in Antlr 4.13.0. This PR also contains a `transformGrammar.py` file for Go because the `this.` syntax in actions must be rewritten to Go syntax. The references for the lexer object have to be modified differently between semantic predicates and side-effect actions. The Go/README.md file was removed because it only clouded how to do a build (just use trgen).

With this change, the grammar now works for Golang, and the tests take just over 1 minute.


teverett:
@kaby76  thanks!

Issue #3522, opened 2023-06-08
[build] Dependabot supports Go, it should be added.


kaby76:
This took me a while to figure out, but it's possible. And, extracting the Antlr tool version string from the go.mod file should be easy.

There are several tricks for this to happen.
* The dependabot.yml file needs to have an entry for Go.
```
  - package-ecosystem: "gomod"
    directory: "/templates/Go"
    schedule:
      interval: "daily"
    allow:
      - dependency-type: "all"
    labels:
      - "infrastructure"
      - "dependencies"
```
* The templates, and trgen, have to, again, be modified. The dependabot code requires any .go files to be parsible. That means that they cannot be templated. But, they have to in order to get the import paths correct in the source code. The trick is to make sure it does not have the .go suffix!! So, we need to replace the ".meta" hack. All templated files must have the .stg suffix and be StringTemplate group template files. (`templates/Go/test.go` => `templates/Go/test.go.stg`). `templates/Go/go.mod` is not templated, so it remains as is for Dependabot to modify.
* The file `templates/Go/build.sh` should have a grep/awk of go.mod to extract the string.

kaby76:
* Workaround of `<embeddedresource>` complete. https://github.com/kaby76/Domemtech.Trash/commit/5dfe2052c706e1ec8ca3e09a02ac745d193af5ed
* Add code to handle stg, st, or plain files. https://github.com/kaby76/Domemtech.Trash/commit/ea39170a3ebfbfe20345d4b90963b0071cc5bca4

PR #3537, opened 2023-06-15
[build] Fix for #3522--Dependabot supports Go, it should be added.

kaby76:
This PR fixes https://github.com/antlr/grammars-v4/issues/3522

In this change, the Dependabot "to do" was updated for monitoring and issuing PRs for the Go target for Antlr. From here on, when Antlr is released, the only thing that needs to be done is to merge the pull request. The build scripts examine the go.mod file for the version of Antlr to use for the "antlr4" tool.

But, to get Dependabot to work on the go.mod file, the templates for trgen have been updated to use StringTemplate's TemplateGroup, Template, or unprocessed file processing, depending on the name of the template file. It was necessary to change trgen specifically for this issue because Dependabot requires go.mod to be a plain file (not a template file), and there cannot be a Test.go file (renamed to "Test.go.st") because it is a template file and Dependabot cannot parse it. This repo now uses version 0.20.27 of the Trash toolkit.

### What has changed and why?

* .config/dotnet-tools.json https://github.com/antlr/grammars-v4/pull/3537/files#diff-7afd3bcf0d0c06d6f87c451ef06b321beef770ece4299b3230bb280470ada2f6. This file updates the version of trgen, which is essential to process all file types, from .stg, to .st, to plain files.
* .github/dependabot.yml https://github.com/antlr/grammars-v4/pull/3537/files#diff-dd4fbda47e51f1e35defb9275a9cd9c212ecde0b870cba89ddaaae65c5f3cd28. This file now has lines to indicate processing of the Golang templates.
* The ".meta hack", which I introduced a week or so ago, has been removed. Templates are supported by the name of the file (".stg" => TemplateGroup(), ".st" => Template(), neither ".stg" nor ".st" => plain file, not templated).
* I renamed most of templated files to `st.<file-name>` since most files are StringTemplate "Template()" files. "st" in the file name can be at the beginning of the file name or at the end. However, the "st" in the name indicates to trgen that the file is to be processed as a Template file. If there is no "st" in the file name, it's a plain file, which is just copied to the "Generated" directory.
* For the Go target template (in "_scripts/templates/Go"), "Test.go" must be renamed to "Test.go.st".
* There are no ".stg" files, but trgen can process StringTemplateGroup files.
* "_scripts/templates/files" is no longer needed because trgen doesn't use it anymore.
* I include some templates for testing PHP, but they don't work because we don't have an Antlr 4.13.0 runtime for PHP. https://github.com/antlr/antlr-php-runtime/issues/40
* The changes have been tested using _scripts/test.sh and _scripts/test.ps1 on abb/ and all grammars under sql/.
* There was a call to the "antlr4" Python script but it had an error in not using the `-v` option. https://github.com/antlr/grammars-v4/blob/1966b6081dc49a24e6641259de0eaa483b8565da/_scripts/templates/Go/build.ps1#L19 That caused a timeout problem in a build.  https://github.com/antlr/grammars-v4/actions/runs/5271307409/jobs/9531995856#step:22:63


teverett:
@kaby76  thanks!

PR #3547, opened 2023-06-21
 [javascript] Field declaration should end with ';' (fixes #3546)

### LINKED ISSUE #3546: [javascript] Field declaration should end with ';'
Code:
```
class Foo {
    bar baz;
}
```

The code isn't correct according to documentation https://tc39.es/ecma262/multipage/ecmascript-language-functions-and-classes.html#prod-ClassElement and parsing should fail.

ris58h:


teverett:
@ris58h  there are some build errors

ris58h:
@teverett I know. I've just left it in undone state.

PR #3557, opened 2023-06-26
[PostgreSQL] Optimize the parseTree's depth of PostgreSQL's a_expr rule

jiefei30:


jiefei30:
I'm not quite sure why the error occurred in geometry.sql
This is just as a reference to solve the problem of stack overflow caused by too deep parse tree, 
so I will close this merge

PR #3587, opened 2023-07-06
Add TypeScript target to Dart2 grammar

rainerschuster:


kaby76:
I think the problem here is that the parser is pretty slow, so I didn't add it to testing. The testing of TypeScript should be tested against a small number of inputs. Then, we can focus on cleaning up the grammar.

Copy a few of the input files to a "small/" directory, then change the desc.xml to test against that for TypeScript.
```
   <targets>CSharp;Dart;Java;TypeScript</targets>
   <test>
      <targets>CSharp;Dart;Java</targets>
      <inputs>examples</inputs>
   </test>
   <!-- slow parsers work only on a small test suite. -->
   <test>
      <targets>TypeScript</targets>
      <inputs>small</inputs>
   </test>
</desc>
```

kaby76:
Please close this PR https://github.com/antlr/grammars-v4/pull/3587. The code has been merge and the rest of the ports added with PR https://github.com/antlr/grammars-v4/pull/3740.

rainerschuster:
Thanks for merging!

PR #3590, opened 2023-07-06
 [CSS] at-keyword token (fixes #3589).

### LINKED ISSUE #3589: [CSS] at-keyword token
According to the [specification](https://www.w3.org/TR/css-syntax-3/#typedef-at-keyword-token) CSS has at-keyword token. In the current css3 grammar it's a rule, not token. Introducing such token will make the grammar more simple and closer to the spec.

ris58h:


ris58h:
It fails on Windows because `\r\n` line ending is expected. What should I do? Add `text eol=crlf` to `*.css` in `.gitattributes`?

teverett:
@ris58h thanks!

Issue #3588, opened 2023-07-06
`ANTLR Tool version 4.13.0 used for code generation does not match the current runtime version 4.11.1` on `mvn clean install`


ris58h:
1. Go to `css3` grammar.
2. `mvn clean install`

```
[INFO] --- antlr4test:1.22:test (default) @ CSS3 ---
[INFO] Evaluating Scenario: Default Scenario
[INFO] Parsing :/Users/ris/projects/grammars-v4/css3/examples/mdn-at-font-face.css
ANTLR Tool version 4.13.0 used for code generation does not match the current runtime version 4.11.1
ANTLR Runtime version 4.13.0 used for parser compilation does not match the current runtime version 4.11.1
ANTLR Tool version 4.13.0 used for code generation does not match the current runtime version 4.11.1
ANTLR Runtime version 4.13.0 used for parser compilation does not match the current runtime version 4.11.1
[INFO] Parsing :/Users/ris/projects/grammars-v4/css3/examples/bootstrap.min.css
```

kaby76:
The root pom.xml contains [one version for Antlr4 (4.13.0)](https://github.com/antlr/grammars-v4/blob/c9e5c8caca0b562805711b22d04a8ce39d012461/pom.xml#L14), but the plugin [sets a different version](https://github.com/antlr/antlr4test-maven-plugin/blob/fbba99e1ecebe03681e1269a0a1fba95f75550c3/pom.xml#L41). The version selected in grammars-v4 is not overriding the version in the plugin as intended. The easiest would be to update the plugin, upgrade here. But that would not fix an entry in the root pom.xml that does not work. Why even have it? [The latest version of the plugin is 1.22](https://search.maven.org/artifact/com.khubla.antlr/antlr4test-maven-plugin/1.22/maven-plugin).

kaby76:
I added an issue requesting a new version of the plugin be published. https://github.com/antlr/antlr4test-maven-plugin/issues/89

The problem is that the version in the grammars-v4/pom.xml is the version of the tool, but the version of the Antlr Java runtime is 4.11.1. The tester should not be using two different version numbers. This, again, is why you should be using the Trash toolkit. It uses one, and only one, version number for the tool and runtime, not only for Java, but for all targets.

[...]

PR #3596, opened 2023-07-10
Fix for #3588 -- Update root pom.xml for testing Antlr 4.11.1

kaby76:
This is a fix for https://github.com/antlr/grammars-v4/issues/3588.

The version of the Antlr tool is specified directly in the root pom.xml. But the version of the runtime is specified indirectly with the antlr4test-maven-plugin.version. The plugin links with a version of runtime specified in the plugin [here](https://github.com/antlr/antlr4test-maven-plugin/blob/fbba99e1ecebe03681e1269a0a1fba95f75550c3/pom.xml#L41).

Consequently, when Dependabot updates the pom.xml, e.g., https://github.com/antlr/grammars-v4/pull/3114 and https://github.com/antlr/grammars-v4/pull/3462, it fails to take into consideration updating the version of the plugin, which must be specially released and published. The state for testing using the plugin is then inconsistent.

The newer test framework via [trgen](https://github.com/kaby76/Domemtech.Trash/tree/main/src/trgen) does not have this problem because [it works from the version in the templates for both tool and runtime](https://github.com/kaby76/Domemtech.Trash/blob/e18338d088afb263aaae63f4de2179766a7f1a29/src/trgen/templates/Java/st.build.sh#L9). The [antlr4 tool](https://github.com/antlr/antlr4-tools) is used, which not only downloads the Antlr .jar, but the java runtime. Even when running the compiled java program, [it refers back to the one and only version string](https://github.com/antlr/grammars-v4/blob/8c81d4c7b2dfde7a802d206ac0fae84c2dd9e841/_scripts/templates/Java/st.test.sh#L35).

I am removing a .error file that I created for abb. The test plugin cannot deal with empty .errors files, so it has to be removed.

There's a problem with the XPath grammars. That is being fixed separately. https://github.com/antlr/grammars-v4/pull/3595

The PR updates the support for testing with the plugin, until a time when people stop using it. Grammars that require the new features of Antlr 4.11.2 or more recent must be removed from the root pom.xml.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,8 +11,10 @@
 		<maven.compiler.target>11</maven.compiler.target>
 		<maven.compiler.source>11</maven.compiler.source>
 
-		<!-- antlr.version is the version of the tool. But the
-		runtime is tied to the version of antlr4test-maven-plugin.version -->
+		<!-- The Antlr tool version is specified with
+		antlr.version. The Antlr runtime is tied to the
+		version of antlr4test-maven-plugin.version. When Antlr
+		is release, the plugin has to be updated. -->
 		<antlr4test-maven-plugin.version>1.22</antlr4test-maven-plugin.version>
 		<antlr.version>4.11.1</antlr.version>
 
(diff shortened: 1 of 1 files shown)


teverett:
@kaby76  thanks!

Issue #3601, opened 2023-07-11
[c] C grammar should be consistent with Cpp grammar in ignoring pre-processor directives


kaby76:
The c grammar has some code to ignore pre-processor directives, but not all. That's just wrong. Sure, ignore all preprocessor directives or none. But, don't parse some directives and not try to parse others. https://github.com/antlr/grammars-v4/blob/master/c/C.g4#L850-L859

PR #3602, opened 2023-07-11
Fix for #3601

kaby76:
Replaced rules for preprocessing directive parsing with those from the cpp/ grammar. Added a test from the GCC testsuite, with some of the preprocessor macros replaced to pass parsing.


kaby76:
Replaced by https://github.com/antlr/grammars-v4/pull/3608

PR #3606, opened 2023-07-12
[MySQL] Avoid confusing label and IPv6 address

vjuranek:
Some label combined with some keywords (when written without a space in between, which is valid for MySQL) can look like beginning of an IPv6 address, e.g.

    aaa:BEGIN

is currectly prased as IPv6 address `aaa:be` and leads to a parsing failure.

Make parsing of IPv6 address more robust and require at least two `:` to make sure there is no confusion with labels.


sql/mysql/Positive-Technologies/MySqlLexer.g4
```diff
@@ -1320,7 +1320,7 @@ STRING_USER_NAME:                    (
                                      );
 IP_ADDRESS:                          (
                                        [0-9]+ '.' [0-9.]+
-                                       | [0-9A-F:]+ ':' [0-9A-F:]+
+                                       | ([0-9A-F]+)? ':' ([0-9A-F]+)? ':' [0-9A-F:]+
```
KvanTTT, line 1323: Please simplify:

```suggestion
                                       | [0-9A-F]* ':' [0-9A-F]* ':' [0-9A-F:]+
```

KvanTTT (changes requested):

sql/mysql/Positive-Technologies/MySqlLexer.g4
vjuranek: ops, thanks for checking and correction, applied!

KvanTTT (approved):

Issue #3603, opened 2023-07-12
[cpp] purespecifiers rule is wrong, not parsed correctly according to Spec


kaby76:
Consider this code:
```
class Abstract
{
        public:
                virtual void pure_virtual() = 0;
                virtual void pure_virtual2() = 0x0;
                virtual void pure_virtual3() = 04;
};
```
The three assignments are for pure specifiers, the later two incorrect. This code is instead parsed as `declarator braceOrEqualInitializer`. This is incorrect because braceOrEqualInitializer is only for _data members_. See p219 of the Spec. "A brace-or-equal-initializer shall appear only in the declaration of a data member. (For static data members, see 9.4.2; for non-static data members, see 12.6.2)."

This means that if the declarator contains parametersAndQualifiers, it is not data, but a method.

Essentially, [the check here is insufficient](https://github.com/antlr/grammars-v4/blob/636a6d9cb79aecd4fb5f894fa88a43a28dd72a70/cpp/CPP14Parser.g4#L646). While it may be fine to check for an illegal pure specifier, it doesn't fix the main problem in that it isn't being parsed at all with `pureSpecifier`.

PR: https://github.com/antlr/grammars-v4/pull/3607

kaby76:
One problem here is that the `0` in `= 0` is tokenized as an `IntegerLiteral`, not `OctalLiteral`, which is [what the rule requires](https://github.com/antlr/grammars-v4/blob/6ce408a25a51f7edb735f481335be429e5e75c3e/cpp/CPP14Parser.g4#L646). So, the `pureSpecifier` should use `IntegerLiteral`. But, then, that would conflict with parsing for an integer field with an initializer: `int x = 0;`. The key is to provide a semantic check that disallows `pureSpecifier` for non-method declarators.

PR #3607, opened 2023-07-12
[cpp] Fix for #3603 -- pure specifiers

kaby76:
This is a fix for #3603.

This PR contains changes for the grammar for pure specifiers. The PR also contains a change to make the grammar target agnostic, and includes, now, a port for the CSharp target.

The problem occurs with this test:
```
class Abstract 
{
	public:
		int x = 0;
		virtual void pure_virtual() = 0;
		virtual void pure_virtual2() = 0x0;
		virtual void pure_virtual3() = 04;
};
```
Here, we see a field with an iniitializer, and three declarations for methods with pure specifiers, two of which are wrong (more on this later).

For each, the parser can choose a derivation involving pureSpecifier or braceOrEqualInitializer. Currently, it is always using braceOrEqualInitializer. The field declaration should not be parsed using pureSpecifier, but the methods should be parsed using pureSpecifier.

Also, because there's an ambiguity, the parse performs slowly. _Just because Antlr allows one to specify an ambiguous grammar, it does not mean that one should!_

The the problem is deciding which of the two derivations to choose. The solution implements disambiguation using a semantic predicate. Note, it is crucial that the predicate occur _at the beginning of the alt_ rather than at the end or somewhere in the middle. Antlr will raise an error in the latter. The predicate simply checks the parse tree and makes sure that the parse involves a function declaration.

I added a test that involves three pure specifiers--two of which are parsed, but not flagged as errors--and a data field.

***NB, semantic checking should be done after a parse to validate the value of the assignment in a pure specifier. It should not be in the parser. But, if people insist, I can add the check in.***


teverett:
@kaby76  thanks

PR #3608, opened 2023-07-12
[c] Fix for #3601 -- fix parsing of all preprocessor directives

kaby76:
This is a fix for #3601.

This PR changes the lexer rules for preprocessing directives to ignore all '#' directives. The rules that were in the grammar were incomplete. The idea of parsing these lines is to get past most errors where people had not performed preprocessing--which is a lot of people. It also is required since the preprocessor added `#line` directives for the compiler to print out line/column error information. Again, this grammar is designed to be used on code after preprocessing.

I also added a test from the GCC testsuit, which contains preprocessor directives, and a couple of macros expanded as to parse.


teverett:
@kaby76 thanks

PR #3614, opened 2023-07-14
[gdscript] Fix #3613 -- add reset() method

### LINKED ISSUE #3613: [gdscript] missing base class lexer Reset() method
For my upcoming change in https://github.com/antlr/grammars-v4/pull/3598 I pointed out that there may be several grammars that are missing the Reset() method. The [gdscript](https://github.com/antlr/grammars-v4/tree/master/gdscript) grammar is one of these.

Why is this important? The driver for the grammar first [reads tokens from the lexer](https://github.com/antlr/grammars-v4/blob/master/_scripts/templates/CSharp/st.Test.cs#L228-L242). It does so independently from the parser. Then, after calling [Reset()](https://github.com/antlr/grammars-v4/blob/8dda450cb19a7c4ad0b0ef9c68196c5a6b688f5d/_scripts/templates/CSharp/st.Test.cs#L241), [it performs a parse](https://github.com/antlr/grammars-v4/blob/8dda450cb19a7c4ad0b0ef9c68196c5a6b688f5d/_scripts/templates/CSharp/st.Test.cs#L262). If Reset() does not work, the parse will fail.

kaby76:
This PR fixes #3613.

The change added the reset() method in Java, Reset() method in CSharp. This PR is required for https://github.com/antlr/grammars-v4/pull/3598

### DIFF
--- a/gdscript/CSharp/GDScriptLexerBase.cs
+++ b/gdscript/CSharp/GDScriptLexerBase.cs
@@ -165,4 +165,13 @@ public abstract class GDScriptLexerBase : Lexer {
 			}
 		}
 	}
+
+	public override void Reset()
+	{
+		Tokens = new LinkedList<IToken>();
+		Indents = new Stack<int>();
+		Opened = 0;
+		LastToken = null;
+		base.Reset();
+	}
 }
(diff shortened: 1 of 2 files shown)


kaby76:
Wait for build fix. https://github.com/antlr/grammars-v4/actions/runs/5554306328

teverett:
@kaby76  thanks

Issue #3618, opened 2023-07-17
[cpp] Using the cpp grammar with the python target


pasto9:
I try to use the cpp grammar with the python target. With the latest fix #3603 antlr does not generate proper python code.
So I tried to fix it and noticed that the new pureSpecifier implementation has the problem that python does not use the this keyword. Therefore [this line](https://github.com/antlr/grammars-v4/blob/d6353f2011d11d34007e46cd3e0d91d695957ae6/cpp/CPP14Parser.g4#L633C61-L633C61) in the grammer produces incorrect code in python.

I think this could be fixed if instead of `this.IsPureSpecifierAllowed()` `$parser.IsPureSpecifierAllowed()` would be used. This way it should generate correct code for python and c# (only tested this one). 

I'm new to python and antlr so I hope this is not a stupid suggestion.

kaby76:
Making a PR for the rest of the targets. https://github.com/antlr/grammars-v4/pull/3619

kaby76:
The Python3 port now works. https://github.com/antlr/grammars-v4/pull/3619

pasto9:
Thank you very much for your work. It helps a lot.

PR #3619, opened 2023-07-17
[cpp] Fix for #3618 -- cpp grammar, add Python3 and others targets

kaby76:
This is a fix for #3618: to fill out several more ports of the [cpp grammar](https://github.com/antlr/grammars-v4/tree/master/cpp) to different targets, including Python3.

* Cleaned up CSharp port.
* Added Cpp port.
* Added Go port. Some of the parser grammar symbols had to be renamed due to conflicts in the generated code.
* Added Python3 port.
* Added TypeScript port.
* Cleaned up the grammar: removed useless parentheses.
* Added a readme.md. For "links" section, Wikipedia seems to be the only reliable general-purpose reference of the C++ language, but I included PLDB even though that website is having problems.


teverett:
@kaby76  thanks!

Issue #3597, opened 2023-07-10
[build] coverage times out


kaby76:
Coverage is computed for all grammars using trcover. This tool can timeout for latge test suites. The tool should work on smaller input sets in these situations, which should be specified in the desc.xml.

PR #3624, opened 2023-07-20
[build] Fix for #3597 -- Timeouts of coverage, remove for now

kaby76:
This is a fix for #3597.

"Coverage" is a computation that outputs a .html file that is an annotated text file of a grammar. It shows how the grammar is tested, and helps identify what rules in a grammar are not being tested.

Unfortunately, coverage computation takes over 300s, which is monitored and killed, then reported as an error. This PR removes the Github Action for now.

Also, coverage should report whether a change to a grammar is being tested, not necessarily show how the entire grammar is tested.

The plan is to update the tool trcover so that it is faster and add in code to test only the changes.


teverett:
Thanks!

PR #3634, opened 2023-07-27
Go 1.20 update

codeFather2:
This request updates the Go grammar, adding support for the new 1.20.* syntax.
**Actual Go spec** - [Go/ref/spec](https://go.dev/ref/spec)

## What's New

* Generic types, interfaces and function
*  Interface constraints
* Underlying type operation


kaby76:
This is great. Thanks for updating this grammar!

I'm not sure what should be done with these unused symbols/rules. Should the grammar be adjusted to be closer to the Spec or remove the rules?
```
$ bash /c/Users/Kenne/Documents/GitHub/g4-checks/unused.sh
No arguments were provided.
Finding unused parser symbols in grammars...
./desc.xml
GoParser.g4:L132: emptyStmt: EOS | SEMI;
                  ^
GoParser.g4:L373: receiverType: type_;
                  ^
```

codeFather2:
@kaby76 I think it's best to remove unused rules. The `receiverType` rule is just a dummy layer for the next rule, and the empty statement doesn't really do anything because `EOS` and `SEMI` are covered by `statementList`.

teverett:
@codeFather2 thanks!

PR #3643, opened 2023-08-02
[Snowflake]Fix TAG are schema-level object

MasterKuat:
As describe in subject, tags are schema-level object (aka name could be a db.schema.name)

CREATE TAG documentation is unclear about name but https://docs.snowflake.com/en/user-guide/object-tagging#what-is-a-tag
clearly say it.


teverett:
@MasterKuat  thanks!

Issue #3646, opened 2023-08-05
[PL/SQL] warning(154): PlSqlParser.g4:6241:0: rule json_function contains an optional block with at least one alternative that can match an empty string


doberkofler:
When compiling grammar/PlSqlParser.g4 I get a warning:

$ antlr4 -Dlanguage=Python3 grammar/PlSqlParser.g4
```bash
warning(154): PlSqlParser.g4:6241:0: rule json_function contains an optional block with at least one alternative that can match an empty string
```

kaby76:
It's caused by the expression [json_query_returning_clause?](https://github.com/antlr/grammars-v4/blob/eba5c4487067292a82cf6b753f33a6b546c7e1df/sql/plsql/PlSqlParser.g4#L6246C64-L6246C92)--the `?` operator cannot be applied to a symbol that is nullable. Unfortunately, it's in the Spec. https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/JSON_QUERY.html#GUID-6D396EC4-D2AA-43D2-8F5D-08D646A4A2D9.

The easiest fix is to change the rule to not use the `?`-operator.

```
$ diff PlSqlParser.g4 ..
6246c6246
<     | JSON_QUERY '(' expression (FORMAT JSON)? ',' CHAR_STRING json_query_returning_clause json_query_wrapper_clause? json_query_on_error_clause? json_query_on_empty_clause? ')'
---
>     | JSON_QUERY '(' expression (FORMAT JSON)? ',' CHAR_STRING json_query_returning_clause? json_query_wrapper_clause? json_query_on_error_clause? json_query_on_empty_clause? ')'
```

The other is to rewrite json_query_on_empty_clause to not derive the empty string.

kaby76:
Can this be closed? I fixed this several weeks ago. T.Y.

PR #3648, opened 2023-08-07
[plsql] Fix for #3646 -- removed warning "rule json_function contains an optional block with at least one alternative that can match an empty string"

kaby76:
This is a fix for #3646.

In this PR, the json_function rule contains the expression [json_query_returning_clause?](https://github.com/antlr/grammars-v4/blob/eba5c4487067292a82cf6b753f33a6b546c7e1df/sql/plsql/PlSqlParser.g4#L6246C64-L6246C92), which causes the warning `"rule json_function contains an optional block with at least one alternative that can match an empty string"`. It turns out that [json_query_returning_clause](https://github.com/antlr/grammars-v4/blob/eba5c4487067292a82cf6b753f33a6b546c7e1df/sql/plsql/PlSqlParser.g4#L6298) is nullable. The Spec contains the same error. https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/JSON_QUERY.html#GUID-6D396EC4-D2AA-43D2-8F5D-08D646A4A2D9

Until the Spec is fixed, the ?-operator on json_query_returning_clause should be removed. The alternative is to rewrite json_query_returning_clause so that is not nullable, a more complicated task.

I added a test to make sure json_query_returning_clause is used. (It was not up to this point. See attached [cover.html](https://github.com/antlr/grammars-v4/files/12293675/cover.txt)).


teverett:
@kaby76  thanks!

PR #3657, opened 2023-08-09
Fix issue at llvm-ir grammar.

NiklasHeneka:
Recently I compiled C++ code to llvmir and experienced issues parsing it.

It seems the grammar is missing the align statement at the returnAttribute rule, which should be a valid returnAttribute according to the [language reference manual](https://releases.llvm.org/15.0.0/docs/LangRef.html#paramattrs).

This also fixed my issue parsing the llvmir code.

### DIFF
--- a/llvm-ir/LLVMIR.g4
+++ b/llvm-ir/LLVMIR.g4
@@ -114,7 +114,10 @@ gc: 'gc' StringLit;
 prefix: 'prefix' typeConst;
 prologue: 'prologue' typeConst;
 personality: 'personality' typeConst;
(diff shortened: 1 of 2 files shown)


KvanTTT (changes requested):

Please add/extend an example that covers suggested changes in [llvm-ir/examples](https://github.com/antlr/grammars-v4/tree/master/llvm-ir/examples).

NiklasHeneka:
I added an example covering the change.

KvanTTT (approved):

Issue #3660, opened 2023-08-11
[plsql] Link in readme should be for pl/sql


kaby76:
There is a [link](https://github.com/antlr/grammars-v4/blob/768b12e1db509aa700a316e3eed1e23e8c4bdb06/sql/plsql/README.md?plain=1#L1) in the [readme.md](https://github.com/antlr/grammars-v4/blob/master/sql/plsql/README.md) file that is wrong. It's a link to ["SQL Language Reference"](https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/Introduction-to-Oracle-SQL.html#GUID-049B7AE8-11E1-4110-B3E4-D117907D77AC), which is not specifically "Database PL/SQL Language Reference". The link should be https://docs.oracle.com/en/database/oracle/oracle-database/23/lnpls/database-pl-sql-language-reference.pdf.

In addition, the readme.md style should follow a more standardized format that I've been adopting elsewhere. A standardized format helps automate reading of the readme file via tools.

PR #3662, opened 2023-08-13
[plsql] Fix for 3660 -- clean up readme

kaby76:
This is a fix for #3660.

In this PR, I changed the readme.md file to clarify the links to the PL/SQL Language Reference. I added a few other links and standardized the format.


kaby76:
Unclear why sql-plus commands were add to the plsql grammar. Only added to the confusion. 

The idea/plan is to split the sql-plus components and plsql components into two grammars and place them side by side in this directory. The doc will be updated with this reorganization.

PR #3663, opened 2023-08-14
[MariaDB] Align to parse exception with MySQL DDL statement including "SELECT 1 + @sum:=1 AS ss;"

harveyyue:


teverett:
@harveyyue  thanks!

PR #3664, opened 2023-08-14
Adding WKT CRS v1 grammar.

fikin:
this is grammar for wkt crs v1.
readme contains references to respected sources.
hope it is fine to be included along the rest.


wkt-crs-v1/wktcrsv1.g4
```diff
@@ -0,0 +1,135 @@
...
+fragment Q: [qQ];
+fragment R: [rR];
+fragment S: [sS];
+fragment T: [tT];
+fragment U: [uU];
+fragment V: [vV];
+fragment W: [wW];
+fragment X: [xX];
+fragment Y: [yY];
+fragment Z: [zZ];
```
KvanTTT, line 135: Please use [caseInsensitive](https://github.com/antlr/antlr4/blob/master/doc/options.md#caseinsensitive) option instead of these fragments.

KvanTTT (changes requested):

wkt-crs-v1/wktcrsv1.g4
fikin: ok, updated it.

wkt-crs-v1/wktcrsv1.g4
```diff
@@ -0,0 +1,135 @@
...
+dYBF:              NUMBER;
+dZBF:              NUMBER;
+rXBF:              NUMBER;
+rYBF:              NUMBER;
+rZBF:              NUMBER;
+mBF:               NUMBER;
+code:              TEXT;
+longitude:         NUMBER;
+angularUnit:       NUMBER;
+value:             NUMBER;
```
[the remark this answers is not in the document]
fikin: these are positional argument and their names are as per specification. 
grammar wise yes, they could be generalized. my initial versions in fact were like that.
but later i ended up by naming them explicitly, which made AST processors code way more readable and less cryptic.

KvanTTT (approved):

KvanTTT:
Thanks for the new grammar and welcome to the world of ANTLR grammars!

PR #3676, opened 2023-08-28
[java20] Fix maven tester for java20

kaby76:
Fixes https://github.com/antlr/grammars-v4/issues/3675 with a few changes to the pom.xml.


teverett:
@kaby76  thanks!

PR #3677, opened 2023-08-30
add support for Cpp target to swift5

neon12345:
Modify the helper classes and add experimental C++ support. It  is not very well  tested yet and the existing tests do not necessarily catch memory errors.


kaby76:
Looks good.

neon12345:
I have found a bug with the bitset. I will fix it and reopen.

PR #3679, opened 2023-08-30
add support for Cpp target to swift5

neon12345:
Modify the helper classes and add experimental C++ support. It is not very well tested yet and the existing tests do not necessarily catch memory errors.


neon12345:
Ok, there is another bug: For "return bitSet.get(text.data()[0])" the full utf8 char must be converted to unicode int. I will add a commit.

teverett:
@neon12345 thanks!

PR #3680, opened 2023-08-30
fix utf8/unicode handling (target Cpp for swift5)

neon12345:


neon12345:
I have to recheck the internal antlr encoding, it seems to be utf32 already?

neon12345:
Seems to be fine, execpt for some ambiguities also in the java version.

teverett:
@neon12345 thanks!

PR #3682, opened 2023-09-01
[Snowflake]Task improvment & alter view fix

MasterKuat:
Begin of refactoring task (at least create)
Fix alter view with object_name instead of id_


teverett:
@MasterKuat thanks!

PR #3683, opened 2023-09-02
Update vba.g4

Xenios91:
Moved deftypeStmt to module elements not block statements, this keyword must be used outside of a sub as it is a module level declaration.


teverett:
@Xenios91  thanks!

Issue #3687, opened 2023-09-04
[pcre] parse of octal is wrong


kaby76:
According to [this](https://stackoverflow.com/questions/77023896/i-cant-trigger-antlr-rule/77033683#77033683), the priority of `backreference` should be moved to the top in order to parse octal literals correctly.

_Note, in my opinion, I don't think the suggested fix is correct. The problem is that the grammar is ambiguous. Just because Antlr4 allows you to fix ambiguity by alt reordering does not mean you should._

_There's an extra level of complexity with the rule `atom` with the addition of "shared_literal", "shared_atom", "literal", etc. The result is that ambiguity._

_I would have kept atom flat with respect to the "shared" rules, i.e., unfold these rules into "atom" then remove the ambiguity--and not by reordering the alts._

bkiers:
This can be closed since the grammar has been updated (and the issue was also resolved along the way): https://github.com/antlr/grammars-v4/pull/3690

kaby76:
This will be fixed with merge of https://github.com/antlr/grammars-v4/pull/3690.

PR #3688, opened 2023-09-04
Fix for #3687 -- parsing of octals

kaby76:
This is a fix for #3687.

This fix changes the order of the alts for `atom` so that octals are parsed correctly. See https://stackoverflow.com/a/77033683/4779853

I've included a basic readme.md, and added a test for an octal r.e.


kaby76:
This issue fixed with grammar rewrite https://github.com/antlr/grammars-v4/pull/3690.

PR #3690, opened 2023-09-04
Update PCRE grammar

bkiers:
The old grammar was based on the 10 January 2012 revision of [www.pcre.org/pcre.txt](http://www.pcre.org/pcre.txt). This PR updates it to the 14 June 2021 revision (including some bug fixes and a more comprehensive test set).


bkiers:
No idea why it still trips over the Go target: I've removed Go from the `desc.xml`... 🤷

kaby76 (commented):
The grammar doesn't work for the Cpp and Dart targets because it uses symbols that conflict with these targets. Please rename "accept" to "accept_", and add all the targets back in. https://github.com/antlr/antlr4/issues/4406

bkiers:
@kaby76 and why does the Go target fail, while it is not in the `desc.xml`?

kaby76:
Same symbol conflict--I just didn't test Go, but did just now. So, the same symbol for all three targets.

bkiers:
But _why_ is the Go target invoked while it is _not_ present in `desc.xml`?

kaby76:
The Github Actions file [unconditionally spawns a VM machine for the Go target](https://github.com/antlr/grammars-v4/blob/443916c7460a8f69e66666873c0088d2b3ec64c8/.github/workflows/main.yml#L15). So the build starts with that. In the Github build it fails to set up the machine correctly. Yes it sets up a machine but may not do the test.

The knowledge about what targets actually work is in the desc.xml file. That is tested [here](https://github.com/antlr/grammars-v4/blob/443916c7460a8f69e66666873c0088d2b3ec64c8/_scripts/test.sh#L311). It should skip the build for Go.

teverett:
@bkiers  @kaby76  good to merge this?

kaby76:
Looks good!

teverett:
@bkiers  thanks!

Issue #3125, opened 2023-02-21
PHP in the test matrix is missing


kaby76:
I was looking at this PR for snowflake, and noticed that the PHP target is not being tested. https://github.com/antlr/grammars-v4/actions/runs/4231262914 It needs to be added. https://github.com/antlr/grammars-v4/blob/8b00eaf4b4eab17bfe5493e96260ee89b6f05efa/.github/workflows/main.yml#L15

I think the main problem was that even though the target may work, in some grammars and test suites, the build takes too long because all input is parsed as a group. I think I should have something in the desc.xml file to say to use individual parsing instead of group parsing, and possible have something that says skip this even though the target "works".

Note:

For the CSharp target, it takes minutes to compile. That is because it encodes the table of ~1000 reserved words directly in the grammar. Instead, it should be implemented as a table lookup when an identifier is lexed. This was common knowledge and practice 40 years ago, but got  lost in the deluge of "software developers" without proper training.

In fact, in _Cooper, K.D. and Torczon, L., 2023. Engineering a Compiler_, page 60, it states very clearly that there is a choice in implementing the keywords as rules in the grammar, or as a table lookup in another rule, such as identifiers. It says that either way, the scanner can be efficient. The trade-off is faster build times vs portability across targets. The grammar is already split, and has a base class, so the later concern is moot.

The PHP target is abnormally slow with this grammar.

KvanTTT:
> That is because it encodes the table of ~1000 reserved words directly in the grammar. Instead, it should be implemented as a table lookup when an identifier is lexed.

If I understand correctly, [Token value comparison operator](https://github.com/antlr/antlr4/issues/1965) resolves this problem. Besides the problem that keywords slows down lexer, they also should be placed to `keyword_as_id` parse rule and users often forget doing that. It's especially actual for `sql` grammars with many tokens, mostly useless.

kaby76:
@KvanTTT I was looking at the parse times for some of the SQL grammars, like plsql. I thought that compile time for C# was 5 minutes, but it looks like it's only 45 seconds. But, the "long compile time" is likely not due to PlsqlLexer.cs since it is only 13k lines, but PlsqlParser.cs, which is 195k lines. The alternative implementation I was thinking of would be in an action for the REGULAR_ID rule. It would look up the string in a hash table and return the token type for the reserved word. But, since the build time is insignificant, no change is needed. Just thinking out loud.

PR #3699, opened 2023-09-06
Fix for #3125 -- Add Antlr4 PHP target testing

kaby76:
This is a fix for #3125 -- the PHP target is not being tested.

Testing of the PHP target was removed long ago, and never added back. In this PR, I added testing for the PHP target to the build. The driver templates for testing were updated so that the version of the Antlr tool is extracted from composer.json. Both Bash and Pwsh templates work, but I only added testing in Github Actions only for the Bash. After all, why duplicate testing and waste more money?

Attached is the test for the PHP target against the entire repo. Only python/python3 and sql/snowflake failed.
[out.txt](https://github.com/antlr/grammars-v4/files/12609475/out.txt)

In the case of python3, the problem is that there is no port for PHP. For snowflake, the grammar contains a symbol collision (PHP id's are case insensitive). But, even renaming one of the symbols, REFRESH_ON_CREATE, the grammar is so slow as to be unusable.

Because I touched python/python3/ to remove PHP testing for it, the TypeScript target broke. That's because there was a change (really a fix) to the Antlr4 TypeScript target runtime API for the constructor of CommonToken. I fixed the code and reset the desc.xml to make sure people know only Antlr 4.13.1 can be used with the grammar.


teverett:
@kaby76  thanks!

PR #3702, opened 2023-09-07
PLSQL: Trim+Both syntax fix & Improved cascade behavior for drop constraints

Naros:
Upstream changes based on DBZ-6876 - https://github.com/debezium/debezium/pull/4839


kaby76:
Please add a test that exercises the ` | TRIM '(' ((LEADING | TRAILING | BOTH)? expression? FROM)? concatenation ')'` change you made. Here is the cover.html for examples/*.sql, and it does not highlight the alt for rule `string_function`. 
[cover.html.txt](https://github.com/antlr/grammars-v4/files/12552826/cover.html.txt), which you changed (not that it was tested before!).

Not that it matters much, but I can't explain why your change causes a small but statistically significant performance degradation around 5% in the tests in hw-examples/. I just spent weeks making a change to address some of the performance issues with this grammar. I'll have to make a note of it.

(AMD Ryzen 2700; 16GB; SSD; Windows 11; hw-examples/*.sql repeated 10 times, outliers removed.)
| target | before your change | your PR |
| --- | --- | --- |
| CSharp | 1.56  &plusmn; 0.05 s | 1.73  &plusmn; 0.03 s |

Naros:
Hi @kaby76, the `package_with_cursor.sql` already has such a test case.  When I went to introduce the drop constraint behavior, the SQL test failed. After investigation, it was because the original grammar wasn't interpreting the body correctly, and it just worked as a pure coincidence.

kaby76:
The desc.xml does not contain testing of examples-sql-script/. Please add it (see below). We don't test using the Maven tester because it's obsolete, uses an old version of Antlr, and is not target-independent.

At a minimum, it needs to run on CSharp and Java.
```
   <test>
      <name>examples-sql-script</name>
      <targets>CSharp;Java</targets>
      <inputs>examples-sql-script</inputs>
   </test>
```
Otherwise, it looks fine.

teverett:
@Naros thanks!

PR #3705, opened 2023-09-12
Add Elixir grammar

bkiers:
Add Elixir grammar + examples


bkiers:
Python, Go and Dart are probably failing because of some target-specific keyword clashes, but I do not know which of the parser rules is the culprit 🤷

kaby76:
For Dart, it's saying it doesn't like [bool](https://github.com/antlr/grammars-v4/actions/runs/6156077774/job/16704074230?pr=3705#step:22:117) and [expressions](https://github.com/antlr/grammars-v4/actions/runs/6156077774/job/16704074230?pr=3705#step:22:130). The usual fix is to rename with a trailing underscore and create a [Github Issue in antlr4](https://github.com/antlr/antlr4/issues).

kaby76:
For the Go target, the parse is taking over 5 minutes. For examples/block.ex, the parser for CSharp takes ~40s, which is pretty slow. There are a few solutions here.
* Declare that the Go target works, and add a test for that target with a small subset of the examples/*.ex files in the desc.xml. An example of how to set up a different test is [here](https://github.com/antlr/grammars-v4/blob/b1fa155ed842cb9c06b79601dfb7873677f1bd76/sql/plsql/desc.xml#L10-L14).
* Declare that the Go target doesn't work, and remove the target from the desc.xml.

teverett:
@bkiers  thanks!

PR #3711, opened 2023-09-16
New grammar -- Add Pegen for reading Python3 grammar

kaby76:
This is a "meta grammar" for CPython's Pegen parser generator. The grammar
describes other grammars, in particular, the Python3 grammar [here](https://github.com/python/cpython/blob/e57ecf6bbc59f999d27b125ea51b042c24a07bd9/Grammar/python.gram). It is used in the Trash toolkit for a few years already.

Why add such a specialized grammar? The plan is to use this to scrape the Python3 syntax from the source in the Python repository, then automatically converted into Antlr4 format.


teverett:
@kaby76  thanks!

Issue #3715, opened 2023-09-18
URL parse error


bendrissou:
I am using the provided [URL grammar](https://github.com/antlr/grammars-v4/blob/master/url/url.g4) to generate an ANTLR parser, which then I use to parse URL inputs.

However, the parser is failing to parse few URL inputs. Upon inspection, the URLs were found to conform to the grammar, as expected, since the URLs were generated from grammar-based generators.

Example of such URLs:
`http://abc.~xyz/path`

I found that the part causing the issue is: `.~`

Which comes from the following grammar segment:

```g4
hostname
   : string ('.' string)*
   
...

STRING
   : ([a-zA-Z~0-9] | HEX) ([a-zA-Z0-9.+-] | HEX)*
   ;
```

In the URL example the dot is wrongly parsed as part of the first string, consequently when the tilde is encountered, it is rejected, as the parser only allows tilde symbol at the start of strings. 

Instead if the dot is parsed as a delimiter, then the tilde would mark the start of a new string. Which would not cause errors.

PR #3718, opened 2023-09-21
Remove ambiguity in the url grammar

bendrissou:
This is to solve an [open issue](https://github.com/antlr/grammars-v4/issues/3715).

The dot character is allowed in strings. So there should not be a dot delimiter between strings. Otherwise, the parser would incorrectly parse the delimiter as part of the first string.

### DIFF
--- a/url/examples/example30.txt
+++ b/url/examples/example30.txt
@@ -1,0 +1,1 @@
+http://www.example.com:8080
--- a/url/url.g4
+++ b/url/url.g4
@@ -52,7 +52,7 @@ host
    ;
 
 hostname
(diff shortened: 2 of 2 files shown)


KvanTTT (changes requested):
Please add an example that covers suggested changes to the [examples dir](https://github.com/antlr/grammars-v4/tree/master/url/examples).

bendrissou:
New example added.

bendrissou:
Hi @KvanTTT 

Any updates?

KvanTTT (approved):

KvanTTT:

Issue #3720, opened 2023-09-21
Minimum Go required is now 1.20, not 1.18


kaby76:
[According to Jim Idle, the minimum version is 1.20](https://github.com/antlr/antlr4/issues/4423#issuecomment-1728710634) not 1.18. It's explicit in the go.mod file. https://github.com/antlr/antlr4/blob/4.13.0/runtime/Go/antlr/v4/go.mod

Time to update this in the templates.

PR #3721, opened 2023-09-21
Fix for #3720 -- Update Go target testing versions.

kaby76:
* The version of Go in the go.mod files should be 1.20. That's the version referenced in the Go Antlr runtime.
* The version of Go installed should be anything greater than or equal to 1.20. We can certainly use a more recent version of Go, e.g., 1.21 which is the current latest. So, for the Github Actions workflow file, I updated to the latest of Go.
* My repo has a Dependabot update that I mistakenly added in. It shouldn't because only these should be done by an Admin for the github.com/antlr/grammars-v4. I undid the change. Eventually, I will need to refork the entire repo from scratch so I don't see spurious commits.
* For now, we don't touch the version of Antlr for the Go target. Right now, it should be 4.13.0. There is no 4.13.1 for Go yet.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -55,7 +55,7 @@ jobs:
       if: ${{ matrix.language == 'Go' }}
       uses: actions/setup-go@v4
       with:
-        go-version: '^1.20.0'
+        go-version: '^1.21.0'
     - name: Test Go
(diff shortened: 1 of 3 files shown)


teverett:
@kaby76  thanks!

Issue #3719, opened 2023-09-21
Update testing of Cpp target to 4.13.1


kaby76:
As far as I know, I think the 4.13.1 code is fine to use now. [The template should be updated](https://github.com/antlr/grammars-v4/blob/d9a6fe5d2720eb9b5253c44013e1ff2fefc096a1/_scripts/templates/Cpp/st.CMakeLists.txt#L7). Dependabot doesn't know how to update this version string, so it has to be done manually.

kaby76:
* There doesn't seem to be any Cpp code changes in https://github.com/antlr/antlr4/releases
* There doesn't seem to be any Cpp code diffs under github.com/antlr/runtime/Cpp for 4.13.1. `git diff 7ed420f 0ae86c2 -- .`. The tool itself was changed slightly, so it is probably good to just make sure it all works for 4.13.1.
* Note, there is only one grammar that requires 4.13.1 and that is python/python3. The desc.xml lists that it should only be tested for 4.13.1 but it seems it's tested anyway--using 4.13.0. I probably don't do any integrity checking that the version specified in the desc.xml is the same as that in the templates. That's a bug.
* Everything that was working still seems to be working (I'm still running the testing for Cpp on the whole repo).

PR #3722, opened 2023-09-21
Fix for #3719 -- Update version for Cpp testing

kaby76:
Antlr 4.13.1 has been released. Although there are no changes in the Cpp runtime, the tool changed slightly. Also, python/python3/ requires Antlr 4.13.1, so we need to update the version for the Cpp target. Dependabot cannot monitor the version of Antlr in a CMake file, so we have to do this manually. To make sure everything is okay, I tested the target across all grammars, and it seems kotlin/kotlin-formal does not work. So, that grammar/target was removed.


teverett:
@kaby76  thanks!

Issue #3728, opened 2023-09-25
JavaScript parser out of range exception


TimurBinogerov:
I use antlr4 to parse JavaScript expressions in C# app. I've encountered out of range exception during parsing something like "``` (||```". It occurs in `JavaScriptParserBase.cs` `lineTerminatorAhead()` method. I assume there is missprint in decrement Token index by 2 (instead of 1) or shoud be range check.

kaby76:
Note: For input `(||`, the parser crashes on all targets (CSharp, Go, Java, JavaScript) except Python3. The base class code for each target is supposed to be direct translations of each other, so they should all function exactly the same, but they don't. But, the parsers should not crash.

kaby76:
This code should test whether [possibleIndexEosToken](https://github.com/antlr/grammars-v4/blob/18d8d8962f2876f555105042901b39d210c746e9/javascript/javascript/CSharp/JavaScriptParserBase.cs#L93) is less than zero, and return false if so.

For Python3, [the base class code looks like this](https://github.com/antlr/grammars-v4/blob/18d8d8962f2876f555105042901b39d210c746e9/javascript/javascript/Python3/JavaScriptParserBase.py#L59-L60). However, it looks like the Python3 target is wrong. It *must* check whether the index is out of range, [but it does not do so](https://github.com/antlr/antlr4/blob/0eb38a02d7aa67147bdba23287501d4e5540dfc1/runtime/Python3/src/antlr4/BufferedTokenStream.py#L80-L81)! For a comparison, see [CSharp](https://github.com/antlr/antlr4/blob/0eb38a02d7aa67147bdba23287501d4e5540dfc1/runtime/CSharp/src/BufferedTokenStream.cs#L257-L261) or [Java](https://github.com/antlr/antlr4/blob/0eb38a02d7aa67147bdba23287501d4e5540dfc1/runtime/Java/src/org/antlr/v4/runtime/BufferedTokenStream.java#L185-L188).

kaby76:
https://github.com/antlr/antlr4/issues/4427

kaby76:
This has been fixed.

PR #3729, opened 2023-09-26
Fix for #3728 -- crash in runtime code with an out of bounds get() call

kaby76:
This is a fix for #3728 -- JavaScript parser out of range exception. The problem also occurs with ecmascript, so I added changes over there.

With input '(||', the parse crashes on a get() method call because the index computed in the support code for this grammar does not check if the index is out of bounds.

This change fixes range checking for get() calls in [javascript/javascript/](https://github.com/antlr/grammars-v4/tree/master/javascript/javascript). The responsibility should be on caller, not callee.

Testing for the upper bound index is unnecessary because it is computed from the current token index, looking backwards in the token stream.


teverett:
@kaby76  thanks!

Issue #3730, opened 2023-09-26
[Build] Update trgen tool version to support "globstar" patterns


kaby76:
Currently, the desc.xml allows one to indicate the *directory* containing files to test using the `<inputs>` element (e.g., `<inputs>examples</inputs>`). But, this has the problem of parsing files that one may not want to parse in the directory, such as a readme.md file. Further, to create a quick, small subset of the tests, one has to copy the files to a new directory. The old "Maven Antlr Tester" (which appears to be unsupported at this point) had this capability.

Trgen will support "globstar" matching shortly. This will allow the tests to specify exactly what to parse without creating directories of subsets of tests, and bypass readme.md files.

PR #3731, opened 2023-09-26
Fix for #3730 -- update Trash tools and templates, bump testing of Java target to Antlr 4.13.1

kaby76:
This is a fix for #3730 -- use the latest trgen, update the templates for testing, and update testing of Java with Antlr 4.13.1.

For https://github.com/antlr/grammars-v4/pull/3587 , this PR for porting the grammars-v4/dart2 grammar TypeScript was never completed. The code works fine, but that testing the TypeScript target is very slow. To work around this problem, I suggested creating a subset of the examples/ directory, and testing the target on that small subset. It was never finished.

The better way to solve the testing problem is to support using "globstar" patterns. This was supported in the old Antlr Maven tester, but never implemented in trgen. This PR now uses a version of trgen that supports "globstar". The templates are updated to take advantage of the new trgen. And, a couple of grammars were updated (abb, ocl) to test the new feature.

For example, you can specify all `examples/**/*.g4`, recursively in examples/ and all sub-directories, in the antlr4 grammar:
```
   <targets>CSharp;Cpp;Java;JavaScript;Python3</targets>
   <grammar-files>ANTLRv4Lexer.g4;ANTLRv4Parser.g4</grammar-files>
   <inputs>examples/**/*.g4</inputs>
</desc>
```

Antlr 4.13.1 was just published. For most targets, Dependabot updates critical files that reference the Antlr runtime. But, for Java, it is still manually updated. This PR updates the version of Antlr for the Java target.

The plan is to create a new PR that replaces https://github.com/antlr/grammars-v4/pull/3587 so the problem can be fixed and closed quickly.


teverett:
@kaby76  thanks!

PR #3736, opened 2023-09-28
[MariaDB] Allow whitespace in user name specification

joschi:
The user name specification in MariaDB (and MySQL) allows for a whitespace character between the local user name and the host name (incl. the '@' character) afterwards.

This means that the following strings are valid user name specifications which can be used in CREATE and GRANT statements:
```
`user`@`hostname`
`user` @'hostname'
"user"         @'example.com'
userfoo @localhost
```
References:

- https://mariadb.com/kb/en/grant/
- https://mariadb.com/kb/en/create-user/#account-names
- https://github.com/antlr/grammars-v4/pull/3737


teverett:
@joschi  thanks!

PR #3737, opened 2023-09-28
[MySQL] Allow space in user name specification

joschi:
The user name specification in MySQL (and MariaDB) allows for a whitespace character between the local user name and the host name (incl. the '@' character) afterwards.

This means that the following strings are valid user name specifications which can be used in CREATE and GRANT statements:
```
`user`@`hostname`
`user` @'hostname'
"user"         @'example.com'
userfoo @localhost
```

References:

- https://dev.mysql.com/doc/refman/8.0/en/account-names.html
- https://github.com/antlr/grammars-v4/pull/3736


teverett:
@joschi  thanks!

PR #3738, opened 2023-09-28
[Snowflake]Add keyword and fix priority for sequence in table creation

MasterKuat:
In a statement like : CREATE OR REPLACE TABLE TESTSEED (IDENT int DEFAULT SCHEM.SEQID.NEXTVAL,mycol string);

SCHEM.SEQID.NEXTVAL should match first object_name DOT NEXTVAL instead of primitive_expression DOT (object_name DOT NEXTVAL)

We could also replace _id by object_name in primitive_expression but it's could lead to a lot of issue.

### DIFF
--- a/sql/snowflake/SnowflakeParser.g4
+++ b/sql/snowflake/SnowflakeParser.g4
@@ -3687,7 +3687,8 @@ expr_list_sorted
     ;
 
 expr
-    : primitive_expression
+    : object_name DOT NEXTVAL
+    | primitive_expression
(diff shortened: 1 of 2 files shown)


teverett:
@MasterKuat  there is a merge conflict to resolve

MasterKuat:
Hi @teverett, merge conflict fixed.
Thank you

teverett:
@MasterKuat thanks!

PR #3739, opened 2023-09-29
[Snowflake]New keyword in identity/Sequence

MasterKuat:
Bundle [2023_06](https://docs.snowflake.com/en/release-notes/bcr-bundles/2023_06/bcr-1225) introduce ORDER / NO ORDER in sequence / dentity.

https://docs.snowflake.com/en/sql-reference/sql/create-table
https://docs.snowflake.com/en/sql-reference/sql/create-sequence
https://docs.snowflake.com/en/sql-reference/sql/alter-sequence


teverett:
@MasterKuat  thanks!

Issue #3746, opened 2023-10-02
[caql] The grammar does not have a proper start rule, doesn't parse examples out of the manual, etc.


kaby76:
* https://github.com/antlr/grammars-v4/tree/master/caql does not have an EOF-terminated start rule. A grammar that does not use an EOF-terminated start rule can cause a lot of problems.
* The grammar has a readme that gives little information for maintenance, i.e., links to the spec or doc, author of the grammar, where the test (only one) comes from, etc.
* Grabbing several of the examples from https://docs.circonus.com/caql/reference/ shows the grammar doesn't work. One of the problems is that there is a separate grammar for specific functions, which is just wrong. The syntax is very specific in the manual: `function <- name '(' args ',' kwargs ')'  '{' sources '}'`.
* Rule METRIC_NAME shadows LABEL_NAME. The author does not understand Antlr lexer rules at all.
* I don't know why, but this grammar is split. It doesn't have to be; it just makes things more complicated than they have to be.

PR #3751, opened 2023-10-04
[caql] Fixes for #3746

kaby76:
This is a fix for #3746.

* A readme.md was added so the grammar can be maintained using the links to the online manual.
* The grammar was combined. No need to make this more complicated than it is.
* Grammar for function calls were changed to that explicitly stated in the manual. There is no need for AGGREGATION_OPERATOR and FUNCTION rules, which are lists of functions. Again, as stated in the online manual, a function is just a `name`. One should not encode function names explicitly in a parser/lexer grammar. It makes the grammar inflexible.
* Added UUID lexer rule. The manual constantly refers to uuid. Although the syntax is not specified in the manual, I took the liberty of adding the definition from [rfc4122](https://datatracker.ietf.org/doc/html/rfc4122).
* Fixed the start rule. All grammars should use an EOF-terminated start rule. It forces the parser to consume to the end of file, and helps identify the start rule for both tools and humans. Otherwise, we have to peek into desc.xml and pom.xml.
* Added test input files "s[-7].txt". These tests are from the online manual.
* Updated the pom.xml and desc.xml with new start rule.


teverett:
@kaby76  thanks!

PR #3753, opened 2023-10-06
remove quote arround COMPRESSION keyword

PHaroZ:
The keyword `COMPRESSION` allows a quoted version like `'COMPRESSION'` (or even `'COMPRESSION"`) since [this commit](https://github.com/antlr/grammars-v4/commit/1079f9ba5fc9b1203da18a38ddb2899708c11d71).

According to [the official documentation](https://dev.mysql.com/doc/refman/5.7/en/create-table.html) this is not permitted. And this break parsing when a STRING_LITERAL is equals to `'COMPRESSION'`, like in the snippet `compression int DEFAULT '100' COMMENT 'compression'`.


KvanTTT (approved):

PR #3755, opened 2023-10-08
VBA Grammar updated

Xenios91:
Added support for line numbers. Line numbers do not appear to be common anymore, however, they are part of the VBA language. A line number acts just like a line label except it is a digit and the colon after it is optional.

The goto statement has more information

https://learn.microsoft.com/en-us/office/vba/language/reference/user-interface-help/goto-statement


Xenios91:
got a bug to fix i just found

Xenios91:
should be good now

KvanTTT (changes requested):
Please add an example to [examples](https://github.com/antlr/grammars-v4/tree/master/vba/examples) that covers suggested changes.

Xenios91:
Should be good now with https://github.com/antlr/grammars-v4/pull/3755/commits/34108570d68dc22dcc7c35074ca9469a3583cd4b

KvanTTT (approved):

PR #3761, opened 2023-10-10
[javascript] allow yield as identifier

meraedit:


kaby76:
The PHP target is not a good target. It is slow and has bugs. Please remove the PHP target for this grammar by adjusting the desc.xml. The PR should then pass.
```
$ git diff .
diff --git a/c/desc.xml b/c/desc.xml
index 26bde4ce..b14a37d5 100644
--- a/c/desc.xml
+++ b/c/desc.xml
@@ -1,4 +1,4 @@
-   <targets>CSharp;Cpp;Dart;Go;Java;JavaScript;PHP;Python3;TypeScript</targets>
+   <targets>CSharp;Cpp;Dart;Go;Java;JavaScript;Python3;TypeScript</targets>
 </desc>
10/11-06:38:32 ~/issues/g4-3761/c
```

KvanTTT:
@kaby76 removing of PHP target doesn't relate to this PR.

kaby76:
> @kaby76 removing of PHP target doesn't relate to this PR.

Testing of the target is per grammar. There are grammars that work with the PHP taget such as [abb](https://github.com/antlr/grammars-v4/blob/4d3b08fa77c5820a1cdeaf6a016aee7ce1c7fdeb/abb/desc.xml#L4C48-L4C51). And there are grammars that don't work for the PHP target because they don't work or haven't been ported, e.g. [java](https://github.com/antlr/grammars-v4/blob/4d3b08fa77c5820a1cdeaf6a016aee7ce1c7fdeb/java/java/desc.xml#L3). The desc.xml is the only place that formally states whether a grammar works for a target. This grammar is one of those that doesn't work for PHP. There is no easy explanation for why the parser works on MacOS and Windows but goes into an infinite recursion on Ubuntu. Testing does not distinguish a functioning grammar based on the OS.

meraedit:
Created another pull request because the first one was not really correct, also had some reverts... sorry

kaby76:
No problem.

I see that javascript/javascript/desc.xml did not have the PHP target anyway. The Github Action for this PR actually tested the "c" grammar, and that does have a problem. Testing uses "git diff" between your PR and the tip of master. If you made a branch long ago, or was working on "master" that was even just a few days old, the tester would have picked up testing for the "c" grammar.

Testing should not test against the tip of "master" but the base commit where you started making changes.

PR #3763, opened 2023-10-11
Added `SEQ` to REGULAR_ID list in PLSQL

WazzaJB:
Hey,

We were running into an issue where the parser was failing to continue with the parse because variables had been named 'seq', I figured this was because it was a defined token in the lexer. Removing the token from the lexer broke the parser.

Adding the token to the parsers REGULAR_ID as I have done here works great, however it significantly impacts performance of the parser and was wondering if anyone could offer appropriate guidance on how to improve this PR?

This is my first time using Antlr having worked with TypeScripts AST quite extensively. It is a fantastic tool, thank you for your work on it.

### DIFF
--- a/sql/plsql/PlSqlParser.g4
+++ b/sql/plsql/PlSqlParser.g4
@@ -7152,6 +7152,7 @@ regular_id
     | SELF
     | SERIALLY_REUSABLE
     | SET
+    | SEQ
     | SHARDSPACE
     | SIGNTYPE
     | SIMPLE_INTEGER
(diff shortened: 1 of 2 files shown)


KvanTTT (changes requested):
Thanks for the contribution and welcome to ANTLR grammar repository!

But please cover your change by an example in [examples](https://github.com/antlr/grammars-v4/tree/master/sql/plsql/examples) directory.

WazzaJB:
Thanks for the warm welcome! I have pushed a simple example but fear this may not be enough, do you have any guidelines on what makes a good example? I have intentionally kept it simple but it does not demonstrate the performance challenges I have found with my change.

KvanTTT (approved):
I think this example is enough. Thanks again!

Issue #3766, opened 2023-10-11
[build] PHP install and version test is completely missing from github actions file!


kaby76:
I added PHP to the workflow recently with Antlr 4.13.1, but there is no step to install a *specific version* of PHP. And, there is no step to test what version PHP is installed. (It's done for the other targets.) Consequently, I  cannot tell what version of PHP was used in testing, and why this build failed:  https://github.com/antlr/grammars-v4/actions/runs/6472152694/job/17582212209 and https://github.com/antlr/grammars-v4/actions/runs/6456251348/job/17525843527 . The PHP target of the "c" grammar fails on Github Actions, but works on my Ubuntu machine.

PR #3770, opened 2023-10-11
[build] Fix for #3766 -- install specific version of PHP, set max stack size

kaby76:
This is a fix for #3766, where we weren't setting up a specific version of PHP. In addition, the php.ini settings weren't set up for large stack sizes, which is what the c grammar requires.

This change is simple:
* I added to .github/workflow/main.yml an action to "Setup PHP", and another to print out the version of PHP installed. In the setup, we set `xdebug.max_nesting_level=1000` so that the [c grammar](https://github.com/antlr/grammars-v4/tree/master/c) now runs cleanly on all platforms.
* To test the change, I modified the [c grammar](https://github.com/antlr/grammars-v4/tree/master/c) desc.xml by adding a space. The test now works.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -135,6 +135,7 @@ jobs:
       uses: shivammathur/setup-php@v2
       with:
         php-version: '8.2'
+        ini-values: xdebug.max_nesting_level=1000
     - name: Print PHP version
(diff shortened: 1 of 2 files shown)


teverett:
@kaby76 thanks!

Issue #3772, opened 2023-10-12
[javascript] problem with continue statement and newline followed by whitespace


meraedit:
```js
for (var i = 0; i < [].length; i++) {
	var o = {}
	if (false) continue
	o[test]   //	-  'o' is label for the continue statement, which is wrong
		  //	- no viable alternative at input '['
}
```

The result is 

```
continue o;
[test]
```

instead of

```
continue;
o[test]
```

The grammar contains this rule, `Continue ({this.notLineTerminator()}? identifier)? eos`, therefore `o` should not be a label of continue because there is a newline between them.

The same applies to `break` and `return` statements.

bkiers:
It appears to be an issue with the `notLineTerminator()` method: https://github.com/antlr/grammars-v4/issues/3777

bkiers:
Did a couple of tests:

## JS 1
```js
{
  continue
}
```

## Tree 1
```
'- block
   |- { (OpenBrace)
   |- statementList
   |  '- statement
   |     '- continueStatement
   |        |- continue (Continue)
   |        '- eos
   '- } (CloseBrace)
```

## JS 2
{
  continue a
}
```

## Tree 2
'- block
   |- { (OpenBrace)
   |- statementList
   |  '- statement
   |     '- continueStatement
   |        |- continue (Continue)
   |        |- identifier
   |        |  '- a (Identifier)
   |        '- eos
   '- } (CloseBrace)
```

## JS 3
{
  continue
  a
}
```

## Tree 3
'- block
   |- { (OpenBrace)
   |- statementList
   |  '- statement
   |     '- continueStatement
   |        |- continue (Continue)
   |        |- identifier
   |        |  '- a (Identifier)
   |        '- eos
   '- } (CloseBrace)
```

## JS 4
{
  continue
  a.mu()
}
```

## Tree 4
'- block
   |- { (OpenBrace)
   |- statementList
   |  |- statement
   |  |  '- continueStatement
   |  |     |- continue (Continue)
   |  |     '- eos
   |  '- statement
   |     '- expressionStatement
   |        |- expressionSequence
   |        |  '- argumentsExpression
   |        |     |- memberDotExpression
   |        |     |  |- identifierExpression
   |        |     |  |  '- identifier
   |        |     |  |     '- a (Identifier)
   |        |     |  |- . (Dot)
   |        |     |  '- identifierName
   |        |     |     '- identifier
   |        |     |        '- mu (Identifier)
   |        |     '- arguments
[...]

PR #3773, opened 2023-10-12
[javascript] fix problem with continue/break/return statements #3772

meraedit:


javascript/javascript/JavaScriptParser.g4
```diff
@@ -186,15 +186,15 @@ varModifier  // let, const - ECMAScript 6
     ;
 
 continueStatement
-    : Continue ({this.notLineTerminator()}? identifier)? eos
+    : Continue ({!this.lineTerminatorAhead()}? identifier)? eos
```
kaby76, line 189: Antlr4 doesn't have a target independent action (yet). So, `!` cannot work because the syntax for negating a boolean value is different across targets. You'll have to move the "!" into the target-specific code, and name the function appropriately.

meraedit: Thank you! Fixed it

teverett:
@meraedit thanks!

PR #3778, opened 2023-10-14
Fix 'here' method

### LINKED ISSUE #3772: [javascript] problem with continue statement and newline followed by whitespace

bkiers:
Fixes #3772 

There is a bug in the `here` method. If you add a `println` to the method:

```java
protected boolean notLineTerminator() {
    return !here(JavaScriptParser.LineTerminator);
}

private boolean here(final int type) {

    // Get the token ahead of the current index.
    int possibleIndexEosToken = this.getCurrentToken().getTokenIndex() - 1;
    if (possibleIndexEosToken < 0) return false;
    Token ahead = _input.get(possibleIndexEosToken);

    boolean result = (ahead.getChannel() == Lexer.HIDDEN) && (ahead.getType() == type); // <-- Added
    System.out.printf("ahead='%s', result=%s%n", ahead.getText().replace("\n", "\\n"), result); // <-- Added

    // Check if the token resides on the HIDDEN channel and if it's of the
    // provided type.
    return (ahead.getChannel() == Lexer.HIDDEN) && (ahead.getType() == type);
}
```

and then test what the parser rule `continueStatement`:

```
continueStatement
    : Continue ({this.notLineTerminator()}? identifier)? eos
    ;
```

using the Java code:

```java
JavaScriptLexer lexer = new JavaScriptLexer(CharStreams.fromString("{\n  continue\n  a\n}"));
JavaScriptParser parser = new JavaScriptParser(new CommonTokenStream(lexer));
parser.block();
```

then the following is printed:

```
ahead='  ', result=false
ahead='  ', result=false
ahead='  ', result=false
```

Which is incorrect: after the `continue` _is_ a line break,so there should have been printed `ahead='\n', result=true` to the console.

With the provided fix, this is printed (when adding a `println`), and the unit test shows it now works (which failed before this PR).

### DIFF
--- a/javascript/javascript/CSharp/JavaScriptParserBase.cs
+++ b/javascript/javascript/CSharp/JavaScriptParserBase.cs
@@ -72,14 +72,19 @@ public abstract class JavaScriptParserBase : Parser
     /// </param>
     protected bool here(int type)
     {
-        // Get the token ahead of the current index.
-        int possibleIndexEosToken = CurrentToken.TokenIndex - 1;
-        if (possibleIndexEosToken < 0) return false;
-        IToken ahead = ((ITokenStream)this.InputStream).Get(possibleIndexEosToken);
+        // Get the most recently emitted token.
+        IToken currentToken = ((ITokenStream)this.InputStream).LT(-1);
 
-        // Check if the token resides on the Hidden channel and if it's of the
+        // Get the next token index.
+        int nextTokenIndex = currentToken == null ? 0 : currentToken.TokenIndex + 1;
+
+        // Get the token after the `currentToken`. By using `_input.get(index)`,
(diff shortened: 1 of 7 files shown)


teverett:
@bkiers  thanks!

Issue #3780, opened 2023-10-16
[sparql] Parsing error with language subtags like "@ca-ES"


martijuanola:
The following code in SPARQL uses a language tag `ca` with a region subatg `ES`:

````SPARQL
PREFIX : <http://example.com>

SELECT ?x WHERE {
	?x :name "Test"@ca-ES .
}
````

If parsed, the following error is thrown:

``````
4:19 mismatched input '-' expecting {'{', '}', 'OPTIONAL', 'GRAPH', 'FILTER'}
``````

The lexer rule `LANGTAG` is not well formed in [/grammars-v4/sparql/Sparql.g4](https://github.com/grammars-v4/sparql/Sparql.g4) according to the [official SPARQL grammar](https://www.w3.org/TR/sparql11-query/#grammar).

PR #3781, opened 2023-10-16
[sparql] Fix for #3780 -- parsing error with language subtags

martijuanola:


KvanTTT (changes requested):

But please add an example that covers suggested changes to the `sparql/examples` directory.

martijuanola:
Ups! Example added. That should do it!

KvanTTT (approved):

PR #3784, opened 2023-10-17
[bnf] Fix #3779

### LINKED ISSUE #3779: [bnf] Notion of left and right parentheses, braces, and brackets reversed
In the [bnf](https://github.com/antlr/grammars-v4/tree/master/bnf) grammar, the notion of left and right brackets, braces, and parentheses are all reversed. In addition, this is NOT a grammar for BNF (no alts, no kleene operators). In fact, BNF has a very specific syntax, one and only one syntax. This a grammar for a type of EBNF.

* Left square bracket https://www.compart.com/en/unicode/U+005B https://github.com/antlr/grammars-v4/blob/d80a28ac0e246b41b0b619aa1944a85bc8046ae6/bnf/bnf.g4#L111-L112

* Right square bracket https://www.compart.com/en/unicode/U+005D https://github.com/antlr/grammars-v4/blob/d80a28ac0e246b41b0b619aa1944a85bc8046ae6/bnf/bnf.g4#L107-L109

I came across this while trying to use this grammar to refactor the Lua grammar at the end of the [Manual](https://www.lua.org/manual/5.4/manual.html).

kaby76:
This is a fix for #3779 -- the names of the string literals are reversed in meaning.

In this PR, I substituted string literals for the lexer symbols in parser rules. There is no requirement that lexer rule names have to be used in a parser rule, and it is clearer when reading the parser rules to just use the string literals, even if the grammar is later split. The lexer rule names for these single-character string literals are now just the official Unicode character names (e.g., "[Greater_Than_Sign](https://www.compart.com/en/unicode/U+003E)"), with underscores substituted for dashes and spaces in the name. This is tidier, cleaner, more standardized than trying to invent one's own name.

I am planning to reorganize and rename this grammar as "ebnf", and introduce a "bnf" grammar that takes different versions of BNF derived from the original papers in the '50s and '60s on BNF. Some of those papers have hand-written symbols, but the plan is to use Unicode for recognizing these, as well as incorporating the more contemporary versions of BNF that throw away the less-than and greater-than signs, and use quotes around terminals. There seems to be a lot of confusion as to what is BNF.

### DIFF
--- a/bnf/bnf.g4
+++ b/bnf/bnf.g4
@@ -45,7 +45,7 @@ rhs
     ;
 
 alternatives
-    : alternative (BAR alternative)*
+    : alternative ('|' alternative)*
     ;
 
 alternative
@@ -61,15 +61,15 @@ element
     ;
 
 optional_
-    : REND alternatives LEND
+    : '[' alternatives ']'
     ;
 
 zeroormore
-    : RBRACE alternatives LBRACE
+    : '{' alternatives '}'
     ;
 
 oneormore
-    : RPAREN alternatives LPAREN
+    : '(' alternatives ')'
     ;
 
 text_
@@ -77,7 +77,7 @@ text_
     ;
 
 id_
-    : LT ruleid GT
+    : '<' ruleid '>'
     ;
 
 ruleid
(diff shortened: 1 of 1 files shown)


teverett:
@kaby76 thanks!

PR #3791, opened 2023-10-22
[PlSql] Add pragma_declaration to package_obj_body

marty-se:
E.g. https://docs.oracle.com/en/database/oracle/oracle-database/21/lnpls/SERIALLY_REUSABLE-pragma.html states that: "The SERIALLY_REUSABLE pragma can appear in the declare_section of the specification of a bodiless package, or in both the specification and body of a package, but not in only the body of a package."

Also, `pragma exception_init (...)` has been verified to work in the declaration section of a package body, although https://docs.oracle.com/en/database/oracle/oracle-database/21/lnpls/EXCEPTION_INIT-pragma.html doesn't mention it.

### DIFF
--- a/sql/plsql/PlSqlParser.g4
+++ b/sql/plsql/PlSqlParser.g4
@@ -629,7 +629,8 @@ function_spec
     ;
 
 package_obj_body
-    : exception_declaration
+    : pragma_declaration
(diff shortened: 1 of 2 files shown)


KvanTTT (approved):

Issue #3795, opened 2023-10-24
[javascript] of should be a keyword


meraedit:
Is there a reason why 'of' is not a keyword?
There's a comment in the grammar file about the ForOfStatement:  that 'of' is an identifier. and this.p("of") not work in sometime.
When we have an incomplete for of loop (while typing) or with a typo:
```
for (cont e of obj) {  } //typo, should be 'const'
```
that results in a rule iterationStatement failed predicate: {this.p("of")}? error

meraedit:
thank you!

PR #3796, opened 2023-10-24
fix of should be a keyword #3795

meraedit:


KvanTTT (changes requested):
Please add a test to [VarDecl.js](https://github.com/antlr/grammars-v4/blame/master/javascript/javascript/examples/VarDecl.js) or [SpecialIdentifiers.js](https://github.com/antlr/grammars-v4/blame/master/javascript/javascript/examples/SpecialIdentifiers.js) that covers a case when `of` is used as identifier.

meraedit:
of was already in [SpecialIdentifiers.js](https://github.com/antlr/grammars-v4/blame/master/javascript/javascript/examples/SpecialIdentifiers.js) line 5
But this is also about the for-of loop.
For some reason it was not a keyword, although the ecma standard allows it.

KvanTTT:
Oh, definitely, I missed it. Thanks!

KvanTTT (approved):

PR #3797, opened 2023-10-26
Initial commit

RobEin:
I closed and deleted the previous PR, because in the meantime I also finished the latest Python (v3.12.0) parser.

added directories:
- python2_7_18
- python3_12_0

removed directory:
- tiny-python

It would be nice if someone created the directory structure discussed previously, but I don't want to modify it.
[Bring Python3 grammar up to date with current Spec grammar](https://github.com/antlr/grammars-v4/issues/2929#issuecomment-1720149617)


teverett:
@RobEin thanks!

PR #3806, opened 2023-10-31
Revert "[javascript] fix problem with continue/break/return statements #3772"

meraedit:
Reverts antlr/grammars-v4#3773
because https://github.com/antlr/grammars-v4/pull/3778 is a better fix and is already merged


teverett:
@meraedit  thanks!

PR #3811, opened 2023-11-04
[PlSql] Break out unary_logical_operation rule

marty-se:
To simplify processing of multiple unary logical operations within a unary_logical_expression, as well as to avoid ambiguity of NOT tokens within unary_logical_expression.

### DIFF
--- a/sql/plsql/PlSqlParser.g4
+++ b/sql/plsql/PlSqlParser.g4
@@ -6083,7 +6083,7 @@ logical_expression
     ;
 
 unary_logical_expression
-    : NOT? multiset_expression unary_logical_operation*
+    : NOT? multiset_expression unary_logical_operation?
     ;
 
 unary_logical_operation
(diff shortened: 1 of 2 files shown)


sql/plsql/PlSqlParser.g4
```diff
@@ -6083,7 +6083,11 @@ logical_expression
     ;
 
 unary_logical_expression
-    : NOT? multiset_expression (IS NOT? logical_operation)*
+    : NOT? multiset_expression unary_logical_operation*
+    ;
+
+unary_logical_operation
+    : (IS NOT? logical_operation)
```
kaby76, line 6090: Parentheses of `(IS NOT? logical_operation)` should be removed. They are unnecessary. Also, is there a test for this for the parse tree?

marty-se: Thanks, fixed.

Also, I realized that multiple unary logical operations were allowed before this PR - e.g. `foo IS NULL IS NOT NAN`, which doesn't seem right and is also rejected by an Oracle server. So I changed the `*` to a `?` in the `unary_logical_expression` rule. Does that look correct?

There was some example code containing a unary logical operation more or less implicitly, so I've added a file containing some more explicit examples.

kaby76 (approved):

PR #3818, opened 2023-11-13
[PlSql] Handle procedure_spec before variable_declaration

MartinKarlgrenIMI:
To avoid parsing procedure specs without arguments as variable declarations.


KvanTTT (approved):

teverett:
@MartinKarlgrenIMI  thanks!

PR #3825, opened 2023-11-15
[Snowflake] cast expr

gaoshihang:
CAST is a function, we can use cast_expr to represent it, So we can traverse the tree better.

### DIFF
--- a/sql/snowflake/SnowflakeParser.g4
+++ b/sql/snowflake/SnowflakeParser.g4
@@ -3636,6 +3636,7 @@ builtin_function
     | IDENTIFIER
     | FLATTEN
     | SPLIT_TO_TABLE
+    | CAST
     ;
 
 list_operator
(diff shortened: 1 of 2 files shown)


KvanTTT (changes requested):
Welcome to ANTLR grammars repository and thanks for the fix!

But please add/extend an example to the [examples](https://github.com/antlr/grammars-v4/tree/master/sql/snowflake/examples) that covers suggested changes.

gaoshihang:
Hi @KvanTTT , I have added an example to it in the select.sql file, but don't know if it is right.
Please help review, if there is any issue, i will keep updating to this PR. 
Thanks!

gaoshihang:
Hi @KvanTTT , can you help me review on this? please

KvanTTT (approved):

KvanTTT:
Everything looks good. Thanks!

gaoshihang:
Thanks @KvanTTT !

PR #3740, opened 2023-09-29
[dart2] Add ports, readme.md, clean up testing

kaby76:
This PR is a replacement for PR https://github.com/antlr/grammars-v4/pull/3587.

The TypeScript code added here is identical to that in https://github.com/antlr/grammars-v4/pull/3587. I added the other ports for the rest of the targets because they were easy.

The desc.xml was changed to utilize the new "globstar" patterns. Testing for all ports except Java are now a quick check of `examples/*.dart`, and Java the full test suite of `examples/**/*.dart`. A readme.md was added and the author of the Dart2 port cited.

With this merge, please close off https://github.com/antlr/grammars-v4/pull/3587.

PR #3831, opened 2023-11-16
Prerequisite for PR3740

kaby76:
This change alters some template scripts for testing. It is a prerequisite for [PR3740](https://github.com/antlr/grammars-v4/pull/3740).

The change specifically replaces the ["find" in Bash scripts](https://github.com/antlr/grammars-v4/blob/09c559937a745759279ecf73c740c67755a4bd30/_scripts/templates/Java/st.test.sh#L15) and the ["foreach ... Get-ChildItem ... -Recurse" in Powershell scripts](https://github.com/antlr/grammars-v4/blob/09c559937a745759279ecf73c740c67755a4bd30/_scripts/templates/CSharp/st.test.ps1#L13) with an OS- and shell-independent "globstar" tool called trglob.

We would like to specify the files to test thus: `examples/**/*.dart` (for fast targets like Java), and `examples/*.dart` (for targets that are slower).

Currently, if we wanted to test targets with different tests, we would have to partition the tests into several directories. With this change, we can have one directory but now test different inputs depending on a globstar path pattern.

The tool inputs Bash "globstar" patterns. In addition, for backward compatibility, any pattern that specifies a directory implies the contents of the directory and any sub-directories recursively are provided. So, the pattern `examples` or `examples/` or `examples/*` all imply all files recursively in the `examples/` directory. But, `examples/*.txt` does not recurse (assuming there is no directory named with extension `.txt`).

### Changes
* The tool was added to the local ".config/" set of tools for testing. Since the tool was only updated recently, I updated all the version numbers for the testing toolset.
* The templates for testing were updated to use the new tool "trglob".
* The dart2 grammar "desc.xml" was updated to test the new tool with the patterns I describe above.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,67 +3,67 @@
   "isRoot": true,
   "tools": {
     "trcaret": {
-      "version": "0.21.15",
+      "version": "0.21.16",
       "commands": [
         "trcaret"
       ]
     },
     "trcover": {
-      "version": "0.21.15",
+      "version": "0.21.16",
       "commands": [
         "trcover"
       ]
     },
     "trgen": {
-      "version": "0.21.15",
+      "version": "0.21.16",
       "commands": [
         "trgen"
       ]
     },
     "trglob": {
-      "version": "0.21.15",
+      "version": "0.21.16",
       "commands": [
         "trglob"
       ]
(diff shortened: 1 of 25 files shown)


teverett:
@kaby76 thanks

PR #3832, opened 2023-11-17
Reformat of the grammar

mike-lischke:
This patch shows how a grammar would be automatically reformatted and the options used for that.


teverett:
@mike-lischke thanks!

PR #3836, opened 2023-11-20
[Snowflake]Refactoring id and expr

MasterKuat:
This do not change parsing capabilities but change AST for better and easier inspection.
@kaby76: Is it possible to add an AST's test infrastucture ? Should be great


MasterKuat:
I think about something like this (in an exemple file ) : 
```select coalesce(A, -1); //TreeTest:LeafNode = Coalesce ; Path = /*/expr/function_call/list_function```
Some details :

- ```//``` is comment for this grammar in order to not interfear.
- ```TreeTest:``` is a command for test infrastructure asking for AST test
- ```LeafNode``` = AST branch must end with this kind of node
- ```Path``` = Part of branch where LeafNode is

And test could be : Assert(AST branch /*/expr/function_call/list_function/coalesce exist)

Indeed, this could be quite challenging for all grammar 🚀

kaby76:
I see. You are suggesting that we add to the input a self-checking assertion. Or alternatively an input file that describes the input and xquery that evaluates to a boolean for the expected tree. The point is to eliminate all the useless tree changes irrelevant to the test. Yes, this is possible.

MasterKuat:
Hi, @kaby76, @KvanTTT : is there a way to re-run only failed test ? (Timeout)

kaby76:
> Hi, @kaby76, @KvanTTT : is there a way to re-run only failed test ? (Timeout)

Not really. The best one can do is wait for a few hours so the servers are under less pressure.

Issue #3837, opened 2023-11-20
[Test Infrastructure]TypeScript and PHP tests do not failed where it's should


MasterKuat:
Hi @kaby76, seems TypeScript and PHP tests do not fail : https://github.com/antlr/grammars-v4/actions/runs/6927678183

kaby76:
This is not a bug, but expected behavior. [The desc.xml does not contain TypeScript nor PHP](https://github.com/MasterKuat/grammars-v4/blob/48419e74ed7dc0ea898c0a244772faffe8ede583/sql/snowflake/desc.xml#L4).

[A Github Action job is spawned for all targets.](https://github.com/antlr/grammars-v4/blob/4b728d96ef44575336829b8afbea81b0b3cdb824/.github/workflows/main.yml#L13-L15)  I do not see how not spawn a job apriori by reading the desc.xml and deciding to not do the spawn for the target.

There is no way I know of how to not spawn a Github Action for a specific target.

MasterKuat:
Hum..I see.
Thank you for pointing that.

I do not know github workflow but may be we could do something like this : 
1/ premain.yaml (with no matrix) = Get and cumulate <targets> from each desc.xml in each directory from modified file
2/ Run main.yml with matrix as a parameter from premain.yml 
(just an idea from a github noob)

kaby76:
A simple formula is described [here](https://www.kenmuse.com/blog/dynamic-build-matrices-in-github-actions/). I [applied that](https://github.com/antlr/grammars-v4/pull/3839) to the workflow. [I created a simple json output](https://github.com/antlr/grammars-v4/blob/f1ba932e863f78ece5b674b85a5c3f797a22a58a/.github/workflows/main.yml#L16) that is then used for [spawning only CSharp and Java "build-pwsh" jobs](https://github.com/antlr/grammars-v4/actions/runs/6932939936). So, the formula works.

I now only need to go through the desc.xml required by the changes in the PR to gather the targets needed to test.

MasterKuat:
@kaby76 Thank you ❤️

PR #3839, opened 2023-11-20
Fix for #3837

kaby76:


kaby76:
Not working, and I don't have admin privileges to just kill the damn actions.

kaby76:
The build.ps1 code is incorrect for Cpp. [The code uses a variable `$(MAKE)`, an unset variable](https://github.com/antlr/grammars-v4/blob/5b3f1b73e86588721f65b2601b8ccfb3923aec17/_scripts/templates/Cpp/st.build.ps1#L25). It should just use "make".

kaby76:
Apparently a Github Action job "matrix" cannot have an empty list, which is the result of updating clif. I will need to do something to not run into an error.

kaby76:
@teverett This PR is all set. I'll keep an eye out for any problems, but I don't expect any problems.

teverett:
@kaby76  thanks for this!

PR #3848, opened 2023-11-27
fix: attach rule in ICalendar.g4 was not handling binary properly

khawarizmus:
This PR fixes the attache rule to properly parse the `ATTACH` property when it it's value is binary.

The following example now works
```txt
ATTACH;FMTTYPE=text/plain;ENCODING=BASE64;VALUE=BINARY:VGhlIH
 F1aWNrIGJyb3duIGZveCBqdW1wcyBvdmVyIHRoZSBsYXp5IGRvZy4
```
before this the binary value was interpreted as a url


KvanTTT:
Also, please add/extend an example to the [examples](https://github.com/antlr/grammars-v4/tree/master/icalendar/examples) dir that covers suggested changes.

teverett:
@khawarizmus  could you fix the merge conflicts please?

khawarizmus:
@KvanTTT @teverett this pr is ready for review again. Waiting for additional comments from @KvanTTT

icalendar/ICalendar.g4
```diff
@@ -306,16 +306,15 @@ daylightc
 
 // 3.8.1.1 - Attachment
 attach
-    : k_attach attachparam* (
-        COL uri
-        | SCOL k_encoding ASSIGN k_base D6 D4 SCOL k_value ASSIGN k_binary COL binary
-    ) CRLF
-    ;
+   : k_attach attachparam* COL (binary | uri) CRLF
```
KvanTTT, line 309: Please use 4 spaces for indents instead of 3 as it's used for the entire grammar.

KvanTTT (changes requested):

icalendar/ICalendar.g4
khawarizmus: Done 👍

khawarizmus:
Fixed the spaces @KvanTTT @teverett

teverett:
@khawarizmus thanks!

PR #3849, opened 2023-11-27
Fix for 3847 -- fix build.

kaby76:
The following targets fail the build:

* asm/ptx/ptx-isa-1.0 with Cpp: NAN is a symbol conflict for Cpp--apparently now for MacOS. ./icon ./lark .sieve ./smtlibv2 ./wavefront with Go target--symbol conflict apparently now with "start" as a parser rule.
* ./scala with Go target--timeout. This grammar is super slow. Go target should be removed from desc.xml to indicate it doesn't work.
* elixir with PHP target--stack overflow. The PHP has logical errors in the code. I tried to address some of the problems, but it was too much, and needed to work on other things. Fix for #12 -- Allowed memory size of 134217728 bytes exhausted. antlr-php-runtime#34

### DIFF
--- a/apt/apt.g4
+++ b/apt/apt.g4
@@ -3,7 +3,7 @@
 grammar apt;
 
 //  productions
-record: commented=commenterR? rType=TypeR WSS options=optionsR? uri=uriR WSS distribution=wordWithDash components=componentsR WSS? EOF;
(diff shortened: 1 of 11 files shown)


teverett:
@kaby76  thanks!

PR #3874, opened 2023-12-06
[build] Remove "build-bash" testing

kaby76:
I'm removing the "build-bash" because it is duplicate of "build-pwsh" testing. This is in response to the numerous network and Github job queue issues in Github Actions. This change will result in half the number of jobs being executed, and lower costs to maintain this repo.


teverett:
@kaby76  thanks!

Issue #3867, opened 2023-12-02
[build] Concurrency "cancel-in-progress" might need tweaking


kaby76:
Looks like the ["group" for concurrency](https://github.com/antlr/grammars-v4/blob/master/.github/workflows/main.yml#L9) control might need to be adjusted. Looks like [the build was canceled prematurely](https://github.com/antlr/grammars-v4/actions/runs/7063469205/job/19229541478).

PR #3875, opened 2023-12-06
[build] Enforce coding standard for grammars.

kaby76:
This PR changes several things in the build:
1. It adds a check to the static tests to establish a grammar follows the coding standard format for this repo. This check occurs during the normal PR development process. When a PR is submitted, the grammar coding standard is tested. The developer can go ahead and fix the warnings that arise during a build.
2. It performs a reformat and commit for grammars for changed grammars. When a PR is merged, the coding standard format is enforced. The grammars are changed automatically and then checked into the "master" branch. If the .g4's are fine, nothing gets checked into "master". This avoids the problem of someone having to monitor the build and correct the problem later on. This commit happens before performing a test of the grammar against targets.
3. A naive check is included to raise an error if a PR contains deleted "antlr-format" lines. This prevents someone from mucking around with grammars that already follow the standard. If the grammar is new, it is preferred that the grammar not contain the "antlr-format" strings, but it is allowed. The grammar will be modified with the format strings added.
4. It fixes https://github.com/antlr/grammars-v4/issues/3867, where quick successive PR merges result in previous Github Actions canceled.

This change parallelizes the static checks performed on a build. A summary of the check is outputted using [a Github Action workflow command for warning](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-warning-message). The workflow steps are ordered so that the reformat occurs prior to the grammar tests.

Nota bene: We will need to update the Wiki so that it reflects a formatting standard. https://github.com/antlr/grammars-v4/wiki#is-there-a-coding-standard-for-antlr4-grammars


kaby76:
@mike-lischke I'm adding a step for your formatting tool to check if a grammar "fits" the coding standard for grammars. The test simply does the format, then a `git diff` to find if it's changed from what was in the PR. It then annotates the output. For example, here.  https://github.com/antlr/grammars-v4/actions/runs/7119154321 , and scroll down to the "Annotations" section. I could provide better line information in the output, such as "Rule such-and-such doesn't fit the coding standard", but that would be a little more involved. One can simply go to the output from the build and look at the diffs. https://github.com/antlr/grammars-v4/actions/runs/7119154321/job/19383632195#step:15:419

mike-lischke:
I was just confronted with a similar problem in the vite repository. Opened a PR there with only minimal changes and the verification asked me to run my patch through prettier. However they use pnpm which would require me to install that manager just for this check, which is a pain. So, while I believe that this repo should enforce the formatting standard, it should do that in a way that does not force the PR creator to install any additional software, especially if he/she is not working with Node.js usually.

Maybe apply formatting automatically (if that's possible in GH)?

mike-lischke:
This PR is still a draft @kaby76 :-)

kaby76:
> This PR is still a draft @kaby76 :-)

Yes, it's supposed to be because I am not finished. The last commit was to check whether the workflow is syntactically correct. I will reread the code for a final pass.

kaby76:
Ready for review.

mike-lischke:
Oh, to me it looked as if you had merged it already, no idea why. Do you want me to review? I'm afraid shell scripts are greek to me.

teverett:
@kaby76  @mike-lischke  good to merge this?

mike-lischke:
I would if I could, but that's not my area of expertise.

kaby76:
> I would if I could, but that's not my area of expertise.

I think we should just agree that doing a reformat of a grammar on a PR merge should be okay. I'll keep an eye out on the builds to make sure anything bad happens gets corrected ASAP.

teverett:
@kaby76  there is a merge conflict.

kaby76:
> @kaby76 there is a merge conflict.

I'm going to double-check this change on another repo,  https://github.com/kaby76/test-3875, to make sure it is going to work as intended.

mike-lischke:
@kaby76 Do you also check that the formatting options in the grammar are not changed? Otherwise the grammar could be reformatted to a different style. There must be no other options than what exists now.

kaby76:
> @kaby76 Do you also check that the formatting options in the grammar are not changed? Otherwise the grammar could be reformatted to a different style. There must be no other options than what exists now.

The formatter is called with no args other than the file names. https://github.com/kaby76/grammars-v4/blob/20dc8d4fbfeb7c07846d9f3d2e3ebbc2a2c00b7c/_scripts/reformat.sh#L337.

I don't know whether the embedded strings for antlr-format in Antlr comments are read by the formatter in reformatting. Eg abbLexer https://github.com/kaby76/grammars-v4/blob/20dc8d4fbfeb7c07846d9f3d2e3ebbc2a2c00b7c/abb/abbLexer.g4#L1-L3.

mike-lischke:
They are read and override any option given on the command line (and the default options). So, if a patch changes them, the next reformatting will use the changed settings. So it's probably wise to check that these lines always stay the same, unless a change of them is required (in which case the change should be made in all grammars).

kaby76:
> They are read and override any option given on the command line (and the default options). So, if a patch changes them, the next reformatting will use the changed settings. So it's probably wise to check that these lines always stay the same, unless a change of them is required (in which case the change should be made in all grammars).

I will need to add a config.json for the antlr-format call for this repo:  
[config.json](https://github.com/antlr/grammars-v4/files/13692703/config.json). The file was derived from the abb grammar. `-c config.json` will need to be used because the "default" format isn't the standard for the repo. Without the `-c config.json` option, three empty lines get added before the `grammarSpec` for the new grammar.

I ran the formatter over the entire repo with the `-c config.json` option, and got a number of reformattings. (`for i in `find . -name desc.xml`; do pushd `dirname $i`; antlr-format -c ~/test/config.json *.g4; popd; done`) See this diff: 
[out.txt](https://github.com/antlr/grammars-v4/files/13692710/out.txt)

If the embedded option strings change in a PR, there's not a lot I can do using sed, awk, ed, diff, etc. These tools are not suitable to modify the antlr-format option strings. If we ever need to change the standard, all 500 .g4's, or readjust the `$antlr-format` strings that were changed in a PR, we'll need a tool to wipe out the old `$antlr-format` strings and reformat with the new config.json.

mike-lischke:
Ah, good point. New grammars have no formatting options yet (usually), so yes, you need to run the formatter with explicit options.

For checking the embedded options for changes: can't we just check the diff if it contains the format introducer (`antlr-format`) and reject the grammar if that's the case?

kaby76:
> For checking the embedded options for changes: can't we just check the diff if it contains the format introducer (`antlr-format`) and reject the grammar if that's the case?

I could I suppose. I'd have to do the check before [this call](https://github.com/kaby76/grammars-v4/blob/a46cbc5531dcafda32cbcd694cdb8473012affc3/.github/workflows/main.yml#L42). (Alternatively, I could do it in the reformat.sh script.)

A naive check would be something like `git diff . $Befire $After | fgrep '$antlr-format'; if [ $? -eq 0 ]; then echo "fail"; fi`). This would do a "reject" if the commit contained any `$antlr-format` changes.

The problem then becomes: How would I create a PR with a new coding standard?

mike-lischke:
That would be a very special situation and this check would have to be disabled for this single PR. But what about new grammars? They also have changed formatting options.

Maybe it's better to rely on the review process, instead of creating a complicated automatic solution? One more thing to watch out for when reviewing a PR. It's pretty easy to spot if a formatter option line is changed.

kaby76:
I'd rather have the enforced coding standard than rely only on code reviews. There's no official list of checks to do and it's best to automate things. This PR is a great idea (thank you @mike-lischke !!) and for the better.

I can write a tool to erase the option strings with the Trash toolkit prior to a change in the coding standard. No hurries for that but it is easy to write. I'll add in the naive check since that will work mostly.

kaby76:
@teverett It's now all set.

teverett:
@kaby76 @mike-lischke  thanks!

PR #3876, opened 2023-12-06
added MIT license

OleksiiKovalov:
Added text of MIT license and creator info to the PostgreSQL grammar file


KvanTTT (changes requested):
It's also worth mention to have a license header in [PostgreSQLLexer.g4](https://github.com/antlr/grammars-v4/blob/master/sql/postgresql/PostgreSQLLexer.g4) and target files (Java, C#, etc.)

KvanTTT (approved):

teverett:
@OleksiiKovalov  thanks!

PR #3879, opened 2023-12-10
syntax bug fix according to oracle 21 pl/sql language reference

langio:
two syntax bug fixs
1. varray_type_def, add ? after VARYING
2. forall_statement, replace sql_statement with data_manipulation_language_statements and move it to the end


KvanTTT (changes requested):
Welcome to ANTLR grammars repository and thanks for the contribution.

Please add/extend an example that covers suggested changes to the [examples](https://github.com/antlr/grammars-v4/tree/master/sql/plsql/examples) subdirecotry.

kaby76:
Please rebase/fast forward to pick up https://github.com/antlr/grammars-v4/pull/3883.

langio:
I should have put these two bugs on separate PRs that would be clearer, and add an example.

PR #3884, opened 2023-12-11
[Snowflake]Refacto expr & id_

MasterKuat:
- House keeping on id_ and related rules (WIP)
- Reordering expr for better AST


teverett:
@MasterKuat thanks

Issue #3887, opened 2023-12-12
[build] Numerous warnings in Annotations section of a build


kaby76:
There are numerous warnings for the Go and Cpp targets. See https://github.com/antlr/grammars-v4/actions/runs/7183158267

```
Restore cache failed: Dependencies file is not found in /home/runner/work/grammars-v4/grammars-v4. Supported file pattern: go.sum

...

'antlr4::dfa::Vocabulary::EMPTY_VOCABULARY': Use the default constructor of Vocabulary instead.

...

'initializing': conversion from '_Ty' to '_Ty1', possible loss of data

etc.
```

kaby76:
* Cpp runtime contains an error. At least for GNU g++ 9.4.0, [this line](https://github.com/antlr/antlr4/blob/d25d421780286e3de0714831a5f82eca59234249/runtime/Cpp/runtime/src/Vocabulary.h#L23) is unnecessary because static fields are defining occurrences of itself. Commenting out this line results in a program to link perfectly. In addition, adding this code to a driver `auto foo = new antlr4::dfa::Vocabulary(); foo->EMPTY_VOCABULARY;`, which is an applied occurrence of the identifier EMPTY_VOCABULARY, causes the warning to be outputted by the compiler. And again, the program links and works perfectly fine.

PR #3888, opened 2023-12-12
[build, Go target] Partial fix for #3887

kaby76:
Fix as reported here: https://github.com/actions/setup-go/issues/427#issuecomment-1731969319


teverett:
@kaby76 thanks

Issue #3891, opened 2023-12-14
[build] TypeScript broken


kaby76:
Testing of the TypeScript target is broken.

Tests fail with import module problem.
```
$ bash run.sh ../hw-examples/alter_operator.sql
Error [ERR_MODULE_NOT_FOUND]: Cannot find module 'C:\msys64\home\Kenne\issues\g4-3890\sql\plsql\Generated-TypeScript-hw\PlSqlLexerBase' imported from C:\msys64\home\Kenne\issues\g4-3890\sql\plsql\Generated-TypeScript-hw\PlSqlLexer.js
Did you mean to import ../PlSqlLexerBase.js?
    at finalizeResolution (node:internal/modules/esm/resolve:255:11)
    at moduleResolve (node:internal/modules/esm/resolve:908:10)
    at defaultResolve (node:internal/modules/esm/resolve:1121:11)
    at nextResolve (node:internal/modules/esm/hooks:865:28)
    at resolve (C:\Users\kenne\AppData\Roaming\npm\node_modules\ts-node\dist\child\child-loader.js:15:125)
    at nextResolve (node:internal/modules/esm/hooks:865:28)
    at Hooks.resolve (node:internal/modules/esm/hooks:303:30)
    at handleMessage (node:internal/modules/esm/worker:196:24)
    at Immediate.checkForMessages [as _onImmediate] (node:internal/modules/esm/worker:138:28)
    at process.processImmediate (node:internal/timers:478:21) {
  code: 'ERR_MODULE_NOT_FOUND',
  url: 'file:///C:/msys64/home/Kenne/issues/g4-3890/sql/plsql/Generated-TypeScript-hw/PlSqlLexerBase'
}
```

6 days ago, MS released a new package for typescript and things haven't worked well since. Unfortunately I can't find a rollback state that does work.

I am able to get the target to work if I add a ".js" suffix for the file import in the generated .ts file. But that is not a good solution. Filed an Issue in the Antlr4 repo as to why some import have ".js" and others do not. I'm not sure whether there are settings in the tsconfig.json to fix the problem. 

https://github.com/antlr/antlr4/issues/4491

kaby76:
Apparently, this can be fixed if the entire code is placed in a package. The inconsistency in the tool has been fixed. https://github.com/antlr/antlr4/pull/4492

mike-lischke:
[...]

PR #3892, opened 2023-12-14
[build] Remove TypeScript target testing temporarily.

kaby76:
See https://github.com/antlr/grammars-v4/issues/3891 https://github.com/antlr/antlr4/issues/4491


teverett:
@kaby76 thanks

Issue #3900, opened 2023-12-19
[build] remove auto reformat


kaby76:
The build failed due to a git commit with nothing added. It should not fail. https://github.com/antlr/grammars-v4/actions/runs/7261930105/job/19784088145#step:6:14

```
$ git commit -m "Reformat to coding standard."
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean
12/19-08:56:05 ~/issues/g4-3900
$ echo $?
1
```

The return value is currently not ignored because it is the last step in the "run".

***We need to undo the code for reformatting a grammar. Instead, we can just do a check and pass on a warning if the grammar does fit antlr-format standard. People can ignore the warning since it doesn't affect the build success.***

PR #3903, opened 2023-12-20
[Oracle] Support create table with NO MEMOPTIMIZE

ilyasahsan123:

PR #3904, opened 2023-12-20
[build] Fix for #3900--remove auto reformat

kaby76:
@teverett @KvanTTT 

This change is important. It removes the code to reformat grammars and auto check-in those changes. This is what @KvanTTT [wanted](https://github.com/antlr/grammars-v4/discussions/3816#discussioncomment-7911861). Instead antlr-format is run as a check. Aside, the code has a bug [here](https://github.com/antlr/grammars-v4/blob/aff0126d0aff25ecfcdf7d83ca0727f74ee5ce26/.github/workflows/main.yml#L53) in that when "git commit" doesn't contain any changes, the program returns 1, which causes the "setup" job to fail, and all tests for the grammar to not run because [they depend on "setup"](https://github.com/antlr/grammars-v4/blob/aff0126d0aff25ecfcdf7d83ca0727f74ee5ce26/.github/workflows/main.yml#L78).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -51,6 +51,7 @@ jobs:
         git config --local user.name "github-actions[bot]"
         git add --all
         git commit -m "Reformat to coding standard."
+        echo "Committed reformatted code."
     - name: Push changes
       # git push changes to "master" branch.
       # make sure to uncheck "Require a pull request before merging" in gh actions, "Branches" section.
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,40 +28,6 @@ jobs:
       shell: bash
       run: |
         dotnet tool restore
-    - name: Check grammar coding standards
-      continue-on-error: false
-      run: |
-        if [ "${{github.event_name}}" == "pull_request" ]; then
-          Before="${{github.event.pull_request.base.sha}}"
-          After="${{github.event.pull_request.head.sha}}"
-        else
-          Before="${{github.event.before}}"
-          After="${{github.event.after}}"
-        fi
-        # check for bad coding standard in grammar by doing a reformat.
-        bash _scripts/reformat.sh -f diff $Before $After
-    - name: Commit changes
-      # git commit reformatting changes if any, but only if on master branch.
-      # If there were no changes, this is a no-op.
-      if: github.ref == 'refs/heads/master'
-      continue-on-error: false
-      shell: bash
-      run: |
-        git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
-        git config --local user.name "github-actions[bot]"
-        git add --all
-        git commit -m "Reformat to coding standard."
-        echo "Committed reformatted code."
-    - name: Push changes
-      # git push changes to "master" branch.
-      # make sure to uncheck "Require a pull request before merging" in gh actions, "Branches" section.
-      uses: ad-m/github-push-action@master
-      if: github.ref == 'refs/heads/master'
-      continue-on-error: false
-      with:
-        github_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


kaby76:
@teverett This PR should be committed ASAP because the builds are failing in "setup". See my comment [here](https://github.com/antlr/grammars-v4/pull/3903#issuecomment-1864763980). There is a problem with the "badge" showing "success" but the Github Actions badge link is bad, and always shows "ok". I'll have a fix for that shortly.

KvanTTT (commented):
Can't we use [`-allow-empty`](https://git-scm.com/docs/git-commit/2.31.0#Documentation/git-commit.txt---allow-empty) (or also `--allow-empty-message`) flag to force `git commit` always return `0`?

Also, why does the situation with no files to commit occur?

kaby76:
> Can't we use [`--allow-empty`](https://git-scm.com/docs/git-commit/2.31.0#Documentation/git-commit.txt---allow-empty) (or also `--allow-empty-message`) flag to force `git commit` always return `0`?

I will need to test this. My concern is that git will create an empty commit, not a "no-op". I don't think we want a ton of empty commits following the commit of the changes for the PR every time a PR is merged.

> Also, why does the situation with no files to commit occur?

`git add --all-files` returns 0 regardless of whether are are files nor not that are modified or new. That command is executed [here](https://github.com/antlr/grammars-v4/blob/2e2844eb1ffaeb7423e1920bb2eeb948a03c6624/.github/workflows/main.yml#L52).

The next line, `git commit -m "Reformat to coding standard."`, [here](https://github.com/antlr/grammars-v4/blob/2e2844eb1ffaeb7423e1920bb2eeb948a03c6624/.github/workflows/main.yml#L53C9-L53C53), returns 0 for a commit that has files added. If there are no files added, `git` returns 1. It's not something I think is correct, but that's due to whoever wrote the program, believing that a `git commit` without any files--a "no-op"--is an "error". But, it doesn't matter.

The reason [the github actions "run"](https://github.com/antlr/grammars-v4/blob/2e2844eb1ffaeb7423e1920bb2eeb948a03c6624/.github/workflows/main.yml#L49C8-L53) is because that's how Bash works: the last command of a script sets `$?`, and Bash returns that value. 

[The step for the commit](https://github.com/antlr/grammars-v4/blob/2e2844eb1ffaeb7423e1920bb2eeb948a03c6624/.github/workflows/main.yml#L43C10-L53) fails because that's how Github Action steps work: the result of the shell execute is the result of the step. That's why it fails.

KvanTTT:
I got it. My suggestion about `--allow-empty` is bad, but the idea of printing the message `Committed reformatted code` it also not correct. When `git commit` returns `1` it means no commit actually was performed. I suggest printing different messages that depend on the status of execuring `git commit`.

Moreover, I'm against the idea of any automatic commits because they spoil git history, git blame and in fact they are our of control. See [my message in corresponding topic](https://github.com/antlr/grammars-v4/discussions/3816#discussioncomment-7911861).

kaby76:
I can appreciate having a proper "git blame", but that's the only thing I agree with you regarding auto-commits.

I'll take the conversation over to https://github.com/antlr/grammars-v4/discussions/3816 .

kaby76:
I'm undoing the git commit code. Instead, I'm just adding the reformat as a check. It's highly unlikely I could write the code to spoof the UID of the PR originator, so yes, git blames won't identify who made an (awful) change.

kaby76:
All set when the build completes.

kaby76:
A reminder on the fact that the builds keep failing, and they will keep failing until this PR is merged. https://github.com/antlr/grammars-v4/actions/runs/7338971117

Screenshot (77)

KvanTTT (approved):

KvanTTT:
Thanks for this fix!

PR #3909, opened 2023-12-22
fetch_statement bug fix

langio:
Add collection_expression and variable_or_collection rule that used by fetch_statement to resolve scenarios where the variable_name is a collection


sql/plsql/PlSqlParser.g4
```diff
@@ -6310,6 +6315,10 @@ implicit_cursor_expression
     : SQL PERCENT_BULK_EXCEPTIONS ('.' COUNT | '(' expression ')' '.' (ERROR_INDEX | ERROR_CODE))
     ;
 
+collection_expression
+    : collation_name ('(' expression ')' ('.' general_element_part)*)
```
KvanTTT, line 6319: Outer parentheses look redundant:

```suggestion
    : collation_name '(' expression ')' ('.' general_element_part)*
```

KvanTTT (changes requested):

sql/plsql/PlSqlParser.g4
langio: yes, i have fixed it.

KvanTTT (approved):

PR #3911, opened 2023-12-23
Updated type hinting

Xenios91:
Type hinting in VBA should work when its placed after the digit as well.

Documentation here.

https://learn.microsoft.com/en-us/dotnet/visual-basic/programming-guide/language-features/data-types/type-characters

### DIFF
--- a/vba/examples/example09_typeHints.bas
+++ b/vba/examples/example09_typeHints.bas
@@ -1,0 +1,7 @@
+Sub TypehintsDemo() 
+Dim Number
+ Number = #1
+ OtherNumber = &10
+ Debug.Print Number
+ Debug.Print OtherNumber
+End Sub
--- a/vba/vba.g4
+++ b/vba/vba.g4
@@ -418,7 +418,7 @@ killStmt
     ;
 
 letStmt
-    : (LET WS)? implicitCallStmt_InStmt WS? (EQ | PLUS_EQ | MINUS_EQ) WS? typeHint? valueStmt
+    : (LET WS)? implicitCallStmt_InStmt WS? (EQ | PLUS_EQ | MINUS_EQ) WS? typeHint? valueStmt typeHint?
     ;
 
 lineInputStmt
--- a/vba/vba.g4
+++ b/vba/vba.g4
@@ -1877,7 +1877,7 @@ HEXLITERAL
     ;
 
 SHORTLITERAL
-    : (PLUS | MINUS)? DIGIT+ ('#' | '&' | '@')?
+    : (PLUS | MINUS)? DIGIT+
     ;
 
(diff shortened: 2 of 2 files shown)


teverett:
@Xenios91  thanks!

Beakerboy:
@Xenios91, do you have a source that shows that it is okay to place the double type hint ('#') BEFORE the digit? The VBA 7.1 spec only defines the hint occurring after the identifier.

MS-VBAL 7.1, section 3.3.2 defines
```
FLOAT = (floating-point-literal [floating-point-type-suffix]) / (decimal-literal floating-point-type-suffix)
floating-point-type-suffix = '!' / '#' / '@'
```

However, the octal and hex literals do have the '&' before the digits.

Lety me know if this changed between versions. My VBA IDE flags `Number = #1` as a syntax error.

Xenios91:
> @Xenios91, do you have a source that shows that it is okay to place the double type hint ('#') BEFORE the digit? The VBA 7.1 spec only defines the hint occurring after the identifier.
> 
> MS-VBAL 7.1, section 3.3.2 defines
> 
> ```
> FLOAT = (floating-point-literal [floating-point-type-suffix]) / (decimal-literal floating-point-type-suffix)
> floating-point-type-suffix = '!' / '#' / '@'
> ```
> 
> However, the octal and hex literals do have the '&' before the digits.
> 
> Lety me know if this changed between versions. My VBA IDE flags `Number = #1` as a syntax error.

I dont, this is just once again me finding things that work when ran but may not be up to spec.

Beakerboy:
If you feel these “improper” code formats deserve to be in the grammar, would it make sense to push them to their own separate context?

for example:
```
letStmtNew
    : (LET WS)? implicitCallStmt_InStmt WS? (EQ | PLUS_EQ | MINUS_EQ) WS? valueStmt typeHint?    #letStmt
    | (LET WS)? implicitCallStmt_InStmt WS? (EQ | PLUS_EQ | MINUS_EQ) WS? typeHint? valueStmt typeHint?    #letStmtError
    ;
```

if the instances of letStmt are replaced with letStmtNew, ANTLR with autogenerate visitor and listener functions of the name letStmat that still only operate on the correct version, but a new set of functions will work on the weird versions.

Again though, the more I learn about ANTLR, the more I feel there are ways a parser can open an invalid file with the old grammar and fix it on the fly, producing a parse tree that is valid.

Beakerboy:
A hash sign is legal to be placed before an integer or hex digits to indicate a marked file number. However, a marked file number is not assigned as is done in this example, it is produced by the file open statement. So I think there is a little of both here…a number sign is valid, but not as a type hint, and this syntax of assigning ‘#1’ to a variable is a syntax error.

PR #3916, opened 2023-12-30
Support Create Function with Default Collation in Oracle

ilyasahsan123:


ilyasahsan123:
hi @KvanTTT, Could you please take a look? Thanks

KvanTTT (approved):

PR #3924, opened 2024-01-11
Support 'ANY' operation in logical expression

langio:


KvanTTT (approved):

KvanTTT (commented):
I'm not sure if MR title relates to the changed content. Could you clarify why allowing expression closure makes it's possible to "Support 'ANY' operation in logical expression"?

teverett:
@KvanTTT  good to merge this?

KvanTTT:
Not yes, I'm awaiting for response.

langio:
The modification is based on the following two reasons:
    1. The original syntax only supports 'ANY' appearing at the end of the condition， like 'SELECT 1 FROM DUAL WHERE 1=1 AND 1 = ANY(1, 2, 3)'(matched the quantified_expression rule), when the two conditons swap positions, the parser will report an error. 
    2. If modify the quantified_expression rule to support the case mentioned above, it needs to support 'ANY' appearing in different positons of the condition, appearing in any nested parentheses, and any number of occurrences of 'ANY', which happens to be the expressive power of the rule logical_expression. The quantified_expression rule is a part of the logical_expression rule indirectly, and it can be modifited to meet the scenarios mentioned above where 'ANY' appears in various positions. So, i think that modification may be a solution.

teverett:
@langio thanks!

PR #3927, opened 2024-01-12
[SNOWFLAKE]fix small bug

MasterKuat:
- Fix CHANGES clause.
- NULLIF and NVL keyword are already allowed as object name by the "binary_builtin_function" rule
- Refactor add column


teverett:
@MasterKuat thanks

PR #3929, opened 2024-01-15
snowflake - alt cast

mlorek:


kaby76:
The problem is it took too long. There's a 5 minute time out. https://github.com/antlr/grammars-v4/actions/runs/7533854521/job/20507172652#step:24:152

The easiest thing to do is to just yank the target from the desc.xml `<targets>`. https://github.com/antlr/grammars-v4/blob/caf66c3d8a4657a4871a539fa8de9c9a6b5a8b1b/sql/snowflake/desc.xml#L4

The build script captures the output here. https://github.com/antlr/grammars-v4/blob/caf66c3d8a4657a4871a539fa8de9c9a6b5a8b1b/_scripts/templates/Python3/st.test.ps1#L46
But, it should output parse.txt somewhere, and that's probably getting not done.

mlorek:
> The problem is it took too long. There's a 5 minute time out. https://github.com/antlr/grammars-v4/actions/runs/7533854521/job/20507172652#step:24:152
> 
> The easiest thing to do is to just yank the target from the desc.xml `<targets>`.
> 
> https://github.com/antlr/grammars-v4/blob/caf66c3d8a4657a4871a539fa8de9c9a6b5a8b1b/sql/snowflake/desc.xml#L4
> 
> The build script captures the output here.
> 
> https://github.com/antlr/grammars-v4/blob/caf66c3d8a4657a4871a539fa8de9c9a6b5a8b1b/_scripts/templates/Python3/st.test.ps1#L46
> 
> 
> But, it should output parse.txt somewhere, and that's probably getting not done.

thanks for checking!
can the timeout be controlled per target?

teverett:
Is this good to merge?

kaby76:
@teverett Yes, it's good to go.

> can the timeout be controlled per target?

Sorry, it can't. But, you can create a subset of the test inputs. E.g.,
```
   <antlr-version>^4.10</antlr-version>
   <targets>CSharp;Cpp;Dart;Go;Java;JavaScript;PHP;Python3;TypeScript</targets>
   <test>
      <name>fast</name>
      <targets>CSharp;Cpp;Dart;Java;JavaScript</targets>
      <inputs>examples</inputs>
   </test>
    <test>
      <name>slow</name>
      <targets>PHP;Python3;TypeScript</targets>
      <inputs>examples/c*.sql</inputs>
   </test>
</desc>
```

But, the TypeScript target for the grammar doesn't work for some reason.

teverett:
@mlorek thanks!

PR #3933, opened 2024-01-19
Update C.g4

CarlChenXueYi:
Structure can be empty


KvanTTT (changes requested):
Please [add/extend a test](https://github.com/antlr/grammars-v4/tree/master/c/examples) that covers suggested changes.

PR #3936, opened 2024-01-22
[Trino]: Sync new language features from trino 435

yonatand:
largest feature added was inline routine,
added a simple example.


sql/trino/TrinoParser.g4
```diff
@@ -492,7 +515,7 @@ primaryExpression
     | ARRAY_ LSQUARE_ (expression (COMMA_ expression)*)? RSQUARE_                                                 # arrayConstructor
     | value = primaryExpression LSQUARE_ index = valueExpression RSQUARE_                                         # subscript
     | identifier                                                                                                  # columnReference
-    | base_ = primaryExpression DOT_ fieldName = identifier                                                       # dereference
+    | base = primaryExpression DOT_ fieldName = identifier                                                        # dereference
```
kaby76, line 518: You cannot use "base" because this causes the C# target to fail to build. It is a reserved word in C#. Rename it back to "base_".

kaby76 (approved):

teverett:
@yonatand  thanks!

PR #3939, opened 2024-01-23
[PlSql] Correctly parse package init blocks with simple statements

MartinKarlgrenIMI:
Before this commit, the following package init block was parsed as a variable declaration:

```
    create or replace package body MyPackage is
    begin
      OtherPackage.DoSomething;
    end;
```

... because variable_declaration is:

`: identifier CONSTANT? type_spec (NOT NULL_)? default_value_part? ';'`

so `identifier` would match "begin" and `type_spec` would match "OtherPackage.DoSomething" (because "BEGIN" is a non-reserved keyword which is allowed as an identifier in some contexts).

Making the package_obj_body match non-greedy fixes this.

### DIFF
--- a/sql/plsql/PlSqlParser.g4
+++ b/sql/plsql/PlSqlParser.g4
@@ -667,7 +667,7 @@ create_package_body
     : CREATE (OR REPLACE)? (EDITIONABLE | NONEDITIONABLE)? PACKAGE BODY (schema_object_name '.')? package_name (
         IS
         | AS
-    ) package_obj_body* (BEGIN seq_of_statements (EXCEPTION exception_handler+)?)? END package_name?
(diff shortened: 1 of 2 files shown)


MartinKarlgrenIMI:
Not sure if there's a better solution than making the `package_obj_body` match non-greedy...

KvanTTT (approved):

KvanTTT (commented):
> Not sure if there's a better solution than making the package_obj_body match non-greedy...

Yes, I also don't like this solution. I'll try to think about better solution.

KvanTTT:
Ok, makes sense. Could you please rebase your on master and resolve conflicts?

MartinKarlgrenIMI:
> Ok, makes sense. Could you please rebase your on master and resolve conflicts?

Sure, done.

KvanTTT (approved):

PR #3940, opened 2024-01-24
Whitespace is also optional.

Beakerboy:
Previously, the CLASS token was changed to optional, but the whitespace was left out. I’ve included an example form module for validation.


teverett:
@Beakerboy  thanks!

PR #3951, opened 2024-01-30
[Snowflake] Updated Grammar for stages and stage paths; added support for virtual columns

malcolm-green:
- refactored stages as there is no syntactic difference between an external stage and a named staged
- added support for virtual columns in Create Table ddl's
- Added LINEAR keyword for clustered tables
- Added d and s variables to schema_name to be consistent with object_name
- Added package to POM so that the Java code generates in a package that can be overridden


mlorek (approved):

teverett:
@malcolm-green  thanks!

PR #3953, opened 2024-01-31
[VB6] Fix two corner cases

Bluenaxela:
In some VB6 code I was dealing with I found two corner cases that are apparently valid in real VB6 that didn't parse successfully in this grammar.

1) A "DO .... LOOP WHILE" type of loop wasn't allowed to be empty, when it should be
2) ELSEIF lines don't require a newline before the block, whitespace will do.

A simple example file is included.


teverett:
@Bluenaxela  thanks!

PR #3955, opened 2024-02-01
update SQLiteLexer.g4 to support both 0x and 0X for hexadecimal value

zxt5:
Update the SQLiteLexer.g4 to support parsing hexadecimal integers beginning with either '0x' or '0X'.


KvanTTT (changes requested):

But your change looks irrelevant since the parser contains enabled [`caseInsensitive = true;`](https://github.com/antlr/antlr4/blob/dev/doc/options.md#caseinsensitive) at the header.

I'm closing.

PR #3958, opened 2024-02-06
Rewrote JSX element parsing to address various issues.

aleppard:
This PR rewrites the JSX element parser by adding modes for JSX opening elements, JSX closing elements and JSX children (and removing the ATTVALUE mode that is no longer required). Note that the previous HTML TAG mode wasn't actually being called.

This fixes a few issues that I came across as documented in https://github.com/antlr/grammars-v4/issues/392

Including:

1. The regular expression no longer matches opening and closing JSX elements:

```return (<div><div></div></div>);```

2. Javascript keywords are no longer matched in HTML text:

```<div>from</div>```

### DIFF
--- a/javascript/jsx/JavaScriptParser.g4
+++ b/javascript/jsx/JavaScriptParser.g4
@@ -6,6 +6,7 @@
     added ECMAScript 6 support, cleared and transformed to the universal grammar.
  * Copyright (c) 2018 by Juan Alvarez (contributor -> ported to Go)
  * Copyright (c) 2019 by Student Main (contributor -> ES2020)
+ * Copyright (c) 2024 by Andrew Leppard (www.wegrok.review)
  *
  * Permission is hereby granted, free of charge, to any person
  * obtaining a copy of this software and associated documentation
@@ -71,7 +72,7 @@ statement
     ;
 
 block
-    : '{' statementList? '}'
+    : OpenBrace statementList? CloseBrace
     ;
 
 statementList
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: javascript/jsx/JavaScriptParser.g4 [hunk 1]
```diff
@@ -172,7 +173,7 @@ breakStatement
 
 returnStatement
     : Return ({this.notLineTerminator()}? expressionSequence)? eos
-    | Return '(' htmlElements ')' eos
+    | Return '(' jsxElements ')' eos
```
kaby76, line 176: Is there a reason why some string literals are unfolded, like '{', but not others, like '('? Many you unfolded, e.g., line 136 with `'=' => Assign`.

aleppard: Yes.

I've added the '{' and '=', etc symbols to the new lexer modes; which means there are now multiple '{', '=' symbols. This means we can no longer match these symbols by their symbol characters in the parser, because the parser doesn't know if '{' is OpenBrace or JsxOpeningElementOpenBrace. I made these changes based on errors from antlr. For others like '(' I haven't had to touch.

teverett:
@kaby76  good to merge this?

kaby76:
@teverett It is fine to merge.

teverett:
@aleppard  thanks!

aleppard:
Thanks all!

teverett:
@aleppard  thanks!

PR #3962, opened 2024-02-08
snowflake - added interval cast

mlorek:


sql/snowflake/SnowflakeLexer.g4
```diff
@@ -438,6 +438,7 @@ INT          : 'INT';
 INTEGRATION  : 'INTEGRATION';
 INTEGRATIONS : 'INTEGRATIONS';
 INTERSECT    : 'INTERSECT';
+INTERVAL     : 'INTERVAL';
```
KvanTTT, line 441: Can it be an identifier? Is so, please add this token to [`non_reserved_words`](https://github.com/antlr/grammars-v4/blob/master/sql/snowflake/SnowflakeParser.g4#L3472) list.

KvanTTT (approved):

PR #3963, opened 2024-02-08
semver does not allow digit '0' in version numbers

brunoos:
Current versions does not allow '10.1.1' or '1.102.1' since it allows only a sequence of positive digits (not '0').


teverett:
@brunoos  could you add an example for this please?

brunoos:
Sure, I added some examples with digit '0' in the version.

teverett:
@brunoos  thanks!

PR #3968, opened 2024-02-15
Remove form1.vb from vb6 examples

chenzimin:
Remove form1.vb from vb6 examples since vb6 does not support classes.


teverett:
@chenzimin  thanks!

PR #3971, opened 2024-02-17
Support Oracle 23 "IF [NOT] EXISTS"

Naros:
Support Oracle 23 "IF [NOT] EXISTS" syntax - taken from https://github.com/debezium/debezium/pull/5276


KvanTTT (approved):

sql/plsql/PlSqlParser.g4
```diff
@@ -1580,7 +1580,7 @@ hier_attr_name
     ;
 
 create_index
-    : CREATE (UNIQUE | BITMAP)? INDEX index_name ON (
+    : CREATE (UNIQUE | BITMAP)? INDEX index_name (IF NOT EXISTS)? ON (
```
KvanTTT, line 1583: I'm not an expert in Oracle 23 but shouldn't `NOT` be optional?

```suggestion
    : CREATE (UNIQUE | BITMAP)? INDEX index_name (IF NOT? EXISTS)? ON (
```

Naros: Hi @KvanTTT, I don't believe so.  My understanding is that they're marketing it as an "IF [NOT] EXIST" clause, a clause that applies to both CREATE and DROP statements.  In the case of `CREATE` all reference material I've seen indicates the use of `IF NOT EXISTS` while for `DROP` the reference material indicates `IF EXISTS`.  

I checked the [grammar syntax for create](https://docs.oracle.com/en/database/oracle/oracle-database/23/sqlrf/CREATE-TABLE.html#GUID-F9CE0CC3-13AE-4744-A43C-EAC7A71AAAB6) and the [grammar syntax for drop](https://docs.oracle.com/en/database/oracle/oracle-database/23/sqlrf/DROP-TABLE.html#GUID-39D89EDC-155D-4A24-837E-D45DDA757B45) and those seem to corroborate my understanding.

PR #3976, opened 2024-02-20
Eiffel grammar

miguelosilva:
Eiffel language grammar (compatible with current language definition).
Contains Eiffel examples and a simple TokenStreamRewriter based eiffel-2-eiffel generator.


kaby76:
You should add a [pom.xml](https://github.com/antlr/grammars-v4/blob/master/abb/pom.xml) and insert it into the [root pom.xml](https://github.com/antlr/grammars-v4/blob/master/pom.xml). However, you will need to add a [desc.xml](https://github.com/antlr/grammars-v4/blob/master/abb/desc.xml), and only set the target to Java.

teverett:
@miguelosilva  thanks!

eiffel/Eiffel.g4
```diff
@@ -0,0 +1,643 @@
...
+THEN: [tT][hH][eE][nN];
+TRUE: [tT][rR][uU][eE];
+//TUPLE: [tT][uU][pP][lL][eE];
+UNDEFINE: [uU][nN][dD][eE][fF][iI][nN][eE];
+UNIQUE: [uU][nN][iI][qQ][uU][eE];
+UNTIL: [uU][nN][tT][iI][lL];
+VARIANT: [vV][aA][rR][iI][aA][nN][tT];
+VOID: [vV][oO][iI][dD];
+WHEN: [wW][hH][eE][nN];
+XOR: [xX][oO][rR] | '⊻';
```
KvanTTT, line 615: Is it possible to rewrite these tokens using [`caseInsensitive`](https://github.com/antlr/antlr4/blob/dev/doc/options.md#caseinsensitive) lexer option?

PR #3978, opened 2024-02-23
Upgraded supported TypeScript version to v2.7.

aleppard:
This PR upgrades compatibility to TypeScript v2.7. It also adds the `yield*` operator to JavaScript and JSX.

First some minor parsing improvements:

You can now name anonymous functions, e.g.

```
object = {
     field: function myFunction(arg1) { return arg1; }
}
```

Assertion typing (casting) is now supported in `for` statements, e.g.

```
let object = {"field1":5, "field2:"value"};
for (let [field, value] of Object.entries(object) as [string, any]) {
}
```

You can now export type aliases, e.g.

```
export type MyType = {
    field: number, 
    field2: string, 
}
```

(and they don't need to end with a semi-colon).

Union, intersection type aliases can now start with a leading | or &, e.g.

```
type EmployeeType =
     | "employee"
     | "manager";
```

See https://github.com/microsoft/TypeScript/pull/12386 

Tuples (and function parameter lists) can have a trailing comma:

```
[string, number,]
```

See https://github.com/Microsoft/TypeScript/issues/28893 

Upgrades based on TypeScript versions:

**TypeScript v0.9**

You can now declare namespaces, e.g.

```
declare namespace myLib {
}
```

See https://www.typescriptlang.org/docs/handbook/declaration-files/by-example.html

**TypeScript v1.8**

Boolean, number and string literal types are now supported, e.g.

```
export type CharityStatsJson = {
    type: "single" | "multi"; 
    type: 1 | 2; 
    type: false | 1 | 2; 
```

See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-1-8.html#type-parameters-as-constraints

**TypeScript v2.0**

Added non-null assertion operator, e.g.

```
let s = e!.name; // Assert that e is non-null and access name
```

See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator 

Added `undefined` type, e.g.

```
let y: number | undefined;
```

See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#null--and-undefined-aware-types

Added `never` type, e.g.

```
function error(message: string): never {
  throw new Error(message);
}
```

See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#the-never-type

**TypeScript v2.1**

Added `keyof` and indexed access types, e.g.

```
interface Person {
  name: string;
  age: number;
  location: string;
}
type Key = keyof Person; // "name" | "age" | "location"
type Age = Person["age"]; 
```

See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-1.html#keyof-and-lookup-types

**TypeScript v2.2**

Added `object` type, e.g.

```
declare function create(o: object | null): void;
```

See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-2.html#object-type

**TypeScript v2.3**

Added `yield*` operator to JavaScript, TypeScript & JSX.

See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-3.html#generators

Added generic parameter defaults, e.g.

```
function genericParameterWithDefault<T = DefaultType>(field: T) {}
```

See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-3.html#generic-parameter-defaults

**TypeScript v2.7**

Added `unique symbol` type, e.g.

`declare const Foo: unique symbol;`

See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#unique-symbol

### DIFF
--- a/javascript/javascript/JavaScriptLexer.g4
+++ b/javascript/javascript/JavaScriptLexer.g4
@@ -169,6 +169,8 @@ Try        : 'try';
 As         : 'as';
 From       : 'from';
 Of         : 'of';
+Yield      : 'yield';
+YieldStar  : 'yield*';
 
 /// Future Reserved Words
 
@@ -182,7 +184,6 @@ Import  : 'import';
 
 Async : 'async';
 Await : 'await';
-Yield : 'yield';
 
 /// The following tokens are also considered to be FutureReservedWords
 /// when parsing strict mode
--- a/javascript/javascript/JavaScriptParser.g4
+++ b/javascript/javascript/JavaScriptParser.g4
@@ -200,7 +200,7 @@ returnStatement
     ;
 
 yieldStatement
-    : Yield ({this.notLineTerminator()}? expressionSequence)? eos
+    : (Yield | YieldStar) ({this.notLineTerminator()}? expressionSequence)? eos
     ;
 
 withStatement
@@ -314,8 +314,9 @@ arrayLiteral
     : ('[' elementList ']')
     ;
 
+// JavaScript supports arrasys like [,,1,2,,].
 elementList
-    : ','* arrayElement? (','+ arrayElement)* ','* // Yes, everything is optional
+    : ','* arrayElement? (','+ arrayElement) * ','* // Yes, everything is optional
     ;
 
 arrayElement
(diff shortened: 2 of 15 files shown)


teverett:
@aleppard  thanks!

PR #3982, opened 2024-02-26
[javascript] problem with continue statement and newline followed by whitespace

meraedit:
OR continue followed by comment #3772


KvanTTT (changes requested):
There is an error with Go runtime. Is it possible to fix it? If so, could you please fix it?

Unfortunately, I can't merge PR with red CI.

PR #3988, opened 2024-02-27
[Snowflake]Refacto Group by + sample

MasterKuat:
Refactoring group by clause as now we should be able to retreive same element in select and group by in AST : 
 
image

BTW, this allow also to handle group by element like $1 column from VALUES table.


KvanTTT (approved):

teverett:
@MasterKuat  thanks!

PR #4001, opened 2024-03-06
Improved support for computed arguments in both JavaScript and TypeScript.

aleppard:
Improved support for computed arguments in both JavaScript and TypeScript:

```
let reference = "...";
let object = {
  [reference](a){},
}
```
and
```
let y = ([reference]: any) => x;
```
and
```
let object = {
  set a([reference]=123){},
};
```

Can now use keywords as function parameter names, e.g 'this' here:

```
let a = {
   get: function(this: SVGElement) {
   }
}
```

### DIFF
--- a/javascript/javascript/JavaScriptParser.g4
+++ b/javascript/javascript/JavaScriptParser.g4
@@ -412,6 +412,7 @@ initializer
 
 assignable
     : identifier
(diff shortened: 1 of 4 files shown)


teverett:
@aleppard  thanks!

PR #4002, opened 2024-03-09
[PlSql] Generate parser and lexer in non-default package

lbovet:
When building a jar, the generated code should appear in a named Java package. Otherwise the classes cannot be imported from a named package.

### DIFF
--- a/sql/plsql/Java/org/antlr/grammars/plsql/PlSqlLexerBase.java
+++ b/sql/plsql/Java/org/antlr/grammars/plsql/PlSqlLexerBase.java
@@ -1,3 +1,5 @@
+package org.antlr.grammars.plsql;
+
 import org.antlr.v4.runtime.*;
 
 public abstract class PlSqlLexerBase extends Lexer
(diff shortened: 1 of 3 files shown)


kaby76:
This cannot work with the given tester [trgen](https://github.com/kaby76/Domemtech.Trash/tree/main/src/trgen).

The [old Maven tester](https://github.com/antlr/antlr4test-maven-plugin) is not used in the Github Actions build. It only works with Java, not the other 8 targets. And, it hasn't been updated for a few years.

The trgen test driver generator does not have a way to specify the package/namespace because the concept doesn't translate well across all the targets. I can add this, but it will take some time to get this right because specifying a package or namespace will be target-specific.

lbovet:
Okay, we will find another way to build a usable JAR. Indeed, it would be handy to have ready-to-use parser classes in a maven artifact instead of source code dependency.

kaby76:
I can do this easily with a few modifications to trgen, but the modifications would mean a seismic shift in what the repo contains: grammars and support code wouldn't be .g4's, .cs's, .java's, .go's, etc, but template files that get per-target modifications. People will then need to use trgen or a similar tool to generate the parser you want. I don't think the admins here would like that idea. The best solution will be Antlr5 and wasm, with a target-specific layer to access the parser.

PR #4010, opened 2024-03-12
[build] Fix for #4008 -- add testing of TypeScript target back

### LINKED ISSUE #4008: [build] Re-add TypeScript testing
Testing of the TypeScript target [was removed](https://github.com/antlr/grammars-v4/pull/3892).

Removing testing for the target was a little rash. Instead, we can currently test grammars that do not use [superClass](https://github.com/antlr/antlr4/blob/master/doc/options.md#superclass), i.e., do not require target-specific source code. For grammars that have, make sure the TypeScript target is commented out.

kaby76:
Fixes #4008.

* Add TypeScript testing.
* Add "Antlr4ng" TypeScript for testing that target in the future, as it is superior to Antlr 4.13.1. For example, inclusion of base classes is not possible with 4.13.1. And, it is not clear whether we will have another release of Antlr4, which would fix the issue. https://github.com/antlr/antlr4/pull/4492
* Remove testing of grammars that have TypeScript base class code as 4.13.1 does not work with these grammars.

### DIFF
--- a/_scripts/what-to-test.sh
+++ b/_scripts/what-to-test.sh
@@ -1,7 +1,7 @@
 #
 #set -x
 declare -A targets
-for t in "Cpp" "CSharp" "Dart" "Go" "Java" "JavaScript" "PHP" "Python3" "TypeScript"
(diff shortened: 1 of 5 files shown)


teverett:
@kaby76 thanks!

PR #4012, opened 2024-03-12
[PlSql] Various fixes

lbovet:
Submitting several small fixes in one PR. I hope it's ok.

- [PlSql] Use condition in CASE WHEN clause instead of expression
In the "search form", the WHEN clause evaluates a condition.
- [PlSql] Identifier as LISTAGG delimiter
Although undocumented, variables or other identifiers can be used as delimiter
- [PlSql] DEPRECATE Pragma
Was missing
- [PlSql] Add COLLATE as operator
Was missing
- [PlSql] Fix NOT NULL position in associative array declaration
Was at wrong place
- [PlSql] Allow unicode letters in identifiers
User-defined identifiers character set is actually the [database character set](https://docs.oracle.com/en/database/oracle/oracle-database/18/lnpls/plsql-language-fundamentals.html#GUID-321084DE-3BDD-484B-AE85-7A991B67C51A). It is probably a safe guess to accept only letter symbols.

With these last fixes, I am able to parse three decade-old application codebases summing up 600k lines code!

### DIFF
--- a/sql/plsql/PlSqlLexer.g4
+++ b/sql/plsql/PlSqlLexer.g4
@@ -418,6 +418,7 @@ DELETEXML                      : 'DELETEXML';
 DEMAND                         : 'DEMAND';
 DENSE_RANKM                    : 'DENSE_RANKM';
 DEPENDENT                      : 'DEPENDENT';
+DEPRECATE                      : 'DEPRECATE';
 DEPTH                          : 'DEPTH';
 DEQUEUE                        : 'DEQUEUE';
 DEREF                          : 'DEREF';
--- a/sql/plsql/PlSqlLexer.g4
+++ b/sql/plsql/PlSqlLexer.g4
@@ -2511,7 +2511,7 @@ SPACES: [ \t\r\n]+ -> channel(HIDDEN);
 
 fragment NEWLINE_EOF    : NEWLINE | EOF;
 fragment QUESTION_MARK  : '?';
-fragment SIMPLE_LETTER  : [A-Z];
+fragment SIMPLE_LETTER  : [\p{Alpha}\p{General_Category=Other_Letter}];
 fragment FLOAT_FRAGMENT : UNSIGNED_INTEGER* '.'? UNSIGNED_INTEGER+;
 fragment NEWLINE        : '\r'? '\n';
 fragment SPACE          : [ \t];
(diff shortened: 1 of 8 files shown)


sql/plsql/PlSqlLexer.g4
```diff
@@ -418,6 +418,7 @@ DELETEXML                      : 'DELETEXML';
 DEMAND                         : 'DEMAND';
 DENSE_RANKM                    : 'DENSE_RANKM';
 DEPENDENT                      : 'DEPENDENT';
+DEPRECATE                      : 'DEPRECATE';
```
KvanTTT, line 421: I guess it can be an identifier. If so, I suggest adding it to [`regular_id`](https://github.com/antlr/grammars-v4/blob/master/sql/plsql/PlSqlParser.g4#L7442C1-L7442C11) (or `non_reserved_keywords_pre12c`, `non_reserved_keywords_in_12c`, `non_reserved_keywords_in_18c`).

sql/plsql/PlSqlLexer.g4
```diff
@@ -2511,7 +2512,7 @@ SPACES: [ \t\r\n]+ -> channel(HIDDEN);
 
 fragment NEWLINE_EOF    : NEWLINE | EOF;
 fragment QUESTION_MARK  : '?';
-fragment SIMPLE_LETTER  : [A-Z];
+fragment SIMPLE_LETTER  : [\p{Alpha}\p{General_Category=Other_Letter}];
```
KvanTTT, line 2515: Could you please add a test on identifiers that consist of letters from these caterogies?

KvanTTT (changes requested):

sql/plsql/PlSqlLexer.g4
lbovet: There is one in the commit (https://github.com/antlr/grammars-v4/pull/4012/commits/7e2337b1f687190d67791dd36cc151889840dd63): sql/plsql/examples/lexer02.sql

sql/plsql/PlSqlLexer.g4
lbovet: Amended commit https://github.com/antlr/grammars-v4/pull/4012/commits/d875c67b00330d87aa0b15f335d39a1b18ee1631

KvanTTT (approved):

PR #4016, opened 2024-03-14
[build] Add package to Java target generated parsers

lbovet:
As alternative to #4002.
This moves the generated classes into a Java package (and not keep them in the unnamed package) without disturbing the trgen process.
This way, one can build a reusable JAR containing the lexer and parser.


lbovet:
Not yet working as expected, closing

PR #3980, opened 2024-02-24
Update to dotnet version 8 (released in November '23).

kaby76:
Fix for https://github.com/antlr/grammars-v4/issues/3979. This change updates the version of dotnet to version 8. This will be required for new Trash tools.

PR #4019, opened 2024-03-17
Enable testing of grammars with Antlr4ng

kaby76:
This PR updates the grammars and build to test using the Typescript tool and runtime Antlr4ng for certain Antlr4 grammars. The target `Antlr4ng` is added to the grammar's desc.xml where the target works.

[Antlr4ng](https://github.com/mike-lischke/antlr4ng) is @mike-lischke 's updated TypeScript target, improving on many issues with the Antlr4 TypeScript target version 4.13.1. The reason it is added is because there will likely not be any further releases of Antlr4. In fact, for grammars that have TypeScript target-specific support code, Antlr 4.13.1 TypeScript runtime does not work with newer versions of tsc/Node because the generated parser and lexer cannot run ([it cannot find the compiled base class files](https://github.com/antlr/antlr4/issues/4491)). The only hope is to use Antlr4ng.

This PR made the following changes.
* Installed Node 21.7.1 for all OSes for the CI workflow. This is required to run on Mac ([dyld[15293]: Library not loaded: '/usr/local/opt/icu4c/lib/libicui18n.73.dylib'](https://github.com/antlr/grammars-v4/actions/runs/8319594275/job/22763188451#step:24:609)). In addition, I saw a performance gain with the newer node, but I don't see the gain here in the Github Actions.
* Added Antlr4ng templates for trgen. These templates mirror what is done with the Antlr 4.13.1 TypeScript templates, except that here we need to use the Antlr4ng-specific tool and runtime. I have the package.json use specifically version 3.0.4 of the runtime and 2.0.0 of the tool. Note, *the package-lock.json file is NOT checked in because: (1) it is generated from a specification: package.json; (2) when I build a TS application, the package-lock.json gets updated. I do NOT want to check this change as a side-effect from a build. There have been many "answers" on SO that insist to check in package-lock.json. I emphatically reject this idea! Package.json is a specification; package-lock.json is an implementation. Fortunately, in the world of C#, there is no such stupid concept with .csproj's because every one knows you don't check in generated files.*
* Updated the CSharp targets with net8.0. I already updated the Dotnet tool chain with version 8 [here](https://github.com/antlr/grammars-v4/pull/3980). This PR completes the switchover by modifying the trgen templates to use net8.
* Fixed a problem with "clean.sh" in templates. The Antlr .jar is used to tell what files are generated and what should be removed. Note, this script is never used in testing. The change is not perfect because I hardwire the name of the Antlr tool, so I will still need to figure out a more permanent solution.
* Added in `Antlr4ng` into `<targets>` sections in desc.xml. This tells trgen that the grammar works for Antlr4ng. Otherwise, the grammar DOES NOT work for the target. No split grammars with a base class are working yet. The assumption is that it should work. If not, we can ask Mike to make a fix and cut an immediate release. Mike is a much more responsive partner in fixing bugs.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -124,12 +124,10 @@ jobs:
       run: |
          pip install antlr4-tools
     - name: Install JavaScript
-      if: ${{ matrix.language == 'JavaScript' }}
       uses: actions/setup-node@v4.0.2
       with:
-        node-version: '16.13.0'
+        node-version: '21.7.1'
     - name: Test JavaScript
-      if: ${{ matrix.language == 'JavaScript' }}
       run: |
         node --version
     - name: Update paths
--- a/_scripts/templates/Antlr4ng/package.json
+++ b/_scripts/templates/Antlr4ng/package.json
@@ -1,0 +1,22 @@
+{
+  "name": "i",
+  "version": "1.0.0",
+  "description": "",
+  "main": "Test.js",
+  "scripts": {
+    "build": "tsc -p tsconfig.json"
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "antlr4ng": "^3.0.4",
+    "buffer": "^6.0.3",
+    "fs-extra": "^11.1.1",
+    "timer-node": "^5.0.6",
+    "typescript-string-operations": "^1.5.0"
+  },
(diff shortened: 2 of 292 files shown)


teverett:
@kaby76  thanks!

PR #4020, opened 2024-03-19
[build] Add java package in Maven build

lbovet:
As alternative to #4002.
This moves the generated classes into a Java package (and not keep them in the unnamed package) without disturbing the trgen process. This way, one can build a reusable JAR containing the lexer and parser.

The plugin configuration is done globally in the root POM. 
To benefit from it, grammar sub-projects have to adapt their POM and base classes. This PR includes the adaptation for the plsql grammar to illustrate what has to be done.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,6 +19,10 @@
 		<antlr.version>4.11.1</antlr.version>
 
 		<junit.version>4.13.2</junit.version>
+
+        <packageName>org.antlr.grammars.${project.artifactId}</packageName>
+        <packagePath>org/antlr/grammars/${project.artifactId}</packagePath>
+        <packageLine>package ${packageName};</packageLine>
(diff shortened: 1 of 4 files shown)


kaby76 (approved):
This change is fine.

What you are doing, essentially, is creating a template for the code. Ideally, this code (the .g4's, the .java's, the pom.xml) would all be tailored for `packageLine`, if a package is specified for a target. Ideally, there should be a syntax for attributes that would be used in place of `packageLine` (we do--StringTemplates), and that the desc.xml would contain the necessary attributes to generate the code for the target (not yet).

lbovet:
Yes, this is an intermediate solution until there is something better integrated with the multi-target approach.

teverett:
@lbovet thanks!

PR #4021, opened 2024-03-19
Suggesting another alternative for rule postfixExpression

dk2k:
Works in my project
Please let me know if I need to update sample class to trigger this alternative in the rule


KvanTTT (changes requested):
Hi! Thanks for the contribution and welcome to ANTLR grammars repository!

But please cover your change by an example in [examples](https://github.com/antlr/grammars-v4/tree/master/java/java8/examples) directory. You can either add a new file or extend an existing file.

Issue #4013, opened 2024-03-13
[build] Add performance test comparison of a PR with current tip in main branch


kaby76:
Antlr4 accepts almost any grammar, including ambiguous or ones that require full context. This isn't really a good thing in the long run because most people who want to pick up a grammar here are looking for *efficient* grammars.

When a PR is created, we currently do not know whether the change adds new problems into the grammar. While we currently cannot test whether the grammar is ambiguous, we can compare the performance of the current grammar with the current set of example inputs with the new grammar in the PR. Note, we do not want to test new inputs.

When a PR causes a "large degradation" in the performance, we should fail the PR.

PR #4023, opened 2024-03-22
[build] Add performance testing of PRs vs current

kaby76:
This PR, for #4013, helps to solve a significant problem in detecting regressions in grammar performance. Changes to a grammar that solve a syntax problem often result in a performance degradation that is unknown until well after the PR is merged. To detect this problem, a workflow is added to perform a statistical analysis of the performance for a grammar, comparing the grammar before and after a change. While no errors should be raised if the testing fails, the workflow will give a "heads up" that people can look at to see whether the PR introduces a regression.

While it does not solve the greater problem of finding ambiguity and full-context parser fallbacks, these problems can be inferred via a statistical analysis of the performance.

### Changes

* A new workflow is added: "perf.yml".
* The workflow executes _scripts/perf-changed.sh.
* The workflow passes to the Bash script what the PR modifies. For each grammar modified, the script checks the runtime of the grammar before and after the PR.
* The sample size (i.e., the number of times all tests in examples/ are run) is the minimum of 40 or the number of times one can run the test under 20 minutes total (10 m before + 10 m after, for each grammar changed). For example, if the test normally takes 3 minutes, then only 10/3 = 3 iterations can be done. The test driver always outputs a "Total Time", which is stored in files for the PR and before the PR, which is in units of seconds.
* A bar graph showing the mean time and standard deviations for each grammar/before-after PR. The bar graph is output as TTY since artifacts are generally not available. Note, it is important to note that you can have a "statistical significance" while the bar graph shows the mean times roughly the same and error bars overlapping.
* A Welch's test is computed to determine "statistical" significance.
* We also check the ratio of the mean runtimes to determine "practical" significance. A cut off of 5% is assumed.
* If there is both a "statistical" and "practical" significant difference in performance, the workflow outputs a message to indicate there is a performance drop. See https://github.com/antlr/grammars-v4/actions/runs/8466843554/job/23196490098#step:15:359 
* This PR alters the sql/plsql grammar so that the statistical test can be performed. The grammar was reformatted according to the coding style rules for the repo.


kaby76:
A statistical difference does not equate a "practical" difference. In addition to a statistical difference there must be a slowdown of 5%--an arbitrary value, yes.

kaby76:
All set.

teverett:
@kaby76  thanks!

Issue #4028, opened 2024-03-23
PL/SQL grammar cannot recognise identifiers that start with keywords


stephendrew:
Target: C#
Grammar: PL/SQL

I am trying to parse parser.sql_script() for this:

```sql
CREATE OR REPLACE PACKAGE ACCESS_SECURITY AS
-- omitted
END;
```

But it fails due to treating ACCESS as a keyword.

```
Token: [@0,0:5='CREATE',<305>,1:0]
Token: [@1,6:6=' ',<2380>,channel=1,1:6]
Token: [@2,7:8='OR',<1237>,1:7]
Token: [@3,9:9=' ',<2380>,channel=1,1:9]
Token: [@4,10:16='REPLACE',<1484>,1:10]
Token: [@5,17:17=' ',<2380>,channel=1,1:17]
Token: [@6,18:24='PACKAGE',<1257>,1:18]
Token: [@7,25:25=' ',<2380>,channel=1,1:25]
Token: [@8,26:31='ACCESS',<4>,1:26]
Token: [@9,32:32='_',<2372>,1:32]
Token: [@10,33:40='SECURITY',<1563>,1:33]
Token: [@11,41:42='\r\n',<2380>,channel=1,1:41]
Token: [@12,43:44='AS',<67>,2:0]
```

```
Exception of type 'Antlr4.Runtime.NoViableAltException' was thrown.
OffendingToken: [@9,32:32='_',<2372>,1:32]
```

kaby76:
This is not a "symbol conflict". It is a regression in the recognition of identifiers with underscore, introduced by @lbovet with https://github.com/antlr/grammars-v4/pull/4012.
```
$ ./bin/Debug/net8.0/Test.exe xxx.in -tokens
[@0,0:5='CREATE',<305>,1:0]
[@1,6:6=' ',<2380>,channel=1,1:6]
[@2,7:8='OR',<1237>,1:7]
[@3,9:9=' ',<2380>,channel=1,1:9]
[@4,10:16='REPLACE',<1484>,1:10]
[@5,17:17=' ',<2380>,channel=1,1:17]
[@6,18:24='PACKAGE',<1257>,1:18]
[@7,25:25=' ',<2380>,channel=1,1:25]
[@8,26:40='ACCESS_SECURITY',<2378>,1:26]
[@9,41:41=' ',<2380>,channel=1,1:41]
[@10,42:43='AS',<67>,1:42]
[...]

PR #4029, opened 2024-03-25
[PlSql] Fix alphabetic match portability

lbovet:
Hopefully fixes https://github.com/antlr/antlr4/issues/4556 and fixes https://github.com/antlr/grammars-v4/issues/4028


teverett:
@lbovet thanks!

Issue #4035, opened 2024-03-31
[cpp] Add Antlr4ng port for the CPP grammar


kaby76:
As a follow-up to all this Antlr4ng and TypeScript issues, as a "proof-in-concept", the [cpp grammar](https://github.com/antlr/grammars-v4/tree/c7883a594bf622200c561db181848d277fcd49a7/cpp) is ideal to use as a test case. The grammar already contains ports for all the other targets, so adding the Antlr4ng port should be easy. And it is.

kaby76:
Fixed with https://github.com/antlr/grammars-v4/pull/4036

PR #4036, opened 2024-03-31
Fix for #4035 -- add Antlr4ng port for cpp grammar.

kaby76:
This is a fix for #4035. It's based on the TypeScript port with a few simple mods specific for Antlr4ng.


teverett:
@kaby76  thanks!

Issue #4038, opened 2024-04-01
[php] examples/strings.php has a BOM--files cannot have BOM!!


kaby76:
The templates for testing do not consider BOM in a file input. All files cannot have BOM. [php/examples/strings.php](https://github.com/antlr/grammars-v4/blob/c7883a594bf622200c561db181848d277fcd49a7/php/examples/strings.php) contains a BOM. For TypeScript, the file is opened and read via [readFileSync()](https://github.com/antlr/grammars-v4/blob/c7883a594bf622200c561db181848d277fcd49a7/_scripts/templates/Antlr4ng/st.Test.ts#L166) into a string. That should have gotten rid of the BOM but it does not. The encoding parameter is "utf-8", but it does not automatically handle BOM. Instead, it considers it as a character. What a lovely language TypeScript/JavaScript is----NOT!!!

I can fix this file, but for future problems, the templates need to cut this BS out. NO  BOM!!!!

https://stackoverflow.com/questions/33977843/how-to-detect-if-a-file-has-a-utf-8-bom-in-bash

kaby76:
Fixed with https://github.com/antlr/grammars-v4/pull/4039

PR #3994, opened 2024-02-29
Added TypeScript PhpLexerBase for php grammar

tomaspiaggio:
I added the TypeScript class for using the php grammar from the typescript target. It's very similar to the JavaScript one.

PR #4039, opened 2024-04-01
[php] Fix for #4037 and #4038

### LINKED ISSUE #4037: [php] Port the php grammar to Antlr4ng
As a replacement for https://github.com/antlr/grammars-v4/pull/3994 -- which cannot work because the TypeScript Antlr 4.13.1 tool does not work -- add in a port of the grammar to Antlr4ng.

kaby76:
This PR fixes a few problems in the [php grammar](https://github.com/antlr/grammars-v4/tree/c7883a594bf622200c561db181848d277fcd49a7/php) ( #4037 and #4038 ), and replaces PR #3994 .

First, #4038 . There are two test files that have BOM. While I could adjust the test inputs into two piles, one with BOM and the other without, then set up desc.xml to set the encoding specifically for no BOM, it's untested, and likely going to run into problems. Instead, I removed the BOM from the files. I should update the tester so that if there is any BOM, fail. Maybe??

Second, this PR fixes #4037. This is a substitute for #3994 , which cannot work because Antlr 4.13.1 cannot generate a correct PhpLexer.ts file: the generated file contains an incorrect `import` statement. That PR should be closed.


teverett:
@kaby76  thanks

PR #4046, opened 2024-04-11
[javascript] problem with continue statement and newline followed by whitespace

meraedit:
OR followed by comment #3772

Created new pull request because the previous branch was wrong (created from our branch not master)

### DIFF
--- a/javascript/javascript/Cpp/JavaScriptParserBase.cpp
+++ b/javascript/javascript/Cpp/JavaScriptParserBase.cpp
@@ -24,7 +24,7 @@ bool JavaScriptParserBase::next(std::string str)
 
 bool JavaScriptParserBase::notLineTerminator()
 {
-    return !here(JavaScriptParser::LineTerminator);
+    return !lineTerminatorAhead();
 }
 
 bool JavaScriptParserBase::notOpenBraceAndNotFunction()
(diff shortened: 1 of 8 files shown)


javascript/javascript/Cpp/JavaScriptParserBase.cpp
```diff
@@ -24,7 +24,7 @@ bool JavaScriptParserBase::next(std::string str)
 
 bool JavaScriptParserBase::notLineTerminator()
 {
-    return !here(JavaScriptParser::LineTerminator);
+    return !lineTerminatorAhead();
```
kaby76, line 27: Please remove all `here()` method definitions. You removed the applied usages for `here()`, so it does not make any sense to keep the old code around.

kaby76:
The Go target fails because the code is not isomorphic to the other ports.

Please correct line 54.

https://github.com/Servoy/grammars-v4/blob/de604bba9bd805524db7ec0ef721f93675c42909/javascript/javascript/Go/javascript_parser_base.go#L84

This is the equivalent line in the CSharp target.

https://github.com/Servoy/grammars-v4/blob/de604bba9bd805524db7ec0ef721f93675c42909/javascript/javascript/CSharp/JavaScriptParserBase.cs#L106

javascript/javascript/Cpp/JavaScriptParserBase.cpp
meraedit: Thank you, I removed the unused method. I wouldn't have figured out what was the problem with the go parser.

kaby76 (approved):

teverett:
@meraedit  thanks!

PR #4051, opened 2024-04-17
Add 'DISTINCT FROM' operators

danysantiago:
Add `IS DISTINCT FROM` and `IS NOT DISTINCT FROM` operators that where added in SQLite [3.39.0](https://www.sqlite.org/changes.html#version_3_39_0)

### DIFF
--- a/sql/sqlite/SQLiteParser.g4
+++ b/sql/sqlite/SQLiteParser.g4
@@ -249,7 +249,7 @@ drop_stmt
     + -
     << >> & |
     < <= > >=
-    = == != <> IS IS NOT IN LIKE GLOB MATCH REGEXP
+    = == != <> IS IS NOT IS DISTINCT FROM IS NOT DISTINCT FROM IN LIKE GLOB MATCH REGEXP
     AND
     OR
  */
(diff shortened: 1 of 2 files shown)


sql/sqlite/SQLiteParser.g4
```diff
@@ -270,6 +270,8 @@ expr
         | NOT_EQ2
         | IS_
         | IS_ NOT_
+        | IS_ DISTINCT_ FROM_
+        | IS_ NOT_ DISTINCT_ FROM_
```
KvanTTT, line 274: Please simplify:

```suggestion
        | IS_ NOT_? DISTINCT_ FROM_
```

KvanTTT (changes requested):
Please cover your changes by an example in [sqlite/examples](https://github.com/antlr/grammars-v4/tree/master/sql/sqlite/examples) (add a new or extend an existing).

sql/sqlite/SQLiteParser.g4
danysantiago: Done - Thanks

danysantiago:
> Please cover your changes by an example in [sqlite/examples](https://github.com/antlr/grammars-v4/tree/master/sql/sqlite/examples) (add a new or extend an existing).

Of course! - Done

PR #4052, opened 2024-04-17
Add RIGHT and FULL join operators

danysantiago:
Add the `RIGHT OUTER JOIN` and `FULL OUTER JOIN` join operators that where added in SQLite [3.39.0](https://www.sqlite.org/changes.html#version_3_39_0). See also the syntax diagram in https://www.sqlite.org/syntax/join-operator.html

### DIFF
--- a/sql/sqlite/SQLiteParser.g4
+++ b/sql/sqlite/SQLiteParser.g4
@@ -421,7 +421,7 @@ result_column
 
 join_operator
     : COMMA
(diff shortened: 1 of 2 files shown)


KvanTTT (changes requested):

But please cover your changes by an example in [sqlite/examples](https://github.com/antlr/grammars-v4/tree/master/sql/sqlite/examples) (add a new or extend an existing).

KvanTTT (approved):

PR #4053, opened 2024-04-17
Add unicode charset to IDENTIFIER

danysantiago:
SQLite supports unicode characters as identifiers and specifically any character larger than `u007f`. This changes adds the unicode range to the `IDENTIFIER` lexer rule. The supported character ranges where documented early in the tokenizer design doc in https://www.sqlite.org/draft/tokenreq.html

### DIFF
--- a/sql/sqlite/SQLiteLexer.g4
+++ b/sql/sqlite/SQLiteLexer.g4
@@ -222,8 +222,8 @@ IDENTIFIER:
     '"' (~'"' | '""')* '"'
     | '`' (~'`' | '``')* '`'
     | '[' ~']'* ']'
-    | [A-Z_] [A-Z_0-9]*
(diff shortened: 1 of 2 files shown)


KvanTTT (changes requested):
Please cover your changes by an example in [sqlite/examples](https://github.com/antlr/grammars-v4/tree/master/sql/sqlite/examples) (add a new one or extend an existing).

danysantiago:
> Please cover your changes by an example in [sqlite/examples](https://github.com/antlr/grammars-v4/tree/master/sql/sqlite/examples) (add a new one or extend an existing).

Done - Added sample / test with some basic cases, not too elaborated because validating unicode can get a bit crazy. 😅

KvanTTT (approved):

PR #4060, opened 2024-04-20
[sql/derby and sql/tsql] Fix #4047

### LINKED ISSUE #4047: Problems with SQL reformatting
https://github.com/antlr/grammars-v4/blob/753536777d827ccc0c9b108531ea67375c2039ac/sql/tsql/TSqlLexer.g4#L1292

It looks like some information was lost because of the automatic formatting. For example, at the end of this file:
```
    | '\uff00'..'\ufff0'
    // | '\u10000'..'\u1F9FF'  //not support four bytes chars
    // | '\u20000'..'\u2FA1F'
    ;
```
turned into 
```
    | '\uff00' ..'\ufff0'
    ; // | '\u20000'..'\u2FA1F'
```

Someone might want to compare that commit with its parent with all whitespace stripped to find other cases of lost comments and fix the tool so it doesn't remove comments.

To begin with, automatic formatting may not have been the best idea considering there were deliberate decisions made for things like this bit from the T-SQL parse grammar:

```
dateparts_12
    : dateparts_9
    | DAYOFYEAR | DAYOFYEAR_ABBR
    | MICROSECOND | MICROSECOND_ABBR
    | NANOSECOND | NANOSECOND_ABBR
```

kaby76:
Fixes for #4047 

The comments that were deleted by the formatter were recovered, and the updated grammars were reformatted.

The haskell grammar, which [I mentioned in the Issue](https://github.com/antlr/grammars-v4/issues/4047#issuecomment-2054331635), did not need to be changed.

I noticed that the comment `not support[ed] four bytes chars` is actually incorrect. [Indeed, Antlr4 does handle 4-byte Unicode](https://github.com/antlr/antlr4/blob/master/doc/unicode.md#unicode-code-points-in-lexer-grammars). At the time of [the change](https://github.com/antlr/grammars-v4/commit/1914b0561b6295d20eb3595169edc45bfc0dbcb8), in 2016, Unicode for Antlr was not yet implemented. I don't know whether the grammar should contain the Unicode ranges mentioned in the comments.  I am not familiar enough with the language to decide without reading the specs.

### DIFF
--- a/sql/derby/DerbyLexer.g4
+++ b/sql/derby/DerbyLexer.g4
@@ -428,4 +428,6 @@ fragment FullWidthLetter options {
     | '\ua000' ..'\ud7ff'
     | '\uf900' ..'\ufaff'
     | '\uff00' ..'\ufff0'
-; // | '\u20000'..'\u2FA1F'
+    // | '\u10000'..'\u1F9FF'  //not support four bytes chars
+    // | '\u20000'..'\u2FA1F'
+;
--- a/sql/tsql/TSqlLexer.g4
+++ b/sql/tsql/TSqlLexer.g4
@@ -1289,4 +1289,6 @@ fragment FullWidthLetter options {
     | '\ua000' ..'\ud7ff'
     | '\uf900' ..'\ufaff'
     | '\uff00' ..'\ufff0'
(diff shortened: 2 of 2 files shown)


teverett:
@kaby76  thanks!

PR #4061, opened 2024-04-23
[Snowflake]Add support rounding function

MasterKuat:
Add support for rounding function espacially for ROUND with 2 syntax and special parameter for round mode.


KvanTTT (approved):

kaby76:
(Build for [CSharp target failed](https://github.com/antlr/grammars-v4/actions/runs/8885296589/job/24396353662#step:24:67). An old version of Antlr4BuildTasks was being referenced. Something changed in the env to cause the old package to break. Fixed with a new version [here](https://github.com/antlr/grammars-v4/pull/4076).)

Issue #3798, opened 2023-10-26
[bnf] Add new BNF grammar


kaby76:
* The current bnf grammar is really for EBNF. This should be renamed to "ebnf".
* A new bnf grammar should be introduced that parses some of the more popular BNF, but most importantly, parses the original Algol 60 BNF. The main problem here is that there are really two ways people note non-terminals and terminals. In the original BNF, non-terminals are delimited by '<' and '>'. Later on, people noticed that BNF could not describe itself, so the notion of what was delimited was reversed, either using '<' and '>', or more popularly using '"' or '\''.

PR #4074, opened 2024-04-30
[bnf and ebnf] Rename old bnf grammar to ebnf. Add new bnf grammar.

kaby76:
This PR is a partial fix for #3798.

* The [bnf grammar](https://github.com/antlr/grammars-v4/tree/a1cc6752263187a086ae961a47f82284b0a10cdf/bnf) wasn't really BNF, but actually EBNF. This is because [it had kleene operators](https://github.com/antlr/grammars-v4/blob/a1cc6752263187a086ae961a47f82284b0a10cdf/bnf/bnf.g4#L70-L76). The directory containing the old grammar was moved to "ebnf". Otherwise, the original grammar is unchanged.
* Another PR will add a new EBNF grammar along with keeping the old EBNF grammar. The new EBNF grammar will be more flexible in order to handle different popular EBNF.
* I added a new bnf grammar and examples. This grammar also switches between various flavors of BNF, which are described in the references in the Readme.
* The purpose of this grammar is to aide in automated refactoring of BNF grammars.


kaby76:
Waiting for https://github.com/antlr/grammars-v4/pull/4076

kaby76:
I will need to change perf-changed.sh to not fail. The problem was caused by a rename of bnf. However, the "bnf" grammar is new and has nothing to do with the old grammar.

teverett:
@kaby76  thanks!

PR #4079, opened 2024-05-01
Adding new grammar Langium.

kaby76:
This PR adds Langium, a parser generator that is similar in syntax to Antlr4.

The grammar from the source, which is included as an example, was scraped by hand and converted into Antlr4 syntax. The examples are all that I could find, as there doesn't seem to be a lot of users of the tool. Three ports are provided (Antlr4ng, CSharp, and Java), but it should be easy to complete the rest of the ports for those who may be interested.

The "perf-changed.sh" script was changed to not fail. i.e., if there is a problem, don't worry about it. In this PR, a grammar was added, but no previous one existed. So, trying to build the "previous" where none exists gives an error, as you would expect. But, we just want to ignore it.


teverett:
@kaby76  thanks!

PR #4082, opened 2024-05-03
Remove backslash from JSON exponential notation

emileferreira:
The JSON grammar accepts the input `1e\1`, since the character set of the `EXP` lexer rule contains a superfluous backslash. The `-` character [does not need to be escaped if it is the first or last character in a character set](https://github.com/antlr/antlr4/blob/master/doc/lexer-rules.md#lexer-rule-elements). This removes that backslash.


teverett:
@emileferreira  thanks!

PR #4083, opened 2024-05-05
Aterm

teverett:


teverett:
@kaby76 I seem to have something not quite right in this commit.  I see errors like "<identifer> expected".  Would you mind taking a look?

kaby76:
Trgen tries to find a start rule but it cannot because it is not an EOF-terminated start rule. The grammar should be augmented with a rule like `start_ : term_ EOF ;`.

Issue #4088, opened 2024-05-09
[csharp grammar] CSharp and Java producing different error messages for expected tokens


kaby76:
[Someone was trying to use the CSharp port](https://stackoverflow.com/questions/78424669/linking-visitmember-access-and-visitifstatement-methods-of-antlr4#comment138303496_78424669) of the [csharp grammar](https://github.com/antlr/grammars-v4/tree/19de2a44eaef0de599044b76ede8492fb8e07b4d/csharp) (even though [the port is not listed in the desc.xml](https://github.com/antlr/grammars-v4/blob/19de2a44eaef0de599044b76ede8492fb8e07b4d/csharp/desc.xml#L3)).

The port of the grammar is fine, the error recovery reporting code in the CSharp and Java Antlr runtime are not the same. See https://github.com/antlr/antlr4/issues/3700#issuecomment-2102551486 . The solution is to add the port to the desc.xml and remove [the test](https://github.com/antlr/grammars-v4/blob/19de2a44eaef0de599044b76ede8492fb8e07b4d/csharp/examples/issue-2612.txt) for now.

kaby76:
Completed. https://github.com/antlr/grammars-v4/pull/4089

PR #4089, opened 2024-05-09
[csharp] Fix for #4088 (add testing of the CSharp target of csharp grammar)

kaby76:
This PR adds testing of the CSharp port for the [csharp grammar](https://github.com/antlr/grammars-v4/tree/master/csharp), which is a workaround for #4088. The port is fine. The main problem is that the Antlr4 runtime error recovery does not use the same algorithm to report errors, so one of the tests would give a different error message. The solution is to just ignore the file for now. https://github.com/antlr/antlr4/issues/3700#issuecomment-2102551486


teverett:
@kaby76 thanks!

PR #4090, opened 2024-05-11
Enhance PLSQL DDL for Alter Database

ilyasahsan123:
Reference: 
https://docs.oracle.com/en/database/oracle/oracle-database/12.2/sqlrf/ALTER-DATABASE.html#GUID-8069872F-E680-4511-ADD8-A4E30AF67986__I2080163

image


ilyasahsan123:
hi @kaby76, Could you please take a look at this pull request? Thank you

sql/plsql/PlSqlParser.g4
```diff
@@ -4276,7 +4276,7 @@ general_recovery
 
 //Need to come back to
 full_database_recovery
-    : STANDBY? DATABASE (
+    : STANDBY? DATABASE? (
```
kaby76, line 4279: Hmm. "full_database_recovery" seems to require the DATABASE token. https://docs.oracle.com/en/database/oracle/oracle-database/12.2/sqlrf/ALTER-DATABASE.html#GUID-8069872F-E680-4511-ADD8-A4E30AF67986__I2128285 . It is optional and the doc wrong? Also, can't the expression `( ( ... )+ )?` following DATABASE in the rule be `( ... )*`? _(NB: I need to fix the script that detects useless parentheses as it flags the outer parens as useless, but it is not. https://github.com/antlr/grammars-v4/actions/runs/9044581722/job/24853394579#step:15:131 )_

teverett:
@kaby76 this is not quite ready to merge yet?

kaby76:
> @kaby76 this is not quite ready to merge yet?

No, it's not quite ready. I still don't understand why the doc says `DATABASE` is required in `full_database_recovery` but here @ilyasahsan123 changed it to be optional. Please address.

PR #4097, opened 2024-05-17
Adding ruleworks grammar.

whcox603:


whcox603:
Any idea what this failure is about?

Test.ts:199:25 - error TS1003: Identifier expected.

199     const tree = parser.();
     
I am slowly getting a build/test environment setup on my system and I get the same result. I don't see anything that tells what/where the problem is located.

bill

teverett:
@whcox603  i had this exact issue last week.  It needs an EOF.  

Try changing this
`
program
    : TOK_LPAREN program_line* TOK_RPAREN
    ;
`

to 

`
program
    : TOK_LPAREN program_line* TOK_RPAREN EOF
    ;
`

kaby76:
I'll fix the tester to give a betyer warning.

whcox603:
Much better. I updated ruleworks.g4

whcox603:
It looks like the 3 failures are with the Go target, but not much to go on (pun unintended.)

On my system the Go test succeeded. I fixed the conflict with action (renamed to action_).

kaby76:
Dependabot updated the go.mod file to remove the reference to Antlr. So, the code that was there to yank out the version doesn't find one. (The code should have checked for a missing reference so it doesn't try to then do a bogus "go get".)

I updated the Go templates to put the version directly in the build.* scripts. Please sync your PR with master.

whcox603:
Much better. Thanks to you both (@teverett @kaby76) for the assistance.

The coverage is not that great (around 50%), but there were only 3 sample programs provided. I have built the compiler and I will use that to test out some more sample programs. One problem is that the SQL calls only work with VAX rdb, which I don't have access to.

teverett:
What a great contribution.  Thanks @whcox603 .

PR #4098, opened 2024-05-18
[MySQL] Add IF NOT EXISTS support for CREATE TRIGGER

Naros:
Upstream contribution from https://github.com/debezium/debezium/pull/5566


teverett:
@Naros  thanks!

PR #4100, opened 2024-05-18
Whenever

teverett:


kaby76:
I fixed the Go builds. Turns out Dependabot screws up the go.mod files--removing the reference to Antlr, and the only way I get the version of the Antlr Tool and Antlr Go runtime to match.

Please "fast forward" from "master" once https://github.com/antlr/grammars-v4/pull/4102 is merged. Sorry for the mess.

teverett:
> I fixed the Go builds. Turns out Dependabot screws up the go.mod files--removing the reference to Antlr, and the only way I get the version of the Antlr Tool and Antlr Go runtime to match.
> 
> Please "fast forward" from "master" once #4102 is merged. Sorry for the mess.

No big deal.  Thanks for the fix!

teverett:
so @kaby76  any idea why this grammar fails on Windows but not Ubuntu or OSX?

Issue #4101, opened 2024-05-18
[Go target] Dependabot removes the dependency to Antlr4.13.1 and so breaks the Go builds


kaby76:
I cannot depend on Dependabot to update a go.mod file. I will need to hardwire the version in build.* in the templates.

kaby76:
Fixed with https://github.com/antlr/grammars-v4/pull/4102

PR #4102, opened 2024-05-18
Fix for #4101

kaby76:
Dependabot cannot be used to update go.mod. I don't know why it deleted the reference to Antlr but that's catastrophic: I assumed the go.mod file would have the version of Antlr to use, and there was code in the build.* scripts to read the file to get the version of Antlr to use. This is extremely important because the version of the runtime and the tool must match exactly!!

This PR fixes the Go target testing, #4101 

* Moved _scripts/templates/Go/go.mod to st.go.mod. This causes Dependabot to not try to do any updating of the dependencies because I changed the name of the file. trgen will however rename the template to go.mod after generating the driver.
* _scripts/templates/Go/build.* now have the version of Antlr hardwired to 4.13.1. I didn't want to set up a file that would be the "oracle" for the version of Antlr to use--it's not worth it. So, if Antlr gets updated, you will need to update these two files.


teverett:
@kaby76  thanks!

PR #4103, opened 2024-05-19
Here's the full Oracle MySQL grammar

mike-lischke:
This patch not only comes with the grammars, but also with a TypeScript implementation of 2 support classes and a demo.


mike-lischke:
@kaby76 Can you please help with the failing CI tests for this PR? I have no idea what I should change to fix them.

kaby76:
So, there are a couple of problems.

First, I see that you have a package.json and tsconfig.json in the TypeScript/ directory. The problem is that this overrides the package.json and tsconfig.json in the driver templates. https://github.com/antlr/grammars-v4/blob/master/_scripts/templates/TypeScript/package.json https://github.com/antlr/grammars-v4/blob/master/_scripts/templates/TypeScript/tsconfig.json. Trgen will overwrite the default .json files with yours, so it basically cannot find the tool .jar. () I would recommend moving your driver to another directory and just let the tester work with the default .json files.

Also, I see you are using the Antlr4ng fork. This is great because the official TypeScript Antlr target really don't work for split grammars. (The 4.13.1 Tool generates the imports incorrectly for superClass.) That brings up whole issues like: "Whose responsibility is it to import the base class files?" In the CSharp, Java targets, the templates generate the imports for the `options { superClass=...; }` option. But, for Antlr4ng and Cpp, the tool does not generate any includes or imports.

My recommendation is that you use the "Antlr4ng" target in the desc.xml. All that is needed is to rename the "TypeScript/" directory as "Antlr4ng/", and adjust the desc.xml to have `<targets>Antlr4ng</targets>`.

Trgen will place all .g4's and Antlr4ng/ in one, flat directory. So, `import { MySQLBaseLexer, SqlMode } from "../MySQLBaseLexer.js"` should be adjusted to `import { MySQLBaseLexer, SqlMode } from "./MySQLBaseLexer.js"`

You can take care of these problems by modifying the .g4's and .js's, or by having a "transformGrammar.py" to modify the paths for the build.

After testing these changes, I noticed that everything parses except examples/statements.txt with error.

mike-lischke:
Hmm, that sounds a bit limiting. It might be a good way for testing, but what about the end (grammar) user? I would like to provide something they can directly try out. As it is now the user just has to run two simple commands and has a working parser application, from where they can expand. But on the other hand I have no better idea for testing, so I'm unsure how to proceed to satisfy both needs.

> Trgen will overwrite the default .json files with yours, so it basically cannot find the tool .jar. () I would recommend moving your driver to another directory and just let the tester work with the default .json files.

You can probably make your JS/TS testing easier if there was a top level package.json. This can import all required TS/JS runtimes and you can then use them from any grammar folder, without needing to adjust the import paths (similar to what I used in the [benchmark project](https://github.com/mike-lischke/antlr4-runtime-benchmarks/blob/main/package.json#L25-L29)).

> My recommendation is that you use the "Antlr4ng" target in the desc.xml.

I can do that, however, I can see raise eye brows what this means when someone looks for TypeScript actually. But yes, there's more than one TS target. Should I use "Antlr4ng" as target folder name then? This feels kind of awkward, really.

> Trgen will place all .g4's and Antlr4ng/ in one, flat directory. So, import { MySQLBaseLexer, SqlMode } from "../MySQLBaseLexer.js" should be adjusted to import { MySQLBaseLexer, SqlMode } from "./MySQLBaseLexer.js"

That's what I have the biggest trouble with. Using no path here would mean these target specific files would have to be placed side-by-side with the grammars. In the ANTLR4 grammar folder I saw that each target has an own subfolder with the specific files. However, when I keep these files in a separate folder, I need the paths, or the demo won't work. But wait, I could try to use path mapping in tsconfig.json to add other import folders, provided keeping my package.json is ok for you?

mike-lischke:
@kaby76 I changed the paths and desc.xml, but the build still fails. It looks like you are using the wrong antlr4ng runtime version, however _scripts/templates/Antlr4ng/package.json loads the right one.

kaby76:
The name of the "Antlr4ng directory" is missing the lower-case "L". trgen copies all the code in [Antr4ng](https://github.com/mike-lischke/grammars-v4/tree/8ac94eeed0d63706834bcf4bc132c39a962fc672/sql/mysql/Oracle/Antr4ng) into `Generated-Antlr4ng/Antr4ng`, and so it cannot find the parser base class file. The directory contains generated code, too. https://github.com/mike-lischke/grammars-v4/blob/8ac94eeed0d63706834bcf4bc132c39a962fc672/sql/mysql/Oracle/Antr4ng/MySQLLexer.ts#L1

mike-lischke:
I see, thanks. For the tests: how do you execute them? MySQL statement files contain multiple statements, so you have to split them first before you can feed any to the MySQL parser.

kaby76:
The build is fail now because test.ts requires some support software that you don't have in your .json files, as those in Antlr4ng overwrite the driver .json files (trgen does a `cp -r Antlr4ng/* .; rm -rf CSharp/ Antlr4cs/ Java/ Go/ JavaScript/ PHP/ Python3/ Cpp/ Dart/`. It removes all other targets because some compilers like C# try to compile source from every sub-directory. You should put your demo.ts and .json files in Antlr4ng/mydemo.

The tester tests each file from the start production. If the test files need to be split, then you'll need to make an individual test.sh for your test (copy https://github.com/antlr/grammars-v4/blob/1632e26ff88e08fc4d557cf1b776562407e213ab/_scripts/templates/Antlr4ng/st.test.sh, modify) and check that into Antlr4ng/. Or, break up the test files to be an individual test. Or, pick a differ start rule.

kaby76:
And, yes, it's likely I will need to change the command-line call to the Antlr4ng Tool when the tool goes completely to TypeScript. (You may be already there. You seem to be the only one working on "Antlr" proper--"official" version 4 Parr doesn't want to update, and "Antlr5" seems dead.) Right now, the build calls `java -jar` of the .jar in the download directory. https://github.com/antlr/grammars-v4/blob/1632e26ff88e08fc4d557cf1b776562407e213ab/_scripts/templates/Antlr4ng/st.build.sh#L16

mike-lischke:
Oh my, that's getting too far for my taste. So many compromises and extra handling. Can I somehow disable example parsing here? My antlr4ng package uses the same grammars. That should be good enough. Just test if formatting is ok and that's it. Is that possible?

And about the TypeScript targets in general: what about making antlr4ng the only TS runtime here? So you don't need to separate them and use a more self speaking name (`TypeScript`) instead of a package name, which will surely confuse people.

And about antlr4ng-cli: we can head over to the antlr4ng repo and discuss that problem there. Should be easy to fix.

kaby76:
> Can I somehow disable example parsing here? My antlr4ng package uses the same grammars. That should be good enough. Just test if formatting is ok and that's it. Is that possible?

Yes. Add these files.
* Antlr4ng/build.sh:
```
set -x
set -e
rm -rf node_modules package-lock.json
npm install -g typescript ts-node
npm install
java -jar ./node_modules/antlr4ng-cli/*.jar -encoding utf-8 -Dlanguage=TypeScript   MySQLLexer.g4
java -jar ./node_modules/antlr4ng-cli/*.jar -encoding utf-8 -Dlanguage=TypeScript   MySQLParser.g4
exit 0
```
* Antlr4ng/build.ps1
```
$(& npm install -g typescript ts-node ; $compile_exit_code = $LASTEXITCODE ) | Write-Host
if($compile_exit_code -ne 0){
    exit $compile_exit_code
}
$(& npm install ; $compile_exit_code = $LASTEXITCODE ) | Write-Host
if($compile_exit_code -ne 0){
    exit $compile_exit_code
}
$jarFile = Get-ChildItem ./node_modules/antlr4ng-cli/*.jar
$(& java -jar $jarFile.FullName MySQLLexer.g4 -encoding utf-8 -Dlanguage=TypeScript   ; $compile_exit_code = $LASTEXITCODE) | Write-Host
if($compile_exit_code -ne 0){
    exit $compile_exit_code
}
$(& java -jar $jarFile.FullName MySQLParser.g4 -encoding utf-8 -Dlanguage=TypeScript   ; $compile_exit_code = $LASTEXITCODE) | Write-Host
if($compile_exit_code -ne 0){
    exit $compile_exit_code
}
exit 0
```
* Antlr4ng/test.sh
```
exit 0
```
* Antlr4ng/test.ps1
```
exit 0
```


Sorry for the compromises. Unfortunately, Antlr4 did not solve multiple targets well, and it never addressed testing. Also, I had to backward support  the old Java-only Antlr Tester. And, another developer didn't want to use Bash, and instead wrote Powershell scripts for testing.

The driver generator needs a round of new requirements and rewrite.

mike-lischke:
> > Can I somehow disable example parsing here? My antlr4ng package uses the same grammars. That should be good enough. Just test if formatting is ok and that's it. Is that possible?
> 
> Yes. Add these files.

Much simpler: do not list the target in desc.xml.

> Sorry for the compromises. Unfortunately, Antlr4 did not solve multiple targets well, and it never addressed testing. Also, I had to backward support the old Java-only Antlr Tester. And, another developer didn't want to use Bash, and instead wrote Powershell scripts for testing.

No worries, Ken. You are doing a great job here and in other places, for the community. I'm not sure you receive the due respect for that. You have mine at least!

Now that I have disabled testing examples the validation succeeded. I use the grammar on a daily basis and have several public Github projects where it is used (only the formatting is different, as I actually prefer a more compact style), so I'm 100% it is working.

teverett:
@mike-lischke  thanks!

Issue #4105, opened 2024-05-21
Fortran90:  status can be a legit variable, not always a keyword


suehshtri:
What I was doing: parsing some fortran code, as one is want to do.
What I expected: a parse with no errors.
What happened instead: errors
  - no viable alternative at input '(argument_A, ...
  - offending token: 'status'
 
In this case, status is local variable. Looks like it gets caught by STATUS in lexer.

The standard: mentions STATUS Referenced f9x.book, page 111.  

Fortran found in the wild presents the use of a variable named 'status' outside of those contexts in which it is a language word, compileable by gfortran and Intel Fortran compilers.

github refused f90 file.

Reproduction appears to be as easy as 
declaring a variable `integer(4) :: status` in a subroutine.
.

kaby76:
> The standard: mentions STATUS Referenced f9x.book, page 111.

Just to be clear, you are referring to [the N692.pdf draft](https://wg5-fortran.org/N001-N1100/N692.pdf), the last revision before publishing [ISO 1539:1990](https://www.iso.org/standard/6128.html).

Normally, it should be parsed as an [entity-decl](https://github.com/antlr/grammars-v4/blob/438da1ea2c6e0fe837a19ed3dedc4e7dd209358d/fortran/fortran90/Fortran90Parser.g4#L488). It doesn't say in the draft what an `object-name` is, but in [ISO 1539:2023](https://www.iso.org/standard/82170.html), it does: `R804 object-name is name` (p 104) => `R603 name is letter [ alphanumeric-character ] ...
C601 (R603) The maximum length of a name is 63 characters.`

That implies that we should have a rule `objectName : name;` and `name : NAME | STATUS | ......;`. However, with such additional rules, ambiguity (and therefore performance) might become a problem. But, it is something that should be tried.

kaby76:
See https://github.com/antlr/grammars-v4/pull/4106 for a fix.

suehshtri:
> See #4106 for a fix.

I suggested a further test.  I hope it helps.  My test file worked but I quickly created another one to show a related problem.

kaby76:
> I suggested a further test. I hope it helps. My test file worked but I quickly created another one to show a related problem.

Sorry but where are these input? It is not with the comment nor over in the thread for the PR. Please provide.

kaby76:
> Odd. The comments still say "Pending" on 4106. I'll copy them here.

[...]

PR #4106, opened 2024-05-23
[fortran90] Fix "status" as variable name

kaby76:
This is a fix for #4105 

I changed the rule for `objectName` and added the rule `name`, which includes any `NAME` token and all keywords. This implements the spec. I added tests for the fix.

The fix did not significantly degrade the performance, although there appears to be a tiny slowdown due to extra tree nodes being constructed.

times-fortran-fortran90

There is no ambiguity detected for all tests.
```
05/24-06:10:05 ~/issues/g4-4105/fortran/fortran90/Generated-CSharp-large
$ trperf -c a `find ../examples/ -name '*.f90'` 2> /dev/null | sort -u
0
05/24-07:26:25 ~/issues/g4-4105/fortran/fortran90/Generated-CSharp-large
```
(See [trperf](https://github.com/kaby76/Domemtech.Trash/blob/main/src/trperf/readme.md) options on what column specifiers mean.)


kaby76:
Tests generated by this Bash script:
```
for word in `trparse -t ANTLRv4 Fortran90Parser.g4 | trxgrep ' //parserRuleSpec[RULE_REF/text() = "name"]//TOKEN_REF/text()' | grep -v '^NAME$'`
do
	echo $word
	w=`echo $word | tr '[:upper:]' '[:lower:]'`
	cat examples/4105.f90 | sed 's/status/'$w'/g' > examples/4105-$w.f90
done
```

teverett:
@kaby76  thanks!

PR #4111, opened 2024-05-27
Removed copyrights in the MySQL grammars for generated code.

mike-lischke:
Users of the grammars can of course use whatever they want for their generated code.


teverett:
@mike-lischke  thanks

PR #4112, opened 2024-05-28
Fix grammars.json

raf555:
There is a trailing comma in the `grammars.json` causing the lab website to not load the grammars properly due to invalid JSON.

Errors:

Grammars dropdown is not loaded.
image

`grammars` is parsed as string.
image

it's causing exception, but it's not logged so I had to use debugger.
image

### DIFF
--- a/grammars.json
+++ b/grammars.json
@@ -40,9 +40,20 @@
   ]
  },
  {
+  "name": "action",
+  "lexer": "",
+  "parser": "https://raw.githubusercontent.com/antlr/grammars-v4/master/action/action.g4",
+  "start": "file_",
+  "example": [
+   "example1.txt",
+   "example2.txt",
+   "primes.txt"
+  ]
+ },
+ {
   "name": "Ada2005",
-  "lexer": "https://raw.githubusercontent.com/antlr/grammars-v4/master/ada\\ada2005/Ada2005Lexer.g4",
-  "parser": "https://raw.githubusercontent.com/antlr/grammars-v4/master/ada\\ada2005/Ada2005Parser.g4",
+  "lexer": "https://raw.githubusercontent.com/antlr/grammars-v4/master/ada/ada2005/Ada2005Lexer.g4",
+  "parser": "https://raw.githubusercontent.com/antlr/grammars-v4/master/ada/ada2005/Ada2005Parser.g4",
   "start": "compilation",
   "example": [
    "pkg1.adb",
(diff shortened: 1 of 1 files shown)


kaby76:
You are correct. The [grammars.json](https://github.com/antlr/grammars-v4/blob/b91b6dae10165db9000fabd9c8ee3b966da2d4cc/grammars.json) spec file is invalid (using https://jsonlint.com/). This will fix that file, but this file is generated by running [mkindex.py](https://github.com/antlr/grammars-v4/blob/b91b6dae10165db9000fabd9c8ee3b966da2d4cc/_scripts/mkindex.py) periodically then checked in. (It used to be generated automatically on a build, but there were objections to doing that, and I am not an admin on either [grammars-v4](https://github.com/antlr/grammars-v4) or [antlr4-lab](https://github.com/antlr/antlr4-lab) repo, I couldn't work out a Github Artifact solution.)

This can go in, but we need to update the program to generate correct json, fix the program to validate the json, then regenerate and check in the file. There are probably a few grammars in the repo since last generated.

kaby76:
It was hand-edited here: https://github.com/antlr/grammars-v4/commit/438da1ea2c6e0fe837a19ed3dedc4e7dd209358d#diff-26305fd58b600bcacd5b771eb3e9ecbc1c7822293b5e0e319bc5855c3010f7ad. The file must not be hand edited.

Here is the file regenerated: 
[grammars.json](https://github.com/antlr/grammars-v4/files/15467976/grammars.json)

raf555:
Hi @kaby76 thank you for the heads up. I've followed your suggestion and regenerated the file with the provided script.

kaby76:
I don't know why, but your grammars.json file has backslashes in paths. https://github.com/raf555/grammars-v4/blob/cb627672565b4597f2fe19fc0ac0de1c4ff89f3f/grammars.json#L55 It should not.

raf555:
The existing file also has backslashes. I was thinking that it's okay since it's being redirected to the correct content anyway. Just FYI I run the script on windows.

Wdyt? Should I change it with your file instead?

kaby76:
> Just FYI I run the script on windows.

I do too. But, I cannot stand cmd.exe nor pwsh.exe, rather MSYS2 or the original Cygwin. I am rechecking the mkindex.py tool to make sure all backslashes are rewritten and will work on fixing that.

kaby76:
The problem is that the parser and lexer file names are not fixed. [here](https://github.com/antlr/grammars-v4/blob/b91b6dae10165db9000fabd9c8ee3b966da2d4cc/_scripts/mkindex.py#L117) [here](https://github.com/antlr/grammars-v4/blob/b91b6dae10165db9000fabd9c8ee3b966da2d4cc/_scripts/mkindex.py#L119) [here](https://github.com/antlr/grammars-v4/blob/b91b6dae10165db9000fabd9c8ee3b966da2d4cc/_scripts/mkindex.py#L125). They require `.replace('\\', '/')` at the end of the expression.

kaby76:
Looks good. Thanks.

teverett:
@raf555  thanks

PR #4116, opened 2024-05-29
Fix #4115 -- _scripts/mkindex.py does not work on Windows Powershell

### LINKED ISSUE #4115: _scripts/mkindex.py doesn't work on Windows with Pwsh.exe
https://github.com/antlr/grammars-v4/pull/4112#issuecomment-2135214265

kaby76:
This PR fixes #4115. The problem was that on Windows Powershell, file paths use backslashes, not forward slashes. To set the URL for the location of the grammar file, it must use forward slashes.

https://github.com/antlr/grammars-v4/pull/4112 is a companion PR that fixes the grammars.json file itself. This file is used by the lab.antlr.org.


teverett:
@kaby76  thanks

PR #4117, opened 2024-05-29
Fix #4110 -- nameDataRef is incorrect

### LINKED ISSUE #4110: [fortran90] "primary" rule deviates from the Draft Spec, with no explanation, and does not work!
https://github.com/antlr/grammars-v4/issues/4105#issuecomment-2132459469

kaby76:
This PR fixes nameDataRef so that keywords in expressions can be used. See https://github.com/antlr/grammars-v4/issues/4105#issuecomment-2132238398. This workaround doesn't change "primary". That will be done in another commit.

Tests were updated to avoid the "C" in column 1 comment problem: the file should be parsed in fixed or free form, but it cannot be both. The lines are shifted to the right a few spaces to avoid the issue for now.


teverett:
@kaby76  thanks!

PR #4119, opened 2024-05-30
[Snowflake]Session and Password policies

MasterKuat:
Refactoring SESSION POLICY and add PASSWORD POLICY related statements


kaby76:
The build failed cause the "perf" workflow tries to install Octave, but can't. Somewhere in the published packages and OSes that the workflow uses was removed, making it impossible to keep a stable environment. Sync with master once https://github.com/antlr/grammars-v4/pull/4122 is merged.

teverett:
@MasterKuat  thanks!

Issue #4121, opened 2024-05-30
[build] "perf" workflow isn't working because Octave cannot install


kaby76:
Something changed in the Github Actions environment such that Octave cannot install. The workflow uses Octave to perform computations for statistical variance. It cannot function without Octave.

PR #4122, opened 2024-05-30
Fix for #4121 -- remove perf workflow

kaby76:
This is a fix for #4121, where the "perf" workflow isn't working because Octave cannot install. The fix is to remove the workflow for now.

This is a temporary workaround until I have time to figure out why "the environment" (published packages, published OSes, etc.) keep changing when I specifically do not want anything to change as it is a royal pain to get CI working if the rug keeps being pulled from under my feet.


teverett:
@kaby76  thanks!

PR #4126, opened 2024-06-04
Update SQLiteParser.g4 by allowing order_by_expr_asc_desc to be empty.

chengniansun:
The current definition of order_by_expr_asc_desc is right recursive and does not terminate.

### DIFF
--- a/sql/sqlite/SQLiteParser.g4
+++ b/sql/sqlite/SQLiteParser.g4
@@ -587,6 +587,7 @@ order_by_expr
 
 order_by_expr_asc_desc
     : ORDER_ BY_ expr_asc_desc
+    |
 ;
 
 expr_asc_desc


sql/sqlite/SQLiteParser.g4
```diff
@@ -587,6 +587,7 @@ order_by_expr
 
 order_by_expr_asc_desc
     : ORDER_ BY_ expr_asc_desc
+    |
```
kaby76, line 590: 1) Please prove that [order_by_expr_asc_desc](https://github.com/antlr/grammars-v4/blob/dd341526dce78130976cf06c0e6881b42ab26eea/sql/sqlite/SQLiteParser.g4#L588-L590) is infinitely recursive by giving a derivation, or by giving a test that shows non-termination of a parse involving the rule. 
1) The only places where `order_by_expr_asc_desc` is in [window_function](https://github.com/antlr/grammars-v4/blob/dd341526dce78130976cf06c0e6881b42ab26eea/sql/sqlite/SQLiteParser.g4#L561-L570), and that is only referenced in [window_function_invocation](https://github.com/antlr/grammars-v4/blob/dd341526dce78130976cf06c0e6881b42ab26eea/sql/sqlite/SQLiteParser.g4#L510-L515). But, `window_function_invocation` is not referenced anywhere in the grammar, and it is not a start rule. With a comment `'// unknown'` above [window_function](https://github.com/antlr/grammars-v4/blob/dd341526dce78130976cf06c0e6881b42ab26eea/sql/sqlite/SQLiteParser.g4#L561), it seems the author of the grammar didn't complete the implementation of the rule.
1) This is just a note on updating this grammar: (a) There is no readme.md, nothing containing a link to https://sqlite.org/windowfunctions.html#biwinfunc or any of the online documentation; (b) no version of what this grammar supports, e.g., version 3.46.0. (All grammars should have a version number of what it is trying to support.)

chengniansun: My bad. I found this bug in a version I downloaded last year. When I created this PR, I did not notice that the bug has been fixed. 

The fix commit is here, 76b1d66d768902f821b7a7440507bfdb9b6a32f9

Issue #4135, opened 2024-06-10
[r] The R grammar contains an ambiguity in the desc.xml


kaby76:
The [R grammar](https://github.com/antlr/grammars-v4/tree/d2d7b89157946345710254cf6e4121f29cfa011f/r) has two grammars: [R.g4](https://github.com/antlr/grammars-v4/blob/d2d7b89157946345710254cf6e4121f29cfa011f/r/R.g4) and [RFilter.g4](https://github.com/antlr/grammars-v4/blob/d2d7b89157946345710254cf6e4121f29cfa011f/r/RFilter.g4). R.g4 is a combined grammar, which specifies both a lexer and a parser. RFilter.g4 is a parser grammar, which uses the lexer from R.g4. For the moment, only grammar R is tested, which originated from the Maven Tester [pom.xml](https://github.com/antlr/grammars-v4/blob/d2d7b89157946345710254cf6e4121f29cfa011f/r/pom.xml#L43).

The desc.xml does not specify which grammar to test. A new version of trgen now performs a topological sort of the grammar dependencies, and has found that the grammars here are ambiguous in testing: do we test grammar R or do we test grammar RFilter? The desc.xml needs to be updated to say which grammar we are testing.

In addition, the grammar RFilter could be tested, but requires the lexer from one grammar and the parser from another. Neither trgen currently, nor the old Maven Antlr tester handle this case.

PR #4136, opened 2024-06-10
[r] Fix for #4135 -- specify the "R" grammar for testing, skip "RFilter"

kaby76:
This is a fix for #4135 where it is unclear which grammar to test. The current version of trgen miraculously picks "R", but that will not be the case in the next version. Using `grammar-name`, the grammar to test is specified, just as it is in the [pom.xml](https://github.com/antlr/grammars-v4/blob/d2d7b89157946345710254cf6e4121f29cfa011f/r/pom.xml#L43).


teverett:
@kaby76  thanks!

PR #4137, opened 2024-06-10
VBA Conditional Compilation Grammar

Beakerboy:
VBA includes a precompiler of sorts, with its own grammar. The existing vba6 grammar tries to shoehorn it in with the main language, but there are situations where a VBA file with CC statements is not valid VBA until after the precompiler is run.
For example:

```vba
#If WIN64 Then
Function Foo(bar, Optional baz = 1)
#Else
Function Foo(bar, Optional baz = 2)
#End If
End Function
```

This new grammar powers a precompiler which can turn this example into valid VBA given a set of environment variables.

### DIFF
--- a/vba/pom.xml
+++ b/vba/pom.xml
@@ -13,6 +13,7 @@
 
     <modules>
         <module>vba6</module>
+        <module>vba_cc</module>
         <module>vba_like</module>
     </modules>
 </project>
--- a/vba/vba_cc/README.md
+++ b/vba/vba_cc/README.md
@@ -1,0 +1,7 @@
+# Visual Basic Conditional Compilation Grammar for ANTLR4
+
+Derived from the Visual Basic 7.1 language reference
+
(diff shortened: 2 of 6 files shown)


Beakerboy:
I see that there is a failure in only one of the build matricies. I’m more than happy to fix this if someone can tell me what it means.

kaby76:
There was a regression in trgen for the names of the include files to `#include` in the driver program. So, that is likely why it doesn't compile. The last version it did work with is trgen 0.22.0. There may be other problems but you'll need to remove the Cpp target for the grammar if you just want to get it merged as is. I will have a new version sometime in a few days.

Beakerboy:
I’ve removed that target. I can read it in a a few days when I add the VBA 7.1 grammar.

Beakerboy:
Renamed tokens after the ninja edit and readers the CPP target. Good to go.

teverett:
@Beakerboy  thanks!

PR #4139, opened 2024-06-11
Add alternative Oracle constraint syntax

Naros:
Based on the changes https://github.com/debezium/debezium/pull/5612


teverett:
@Naros  thanks!

PR #4144, opened 2024-06-20
Fix #4143 [javascript] Fixed the end detection of the placeholder in template literals.

### LINKED ISSUE #4143: [javascript] Parsing fails if there are curly braces inside a template literal
An error will occur when parsing a statement like the one below.

```javascript
var nested = `Object in template literal ${JSON.stringify({ key: 'value' })}`
```

The error that occurs is as follows:

```
line 1:57 mismatched input '(' expecting TemplateCloseBrace
line 1:73 no viable alternative at input '}'
line 2:0 mismatched input '<EOF>' expecting {BackTick, '${', TemplateStringAtom}
```

I suspect that when there are curly braces within a template literal, the `}` is mistaken for the end of the template literal, causing an error.

tanzaku:
This PR fixes #4143.

The parsing error in #4143 seems to be caused by the handling of curly braces inside template literals.
In the following example described in the issue, the error occurs because the `}` in `{ key: 'value' }` is misidentified as the end of the placeholder.

```javascript
var nested = `Object in template literal ${JSON.stringify({ key: 'value' })}`
```

This PR fixes the problem by managing the number of curly braces and correctly recognizing the end of the placeholder.

### DIFF
--- a/javascript/javascript/CSharp/JavaScriptLexerBase.cs
+++ b/javascript/javascript/CSharp/JavaScriptLexerBase.cs
@@ -30,15 +30,14 @@ public abstract class JavaScriptLexerBase : Lexer
     private bool _useStrictCurrent = false;
 
     /// <summary>
-    /// Keeps track of the the current depth of nested template string backticks.
-    /// E.g. after the X in:
-    ///
-    /// `${a ? `${X
-    ///
-    /// templateDepth will be 2. This variable is needed to determine if a `}` is a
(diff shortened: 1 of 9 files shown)


tanzaku:
Thank you for running workflow.
I fixed the failing tests.

teverett:
> Thank you for running workflow. I fixed the failing tests.

All good

teverett:
@tanzaku  thanks!

PR #4147, opened 2024-06-27
[ICalendar] add trailing CRLF to request-status property

maximilian-bauer:
According to a [recently accepted errata](https://www.rfc-editor.org/errata/eid7945), the `REQUEST-STATUS` property should be parsed with a trailing CRLF, as all other properties already are.

This PR adds the CRLF and also fixes a small typo along the way.


teverett:
@maximilian-bauer thanks!

PR #4167, opened 2024-07-22
[MySQL][Fix] statement is supported as the field name

caicancai:


caicancai:
@kaby76 PATL，thank you

sql/mysql/Positive-Technologies/examples/dml_select.sql
```diff
@@ -248,3 +248,6 @@ FROM
     (SELECT @sum := 0) AS init
 ORDER BY
     some_order_column;
+
+--- statement is supported as the field name
+SELECT REPLACE(statement, ' ','') as statement from your_table;
```
KvanTTT, line 253: Are you sure it couldn't be an identifier? If not, please add `STATEMENT` token to [keywordsCanBeId](https://github.com/antlr/grammars-v4/blob/master/sql/mysql/Positive-Technologies/MySqlParser.g4#L2707C1-L2707C16) rule.

caicancai: I can't guarantee that all versions of MySQL are consistent. I agree with your solution. Thank you for your suggestion.

KvanTTT (approved):

teverett:
@caicancai  thanks!

PR #4168, opened 2024-07-23
[Postgres][Imporve] Change the var keyword to the Java base class ProxyErrorListener

caicancai:


caicancai:
@kaby76 PATL, thank you

kaby76:
Please explain why "var" was changed to the explicit type. Thanks.

caicancai:
> Please explain why "var" was changed to the explicit type. Thanks.

Java 8 does not support the var keyword. The var keyword is supported starting from Java 10. When using some older Java 8 versions, the var keyword may report an error.
image

kaby76:
You are right. Java 10 introduced "var". We don't have a way specifying in the desc.xml the minimum version of Java, or any other target language, required. This is all hardwired in the testing templates.

https://github.com/antlr/antlr4/releases#:~:text=class%20files%20for%20the%20ANTLR,bumping%20up%20from%20Java%207).

The PR looks fine.

teverett:
@caicancai  thanks!

Issue #2991, opened 2023-01-05
We should always test the parse trees generated


kaby76:
Perhaps this is better in the Discussions section, but I just don't understand why we aren't performing tree diffs on every parse, across all targets. It's not going to slow down the tests that much and it really will catch errors in Antlr, e.g., css3, mdn-at-counter-style.css, [explained here](https://github.com/antlr/antlr-php-runtime/pull/34#issuecomment-1369788414).

RossPatterson:
We should at least compare the trees from each target for every test, and ensure that all targets produce the same tree.  We could do that without needing any specific effort in any of the grammars.  Ideally, every test would have an expected parse tree, and we would include that in the comparison as well, ensuring that all targets produce the expected tree.  Unfortunately, there are very few expected tree files in the `grammars-v4` repo right now - I checked them all during my work to improve testing of parse tree files.  The downside of this is that most changes to a grammar would cause tests to fail, until the expected tree was re-generated, manually verified, and committed.

We could arbitrarily assert that the current parse tree of every test that succeeds is by definition correct, and check that tree into the repo as the expected result.  We might even be correct for well over 90% of the tests.

kaby76:
> We should at least compare the trees from each target for every test, and ensure that all targets produce the same tree. We could do that without needing any specific effort in any of the grammars. Ideally, every test would have an expected parse tree, and we would include that in the comparison as well, ensuring that all targets produce the expected tree. Unfortunately, there are very few expected tree files in the `grammars-v4` repo right now - I checked them all during my work to improve testing of parse tree files. The downside of this is that most changes to a grammar would cause tests to fail, until the expected tree was re-generated, manually verified, and committed.
> 
> We could arbitrarily assert that the current parse tree of every test that succeeds is by definition correct, and check that tree into the repo as the expected result. We might even be correct for well over 90% of the tests.

I think the plan is to generate a tree for every parse and compare with the previous. This will be across targets which will likely result in new bugs being discovered. To generate the .tree files, use either the trgen-generated code to remaster the .tree files or create by hand. We'll need to consider what to do with people who are completely phobic about installing dotnet and trgen (works everywhere), and running the driver code for any one of the targets they want in order to remaster the .tree and .errors files.

kaby76:
I'm starting to address this issue. The easy solution is to just check in all the .tree files. But, people will likely have a hard time remastering the files, and some will likely forget to do it. So, the idea to compare the trees across targets is probably the way to go.

[...]

Issue #748, opened 2017-06-07
Format all .tree files


KvanTTT:
Add indents and line breaks to all .tree files.

It's not convenient to view and compare .tree files as single lines. See https://github.com/antlr/grammars-v4/blob/master/vb6/examples/calls/Calls.cls.tree for example.

Also see this issue for detail: https://github.com/antlr/antlr4/issues/1502.

PR #4169, opened 2024-07-23
Add in parse tree testing for every grammar, every input file

kaby76:
This repo currently tests the outputted CST of a parse for a small subset of grammars. With new versions of Antlr coming to fore (different targets, e.g., Wasm, Rust; rewrites, e.g., Antlr4ng), it's important to make sure these produce identical parse trees, checking the correctness of the generated parsers, runtime, and ports. This PR adds in parse tree testing for all test input, for all grammars.

This fixes two issues:
* #748 
* #2991 

All .tree files are removed and replaced with .ipt "Indented Parse Tree" files. These files are human readable, with each node in the parse tree on a separate line and indented proportional to the depth of the node in the CST. The old Maven tester still works, but only for quick sanity checking of grammar changes. The old tester is no longer used for comparing the outputted CSTs. All input files that are parsed now have a corresponding `.ipt` file, which must be kept in sync with the grammar and test file.

The code for formatted parse trees is implemented in the driver code for testing in this repo. Antlr does not have any implementation of formatted parse trees, and likely never will, as it is difficult to get buy-in on any new features.

Going forward, people will need to update the .ipt with a change in the grammar. They can do this by:
* running trgen locally;
* or, running a build and copying and pasting the resulting trees to the .ipt's.

People will need be careful of the newline characters of both input and .ipt files generated by the parser. The .ipt files must be in Unix-style endings for easy "git diff" comparisons. Any input files for testing should probably be in "eol=lf" representation in the git repo. You will probably need to update the .gitattributes file.


kaby76:
I'm going to close this off because it's low-priority. I'd rather see ambiguity and fallback cleaned up. Improving the speed of parsing will be more impactful and help change the pervasive view that Antlr is "garbage", "slow", etc.

Issue #4172, opened 2024-07-26
Various grammars still not following standard layout for Java target ports.


kaby76:
I just noticed that the cobol85/java/ directory exists. I also find other similar non-standard layouts for the Java port.
```
$ find . -name java -type d
./cobol85/Generated-Antlr4ng/java
./cobol85/java
./java
./java/java
./less/testsrc/test/java
./stringtemplate/src/main/java
./swift/swift2/java
./swift/swift2/src/main/java
./z/java
./z/src/main/java
```

The Java target is "Java" not "java". The directory containing Java code should be placed in the directory called "Java/", not "java/", not "src/", etc.

kaby76:
Fixed.

PR #4173, opened 2024-07-26
Fix for #4172 -- various grammars use non-standard directories for Java source.

kaby76:
Fix for #4172.

Currently, there are a few grammars that contain directories for Java source code in some non-standard directories, like "src/", "java/" (should be capital "J" as in "Java/"), "testsrc/", etc. These files should be placed in the Java/ directory because they have nothing to do with the other target ports. Besides, if we want to write the equivalent of these codes in other targets in the future, we'd end up questioning where to put all this stuff.

This PR just avoids problems by moving the Java source code for testing under Java/.


teverett:
@kaby76  thanks

PR #4179, opened 2024-07-28
Structured Definition Language (SDL)

whcox603:
The ANTLR4 grammar is based on the source code for SDL that has been released as freeware.

- formatted using antlrformat
- all examples parse cleanly
- tests passed with all languages
- Some sort of kerfluffle with php, which I believe is my environment (not a test failure)


teverett:
@whcox603  thanks!

PR #4193, opened 2024-08-06
[Snowflake]Improvement DYNAMIC TABLE

MasterKuat:
Review of CREATE and ALTER statement for[ DYNAMIC TABLE](https://docs.snowflake.com/en/sql-reference/sql/create-dynamic-table).

There is also a small refactoring for rls (ROW ACCESS POLICY) on table as these operations are shared with dynamic table.


teverett:
@MasterKuat thanks

PR #4198, opened 2024-08-09
[php] Fix #4196 -- fix alt labels, clean up code.

### LINKED ISSUE #4196: PHP bitshifts with wrong labels
at https://github.com/antlr/grammars-v4/blob/feb9e9e082842d9cafed1e1558e9d1aad7052b87/php/PhpParser.g4#L510
the bitshifts are labelled as `ComparisonExpression` when they are, in fact, `ArithmeticExpression`

kaby76:
* Renamed alt labels to conform to the names in the spec at https://github.com/php/php-langspec/blob/master/spec/19-grammar.md (#4196).
* Removed Python/ directory because there is no such thing as a "Python", only "Python3" at this point.
* Removed unused parser rule `magicMethod`.
* Added `<inputs>` element to make the test file inputs explicit.


teverett:
@kaby76  thanks

PR #4200, opened 2024-08-09
[ecmascript] Fix #4199 -- rename the Python/ target specific directory to Python3/

### LINKED ISSUE #4199: [ecmascript] No such thing as "Python" target; it is called "Python3".
It is not correct to type `antlr4 -Dlanguage=Python *.g4`. There is no such thing as a "Python" target. It can only be `antlr4 -Dlangauge=Python3 *.g4`. So, the directory [Python/](https://github.com/antlr/grammars-v4/tree/feb9e9e082842d9cafed1e1558e9d1aad7052b87/javascript/ecmascript/Python) should be renamed to Python3/.

kaby76:
* Renamed Python/ to Python3 as there is no "Python" target--only "Python3".
* NB: I have not tested the grammar to verify that it works, or to change it to target-agnostic format.


teverett:
@kaby76  thanks

PR #4202, opened 2024-08-10
[build] Fix #4201 -- remove "import" grammars from grammars.json because lab.antlr.org cannot handle imported grammars

### LINKED ISSUE #4201: [build] mkindex.py generates testing for grammars that contain "import", it shouldn't
mkindex.py generates a [grammars.json file for the kotlin grammars](https://github.com/antlr/grammars-v4/blob/feb9e9e082842d9cafed1e1558e9d1aad7052b87/grammars.json#L2067-L2187). It shouldn't because they [contain an import statement](https://github.com/antlr/grammars-v4/blob/feb9e9e082842d9cafed1e1558e9d1aad7052b87/kotlin/kotlin-formal/KotlinLexer.g4#L19). The reason it can't is because lab.antlr.org only has two tabbed text boxes for grammar files: a lexer tab and a parser tab. An imported grammar can't be appended into either of these tabs.

kaby76:
* Fix for #4201 
* Reindexed the grammars (i.e., create new version of grammars.json file) with the latest changes.

### DIFF
--- a/_scripts/mkindex.py
+++ b/_scripts/mkindex.py
@@ -106,6 +106,12 @@ def index_grammars(root : str) -> Sequence[dict]:
                     bar = p3.stdout.readline()
                     if bar:
                         skip = True
+                    p1 = subprocess.Popen(["dotnet", "trparse", "-t", "ANTLRv4", foo], stdout=subprocess.PIPE)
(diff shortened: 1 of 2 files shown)


teverett:
@kaby76  thanks

PR #4206, opened 2024-08-14
[php] Fix #4205 -- fix precedence of 'require' and 'include' operators

### LINKED ISSUE #4205: [php] Left recursive expression parsing is not suited for some rules
I attempted to parse the following as an expression.
`<?php
require_once 'a' . 'b'`

The issue I see is that the associativity determines `(. (require_once ('a')) ('b'))`, so the concatenation is done between `require_once 'a'` and `'b'`. I was expecting to have the concatenation as a sub-tree of `require_once`, between 'a' and 'b' . The rules that are used in the parser are:

`expression
   : (Require | RequireOnce) expression
   | expression op = ('+' | '-' | '.') expression 
   ;`

I suppose this is due to how left recursive parser works, delimiting the recursive rule and the base case rule. I am looking forward for a mean of convincing the parser that the `(Require | RequireOnce) expression` rule is in fact a high priority rule and should not be referred as a base case of the left recursive feature. 

I considered rewriting the PHP parser to extract these cases in another expression rule, but I think it defies the purpose of having everything in one fat expression rule.

kaby76:
* Fixed #4205 by moving 'require' and 'include' alts below arithmetic alts, below the spaceship alt. It is not clear where exactly it should be but this seems "safe".
* Changed the name of alt labels for 'require' and 'include' alts to conform to the spec at https://github.com/php/php-langspec/blob/master/spec/19-grammar.md#grammar-include-expression.
* Added test for 'require' and 'include', in a variety of ways. Included "config.php" to actually test how it parses and works in PHP8.3.


teverett:
@kaby76 thanks

PR #4160, opened 2024-07-10
Improved ANTLR Grammar for R Code in JPlag

TwoOfTwelve:
The JPlag repository has two large examples of R code (can be found here: [JPlag R examples](https://github.com/jplag/JPlag/tree/main/languages/rlang/src/test/resources/de/jplag/rlang)). I encountered several issues with these two files that I tried to address.

The problems were:
- In exprlist ANTLR seemed to prefer the empty option always. I resolved this by removing the empty option and making exprlist optional within expr.
- The grammar didn't allow for line-breaks before expressions, so I added an option "NL expr" for expr.
- There were no linebreaks allowed between an if block and the else keyword

I should mention that I know neither ANTLR nor R well, so it would be nice if someone could look this over.

PR #4208, opened 2024-08-14
R parser improvements

TwoOfTwelve:
In its current state the R grammar is hard to use, since there is one parser rule containing almost everything. Additionally, most tokens are defined implicitly. Therefore the AST elements that are generated can only be distinguished by comparing the strings in the tokens. That requires a lot of hard-coded string literals in the application code.

To make using the AST easier, this PR defines most tokens explicitly and breaks up the parser rules into smaller ones. I tried to do that as general as possible, but I focused on what is needed for [JPlag](https://github.com/jplag/JPlag).
With these changes the AST nodes can be distinguished by their types. Different cases can be checked via the names of the tokens instead of using a string comparison.
Terminal nodes can be identified through named constants.

As far as I can tell the functionality of the grammar has not changed. I would appreciate it if someone could confirm that.

This PR is a follow up to #4160.

### DIFF
--- a/r/R.g4
+++ b/r/R.g4
@@ -49,7 +49,7 @@ $ java TestR sample.R
 grammar R;
 
 prog
-    : (expr)* EOF
+    : (expr (SEMICOLON | NL)* | NL)* EOF
     ;
 
 /*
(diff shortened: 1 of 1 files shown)


kaby76:
The rewrite of `expr` will likely impact the performance. You should check this. Have you considered the alternative that uses alt labeling?

PR #4211, opened 2024-08-19
[Snowflake]Add with tag for stream

MasterKuat:
CREATE STREAM support WITH TAG syntax even if is not in documentation.
Request to update doc https://docs.snowflake.com/en/sql-reference/sql/create-stream is done.


teverett:
@MasterKuat  thanks

Issue #4212, opened 2024-08-20
[kotlin/kotlin] Grammar should be marked available


kaby76:
As noted in https://github.com/antlr/grammars-v4/issues/3965, [kotlin/kotlin-formal is tested](https://github.com/antlr/grammars-v4/blob/2a3a6bdcbe1f9b975591d194d69df22e336dd7da/kotlin/kotlin-formal/desc.xml#L3), but [kotlin/kotlin is not](https://github.com/antlr/grammars-v4/blob/2a3a6bdcbe1f9b975591d194d69df22e336dd7da/kotlin/kotlin/desc.xml#L3). I think it's okay to enable the second grammar. That said, everything I noted in https://github.com/antlr/grammars-v4/issues/3965 is still true.

PR #4213, opened 2024-08-20
[kotlin] Fix 4212 -- add supported targets for grammar.

kaby76:
This PR updates the desc.xml and remasters a few .errors files for the grammar, fixing #4212.


teverett:
@kaby76  thanks

PR #4215, opened 2024-08-21
Feature/r parser labels

TwoOfTwelve:
In its current state the R grammar is hard to use, since there is one parser rule containing almost everything. Additionally, most tokens are defined implicitly. Therefore the AST elements that are generated can only be distinguished by comparing the strings in the tokens. That requires a lot of hard-coded string literals in the application code.

To make using the AST easier, this PR defines most tokens explicitly and adds labels to all cases of "expr".
With these changes the AST nodes can be distinguished by the generated *Context classes.
Terminal nodes can be identified through named constants.

As far as I can tell the functionality of the grammar has not changed. I would appreciate it if someone could confirm that.

For some reason I had to change the prog rule, since it clashed with the `NL+ expr` option in expr. As far as I can tell, this version works and does the same.

This PR is a follow up to https://github.com/antlr/grammars-v4/pull/4160.

### DIFF
--- a/r/R.g4
+++ b/r/R.g4
@@ -49,7 +49,7 @@ $ java TestR sample.R
 grammar R;
 
 prog
-    : (expr (SEMICOLON | NL)* )* EOF
+    : ((SEMICOLON | NL)+ | expr )* EOF
     ;
 
 /*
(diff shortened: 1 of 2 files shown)


teverett:
@TwoOfTwelve  thanks

Issue #4218, opened 2024-08-24
[build] Update find-useless-parentheses.sh


kaby76:
Three bugs have been fixed for the Trash script [find-useless-parentheses.sh](https://github.com/kaby76/g4-scripts/blob/5d94da10c285fbbd69eac3b3a429b8ccec4513fa/find-useless-parentheses.sh). 
* https://github.com/kaby76/g4-scripts/issues/1
* https://github.com/kaby76/g4-scripts/issues/2
* https://github.com/kaby76/g4-scripts/issues/3

Testing here gets an older version. https://github.com/antlr/grammars-v4/blob/2a3a6bdcbe1f9b975591d194d69df22e336dd7da/_scripts/test-static-checks.sh#L360. It should be updated.

MasterKuat:
Hi @kaby76, seems this fix break the test.
Please see https://github.com/antlr/grammars-v4/actions/runs/10629557556/job/29466637942?pr=4223
```
sql/snowflake,useless-parens:
main: line 19: trparse: command not found
Computing useless parentheses in SnowflakeLexer.g4 SnowflakeParser.g4
main: line 21: trparse: command not found

```

kaby76:
Sorry, the script was changed to use a ["global" trparse](https://github.com/kaby76/g4-scripts/blob/5d94da10c285fbbd69eac3b3a429b8ccec4513fa/find-useless-parentheses.sh#L19) (i.e., calling "trparse" from the command line), when it should have stayed with the ["local" trparse](https://github.com/kaby76/g4-scripts/blob/9e70a6392783936ddd51cd270f6ac29fecbefef8/find-useless-parentheses.sh#L19) (calling "trparse" via "dotnet").

Will fix today.

PR #4224, opened 2024-08-30
[build] Fix find useless parentheses script

kaby76:
As noted [here](https://github.com/antlr/grammars-v4/issues/4218#issuecomment-2320520804), so I updated the versions of the Trash toolkit and script to find useless parentheses. A couple of grammars are altered innocuously to perform a sanity check.


teverett:
@kaby76  thanks!

PR #4234, opened 2024-09-12
add_clickhouse_g4

wangyang377:
Based on the G4 file in the clickhouse source code, I made some modifications to adapt to the static SQL parsing of antlr4.

### DIFF
--- a/sql/clickhouse/ClickHouseLexer.g4
+++ b/sql/clickhouse/ClickHouseLexer.g4
@@ -1,0 +1,296 @@
+
+// $antlr-format alignColons trailing, alignLabels true, alignLexerCommands true, alignSemicolons ownLine, alignTrailers true
+// $antlr-format alignTrailingComments true, allowShortBlocksOnASingleLine true, allowShortRulesOnASingleLine true, columnLimit 150
+// $antlr-format maxEmptyLinesToKeep 1, minEmptyLines 0, reflowComments false, singleLineOverrulesHangingColon true, useTab false
+
+lexer grammar ClickHouseLexer;
+
+// NOTE: don't forget to add new keywords to the parser rule "keyword"!
+
+// Keywords
+
+ADD          : A D D;
+AFTER        : A F T E R;
+ALIAS        : A L I A S;
+ALL          : A L L;
+ALTER        : A L T E R;
+AND          : A N D;
+ANTI         : A N T I;
+ANY          : A N Y;
+ARRAY        : A R R A Y;
+AS           : A S;
+ASCENDING    : A S C | A S C E N D I N G;
+ASOF         : A S O F;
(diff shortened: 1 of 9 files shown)


wangyang377:
When I want to directly use the G4 file in the source code, I encounter the following problem.
Snipaste_2024-09-12_10-57-13
Now, everything is done.

caicancai:
@kaby76 PATL, thanks

wangyang377:
@teverett  Please take a look, thank you very much

sql/clickhouse/ClickHouseLexer.g4
```diff
@@ -0,0 +1,296 @@
+
+// $antlr-format alignColons trailing, alignLabels true, alignLexerCommands true, alignSemicolons ownLine, alignTrailers true
+// $antlr-format alignTrailingComments true, allowShortBlocksOnASingleLine true, allowShortRulesOnASingleLine true, columnLimit 150
+// $antlr-format maxEmptyLinesToKeep 1, minEmptyLines 0, reflowComments false, singleLineOverrulesHangingColon true, useTab false
+
+lexer grammar ClickHouseLexer;
```
KvanTTT, line 6: Could you please use [`caseInsensitive`](https://github.com/antlr/antlr4/blob/dev/doc/options.md#caseinsensitive) option and replace fragments with string literals? See [SQL](https://github.com/antlr/grammars-v4/tree/master/sql) grammars as an example.

KvanTTT (changes requested):

sql/clickhouse/ClickHouseLexer.g4
wangyang377: You mean That will be a big work and make little sense

wangyang377: Thank you for your advice. But replacing all fragments will make a big change to the g4 file in clickhouse source code and  makes little sense to the lexer result. Is it necessary to do this format change?

KvanTTT: It's not necessary, but desired since it makes grammar more clear and other grammars in the repository uses this option.

KvanTTT (approved):

sql/clickhouse/ClickHouseLexer.g4
wangyang377: Thanks. I will consider optimizing it when I have the chance.

teverett:
@wangyang377  thanks

pkit:
@wangyang377 just FYI, this is not even close to actual ClickHouse parser.
It's clearly stated in CH sources too: https://github.com/ClickHouse/ClickHouse/tree/master/utils/antlr#this-parser-is-unsupported

PR #4237, opened 2024-09-14
Add license for HTTP grammar

Marti2203:
I had made a simple HTTP grammar a couple of years ago and at that time the rules for license inclusion were not well defined.
I was recently contacted to remediate this.


kaby76:
Looks like a regression for the tester. Will need to fix this before rerunning the tests.

kaby76:
This grammar just does not work. https://github.com/antlr/grammars-v4/issues/4244

The grammar didn't have a proper EOF-terminated start rule. This caused the parser to "accept" the two inputs (test.http and testHeader.http). But, the parse trees were empty. By design, Antlr rewinds the parser to an acceptable state before an error. In addition, no errors are reported.

Marti2203:
Hi, I will fix it then. Unfortunately, I really do not remember from 5 years ago, so I will spend some time fixing it and make the grammar a better shape.

kaby76:
Much appreciated. It's going to take some time to fix trgen and add regression tests so the build doesn't break too often.

kaby76:
@Marti2203 Could you synch with the latest in master? The updates for build regressions should fix the build for this PR. Thanks!

teverett:
@Marti2203  thanks

PR #4242, opened 2024-09-15
added TypeScript target and some modifications

RobEin:
- added TypeScript target
- update to ANTLR 4.13.2
- update to Python 3.12.6


kaby76:
`update to ANTLR 4.13.2`. Could you describe which targets require 4.13.2 here? There is no version 4.13.2 for the CSharp target.

RobEin:
The two following two parsers require ANTLR 4.13.2:
 - JavaScript target for Python 3 (because of depricated features: [#4562](https://github.com/antlr/antlr4/pull/4562))
 - TypeScript target for Python 3 (because of new features: [#4562](https://github.com/antlr/antlr4/pull/4562))

kaby76:
Now I understand. The JavaScript port requires Antlr 4.13.2 because this PR uses the API [Lexer.setMode()](https://github.com/antlr/antlr4/blob/2a7904a595479954ccfb1c78124fc3d7f5bebcb1/runtime/JavaScript/src/antlr4/Lexer.js#L180-L182). Unfortunately, the build locks all grammars that have a JavaScript port at 4.13.2.  So, the old API [mode()](https://github.com/antlr/antlr4/blob/2a7904a595479954ccfb1c78124fc3d7f5bebcb1/runtime/JavaScript/src/antlr4/Lexer.js#L175-L178) outputs a deprecation warning, which is a breaking change. I don't think there are many JS ports that use the old API, but I will need add per-grammar, per-version testing when specified in the desc.xml.

kaby76:
Please synch with "master" to see whether the build is fixed.

teverett:
@kaby76  ready to merge?

kaby76:
Yes, please merge.

teverett:
@RobEin  thanks!

PR #4250, opened 2024-09-19
[MariaDB] Use ifNotExists clause correctly for unique key constraint

ani-sha:


KvanTTT (approved):
I have no objection against changes.

teverett:
@ani-sha  thanks

PR #4251, opened 2024-09-19
Upgrade arden grammar to Arden syntax 3.0 R2 specification

Yngli:
I've updated the grammar for the Arden Syntax to the latest specificiation: Arden Syntax 3.0 R2, which is current in the ballot process by HL7. This specification will soon be accepted by HL7.


teverett:
@Yngli  thanks!

Issue #4243, opened 2024-09-16
]build] Regression


kaby76:
Looks like the build regressed for some of the targets. https://github.com/antlr/grammars-v4/actions/runs/10859932031/job/30139904745?pr=4237#step:24:121

I'll have a fix sometime today. (It was caused by some "copy-paste" code that should have been adjusted.)

PR #4253, opened 2024-09-21
[build] Fix for build regression -- #4243

kaby76:
This is a fix for #4243.

### What changed?

* Trash version 0.23.6 is now used.
* The Github workflow now installs ts-node, typescript, and tsx because these are used in testing Antlr4ng and TypeScript. This change also fixes problems with split grammars with base classes for TypeScript.
* Templates for testing have been updated to correct various problems.
    * Replacing `find` with `trglob`.
    * Replacing global dotnet tool calls with local.
    * Fixing calls to Antlr4ng and TypeScript programs using `npx tsx`.
    * For CSharp target, adding code for outputting ambiguous trees with `-ambig` option, and `trparse --ambig`.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,73 +3,73 @@
   "isRoot": true,
   "tools": {
     "trcaret": {
-      "version": "0.23.3",
+      "version": "0.23.6",
       "commands": [
         "trcaret"
       ]
     },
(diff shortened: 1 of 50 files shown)


kaby76:
This PR is required for several outstanding PRs.

teverett:
@kaby76  thanks

PR #4273, opened 2024-10-04
[dot] Fix semicolon as attribute list separator

dundalek:
Attributes in Graphviz can be separated not only by comma, but also a semicolon. 

Example of a dot file that currently fails to parse, but works with graphviz `dot -Tsvg g.dot > g.svg`:
```dot
graph {
  A [color=blue; shape=hexagon]
}
```

Matching rule in [DOT language grammar](https://graphviz.org/doc/info/lang.html):

### DIFF
--- a/dot/DOT.g4
+++ b/dot/DOT.g4
@@ -59,7 +59,7 @@ attr_list
     ;
 
 a_list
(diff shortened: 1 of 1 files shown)


teverett:
@dundalek  thanks!

PR #4275, opened 2024-10-05
Add TypeScript target for Rust

rafed:
Added TypeScript target for Rust.

I updated desc.xml to test for the quicker files (examples/quick/*.rs) even though it parses the long (examples/long/*.rs) files (a bit slower though). Let me know if that should be changed.


kaby76:
I think the partitioning is fine. The problem with testing is that the time for parsing all input files depends on the OS, hardware, machine load, etc., none of which can be predicted. As long as at least the collection of targets covers the entire set of input files, that's probably the best we can do.

teverett:
@rafed  thanks!

PR #4276, opened 2024-10-05
[build] Add testing for grammar ambiguity.

kaby76:
I'm really excited about this one!!

This change adds testing for grammar ambiguity using trperf. The tool essentially runs the parser with the [ProfilingATNSimulator](https://github.com/antlr/antlr4/blob/2a7904a595479954ccfb1c78124fc3d7f5bebcb1/runtime/CSharp/src/Atn/ProfilingATNSimulator.cs). (It does not test the grammar using Z3/SMT. I am still working on that.)
* A CSharp parser is first generated for the grammar.
* The CSharp parser is built.
* The parser is run via trperf on all input files. trperf collects data on the ambiguity and rule name. This output is filtered using `grep` for non-zero ambiguity, then the list of rule names sorted and made unique. It's all done via the command line, using simple Bash piping.

The output is a little hard to read, but you can see the list of symbols that are ambiguous after all parses.
* abb: dataList
* aql: collect, expr, filter, for_in, let_list, limit, sort
* arden: exprBefore, exprFunction, exprSort
* asl: compound_logical_condition, event_generation, if_statement, logical_condition, operation_invocation, simple_logical_condition, struct_value
* ...
* bcpl: e0, e2, e3, e4, e5, e6, e7, e8, ...
* bnf: xalt, xelem, xlhs, xrhs
* c: blockItem, castExpression, ......
* ...

The entire list of grammars and ambiguity at this point are here: 
[ambig.txt](https://github.com/user-attachments/files/17265993/ambig.txt)

Note, the actual ambiguous parse trees can be obtained using `trparse --ambig`, but that can take a lot of time, so it is not done.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -170,7 +170,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest]
-        test: [ useless-parens, format ]
+        test: [ useless-parens, format, ambiguity ]
     steps:
     - name: Info
       shell: bash
--- a/_scripts/templates/CSharp/st.test-ambiguity.sh
+++ b/_scripts/templates/CSharp/st.test-ambiguity.sh
@@ -1,0 +1,49 @@
+# Generated from trgen <version>
+
+SAVEIFS=$IFS
+IFS=$(echo -en "\n\b")
+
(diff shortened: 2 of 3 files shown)


teverett:
@kaby76  thanks.  This is great.

PR #4277, opened 2024-10-08
[build] Dotnet use of '--' changed

kaby76:
Dotnet 8 functionality changed sometime over the last year with one of the "minor" revisions. Or CommandLineParser changed.   But the cause was irrelevant. `dotnet trgen -- --version` was printing an error not the version of the tool.

Also, `trparse` doesn't support `-t antlr4` anymore ([not since four months ago](https://github.com/kaby76/Trash/blame/main/src/trparse/Grun.cs#L239)), so [test-cover.sh](https://github.com/antlr/grammars-v4/blob/c82c128d980f4ce46fb3536f87b06b45b9619922/_scripts/test-cover.sh#L339) wasn't working perfectly in detecting grammars with actions--and then removing them from consideration. That was changed to `-t ANTLRv4`.

I'm also updating the Trash Toolkit to the latest as it takes care of this with its built-in templates, and fixes the analysis of the .g4s for a grammar.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,189 +3,189 @@
   "isRoot": true,
   "tools": {
     "trcaret": {
-      "version": "0.23.6",
+      "version": "0.23.7",
       "commands": [
         "trcaret"
       ],
       "rollForward": false
     },
     "trcover": {
-      "version": "0.23.6",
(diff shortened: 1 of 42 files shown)


teverett:
@kaby76  thanks!

PR #4278, opened 2024-10-10
[mysql/Oracle] Add ports for sql/mysql/Oracle, and add a better test suite for the grammar.

kaby76:
This is a change to the sql/mysql/Oracle grammar. The grammar was refactored to be in "target agnostic format". Ports were added for Antlr4ng, CSharp, Java, Python3, and TypeScript. (For other targets, I leave as an exercise for anyone who would be interested.)

## Grammar changes
Target agnostic format is the preferred format for grammars-v4 grammars. This allows for a single copy of the .g4 files to be shared across targets, and is very important in order to stop diverging fixes to the grammar between targets.

Target agnostic means that the Antlr actions are in the target-specific language but in a syntax palatable across targets. The only syntax that works is a method call. For example, instead of [BIT_OR_SYMBOL : B I T '_' O R { this.type = this.determineFunction(MySQLLexer.BIT_OR_SYMBOL); };](https://github.com/antlr/grammars-v4/blob/2b4757494d23851b8d71aedc18a048a3e5dd6c46/sql/mysql/Oracle/MySQLLexer.g4#L493-L495), the code must be encapsulated in a method call: `BIT_OR_SYMBOL : B I T '_' O R { this.doBitOr(); };`. Expressions that use operators, e.g., `>=`, `<`, `!`, `&&`, `||`, or constants like "true" and "false" cannot be used directly in Antlr Actions because they can have different syntax across targets. Assignments are also not allowed because the name of the field varies across runtimes.

However, target agnostic format does not work perfectly because there are still different syntax for method codes. A method call for C++ is `this->isThisFine()`, for Python3 `self.isThisFine()`. For C# and Java, the method call can have `this.isThisFine()` or `ThisFine()`, whereas in TypeScript the `this.` must appear as in `this.isThisFine()`. Antlr4 `@header` is required in some targets, but the target-specific code cannot be placed in the grammar. To get around these syntactic issues, transformGrammar.py is used to modify the .g4s for the specific target.

The start rule for the grammar was changed to allow multiple statements per file separated by a semi-colon.

## Base class file names
The base class file names follow the standard used throughout grammars-v4: `<grammar name> ('Parser' | 'Lexer') 'Base' <target-specific file extension>`. [MySQLBaseRecognizer](https://github.com/kaby76/grammars-v4/blob/bb7ae078ce8e120e35d96ed8fab08b231c5df012/sql/mysql/Oracle/TypeScript/MySQLBaseRecognizer.ts#L9) is the parser base class, but [Lexer](https://github.com/antlr/antlr4/blob/ce2557fdbf099d438e8b88e3789218d84dbeffb2/runtime/CSharp/src/Lexer.cs#L23) and [Parser](https://github.com/antlr/antlr4/blob/ce2557fdbf099d438e8b88e3789218d84dbeffb2/runtime/CSharp/src/Parser.cs#L20) both inherit from a Recognizer. I.e., they are all "recognizers."

## Old files saved
The old Antlr4ng grammar and "TypeScript/" code were moved to "original/".

## A new test suite added
The original grammar and tester "demo.ts" tested less than 5% of the grammar. (`dotnet trcover -i "select *, _latin1'😁' from sakila.actor where actor_id = 1;"`) This PR adds the tests from `Positive-Technologies/examples/*.sql`. The driver code required some initialization settings for the lexer and parser, and these are added for each port. (Initialization code should be in the base class constructors.) Many SQL statements were commented out as they would not parse. See comments `#NB` lines in the test files.

**NB: Someone will need to verify that the Oracle grammar is correct and verify that commented-out SQL statements are wrong.**


teverett:
@kaby76 thanks!

mike-lischke:
Note: This change makes it impossible to update the grammar in the future, as it is now very different from the original source. Not a wise decision, if you ask me.

Also, adding lots of methods to make this target agnostic (which doesn't even work in all cases) doesn't help users of this grammar.

It would have been good to invite me to review these changes.

kaby76:
Having multiple copies of the grammar to update is exactly the problem we are trying to avoid!! Unlike the Oracle repos, anyone can change the code in here. At best, this only results in synch problems, which require someone to take time to fix. Worse, the changes "fix" the grammar by adding in more ambiguity and SLL fallbacks, again requiring someone to fix.

Please tell me where the "official" source files are so I can write a script to scrape this grammar from. It does not have to be done by hand. The [readme](https://github.com/antlr/grammars-v4/blob/5ed2a3a3b9c2944a0fc60eaee26ccc1da2e7e26b/sql/mysql/Oracle/original/README.md) provides several links, but (1) listed below seems to be the "official".
1) https://marketplace.visualstudio.com/items?itemName=Oracle.mysql-shell-for-vs-code ==> https://github.com/mysql/mysql-shell-plugins ==> via `find . -name '*.g4' ==> https://github.com/mysql/mysql-shell-plugins/blob/332ced04ad6df3785ba781ce7dbb327598933699/gui/frontend/src/parsing/mysql/MySQLParser.g4
2) https://github.com/mysql/mysql-server/blob/8.0/sql/sql_yacc.yy
3) https://github.com/mysql/mysql-workbench ==> via `find . -name '*.g4'` ==> https://github.com/mysql/mysql-workbench/blob/c2b118c69a47b10ee6950b9664dcc2c75b3d296c/library/parsers/grammars/MySQLParser.g4

Your Antlr4ng port is fine, but people expect the rest of the targets to be filled out. At some point, I will add the rest for this grammar. While target agnostic isn't perfect, it is the only thing that we have at the moment that avoids the synch problem. Hopefully, Antlr4ng/ANTLRng will fix the multi-target problem. I've given up with Antlr5.

mike-lischke:
Don't get me wrong, I very much appreciate the work you are doing here! I just wish you had me included as a reviewer. I would have pointed out a few other things that are not working well, like allowing more than query to parse in one go. Don't you think we had this considered long before, if it was a viable solution? It does not work with custom delimiters.

Anyway, doing an automated conversion is probably the best way to handle this situation. About the sources:

1. This is just a file delivered as part of the MSG extension. Nobody should using that for getting code.
2. This is not an ANTLR4 grammar. It's the YACC grammar which serves as a role model to make sure the ANTLR grammar supports all features from the MySQL server.
3. This is only listed for historical reasons to make the actual origin clear.

The real source is the Github repository of the [MySQL Shell Plugins](https://github.com/mysql/mysql-shell-plugins/tree/master/gui/frontend/src/parsing/mysql), of which the MSG extension is a part of.

Regarding ANTLR5: I completely agree.

kaby76:
@mike-lischke Thanks. I'll keep you in the loop with all this. I'll work out a way to generate ports from the source over in the MySQL Shell Plugins. I am still cleaning up the sql/postgresql grammar (removing all ambiguities and fallbacks and adding the ports for every target). --Ken

PR #4280, opened 2024-10-12
Fix case *_ error

Willie169:
Fix that `case [a, *_] if a == 0:` throws error `rule soft_kw__not__wildcard failed predicate: {this.isnotEqualToCurrentTokenText("_")}?`

### DIFF
--- a/python/python3_12/PythonParser.g4
+++ b/python/python3_12/PythonParser.g4
@@ -27,6 +27,11 @@ THE SOFTWARE.
   *
   */
 
+  /*
+   * Contributors :
+   * [Willie Shen](https://github.com/Willie169) : Fix that `case [a, *_] if a == 0:` throws error `rule soft_kw__not__wildcard failed predicate: {this.isnotEqualToCurrentTokenText("_")}?`
+  */
+
 parser grammar PythonParser; // Python 3.12.6  https://docs.python.org/3.12/reference/grammar.html#full-grammar-specification
 options {
     tokenVocab=PythonLexer;
@@ -427,8 +432,7 @@ maybe_star_pattern
     | pattern;
 
 star_pattern
-    : '*' pattern_capture_target
-    | '*' wildcard_pattern;
+    : '*' NAME;
 
 mapping_pattern
     : LBRACE RBRACE
--- a/python/python3_12/changes.md
+++ b/python/python3_12/changes.md
@@ -5,3 +5,8 @@ Type comments will now be tokenized as plain comment tokens.
 
 Line continuation for string literals (backslash followed by a newline) is no longer resolved.
 (backslash+newline is no longer removed from string literals)
+
+---
+Oct. 18, 2024
+---
+Fix that `case [a, *_] if a == 0:` throws error `rule soft_kw__not__wildcard failed predicate: {this.isnotEqualToCurrentTokenText("_")}?`
--- a/python/python3_12/tests/test_match_case.py
+++ b/python/python3_12/tests/test_match_case.py
@@ -1,0 +1,16 @@
+# COMMAND LINE:
+# grun Python file_input -tokens test_match_case.py
+#
+# EXPECTATIONS:
+#   - [@63,234:234='*',<'*'>,12:13]
+#   - [@64,235:235='_',<NAME>,12:14]
+#   - no error message
+
+a, *b = [1, 2, 3, 4]
+match b:
+    case [2]:
+        print("0")
+    case [f, *_] if f==2:
+        print("1")
+    case _:
+        print("2")


Willie169:
Wait a minute. I forgot to test targets other than Java and add test for the fixed error.

kaby76 (commented):
Please add a .py file to [examples/](https://github.com/antlr/grammars-v4/tree/c82c128d980f4ce46fb3536f87b06b45b9619922/python/python3_12/examples) that tests the changed rules for this PR.

kaby76:
None of the examples/*.py tested `as_pattern`, `closed_pattern`, `star_pattern`, or `double_star_pattern` before. Yes, please add these so I can test the ambiguity. (The grammar before your changes has ambiguities.).

Willie169:
Ok, I will write examples for them

Willie169:
I've added examples from [Python 3.12 Standard Lib](https://github.com/python/cpython/tree/3.12/Lib) and written by me and ChatGPT and added Cpp targets, but I am currently not sure about the validity of targets other than Java.

Willie169:
test not passed: _bootstrap_external.py, _test_eintr.py, dump.py,test__interpchannels.py, test__interpreters.py,test_array.py, test_builtin.py, test_clinic.py,test_compile.py, test_exceptions.py, test_frame.py,test_fstring.py, test_imaplib.py, test_logging.py,test_opcache.py, test_os.py, test_posix.py,test_regrtest.py, test_str.py, test_subprocess.py,test_sys.py, test_tabnanny.py, test_traceback.py,test_type_params.py, test_venv.py

python/python3_12/Cpp/transformGrammar.cpp
```diff
@@ -0,0 +1,67 @@
...
+    inputFile.close();
+    outputFile.close();
+}
+
+int main() {
+    fix("PythonLexer.g4");
+    fix("PythonParser.g4");
+
+    return 0;
+}
```
kaby76, line 67: Please rewrite this as a Python (v3) script. The reason is that the build currently checks for file [transformGrammar.py](https://github.com/antlr/grammars-v4/blob/c82c128d980f4ce46fb3536f87b06b45b9619922/_scripts/templates/Cpp/st.build.sh#L3) (in ps1: https://github.com/antlr/grammars-v4/blob/c82c128d980f4ce46fb3536f87b06b45b9619922/_scripts/templates/Cpp/st.build.ps1#L11-L13. The build does not know how to alter the grammar prior to running CMake. If it does not have a transformGrammar.py, then the grammar will not work in testing, and the Cpp target must be removed.

There are a multitude of examples in grammars-v4.
```
$ find . -name transformGrammar.py | grep Cpp
./bison/Cpp/transformGrammar.py
./cpp/Cpp/transformGrammar.py
./dart2/Cpp/transformGrammar.py
./fortran/fortran77/Cpp/transformGrammar.py
./fortran/fortran90/Cpp/transformGrammar.py
./golang/Cpp/transformGrammar.py
./gvpr/Cpp/transformGrammar.py
./javascript/javascript/Cpp/transformGrammar.py
./lua/Cpp/transformGrammar.py
./php/Cpp/transformGrammar.py
./python/python/Cpp/transformGrammar.py
./python/python3/Cpp/transformGrammar.py
./rust/Cpp/transformGrammar.py
./swift/swift5/Cpp/transformGrammar.py
10/13-09:29:45 ~/issues/g4-mysql-target-agnostic
```

kaby76: The alternative is to create "transformGrammar.py" which performs the compile of the .cpp file using g++, a dependency which is probably ok even on Windows. But, you will need to test.

kaby76: The final alternative that I can think of is to alter the templates st.build.ps1 and st.build.sh scripts (in _scripts/templates/Cpp and maybe the other targets) to perform the g++/run. Wish Antlr handled actions a little better.

Willie169: Thanks. I didn't noticed that, I think I will simply rewrite the transformGrammar with Python3.

RobEin:
> Fix that `case [a, *_] if a == 0:` throws error `rule soft_kw__not__wildcard failed predicate: {this.isnotEqualToCurrentTokenText("_")}?`

I was able to reproduce the error.
This looks like a target independent bug.
I'm working on it.

RobEin:
Adding a CPP port is fine, although I am working on such a port as well.

The problem is that you are trying to overwrite the newer one with an older version, which causes a regression in many files.
You want to overwrite Python 3.12.6 with an older Python 3.12.1.
[e.g. README.md](https://github.com/antlr/grammars-v4/pull/4280/files#diff-4faa9428dc1c154bb7a11a4da87ea054355a476ea11b198f9a2bca6e3ab8c768)

I don't see where you fixed the `rule soft_kw__not__wildcard` failed predicate error.

Please explain the many changes in [PythonParser.g4](https://github.com/antlr/grammars-v4/pull/4280/files#diff-4a0dfe755315378adfd968b8299143c1ff1ba1af8d93bd3e242f14f89c616d71).

Willie169:
The real change that solve the problem is:
```
star_pattern
    : STAR NAME;
```
Other changes are useless and you can change them back. <br>
The escape character in f-string f"\" " and f'\'' can't be recognized, to reproduce:
```
f"\"", f'\''
```
I haven't solved it yet.\
Btw, I haven't checked whether the Cpp port works but I will be busy on exams since today's midterm and probably until the 2nd mock test, so feel free to take over or write another one.

Willie169:
The real change that solve the problem is:
```
star_pattern
    : STAR NAME;
```
Other changes are useless and you can change them back. <br>
The escape quotes character in f-string can't be recognized, to reproduce:
```
f"\"", f'\''
```
I haven't solved it yet.\
Btw, I haven't checked whether the Cpp port works but I will be busy on exams since today's midterm and probably until the 2nd mock test, so feel free to take over or write another one.

kaby76:
Some fixes are in order.
* Undo the deletion of the TypeScript target and re-add the 4.13.2 requirement in desc.xml. The TypeScript port can't work without the changes that happened with Antlr 4.13.2.
* The examples/ directory needs to be reorganized with some structure: Python Standard Lib code should go into its own directory called `examples/Lib`; ChatGPT-generated code should go into its own directory called `examples/ChatGPT`.

RobEin:
Looks like I managed _to_ find and fix the "**_escape quotes character in f-string_**" bug.
I'll test it for a few more days.

Please change only the things related to the error "**_soft_kw__not__wildcard failed predicate_**" in PythonParser.g4 in the new PR.
I did not use token names in the parser grammar so that it resembles the official [python.peg](https://docs.python.org/3/reference/grammar.html) grammar as much as possible, and I think that the grammar is also more readable.
Please make the new PR based on the latest [grammars-v4 repo](https://github.com/antlr/grammars-v4).

Take PythonLexerBase.cpp out of the current PR for the time being, on the one hand because you haven't tested it, on the other hand, because the [Cpp runtime](https://github.com/antlr/antlr4/tree/dev/runtime/Cpp) must be completed first.
For example, the following methods are missing from [Lexer.cpp](https://github.com/antlr/antlr4/blob/dev/runtime/Cpp/runtime/src/Lexer.cpp):
- getModeStack
- getMode
- setMode
	
Once you have the runtime modification, you can test PythonLexerBase.cpp.
And you can only make a PR about PythonLexerBase.cpp if the modified Cpp runtime has already been published.
Please let me know if you can do it.

Willie169:
I will make a PR about the added star_pattern and delete the cpp for now. Thanks.

python/python3_12/PythonParser.g4
```diff
@@ -427,8 +432,7 @@ maybe_star_pattern
     | pattern;
 
 star_pattern
-    : '*' pattern_capture_target
-    | '*' wildcard_pattern;
+    : '*' NAME;
```
kaby76, line 435: There's not a single test for this grammar change.
```
10/21-09:31:22 ~/issues/g4-4280/python/python3_12/Generated-CSharp-0
$ ./bin/Debug/net8.0/Test.exe ../examples/*.py -tree 2>&1 | grep star_pattern
10/21-09:32:57 ~/issues/g4-4280/python/python3_12/Generated-CSharp-0
$
```

Willie169: I added test, thanks.

teverett:
@Willie169 thanks!

PR #4283, opened 2024-10-15
[build] Fix ClickHouse pom typo

be-happy4:
There's a build error, and it's easy to fix.

before:
5C4DA950A021668DFCF2377934810BEC

after:
QQ_1728973465996


teverett:
@be-happy4  thanks!

PR #4284, opened 2024-10-15
[build] Fix "mvn test" from root directory for java and kotlin.

kaby76:
* "mvn test" was not working in the root directory for java/ and kotlin/kotlin.
* I don't know why, but running "mvn test" from the java/ directory worked fine. That is how I originally tested it when I made the change to add java20.
* kotlin/kotlin actually has two scenarios in the pom.xml that were not programmed into desc.xml. This is now fixed. The problem was that kotlin/kotlin/examples/script/*.kt required a different entry point to the grammar. I'm not sure why I remastered the tests instead of just fixing the problem by adding the two test scenarios.
* sql/clickhouse still fails but a fix is in the works. https://github.com/antlr/grammars-v4/pull/4283
* The output from `mvn test > out.txt 2>&1` is here: [out.txt](https://github.com/user-attachments/files/17379482/out.txt)


teverett:
@kaby76  thanks!

Issue #4286, opened 2024-10-21
Strange Python parser for Lua 5.4


boolangery:
Hey,

I generated the Python3 parser code with the Lua grammar in https://github.com/antlr/grammars-v4/tree/master/lua

In the generated code, reference to [LuaLexerBase.py](https://github.com/antlr/grammars-v4/blob/master/lua/Python3/LuaLexerBase.py) are generated like this:

```python
    def COMMENT_action(self, localctx:RuleContext , actionIndex:int):
        if actionIndex == 0:
             this.HandleComment(); 
```

It obviously crashes on `this`.

Any help appreciated

kaby76:
Actions (`{ ... }` or `{ ... }?`) in Antlr4 grammar files is target specific. In order to have the grammar file work for all targets, it has to be in "target agnostic format".

"Target agnostic format" means that the Antlr actions are target-specific (Python3 or C# or Java ...) but in a minimal syntax to be compiled for the target. It is syntactically and semantically a method call. Unfortunately "target agnostic format" does not work perfectly because there is still minor variations. For example, to call a method on a class in C++, `this.` has to be instead `this->`.

To get around this last syntactic issue for a target, you need to apply [transformGrammar.py](https://github.com/antlr/grammars-v4/blob/2b4757494d23851b8d71aedc18a048a3e5dd6c46/lua/Python3/LuaLexerBase.py) to the grammar files. `python3 transformGrammars.py`

Make sure to read the [Wiki](https://github.com/antlr/grammars-v4/wiki).

boolangery:
Thank you for your quick answer, then problem solved!

PR #4290, opened 2024-10-24
Fixed the SHEBANG detection

Webstonesl:
In Rust a shebang is only if the '#!' is not followed by a '['. I added an action to fix it. If it is possible to do it with normal ANTLR, I don't know how.


Webstonesl:
There was a small issue regarding the Cpp and CSharp language specific changes. Should be fixed now.

rust/RustLexer.g4
```diff
@@ -133,7 +133,9 @@ OUTER_BLOCK_DOC:
 
 BLOCK_COMMENT_OR_DOC: ( BLOCK_COMMENT | INNER_BLOCK_DOC | OUTER_BLOCK_DOC) -> channel (HIDDEN);
 
-SHEBANG: {this.SOF()}? '\ufeff'? '#!' ~[\r\n]* -> channel(HIDDEN);
+SHEBANG: {this.SOF()}? '\ufeff'? '#!' {this._input.LA(1) != 97}? ~[\r\n]* -> channel(HIDDEN);
```
kaby76, line 136: Strictly speaking, this is not "target agnostic" (https://github.com/antlr/grammars-v4/issues/4286#issuecomment-2427584163) because `!=` is a target-specific operator. The entire expression should go in a base class, which is already done for both [lexer](https://github.com/antlr/grammars-v4/blob/9f95ea41e82e347a841b01efa178605fc832925b/rust/RustLexer.g4#L29) and [parser](https://github.com/antlr/grammars-v4/blob/9f95ea41e82e347a841b01efa178605fc832925b/rust/RustParser.g4#L29). Please add a "is...()" method, reference that here, and implement the methods for each of the targets.

PR #4296, opened 2024-10-28
Changes withElement definition to be closer to CQL3 specification

TonyRippy:
This PR adds support the following:

1. Specifying `COMPACT STORAGE`.
2. Specifying the order of more than one clustering key.
3. Allowing for additional table options after `WITH CLUSTERING ORDER`.

All the above are allowed in CQL3. See the relevant section of the spec: https://cassandra.apache.org/doc/stable/cassandra/cql/ddl.html#create-table-statement

I believe the following changes are mostly backwards-compatible, with the exception of `clusteringOrder` moving from `withElement` to `tableOptions`.

### DIFF
--- a/cql3/CqlParser.g4
+++ b/cql3/CqlParser.g4
@@ -391,15 +391,16 @@ createTable
     ;
 
 withElement
-    : kwWith tableOptions? clusteringOrder?
-    ;
-
-clusteringOrder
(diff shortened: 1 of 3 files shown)


kaby76 (approved):
Looks good.

TonyRippy:
@teverett, would you mind taking a look when you get a chance?

teverett:
@TonyRippy  thanks!

Issue #4291, opened 2024-10-24
[Postgresql] Grammar is not idiomatic ANTLR


masonwheeler:
If I had to guess, it looks like [this Postgres parser grammar](https://github.com/antlr/grammars-v4/blob/master/sql/postgresql/PostgreSQLParser.g4) was produced by someone taking a grammar designed for a different parser generator and translating it as literally as possible.  But it's got a lot of content that is very bad ANTLR.

For example, you see this pattern a lot:
```
from_clause
    : FROM from_list
    |
    ;
```

Having a rule end in `| ;` means it will always be considered a valid match, even if it contains no content.  This is painful to work with, because now in the visitor, on a rule that contains this as a sub-rule, you can't simply say `if (context.from_clause() != null)` to see if you have a real match; the code to check for it is significantly messier.

The idomatic way to do this in ANTLR is to define it as:

```
from_clause
    : FROM from_list
    ;
```

and then have any rule that uses it invoke it as `from_clause?`.

Would it be possible to clean this grammar up a bit?

kaby76:
Unfortunately, what gets implemented often isn't checked very well. The empty-[alternative](https://github.com/antlr/grammars-v4/blob/9f95ea41e82e347a841b01efa178605fc832925b/antlr/antlr4/ANTLRv4Parser.g4#L295-L297) seems to be an easy thing to detect and fix via script. I'll write a script to detect this and add it to the build. I'll also write a script to fix the problem. This stuff should not be done by hand.

The other problem is an ambiguity. [It is detected in my recent changes for detecting ambiguity over the grammars in the repo](https://github.com/antlr/grammars-v4/pull/4276#issue-2567590784). This script finds the smallest example in the test suite with the ambiguity.

```
10/25-07:51:29 ~/issues/g4-4280/sql/postgresql/Generated-CSharp
$ min=99999
minf=''
for f in ../examples/*.sql
do
    dotnet trperf $f -c ar 2>/dev/null | grep from_list | grep -v '^0' -q
    if [ $? -eq 0 ]
    then
        c=`wc -l $f | awk '{print $1}'`
        if [ $c -lt $min ]
        then
            min=$c
            minf=$f
        fi
    fi
done
echo $min " " $minf
4   ../examples/non_ansi_join.sql
10/25-07:57:06 ~/issues/g4-4280/sql/postgresql/Generated-CSharp
$ dotnet trparse --ambig ../examples/non_ansi_join.sql | dotnet trtree -a
[...]

PR #4299, opened 2024-11-01
[postgresql] Fix for #4291 -- non-idiomatic usage of "(foo | )" instead of "foo?"

kaby76:
This is a change to fix one of the types of problems with the [postgresql grammar](https://github.com/antlr/grammars-v4/tree/199a5121ece05d2f2e7eca330d0738220499e80c/sql/postgresql).

As noted in #4291, the grammar contains numerous non-idiomatic rule definitions. In particular, the grammar contains rules with empty alternatives, appearing to be derived from a grammar for another parser generator. Indeed, that conclusion is correct: this grammar is derived from the Bison grammar at https://github.com/postgres/postgres/blob/9be4e5d293b554d8a0800790c57fc707a3b5cf0f/src/backend/parser/gram.y/. Bison/yacc does not support an optional operator EBNF. But, Antlr does.

This change refactors empty production rules to non-empty rules and adds the `?`-operator to the applied occurrences of the parser symbol. The names of these parser rules are changed to not have the "opt_" prefix, as they don't derive empty (generally) anymore.

The scripts to make these changes have been added to the sub-directory `fixups/`. `detect.sh` finds all productions that have an empty top-level alt. `fix.sh` refactors these productions by removing the empty alt, and adding the `?`-operator to the applied occurrences. `rename.sh` renames rules that are not nullable.

These changes do not affect the performance.

There are two new warnings from the Antlr Tool (` rule stmt_case contains an optional block with
 at least one alternative that can match an empty string` and `rule stmt_return contains an optional block with at least one alternative that can match an empty string`). These do not affect the performance or correctness of the parser. These will be delt with in a later fix to correct other warnings (`non-fragment lexer rule AfterEscapeStringConstantMode_NotContinued can match the empty string` and `non-fragment lexer rule AfterEscapeStringConstantWithNewlineMode_NotContinued can match the empty string`). In addition, the ambiguity in the grammar results in terrible performance, which also needs to be corrected.

I've updated the readme.md to note the actual grammar source and issues.

### DIFF
--- a/sql/postgresql/PostgreSQLParser.g4
+++ b/sql/postgresql/PostgreSQLParser.g4
@@ -188,13 +188,13 @@ callstmt
     ;
 
 createrolestmt
-    : CREATE ROLE roleid opt_with optrolelist
+    : CREATE ROLE roleid with_? optrolelist
     ;
 
-opt_with
+with_
     : WITH
     //| WITH_LA
-    |
+    
     ;
 
 optrolelist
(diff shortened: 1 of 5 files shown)


sql/postgresql/PostgreSQLParser.g4
```diff
@@ -325,7 +325,7 @@ iso_level
     | SERIALIZABLE
     ;
 
-opt_boolean_or_string
+boolean_or_string_
```
kaby76, line 328: `opt_boolean_or_string` was renamed by "rename.sh" because the symbol is *not* nullable. It's completely unclear why it was named with a leading "opt_"!! The Bison grammar has problems.

PR #4301, opened 2024-11-01
[postgresql] Fix for #4300 -- ambiguity in from_list

### LINKED ISSUE #4300: [postgresql] Remove from_list/non_ansi_join ambiguity
In the postgresql grammar, there is a rule for [from_list.](https://github.com/antlr/grammars-v4/blob/199a5121ece05d2f2e7eca330d0738220499e80c/sql/postgresql/PostgreSQLParser.g4#L3192-L3195). It has two alts:
1) non_ansi_join, which is defined in the rule just below `from_list`. non_ansi_join is defined as [non_ansi_join: table_ref (COMMA table_ref)+;](https://github.com/antlr/grammars-v4/blob/199a5121ece05d2f2e7eca330d0738220499e80c/sql/postgresql/PostgreSQLParser.g4#L3197-L3199).
1) `table_ref (COMMA table_ref)*`, which is absolutely identical to that in `non_asi_join`.

Obviously, this is ambiguous. The question is: When and why was it introduced when [gram.y defines from_list with no alts](https://github.com/postgres/postgres/blob/511db38ace2690f19816465baed07cefe535bfec/src/backend/parser/gram.y#L7212-L7215)?

As far back as 1999, gram.y always defined `from_list` as it is today. The Antlr grammar rule was changed with https://github.com/antlr/grammars-v4/pull/3008, but the change for the rule is completely wrong because it adds absolutely no new syntax (other than require more than one table_ref) and should have been flagged. If anything, this needs to be placed in semantic analysis. Antlr is not PEG!

kaby76:
This PR fixes a grammar ambiguity introduced with https://github.com/antlr/grammars-v4/pull/3008.

Consider input `select x from t c1, t c2;`. The substring `t c1, t c2` can be parsed two ways:
* `from_list` => `non_ansi_join` => `table_ref ...` ==> `t c1, t c2`
* `from_list` => `table_ref ...` ==> `t c1, t c2`

As I point out in the write-up for #4300, from_list is defined with two alts that overlap. The two alts did not fix anything. It simply added to the ambiguity of the grammar and made the parser perform worse.

I discovered this using [trparse --ambig](https://github.com/kaby76/Trash/tree/main/src/trparse).

### DIFF
--- a/sql/postgresql/PostgreSQLParser.g4
+++ b/sql/postgresql/PostgreSQLParser.g4
@@ -3190,12 +3190,7 @@ from_clause
     ;
 
 from_list
-    : non_ansi_join
-    | table_ref (COMMA table_ref)*
-    ;
-
-non_ansi_join
-    : table_ref (COMMA table_ref)+
+    : table_ref (COMMA table_ref)*
     ;
 
 table_ref


teverett:
@kaby76  thanks!

Issue #4302, opened 2024-11-01
[postgresql] Ambiguity in opt_analyze due to erroneous modification of stmt and stmtmulti.


kaby76:
Consider this input:
```
VACUUM ANALYZE brin_test;
```

This can be parsed in two ways:
* (root (stmtblock (stmtmulti (stmt (vacuumstmt (VACUUM "VACUUM") (opt_full) (opt_freeze) (opt_verbose) (opt_analyze (analyze_keyword (ANALYZE "ANALYZE"))) (opt_vacuum_relation_list (vacuum_relation_list (vacuum_relation (qualified_name (colid (identifier (Identifier "brin_test") (opt_uescape)))) (opt_name_list)))))) (SEMI ";"))) (EOF ""))
* (root (stmtblock (stmtmulti (stmt (vacuumstmt (VACUUM "VACUUM") (opt_full) (opt_freeze) (opt_verbose) (opt_analyze) (opt_vacuum_relation_list))) (stmt (analyzestmt (analyze_keyword (ANALYZE "ANALYZE")) (opt_verbose) (opt_vacuum_relation_list (vacuum_relation_list (vacuum_relation (qualified_name (colid (identifier (Identifier "brin_test") (opt_uescape)))) (opt_name_list)))))) (SEMI ";"))) (EOF ""))

The problem here is in opt_analyze in the context of a stmt.
```
stmtmulti
    : (stmt SEMI?)*
    ;

stmt
    : ...
    ;

opt_analyze
    : analyze_keyword
    |
    ;
```
The problem is that `stmtmulti` allows for an optional `SEMI`. This is a bad idea because it directly leads to this problem: Where does the `VACUUM` statement end?

[...]

PR #4303, opened 2024-11-02
[postgresql] Fix for #4302 -- ambiguity in stmtmulti

kaby76:
This is a fix for #4302.

Consider the input `VACUUM ANALYZE brin_test;`. This can be parsed with the current grammar in two ways:
* `stmtmulti` => `stmt SEMI` ==> `VACUUM ANALYZE brin_test` `;`.
* `stmtmulti` => `stmt stmt SEMI` ==>`VACUUM` `ANALYZE brin_test` `;`. It is very likely this is meant to be one statement.

The problem is how stmt are separated. In the official gram.y grammar, semicolon has to separate the two. Otherwise it should be considered as one stmt.

This change modifies the rules for meta-commands, which are not part of SQL, and the rules in the parser associated with SQL command separation. The changes in https://github.com/antlr/grammars-v4/pull/2365 modified these rules in an incorrect way, introducing ambiguity in the grammar.

1) The lexer rules for meta-commands are modified to return a SEMI, because that is essentially how they function. The meta-command `\'` is special. it *must* return a SEMI because that is how it is defined in the documentation at https://www.postgresql.org/docs/current/app-psql.html#APP-PSQL-META-COMMAND-SEMICOLON. They don't tell you there that meta-commands work this way also, but they do. As before, there is no attempt to "parse" the meta-commands themselves.
1) SQL commands *must* be separated by a SEMI, as per the official Bison grammar for PostgreSQL.

The ambiguity for `opt_analyze` is eliminated. `dotnet trperf -i 'VACUUM ANALYZE brin_test;' -c ard | grep -v '^0'` yields no output (as `trperf` outputs in column 1 the number of ambiguity detected, and `grep` removes all lines that begin with '0').

I should note that the parse time for `examples/*.sql` is not faster (37s after vs 44s before).

_Also, I really do not understand why there are error listeners defined for this grammar. These have nothing to do with the grammar and must be removed._


teverett:
@kaby76  thanks!

Issue #4304, opened 2024-11-02
[postgresql] Why is sub-parser created for function bodies?


kaby76:
The PostgreSQL grammar defines error listeners [LexerDispatchingErrorListener.cs](https://github.com/antlr/grammars-v4/blob/master/sql/postgresql/CSharp/LexerDispatchingErrorListener.cs), [ParserDispatchingErrorListener.cs](https://github.com/antlr/grammars-v4/blob/master/sql/postgresql/CSharp/ParserDispatchingErrorListener.cs), [LexerDispatchingErrorListener.java](https://github.com/antlr/grammars-v4/blob/master/sql/postgresql/Java/LexerDispatchingErrorListener.java), and [ParserDispatchingErrorListener.java](https://github.com/antlr/grammars-v4/blob/master/sql/postgresql/Java/ParserDispatchingErrorListener.java). But these error listeners have absolutely nothing to do with the grammar, don't reference any symbols from the grammar, and should therefore not be in the grammar.

kaby76:
The code is used in a special sub-parser. Not entirely clear why it's a sub-parser using the same PostgreSQL grammar. I will need to investigate this further. For now, I won't remove the code until I understand why it's being done this way rather than directly.

kaby76:
* The sub-parser was added with this commit, https://github.com/antlr/grammars-v4/commit/0c85145aa0b5518688a9d226354d40183b954ff4#diff-191019bcbd2ccd918ba1d7e6610ca830bc526d04af2ae6e6d3c5356e96274b76, and merged as part of https://github.com/antlr/grammars-v4/pull/2395, but with zero explanation why a sub-parser is created, presumably to mutate the parse tree.

kaby76:
Consider the input in aggregates.sql: https://github.com/antlr/grammars-v4/blob/199a5121ece05d2f2e7eca330d0738220499e80c/sql/postgresql/examples/aggregates.sql#L778-L798

This input contains text that is a function body. However, the body syntax depends on a literal *after* the input text: https://github.com/antlr/grammars-v4/blob/199a5121ece05d2f2e7eca330d0738220499e80c/sql/postgresql/examples/aggregates.sql#L798C4-L798C20

Is a sub-parser created because the parser has a difficult time synching back up once an error is detected?

kaby76:
[...]

PR #4305, opened 2024-11-02
[postgresql] Fix for #4304

kaby76:
This change removes the error listener port code from the grammar because they have nothing to do with the grammar.


kaby76:
Closing pending investigate as to why the parser base class creates an entirely new parser during the parse.

Issue #4306, opened 2024-11-02
[postgresql] a_expr_qual/a_expr_qual_op is ambiguous.


kaby76:
Consider the test input file [collate.sql](https://github.com/antlr/grammars-v4/blob/199a5121ece05d2f2e7eca330d0738220499e80c/sql/postgresql/examples/collate.sql). The parse is ambiguous for a_expr_qual_op. See the two parse trees in [ambig.txt](https://github.com/user-attachments/files/17607267/ambig.txt).

This ambiguity is a major contributor to the performance issues in the grammar. This file takes ~1.3s to parse cold start (~0.4s warm start), and it is only 276 lines long!

Note, I chose this file because it had the least number of non-zero ambiguities, which is 1. If you try to take a test input with a large number of ambiguities, `trparse --ambig` will take a long time. To find the file with the least non-zero number of ambiguities, I did `$ dotnet trperf -c aF ../examples/*.sql | grep -v '^0' | awk '{sum[$2] += $1} END {for (key in sum) print sum[key], key}' | sort -k1 -n | head`.

kaby76:
### Analysis

In the input file [colate.sql](https://github.com/antlr/grammars-v4/blob/199a5121ece05d2f2e7eca330d0738220499e80c/sql/postgresql/examples/collate.sql). https://github.com/antlr/grammars-v4/blob/199a5121ece05d2f2e7eca330d0738220499e80c/sql/postgresql/examples/collate.sql#L153

This seemingly simple expression involving `||` can be parsed in two ways because `qual_op` appears in two places in expressions:
* https://github.com/antlr/grammars-v4/blob/199a5121ece05d2f2e7eca330d0738220499e80c/sql/postgresql/PostgreSQLParser.g4#L3557-L3559
* https://github.com/antlr/grammars-v4/blob/199a5121ece05d2f2e7eca330d0738220499e80c/sql/postgresql/PostgreSQLParser.g4#L3641-L3643

This is all fine, but the problem is the context target_el. https://github.com/antlr/grammars-v4/blob/199a5121ece05d2f2e7eca330d0738220499e80c/sql/postgresql/PostgreSQLParser.g4#L4100-L4103

We see here that `||` can be treated as a unary, postfix op, or a binary op. *This seems quite wrong because the only postfix ops are `!` and `!!`.* https://www.postgresql.org/message-id/BE2DF53D-251A-4E26-972F-930E523580E9%40enterprisedb.com

[...]

PR #4307, opened 2024-11-03
[postgresql] Fix for #4306 -- ambiguity in target_el

kaby76:
This is a fix for #4306.

Consider input `select x || y from collate_test10;`. This substring `x || y`, can be parsed two ways:
* `target_el` => `a_expr identifier` ==>  `x ||` `y`.
* `target_el` => `a_expr` ==> `a_expr_qual_op` => `a_expr_unary_qualop qual_op a_expr_unary_qualop` ==> `x || y`. This is the likely preferred derivation.

The fix is to add a semantic predicate before an alt in the rule `a_expr_qual`, which specifies the syntax for a unary postfix operator. Only `!` and `!!` should be allowed.

However, PostgreSQL allows new operators to be declared. This will need to be fixed later on.

### DIFF
--- a/sql/postgresql/CSharp/PostgreSQLParserBase.cs
+++ b/sql/postgresql/CSharp/PostgreSQLParserBase.cs
@@ -120,4 +120,11 @@ public abstract class PostgreSQLParserBase : Parser
         parser.AddErrorListener(listener_parser);
         return parser;
     }
+
+    public bool OnlyAcceptableOps()
+    {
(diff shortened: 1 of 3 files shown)


kaby76:
PostgreSQL allows one to declare arbitrary operators with various args and precedence. I'm going to punt on this for now, but it should be fixed properly.

teverett:
@@kaby76  thanks!

PR #4314, opened 2024-11-07
[postgresql] Fix for #4311 -- remove Antlr "locals", stop calling sub-parser.

### LINKED ISSUE #4311: [postgresql] Grammar contains target-specific declarations via locals.
Consider this input.
```
create function explain_query_json(query_sql text)
returns table (explain_line json)
language plpgsql as
$$
begin
  set enable_seqscan = off;
  set enable_bitmapscan = on;
  return query execute 'EXPLAIN (ANALYZE, FORMAT json) ' || query_sql;
end;
$$;
```
I was trying to print out the parse tree for this using `toStringTree()`. The output isn't correct because the function body is still just a string.

Instead, it creates a field for the node type `func_as` using a ["locals declaration"](https://github.com/antlr/grammars-v4/blob/5ed2a3a3b9c2944a0fc60eaee26ccc1da2e7e26b/sql/postgresql/PostgreSQLParser.g4#L2048). This is not the right way to do this because it makes the .g4 target-specific rather than target-agnostic. The proper way is to use the [contextSuperClass option](https://github.com/antlr/antlr4/blob/master/doc/options.md#contextsuperclass). The alternative is to redo the tree after parsing the normal PostgreSQL input. This is probably why the [Go port doesn't work](https://github.com/antlr/grammars-v4/blob/5ed2a3a3b9c2944a0fc60eaee26ccc1da2e7e26b/sql/postgresql/desc.xml#L4) because the order of the "type field-name" is reversed for Go. This can be fixed via the transformGrammar.py hack, but it doesn't solve the toStingTree/serialization problem. 

Per-language specific "toStringTree()" code is not attempted. Also, when the parse tree is serialized, Trash doesn't know anything about this. `ToStringTree()` is not the only thing to fix.

I also see that this code always parses the function body as always PlSQL, rather than case on the language first ([Java](https://github.com/antlr/grammars-v4/blob/5ed2a3a3b9c2944a0fc60eaee26ccc1da2e7e26b/sql/postgresql/Java/PostgreSQLParserBase.java#L66), [CSharp](https://github.com/antlr/grammars-v4/blob/5ed2a3a3b9c2944a0fc60eaee26ccc1da2e7e26b/sql/postgresql/CSharp/PostgreSQLParserBase.cs#L59), [Go](https://github.com/antlr/grammars-v4/blob/5ed2a3a3b9c2944a0fc60eaee26ccc1da2e7e26b/sql/postgresql/Go/postgresql_parser_base.go#L136)).

kaby76:
This PR fixes #4311. The problem is that the .g4 grammar file used `locals [ ... ]`. This cannot work for a multi-targeted grammar. In addition, the grammar tried to parse function bodies as always Pl/SQL. It ignores the fact that it could be something else.

* The `locals [ ... ]` declaration has been removed.
* The sub-parse of the function bodies was removed. I added `//NB` comments to re-implement the sub-parse in the correct way later.

The grammar is now twice as fast to parse the test suite in examples/, likely because it is not parsing pl/pgsql. To consider: is it better to create the parser once and use Reset() instead of creating a parser for each pl/pgsql input?

### DIFF
--- a/sql/postgresql/CSharp/PostgreSQLParserBase.cs
+++ b/sql/postgresql/CSharp/PostgreSQLParserBase.cs
@@ -56,14 +56,21 @@ public abstract class PostgreSQLParserBase : Parser
         if (func_as != null)
         {
             var txt = GetRoutineBodyString(func_as.func_as().sconst(0));
-            var ph = getPostgreSQLParser(txt);
             switch (lang)
             {
                 case "plpgsql":
-                    func_as.func_as().Definition = ph.plsqlroot();
+                    // Mutate tree.
+                    // NB: cannot use locals this way because
+                    // it does not work with ToStringTree().
+                    // var ph = getPostgreSQLParser(txt);
+                    // func_as.func_as().Definition = ph.plsqlroot();
                     break;
                 case "sql":
-                    func_as.func_as().Definition = ph.root();
+                    // Mutate tree.
+                    // NB: cannot use locals this way because
+                    // it does not work with ToStringTree().
+                    // func_as.func_as().Definition = ph.root();
+                    // ph.root();
                     break;
             }
         }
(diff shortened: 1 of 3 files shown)


teverett:
@kaby76  thanks!

Issue #4315, opened 2024-11-08
[postgresql] The PostgreSQL grammar contains the PL/SQL grammar. You cannot do this!


kaby76:
The PostgreSQL grammar contains two entry points, and rules for another grammar (PL/SQL). This is the wrong way to implement function body parsing because the two grammars have different lexer requirements. (This is one reason why I could not understand why the lexer and parser grammar contained conflicting keyword sets.

For example if you look at the [scan.l](https://github.com/postgres/postgres/blob/027124a872d7b5dfddc69590af42f626b1727dba/src/backend/parser/scan.l) and [kwlist.h](https://github.com/postgres/postgres/blob/027124a872d7b5dfddc69590af42f626b1727dba/src/include/parser/kwlist.h), there is absolutely no mention of `QUERY`. It isn't even referenced in [gram.y](https://github.com/postgres/postgres/blob/027124a872d7b5dfddc69590af42f626b1727dba/src/backend/parser/gram.y). However, if you look at [pl_gram.y](https://github.com/postgres/postgres/blob/027124a872d7b5dfddc69590af42f626b1727dba/src/pl/plpgsql/src/pl_gram.y), you find rules that use K_QUERY. There is no string `QUERY` recognized for PostgreSQL!!

All the rules for PL/SQL (or is that PL/pgSQL?) must be removed, and the PostgreSQL grammar reflect how it is done with gram.y in the PostgreSQL server.

kaby76:
This can be fixed by removing the `plsqlroot` rule, then interatively removing all unused parser rules via [find-unused-parser-symbols.sh](https://github.com/kaby76/g4-scripts/blob/475af8a62335d36703541bed3a251c6e3129a5a6/find-unused-parser-symbols.sh). Note, bare_label_keyword is not referenced, but it should be, but the PostgreSQL grammar is wrong.

PR #4316, opened 2024-11-09
[postgresql] Fix for #4308, #4315 -- ambiguity and plpgsql grammar.

### LINKED ISSUE #4308: [postgresql] collabel is ambiguous; postgresql contains some plsql rules.
Consider input string `SELECT 'trailing' AS first;` in [comments.sql](https://github.com/antlr/grammars-v4/blob/199a5121ece05d2f2e7eca330d0738220499e80c/sql/postgresql/examples/comments.sql#L5). This is ambiguous because `first` has three different possible trees:
* collabel => identifier => plsql_unreserved_keyword => FIRST_P
* collabel => plsql_unreserved_keyword => FIRST_P
* collabel => unreserved_keyword => FIRST_P

This is caused because of the rule https://github.com/antlr/grammars-v4/blob/199a5121ece05d2f2e7eca330d0738220499e80c/sql/postgresql/PostgreSQLParser.g4#L4233-L4240

There is quite a bit of overlap across each of the alts.

Over in the original gram.y, the rule is https://github.com/postgres/postgres/blob/027124a872d7b5dfddc69590af42f626b1727dba/src/backend/parser/gram.y#L17560-L17565

kaby76:
This PR fixes #4308 and #4315. The issue was caused by adding in rules for PL/PGSQL into PostgreSQL. These rules were added to parse function bodies, but the changes also affected parsing of PostgreSQL. The problem is that the lexing rules overlap between the two languages. You cannot simply add two non-disjoint lexer rules together because you then have an enormous task in trying to clarify the PostgreSQL parser rules, least of which is the ambiguity. This is probably why there's a collection of tests that fail in the [examples.errors/](https://github.com/antlr/grammars-v4/tree/5ed2a3a3b9c2944a0fc60eaee26ccc1da2e7e26b/sql/postgresql/examples.errors) directory.

* All parser rules for PL/PGSQL were removed.
* As much as I can find, all the lexer symbols for PL/PGSQL were removed from the lexer grammar.
* The desc.xml was updated to specify which files to parse (i.e., *.sql).
* `collabel` was renamed to `colLabel` to be closer to the official source for the grammar, [gram.y](https://github.com/postgres/postgres/blob/9be4e5d293b554d8a0800790c57fc707a3b5cf0f/src/backend/parser/gram.y).
* There are a few overlaps with [another PR](https://github.com/antlr/grammars-v4/pull/4314) [ I made.](https://github.com/antlr/grammars-v4/pull/4314) I'll fix the conflicts after https://github.com/antlr/grammars-v4/pull/4314 is first merged.

The ambiguity no longer exists. `trparse --ambig -i "SELECT 'trailing' AS first;" | trtree -a` yields no ambiguity trees.

The plan is to re-add parsing of function bodies with a completely separate grammar for pl/pgsql later.


teverett:
@kaby76 thanks

Issue #4318, opened 2024-11-09
[postgresql] Ambiguities in simple_select_primary and into_clause.


kaby76:
Consider input `SELECT INTO tableam_tblselectinto_heapx FROM tableam_tbl_heapx;`. This is ambiguous in two places.
```
$ trperf -c ard -i 'SELECT INTO tableam_tblselectinto_heapx FROM tableam_tbl_heapx;' | grep -v '^0'
Time to parse: 00:00:00.1027274
1       simple_select_pramary   726
1       into_clause     742
```
Here, `trperf` finds ambiguity with [simple_select_pramary](https://github.com/antlr/grammars-v4/blob/5ed2a3a3b9c2944a0fc60eaee26ccc1da2e7e26b/sql/postgresql/PostgreSQLParser.g4#L2974) and [into_clause](https://github.com/antlr/grammars-v4/blob/5ed2a3a3b9c2944a0fc60eaee26ccc1da2e7e26b/sql/postgresql/PostgreSQLParser.g4#L3007).

The ambiguous parse trees are found using `trparse --ambig`.
```
11/09-10:02:40 ~/issues/g4-current/sql/postgresql/Generated-CSharp
$ trparse --ambig  'SELECT INTO tableam_tblselectinto_heapx FROM tableam_tbl_heapx;' | trtree -a
CSharp 0 SELECT INTO tableam_tblselectinto_heapx FROM tableam_tbl_heapx; success 0.1987994
SELECT INTO tableam_tblselectinto_heapx FROM tableam_tbl_heapx;.726: (root (stmtblock (stmtmulti (stmt (selectstmt (select_no_parens (select_clause (simple_select_intersect (simple_select_pramary (SELECT "SELECT") (into_clause (INTO "INTO") (opttempTableName (qualified_name (colid (identifier (Identifier "tableam_tblselectinto_heapx")))))) (from_clause (FROM "FROM") (from_list (table_ref (relation_expr (qualified_name (colid (identifier (Identifier "tableam_tbl_heapx")))))))))))))) (SEMI ";"))) (EOF ""))
[...]

Issue #4309, opened 2024-11-04
[postgresql] ambiguity in into_clause


kaby76:
Consider input `select * into attest2 from atacc1;`. The substring `into attest2` can be parsed two ways:
* `into_clause` => `INTO opt_strict opttempTableName` ==> `into attest2`.
* `into_clause` => `INTO into_target` ==> `into attest2`.

The original gram.y rule is: https://github.com/postgres/postgres/blob/027124a872d7b5dfddc69590af42f626b1727dba/src/backend/parser/gram.y#L13025-L13039

The rule for into_clause currently is: https://github.com/antlr/grammars-v4/blob/199a5121ece05d2f2e7eca330d0738220499e80c/sql/postgresql/PostgreSQLParser.g4#L3012-L3015

There is a clear divergence in the Antlr grammar with the addition of the rule into_target.

PR #4319, opened 2024-11-09
[postgresql] Fix for #4309, #4315, #4318.

kaby76:
This is a fix for #4309, #4315, and #4318.

This change corrects a lot of issues with the PostgreSQL grammar. The overall issue was that there were still a number of pl/pgsql grammar rules in this grammar even after https://github.com/antlr/grammars-v4/pull/4316. As I mentioned, adding the grammar for pl/pgsql directly into the PostgreSQL grammar causes a lot of problems. ***It is very important to learn from this: you should not merge in the grammar for pl/pgsql into PostgreSQL, unless you are very, very careful. For example, you would have to have every lexer by default work in a different lexer mode. The official parser for PostgreSQL does not try to combine grammars. Why should the Antlr grammar deviate from this?***

The changes bring the grammar to be back in line with the official PostgreSQL grammar [gram.y](https://github.com/postgres/postgres/blob/9be4e5d293b554d8a0800790c57fc707a3b5cf0f/src/backend/parser/gram.y). All the tests pass, but more importantly, the ambiguities that I mentioned are now gone. This is because the official Bison grammar is LALR(1), so it shouldn't be ambiguous.

It's not clear if I still have all the pl/pgsql grammar removed from the postgresql grammar. There was one test that uses embedded SQL identifiers, but it shouldn't. So PLSQLVARIABLENAME is still defined and used.

I noticed that the lexer grammar contained numerous "built-in" function names. These were added erroneously. I added the comments from gram.y for the identifier classes into the parser grammar.


kaby76:
(Need to resolve conflicts once https://github.com/antlr/grammars-v4/pull/4316 is merged.)

Issue #4324, opened 2024-11-12
[postgresql] Method checkLA(int c) is not target agnostic.


kaby76:
I am now writing every port for the PostgreSQL grammar. But, I'm finding that the grammar is not Target Agnostic

In PostgreSQLLexer.g4, [checkLA('-')](https://github.com/antlr/grammars-v4/blob/0120854f3990ccac49047b67bcba17a3cd790ef7/sql/postgresql/PostgreSQLLexer.g4#L130C27-L130C39) cannot be target agnostic because not all languages can automatically cast a character literal (or string literal) `'-'` to an integer (e.g., JavaScript).

All actions in the .g4's must be parameterless method calls.

PR #4323, opened 2024-11-11
[postgresql] Fixes for #4321, #4322

kaby76:
This PR fixes #4321 and #4322.

For #4321, the grammar rule `typename` was refactored from the gram.y. The second alt was removed because it added ambiguity.

For #4322, `a_expr_caret` was corrected to follow the usual Floyd operator precedence grammar pattern. The input `SELECT 1 * 2 ^ 3 + 4;` is no longer ambiguous.

Several other rules were corrected to remove some of the lingering pl/pgsql grammar rules, which caused ambiguity in `col_name_keyword` and `c_expr`.

As I progress in the clean up of this grammar, keep an eye on the list of rules with ambiguity. Before I started a couple of weeks ago, the list of ambiguous rules was this:
```
a_expr_add
a_expr_and
a_expr_compare
a_expr_qual_op
alter_table_cmd
c_expr
col_name_keyword
colid
collabel
createfunc_opt_list
frame_bound
from_list
func_type
group_by_item
into_clause
opt_analyze
opt_target_list
optschemaeltlist
reindexstmt
select_with_parens
selectstmt
stmt
table_ref
tableelement
type_function_name
zone_value
```

It is now this:
```
a_expr_and
a_expr_compare
c_expr
frame_bound
group_by_item
select_with_parens
selectstmt
table_ref
```

The readme was updated to reflect fixes to the grammar that removed Antlr Tool warnings.

PR #4325, opened 2024-11-13
[postgresql] Fix for #4324; add ports.

kaby76:
This is a fix for #4324. The grammar was not "target agnostic" and there were just two targets implemented. With this change almost all targets are supported. I didn't bother with the implementation of the PHP port because it is very slow in general and contains errors in ATN Config hash value and equality computions. This problem becomes accute with ambiguity and fallbacks. I also didn't write the Swift port because I don't develop on MacOS.

_Target Agnostic_ is a specific coding style for .g4 files to allow any target to be generated from _one copy_ of the .g4 files: All `@members` and `@headers` are replaced with a comment; Actions `{ ... }` and `{ ... }?`contain only parameterless method calls with a `this.` prefix; `transformGrammar.py` is used to modify the .g4s just prior calling the Antlr Tool specific for the target because there are still minor variations for targets from the established format. The purpose of Target Agnostic is to avoid the major problem of keeping multiple copies of the .g4 files in synch across ports.

The following ports were added: Antlr4ng, Cpp, Dart, Go, JavaScript, Python3, TypeScript. Existing ports are: CSharp, Java. Note, the Go port became out of synch because it was not plugged into the build. It now works.

#### Notes
* The spaces on either side of the method call in an action have been removed. Python requires code to begin in very specific columns in order to nest within an `if`-statement.
* The readme has been updated with a description of how to use transformGrammars.py.
* All the methods in actions have been updated.
* There is change in one of the rules from https://github.com/antlr/grammars-v4/pull/4323. This "disappears" when performing the merge after the other PR.
* I'm unclear why there are lexer rules that match empty. I had to remove the empty curly actions `{ }` in these rules because the code does not compile for Python3.


kaby76:
(Finally a clean build. Failures had nothing to do with this PR. It's all a matter of chance. See https://github.com/antlr/grammars-v4/issues/4326.)

teverett:
@kaby76  thanks!

PR #4327, opened 2024-11-14
[build] Fix for #4326 -- GitHub Action VMs running MacOS are unreliable.

### LINKED ISSUE #4326: [build] Macos VM runners are unreliable.
The Github runners for Macos are unreliable. We get crashes, hangs, and time-outs for all sorts of stages of the build. Not sure the best way to proceed here.

kaby76:
This PR fixes #4326.

With this change, the parse step is bypassed for MacOS. The bnf grammar was updated to test the PR.

Unfortunately, this won't fix the spurious hangs of the MacOS runners that we see. Google has a problem with the runner code and/or is using too few servers, or is using the slowest servers on the planet. However, bypassing the MacOS tester is a small price to pay to increase the chance for the tests to be completed in a reasonable amount of time. We need a compilation and link for C++ under the MacOS environment because it is the rate-limiting factor in the code we write for the C++ target.

Just yesterday, in antlr/antlr4, a PR was submitted in which 39 runners completed in a [minute or two](https://github.com/antlr/antlr4/actions/runs/11864106553/job/33066903765) on Windows and Ubuntu. However, the PR sat around for a MacOS test to complete in [1h 25m](https://github.com/antlr/antlr4/actions/runs/11864106553/job/33066900407), with the setup of PHP taking the majority of the time at 1h 23m!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -117,7 +117,7 @@ jobs:
     - name: Upgrade Pip.
       run: |
         python -m ensurepip --upgrade
-    - name: Test Pip.
+    - name: Test Pip
       run: |
         pip --version
     - name: Install Antlr tool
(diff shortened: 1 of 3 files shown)


teverett:
@kaby76 thanks.

PR #4333, opened 2024-11-19
Improve objc grammar: better expr syntax; fix some issues

DJBen:
**Overview**
- Borrowed C expression syntax: The expression syntax now is parsed in order of the operator precedence.
- Fixed an issue that array literal containing assignment isn't recognized: `@[ foo = bar ];` is valid
- Fixed an issue that spaces after `@` in `@   try {} @catch (...)` isn't allowed. Surprisingly they are valid.
- Fixed an issue that ellipsis in `@catch` isn't allowed.
- Add `direct` in `@property` attributes
- Stricter numeral type parsing: `long long long int` used to be valid, but now it isn't.

**Testing**
- [x] Test cases are added


teverett:
@DJBen  thanks!

Issue #4332, opened 2024-11-18
[postgresql] expressions with '=' operator ambiguous.


kaby76:
Consider input `select * from tmptz where f1 at time zone 'utc' = '2017-01-18 00:00';` from [examples/timestamptz.sql](https://github.com/antlr/grammars-v4/blob/97bd28855acf37204623a71eb9d5ce3f9f62f80a/sql/postgresql/examples/timestamptz.sql#L484C1-L484C44). This is ambiguous.
```
11/18-06:10:41 ~/issues/g4-current/sql/postgresql/Generated-CSharp
$ trparse --ambig -i "select * from tmptz where f1 at time zone 'utc' = '2017-01-18 00:00';" | trtree -a
CSharp 0 string success 0.1539677
[...]

PR #4334, opened 2024-11-19
[postgresql] Fix for #4332 -- fix ambiguity with '=' (rewrite Floyd operator precedence grammar)

kaby76:
This is a fix for #4332.

The problem raised in the issue was the ambiguity with `select * from tmptz where f1 at time zone 'utc' = '2017-01-18 00:00';`. But, the problem was the haphazard implementation of the operator precedence grammar for `a_expr`.

I rewrote the rules associated with `a_expr` to use the Antlr style for expressions, fixing the ambiguity. In addition, the parse trees for expressions are much more concise. I will need to unfold c_expr into a_expr and refactor the order of alts to correspond to the correct precedence--and remove ambiguity that I introduced with the rewrite of a_expr. This is important because Antlr does not implement precedence like Bison. Some of the alts for a_expr were removed because they overlapped with those in c_expr.

Not only are the trees much smaller, but it is also a little faster, taking ~7s for the examples/ test suite vs. ~8s before this PR.

**To do: Add expression tests with parse trees.**

### DIFF
--- a/sql/postgresql/PostgreSQLParser.g4
+++ b/sql/postgresql/PostgreSQLParser.g4
@@ -3477,9 +3477,28 @@ escape_
     
     ;
 
-//precendence accroding to Table 4.2. Operator Precedence (highest to lowest)
+// Precedence according to Table 4.2. Operator Precedence (highest to lowest)
+// https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-PRECEDENCE
+// (captured on Nov 19, 2024).
+//
+// Operator/Element                Associativity   Description
+// .                               left            table/column name separator
+// ::                              left            PostgreSQL-style typecast
+// [ ]                             left            array element selection
+// + -                             right           unary plus, unary minus
+// COLLATE                         left            collation selection
+// AT                              left            AT TIME ZONE, AT LOCAL
+// ^                               left            exponentiation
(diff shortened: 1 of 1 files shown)


kaby76:
The problem is that Antlr cannot handle operator precedence parsing if one of the symbols used in the RHS does not follow the standard pattern recognized by Antlr (that recursion occurs with first and last occurrences of the symbol only).

Consider this grammar.
```
grammar xxx;

start: e (';' e)* EOF;

e : e '::' t
 | e '||' e
 | ID
 ;
t : e;
f : f '::' f
 | f '||' f
 | ID
 ;
z : ID;
ID : [a-z]+;
WS : [ \t\r\n] -> skip;
```

This grammar is ambiguous because of the symbol `t` in the RHS of the recursive rule. This results in the following ATN for `e`.

graphviz (19)

Now, compare that with the ATN for f (with a proper recursion).

graphviz (20)

See:
* https://github.com/antlr/antlr4/issues/4460
* https://github.com/antlr/antlr4/issues/2525
* https://github.com/antlr/antlr4/pull/4480

kaby76:
Note, Antlr may not rewrite left-recursion correctly. In that case, I may have to explicitly set the precedence. https://stackoverflow.com/q/30001506/4779853

kaby76:
Closing for now until I have some free time for this.

PR #4344, opened 2024-11-27
Added EDITIONABLE clause to create_type

madams51703:
Added EDITIONABLE clause to create_type.


teverett:
@madams51703  thanks!

PR #4345, opened 2024-11-29
[mysql/oracle] Add remaining ports to support all targets.

kaby76:
This PR implements the rest of the ports of the mysql/oracle grammar. This is a first step in generating each of the ports automatically from the official source ([here](https://github.com/mysql/mysql-shell-plugins/tree/8928ada7d9e37a4075291880738983752b315fee/gui/frontend/src/parsing/mysql)).

These ports are necessary because people want to work with a grammar with a minimum of effort. Asking them to implement a port of the grammar is time-consuming and difficult.

In fact, this PR was difficult. The implementation of the two tokens emitted with [DOT_IDENTIFIER rule implementation](https://github.com/antlr/grammars-v4/blob/3f8074f7730c272ddfd2d77df0ae83b46422fd98/sql/mysql/Oracle/MySQLLexer.g4#L350-L352) was outrageous because every target has its idiosyncrasies. (It probably would have been easier to have "emitDot()" emit both the DOT_SYMBOL and IDENTIFIER. I think I've seen this issue in other grammars.) The use of base class methods in the Antlr4ng source is inconsistent. (Actions need to be implemented as base class methods for target-agnostic.)

Previously, from the original Antlr4ng port, initialization was done in the *driver*: https://github.com/antlr/grammars-v4/blob/00a96c93ee9891cb3ebd3510c5244a3e3cc5660c/sql/mysql/Oracle/original/TypeScript/demo.ts#L23-L27

This is actually a bad place to put initialization because it requires the user to read the driver code or readme. Most people who use this repo don't do either. Driver codes are not part of the grammar. Initialization should be done in a constructor. NB: The Go port required a workaround.

### What was done?
* The existing ports (Antlr4ng CSharp Java Python3 TypeScript) were fixed to set defaults in constructors. _For the Go port, the constructor for the base class is not called. To get around this, many of the fields were place in a static var and `init()` defined to initialize the static var._
* The custom `Test.*` files were removed.
* `StackQueue<>` was removed and all ports now use a standard queue data structure.
* SqlMode and SqlModes were placed in their own files.
* The Cpp, Dart, Go, and JavaScript ports were added.
### Performance
[As I mentioned in a discussion](https://github.com/antlr/grammars-v4/discussions/4341), I have been writing scripts to output a graphical comparison of the relative speed across targets. Here is the graph for this grammar.

times
[data.zip](https://github.com/user-attachments/files/17992537/data.zip)

[As mentioned previously](https://github.com/antlr/grammars-v4/discussions/4341#discussioncomment-11427170), the large variance in the times for Java are due to anti-virus and disk caching for the first run of the parser app. (There are over 800 .class files generated by the Java compiler!)

### To do
* I should probably have a metric for "tokens parsed per second" to compare different grammars, e.g., mysql/oracle vs. postgresql. And, I should also devise other metrics, such as total parse tree size for the test suite.
* `charSets` is still not implemented.
* General clean up, tighten up, and sort the base class methods.

@mike-lischke This is a good first step in getting the ports generated via a script from the source. This grammar in target-agnostic format works fine. As with all grammars in this repo, Python3 is pretty slow. But, it does work.


mike-lischke (approved):
You put a lot of work into this. Are you planning to do the same for all grammars? That could be a job that takes several years :-)

mike-lischke:
Very nice to see the performance charts! Interesting numbers and interesting to see that Dart does better than C++ (although only by a small margin).

About the TypeScript code: this was meant to be test/example code. The users of the grammar have to implement the base classes on their own anyway, to match their environment. I think it makes not so much sense to port this demo code to all supported targets. And to serve as a model/template a single implementation is all what's needed.

In any case the effort to make this work cross-target is enormous and I want to get rid of this kind of target specific actions altogether in my TS port of ANTLR4. From today's perspective it was a bad decision to allow native code go into a grammar, but hey, this is how things go sometimes.

kaby76:
Yes, there's a lot of work to do here to make ports for every split grammar. But, I think time would be better spent on understanding how to detect and remove grammar ambiguity and fallbacks, and apply it to the TIOBE-rated programming languages. Antlr is wonderful in accepting all sorts of bad grammars, but it also is why people tend to give it a bad rap. (You should see the trash-talk about Antlr in Reddit.) Looking forward to Antlrng picking up where Antlr4 stops.

teverett:
@kaby76  thanks!

PR #4360, opened 2024-12-17
[plsql] Add TypeScript port back in; it works.

kaby76:
This PR just changes the desc.xml to do testing. The TypeScript port works just fine.


teverett:
@kaby76  thanks!

PR #4362, opened 2024-12-17
[plsql] Added Antlr4ng, Cpp, Dart, Go ports. Dart-specific .g4 files removed.

kaby76:
This PR added a bunch of ports for the plsql grammar. I cleaned up some of the target-specific code, reorganized the tests. This is so I can do performance testing, which is also included in this PR.

* Added Antlr4ng port. Note, the grammar did have "postinclude", which was hardwired to C++. This can't work with other ports, especially Antlr4ng, which also requires "import" of critical base class files. Unfortunately, C++ and Antlr4ng both do not conform to the standard adopted for all the other ports: if the grammar has an `options { superClass=foobar; }`, the generated parser and lexer classes have the necessary "imports" for the base class, so that no additional syntax is needed. Requiring per-target syntax include is a complete mess, introducing yet another way to break the use of the grammar across targets.
* Adjusted the .g4's so that there are no symbol conflicts across all ports.
* Added the Cpp port. *This requires an update to Trash trgen for complete rendering of various files in sql/plsql/Cpp. The problem is that rendering is only done for templates in the template source directory, or in the templates .zip file. I'm not sure why I only copy files from the target-specific directory. They, too, should be considered as templates.*
* Added the Go port.
* Fixed the Dart port. ***This port had forked copies of the grammar files. I must admit: I was guilty of creating the fork with https://github.com/antlr/grammars-v4/pull/2223. At the time, I couldn't think of a better solution. (Now we do with trasnsformGrammar.py.) But, these files were never updated since the fork in 2021. So, it is likely hundreds of changes were never applied to the forked .g4s. This is exactly why grammar forking should not be done.***
* Reorganized examples so we get a broad, fast test suite for all targets in `examples/`, and a select number of other tests in `more/`.
* Added performance graph and updated readme. The Bash scripts that did the analysis are included as well as a copy of the data.zip.


teverett:
@kaby76  thanks!

PR #4363, opened 2024-12-17
[Plsql] pivot_clause and unpivot_clause can have own alias

jiefei30:
```sql
SELECT s2.id, name, subject, score
FROM scores2 s1
    UNPIVOT (
    score FOR subject IN (
        chinese_score AS 'Chinese',
        math_score AS 'Math',
        english_score AS 'English'
    )
) s2;
```
Here it is, pivot_clause and unpivot_clause can have own alias.
Previous query and its pivot_clause can both have an alias, but pivot_clause's alias will replace previous query alias.
which meas
```sql
SELECT s1.id, name, subject, score
FROM scores2 s1
    UNPIVOT (
    score FOR subject IN (
        chinese_score AS 'Chinese',
        math_score AS 'Math',
        english_score AS 'English'
    )
) s2;
```
if you use s1 to be the identifier will be wrong.
i tried these sql on livesql.oracle.com

### DIFF
--- a/sql/plsql/PlSqlParser.g4
+++ b/sql/plsql/PlSqlParser.g4
@@ -5988,7 +5988,7 @@ flashback_query_clause
     ;
 
 pivot_clause
-    : PIVOT XML? '(' pivot_element (',' pivot_element)* pivot_for_clause pivot_in_clause ')'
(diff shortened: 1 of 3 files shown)


teverett:
@jiefei30  could you add a test case for this please?

jiefei30:
@teverett I added two cases in /example folder : pivot14.sql, pivot15.sql

teverett:
@jiefei30  thanks!

PR #4365, opened 2024-12-18
[build] Update Trash and node.js.

kaby76:
This PR is to update the Trash Toolkit for https://github.com/antlr/grammars-v4/pull/4362. I also updated the version of `node.js` installed because GitHub is updating `ubuntu-latest`, which is causing the builds to crash in another repository.


teverett:
@kaby76  thanks!

PR #4377, opened 2025-01-09
[MySql] Synchronize grammar and tests; remove Mariadb grammars

ani-sha:
relates to https://github.com/debezium/debezium/pull/6074


kaby76:
The build fails because of setup-dotnet does not work. The problem is that setup-dotnet is trying to install in a directory that does not exist. I have reported this issue to the setup-dotnet people (https://github.com/actions/setup-dotnet/issues/565). It occurs with other platforms as well (https://github.com/actions/setup-dotnet/issues/571).

There is a workaround, but I'll have to add that here in this repo.

teverett:
@ani-sha  thanks!

PR #4378, opened 2025-01-09
[builds] Remove dotnet install, assume that it exists in VM images for builds.

kaby76:
[setup-dotnet is not robust](https://github.com/actions/setup-dotnet/issues/565). Assume dotnet exists and is on path in all VMs. This *should* fix the build issues in https://github.com/antlr/grammars-v4/pull/4377 related to [setup-dotnet](https://github.com/actions/setup-dotnet).


teverett:
@kaby76  thanks

PR #4388, opened 2025-01-21
[Swift5] grammar fixes

samatanna:
1. Fix closure_parameter rule, add optional UNDERSCORE
Closures don't have external parameter names, but underscore is allowed.
The UNDERSCORE token definition has been moved before the IDENTIFIER token in the lexer, to use UNDERSCORE token in rule.
2. Fix ambiguity of '?' token
3. Add STRUCT token to identifier rule


codeFather2 (approved):

codeFather2 (approved):

teverett:
@samatanna  thanks!

PR #4406, opened 2025-02-03
[lua and z] Changes to fix ambiguity in Z grammar, and unnecessary "using" in lua.

kaby76:
The lua grammar contained a "using" to something that is not needed for the CSharp target of the grammar. If not removed, a build error will occur for the new Antlr-ng tool.

For the Z grammar, there is an additional preprocessor grammar. The desc.xml is required to disambiguate which grammar for building and testing.


teverett:
@kaby76 thanks!

PR #4411, opened 2025-02-11
[cobol85, glsl] Fix additional ambiguities of which grammar to test.

kaby76:
Trgen requires you to make a choice of what grammar to test when it is ambiguous. This happens when there are preprocessor grammars. This PR changes the desc.xml to say what grammar to test.


teverett:
@kaby76  thanks

PR #4408, opened 2025-02-07
Fixes to bring sql/mysql/Oracle up to date.

kaby76:
This PR is a change to bring the [sql/mysql/Oracle grammar](https://github.com/antlr/grammars-v4/tree/master/sql/mysql/Oracle) up to date with the one in the [Oracle MySQL Shell for VS Code extension](https://github.com/mysql/mysql-shell-plugins/tree/f35d371f3290fefc224ed474e1d9766adeec71c2/gui/frontend/src/parsing/mysql). Many of the changes are reformatting. This was done because this grammar will be automatically generated from the Oracle repo sources, which will be pre-formatted with one line per Antlr4 grammar action. I'm not sure when I'll get around to writing the script that scrapes the grammar directly from the Oracle repo, but soon.

PR #4383, opened 2025-01-16
[MySql] Heatwave, TokuDB and Azure MySQL extensions

ani-sha:
related to https://github.com/debezium/debezium/pull/6074

PR #4412, opened 2025-02-12
Support MySQL GRANT SENSITIVE_VARIABLES_OBSERVER

kgalieva:
Adding support for `GRANT SENSITIVE_VARIABLES_OBSERVER` for MySQL
https://dev.mysql.com/doc/refman/8.4/en/privileges-provided.html


kaby76:
@mike-lischke Is this change correct to the Oracle grammar? I just updated the grammar to the lastest from the vscode extension. https://github.com/antlr/grammars-v4/pull/4408

mike-lischke:
This PR should be rejected. It adds a [dynamic privilege](https://dev.mysql.com/doc/refman/8.4/en/privileges-provided.html#privileges-provided-dynamic) to the grammar. The `grant` statement not only allows to specify predefined privileges, but simple identifiers. Hence it allows input like `grant SENSITIVE_VARIABLES_OBSERVER  to mike@mysql.com` and does a runtime lookup for the given priv name.

The upstream fix is also wrong btw. @kgalieva.

mike-lischke:
Side note: this PR is not for the Oracle grammar, but the one from Positive Technologies. I'm not sure this repository should keep multiple MySQL grammars.

kaby76:
> Side note: this PR is not for the Oracle grammar, but the one from Positive Technologies. I'm not sure this repository should keep multiple MySQL grammars.

Sorry, I see that now. I only saw that a test was modified over in the Oracle grammar, and it parses fine.
```
$ ./bin/Debug/net8.0/Test.exe -input 'GRANT SENSITIVE_VARIABLES_OBSERVER ON *.* TO `admin`@`%`;' -tree
(queries (query (simpleStatement (accountManagementStatement (grantStatement GRANT (roleOrPrivilegesList (roleOrPrivilege (roleIdentifierOrText (roleIdentifier (pureIdentifier SENSITIVE_VARIABLES_OBSERVER))))) ON (grantIdentifier * . *) TO (grantTargetList (userList (user (userIdentifierOrText (textOrIdentifier (identifier (pureIdentifier `admin`))) (userVariable @ (textOrIdentifier (identifier (pureIdentifier `%`))))))))))) ;) <EOF>)
CSharp 0 string0 success 0.103722
Total Time: 0.2760364
02/13-06:20:05 ~/issues/g4-current/sql/mysql/Oracle/Generated-CSharp
```

I agree. It might be time to archive the PT grammar, or just remove it.

mike-lischke:
@teverett @KvanTTT What do you think? Shall we remove the PT MySQL grammar, in favor of the Oracle grammar? I know the PT grammar also supports MySQL 5.6 and 5.7, but these versions are EOL already and the PT grammar is only partially correct. If people really want support for these versions, they can download an earlier version of the current Oracle grammar from https://github.com/mysql/mysql-workbench/blob/8.0/library/parsers/grammars/.

KvanTTT:
I'm not sure, because other users still use these grammars and even make pull requests. For instance, the latest was on Jan 16: https://github.com/antlr/grammars-v4/pull/4383

Please ask them at first (@ani-sha) and create a related topic to make it noticable.

mike-lischke:
I have a conflict of interests here, as my grammar would prevail, so maybe someone else is the better choice to drive this process. This PR here as well as the one you mentioned @KvanTTT are both wrong (I explained that above).

All I can do here is to suggest to remove the PT grammar, because it is not correct. It doesn't matter that someone still files PRs for it (it shows only the grammar needs changes to be correct). What happens instead it that PRs like this and #4383 add things that don't belong in this grammar. It's a MySQL grammar, not a TokuDB or Azure grammar. So with every such PR the grammar gets more and more incorrect, not to mention that it is way behind the actual MySQL time line (we are already working on 9.x).

teverett:
I have no concerns with keeping the PT grammar.  If it's wrong, perhaps it could be corrected?

KvanTTT:
I have no objection regarding removing, but please create an issue regarding the topic to notify users explicetly.

ani-sha:
Hi @KvanTTT, thanks for tagging me in this discussion. I would like to highlight that [Debezium](https://debezium.io/) has been using the Positive Technologies MySQL grammar since at least 2017. To keep up with the evolving MySQL syntax, our maintainers (and community members) actively contribute updates to the upstream PT grammar whenever we make any changes in our grammar.

>  It doesn't matter that someone still files PRs for it (it shows only the grammar needs changes to be correct). What happens instead it that PRs like this and https://github.com/antlr/grammars-v4/pull/4383 add things that don't belong in this grammar. It's a MySQL grammar, not a TokuDB or Azure grammar. So with every such PR the grammar gets more and more incorrect.

In the past, we haven’t received feedback on our PRs regarding these concerns. We didn’t anticipate that the TokuDB or Azure grammar PRs would be merged here, but since they were, we contributed to ensure support for all MySQL grammar variations. That said, we are open to refining our contributions and providing specific semantic predicates where needed.

> Please create an issue regarding the topic to notify users explicetly and get more other opinions. 

That would be helpful. I will need to discuss these changes with my team.

mike-lischke:
Please note: it was not my intention to use this issue to decide about the future of the PT grammar. All I wanted was to get a general yes/no input from important people in this process.

Of course there must be a good deprecation path etc. to phase out the grammar. Ideally one of the owners of it should take the stab and start this process.

teverett:
@KvanTTT ok to merge this?

teverett:
@ani-sha @KvanTTT  are we waiting on feedback from @ani-sha  before merging?

ani-sha:
@teverett, please keep me informed about the deprecation process for Positive Technologies MySQL grammar. Once your deprecation process starts, my team has decided to transition to Oracle MySQL grammar, but we need to factor in your timeline. If anything occurs, please tag me so we can stay updated with the latest information. 

Also, I have a question: If you plan to remove PT grammar, what about [MariaDB](https://github.com/antlr/grammars-v4/tree/master/sql/mariadb) grammar, which is built on top of PT grammar?

mike-lischke:
@ani-sha The MariaDB is not based on any MySQL grammar, AFAIK. It's an own grammar defining everything from scratch.

ani-sha:
@Mike-Lischke Well, I don't think it was created entirely from scratch. If you compare the MariaDB grammar with the PT grammar, it seems to be based on that. @Naros, isn't that right?

mike-lischke:
@ani-sha It's probably copied, but it does not depend on the PT grammar, so removing that grammar has no impact on the MariaDB grammar.

@all I took the freedom and created #4429 to formally start the deprecation process. Please add your thoughts about my proposal there.

kaby76:
In a scenario where MariaDB reuses the MySQL grammar, the Trash tester would require a few changes. In a .g4, the `import` statement does not contain any path information (it doesn't even have ".g4" in the grammar file name), so path info is placed on the Antlr tool command-line. The tester, which works across all OSes and all targets/ports, [uses a "config file"](https://github.com/mike-lischke/antlr-ng/issues/55) with the desc.xml. It would require a change to support `import` path information, but it's an easy change. (What is more difficult to understanding how Antlr "import" works.)

mike-lischke:
I explained the current import process in detail in my [feature issue](https://github.com/mike-lischke/antlr-ng/issues/51) to extend it. This is, however, not related to this issue here.

Naros:
> @Naros, isn't that right?

Hi @ani-sha, it was, particularly given that up to that point, we had been contributing changes for both MySQL and MariaDB into the MySQL grammar for PT for quite some time based on the feedback we received from the maintainers here.

As for these PR changes, if the maintainers believe that the rules aren't correct, we should at least take their suggestions back to our code base and test them. If those suggestions pass, we should adjust not only our local grammar but also this PR so that the code remains aligned, at least until the future of the PT grammar is decided.

As to the proposal for deprecation, I'll leave my opinions in the #4429 issue.

kaby76:
> we had been contributing changes for both MySQL and MariaDB into the MySQL grammar for PT for quite some time based on the feedback we received from the maintainers here.

It sounds like the [mariadb grammar](https://github.com/antlr/grammars-v4/tree/a9c0af477e3047be24abf525043db4e3b71d5f08/sql/mariadb) is a copy of the [PT grammar](https://github.com/antlr/grammars-v4/tree/a9c0af477e3047be24abf525043db4e3b71d5f08/sql/mysql/Positive-Technologies). I'll take a look to see what a grammar diff shows and create an Issue to track fixing the mariadb grammar.

Antlr4 `import` has always existed, but it wan't possible to do cross-grammar import in this repo because neither the old Maven tester nor the Trash tester supported cross-grammar import. But, I'm implementing this now for the Trash tester, and I have a [grammar for TrapC](https://github.com/kaby76/grammars-v4/tree/5a3e69b1eed01b1847d1705d7f7bbff040e59f73/trapc) that uses it. This should help reduce grammar forking.

mike-lischke:
@teverett Since this patch is for the PT grammar I have no objections to merge it.

teverett:
@kgalieva  thanks!

PR #4423, opened 2025-02-24
[PlSql] Support `DEFAULT ON NULL` in alter table modify column properties

Naros:
Introduces support for `ALTER TABLE MODIFY COLUMN ... DEFAULT ON NULL ...` syntax, shown here:

image

Reference: [source](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/ALTER-TABLE.html#GUID-552E7373-BF93-477D-9DA3-B2C9386F2877__CJAIJADJ)


teverett:
@Naros thanks!

PR #4424, opened 2025-02-24
The ANTLRv4 grammar is now equivalent to the ANTLRv3 grammar

mike-lischke:
- The LexBasic.g4 grammar has been removed and the few used rules have been taken over into ANTLRv4Lexer.g4. This lowers complexity and increases readability of error messages (e.g. missing ';' instead of missing SEMI).
- The wrong definition of element options for action (and sempred) blocks has been fixed.
- The accepted token types list for element options has been fixed.
- A predefined token file is now part of the set to make sure tokens appear in the right order (for tree walkers that were ported from ANTLRv3).

This grammar is now used in antlr-ng for generating parsers, so it should also appear as the official grammar (since the one in ANTLRv4 is actually a v3 grammar).


kaby76:
We're going to need the predefined.tokens file in grammars-v4/antlr/antlr4/predefined.tokens since it's source code for the grammar and antlr4 fails without the file. Github Actions says success, but the build failed. It will be fixed with the antlr-ng PR https://github.com/antlr/grammars-v4/pull/4385. Thanks.

mike-lischke:
Oh, I didn't know tokens files are ignored. Should I force-add it or do you have a different solution?

kaby76:
The .tokens file isn't ignored apparently. The old "antlr4" tool requires the file, fails with `error(114): ANTLRv4Lexer.g4:54:17: cannot find tokens file .\predefined.tokens` if it's not there.

teverett:
@mike-lischke  thanks!

PR #4425, opened 2025-02-25
[build] Update to version 0.23.16 of Trash Toolkit.

kaby76:
This build is mainly to fix one issue: https://github.com/antlr/grammars-v4/issues/4434 This PR updates the Trash Toolkit to the latest available, version 0.23.16. In doing so, it also includes a couple of new features. One is related to the discussion on [maintaining the PT MySQL](https://github.com/antlr/grammars-v4/issues/4429#issuecomment-2692719019). It also updates the build to include the new Antlr-ng generator because that is required for testing the Antlr4ng runtime. The runtime no longer wraps the old Java parser generator in the runtime.

The update fixes several problems with the grammar dependency analysis in trgen, which caused testing for https://github.com/antlr/grammars-v4/pull/4424 to report "success" when it really failed; It should have warned that the changes to [antlr/antlr4](https://github.com/antlr/grammars-v4/pull/4424) contained errors, e.g., the use of ["this" in the Python3 fork of the grammar](https://github.com/antlr/grammars-v4/blame/a27f0e23507876c4f7babe8d6edd754cd71f5ff9/antlr/antlr4/Python3/ANTLRv4Lexer.g4#L114), and the parse .errors file for three.g4 needed to be remastered.

Trgen works for all of grammars-v4, except for sql/hive/v3. It turns out that that grammar is invalid: https://github.com/antlr/grammars-v4/issues/4428

In addition, this PR adds "cross-import grammar" testing. By that, one can write a grammar in one directory that uses Antlr `import` statements to override another grammar. To illustrate and test this capability, I added the trapc grammar. https://github.com/antlr/grammars-v4/compare/master...kaby76:grammars-v4:g4-4430

PR #4435, opened 2025-03-06
[antlr/antlr4] Implement target agnostic ports for antlr4.

### LINKED ISSUE #4434: [antlr/antlr4] Ports broken. Tests broken.
With the recent change in the grammar (https://github.com/antlr/grammars-v4/pull/4424), ports and tests are broken.

kaby76:
This PR fixes #4434 for the [antlr/antlr4 grammar](https://github.com/antlr/grammars-v4/tree/master/antlr/antlr4) (where I got the grammar working, but had to remove the Cpp and Python3 ports temporarily). I added new implementations for the Cpp and Python3 ports along with the others, including the Cpp (where the trgen tester was broken). The build is much better, fixing trgen with an update to the latest of the toolkit. Finally, I fixed the Powershell tester script so MacOS builds don't hang so often. The main change is #4434, but the rest of the changes are needed just to make sure all the ports of antlr/antlr4 work.

Recapping what happened over the last couple of weeks: Mike Lischke fixed the antlr/antlr4 grammar with PR #4424. But, the PR never got a clean bill of health from testing because the tester itself was broken! One problem was that the .errors files needed to be remastered. A second problem was that Mike broke the Python3 and Cpp ports, e.g., copying over some Java code from the original grammar to the forked grammar but then not updating the target language syntax. The third problem was that neither of the first two problems were reported because the build was broken, crashing in trgen but not reporting any error. To get things back on track, I first started fixing the build, more or less, with https://github.com/antlr/grammars-v4/pull/4425. I removed the [antlr/antlr4 grammar](https://github.com/antlr/grammars-v4/tree/master/antlr/antlr4) Cpp and Python3 ports, not only because they were broken, but because making copies of a grammar for a particular port is a really bad practice (we tend to get version skews in the grammar). But, that didn't fix everything.

This PR, #4435, completes the fixes:
* Updated Trash to version 0.23.17.
* Added the rest of the ports to the grammar. This required changes to the grammar to make it "target agnostic." *NB: I did not add a Go port because it goes back to the problem with the Go target not implementing a class structure correctly.*
    * Antlr Dart runtime does not have a method to get the size of the mode stack. I worked around this problem by doing popMode()/pushMode() within a try-block.
    * Python cannot have circular imports, so there is no way to reference mode constants in the lexer base class. To work around this problem I used hardwired integer constants in the code.
* Updated the test suite to include in examples/ a copy of all existing .g4's from [grammars-v4](https://github.com/antlr/grammars-v4). This expands on the list of tests from 16 .g4's to 554 .g4's. This is best because the new antlr-ng tool uses the grammars-v4/antlr/antlr4 grammar!
* Updated the testing scripts on how to compute what grammars to test, so that the MacOS doesn't hang so much after adding all the .g4's to the example/ test suite.


teverett:
@kaby76  thanks!

Issue #4429, opened 2025-02-27
Deprecate the MySQL Grammar from Positive Technologie


mike-lischke:
It has been suggested in the past a few times already and it was discussed in issue #4412: keeping two different MySQL grammars leads to confusion and unnecessary work. There are currently two grammars:

- [Oracle](https://github.com/antlr/grammars-v4/blob/master/sql/mysql/Oracle) - A complete MySQL grammar including the latest language features of MySQL. This grammar is directly derived from the MySQL Yacc server grammar and is updated twice a year to stay in sync with the MySQL server enhancements.
- [Positive Technologies](https://github.com/antlr/grammars-v4/blob/master/sql/mysql/Positive-Technologies) - A grammar created from the official documentation (5.6, 5.7, 8.0)

The first one is the preferred one as it includes all language features starting with MySQL 8.0, while the other one still supports version 5.6 and 5.7 (which are already EOL'd).

We can see occasional pull requests to add newer language features to the PT grammar, but they are always only punctual, not going the full way (and sometimes they are plain wrong).

Since the Oracle grammar supports the full MySQL language properly, is released under a very permissive license and is maintained by Oracle, it makes sense to keep that grammar as the only one for MySQL and deprecate the grammar from Positive Technologies.

I'm not a maintainer of this repository, so I can only suggest a deprecation process:

1. Add a prominent text about the deprecation to the readme in the `sql/mysql` folder and to the readme in the `sql/mysql/Positive-Technologies`.
2. Add a prominent note in the grammars (lexer/parser) that also inform about the deprecation.
3. Reject all future PRs for this grammar with a note that it is deprecated.
4. Give it a few months (maybe 3?) to let this sink in.
5. Remove the PT grammar and restructure the `mysql` folder and its associated tests.

It's important to stick to such a plan, once decided, to make downstream planning possible. Users of the PT grammar have to plan some time for migration.

An additional option could be to add a `DEPRECATED` folder in the root of this repository and move this and future deprecated grammars to that. Grammars in this folder are not tested and no PRs will be accepted for them.

Naros:
Hi @mike-lischke, thank you for starting this discussion. I think it's indeed good to have, despite the outcome.

First, I agree that deprecating the PT grammar makes the most long-term sense. Given that it's driven by the upstream grammar used by the MySQL engine, this has the absolute benefit of implementing anything supported by the engine in its complete form in the grammar, which addresses all the shortcomings with the PT grammar led by the community.

[...]

PR #4438, opened 2025-03-12
Deprecate the Positive Technologies MySQL Grammar

mike-lischke:
With this patch the PT grammar will taken out of the normal maintenance. It is no longer supported and no PRs should be accepted for it. See also #4429.


kaby76:
My updates are here: https://github.com/kaby76/grammars-v4/tree/g4-4438

mike-lischke:
@kaby76 You should be able to directly push to this PR. Edits are allowed here.

mike-lischke:
Making this PR a draft, to be handled after the deprecation period of the PT grammar.

PR #4439, opened 2025-03-12
[PlSql] Separate case_statement from case_expression

MartinKarlgrenIMI:
To conform with Oracle's documentation for
searched_case_statement / simple_case_statement:
https://docs.oracle.com/en/database/oracle/oracle-database/23/lnpls/CASE-statement.html

searched_case_expression / simple_case_expression: https://docs.oracle.com/en/database/oracle/oracle-database/23/sqlrf/CASE-Expressions.html

This makes the rules more specific, to disallow seq_of_statements within an expression, for example.

Also fixes a bug where a column alias would be incorrectly parsed as label_name for the CASE in this example:

SELECT CASE WHEN 1=1 2 ELSE 3 END COLUMN_ALIAS FROM DUAL;


teverett:
@MartinKarlgrenIMI  thanks

PR #4441, opened 2025-03-14
[blueprint] Add blueprint grammar

xndcn:
Blueprint is Android build script similar to Bazel BUILD files.
It is documented here: https://source.android.com/docs/setup/reference/androidbp
and there is an official parser:
https://android.googlesource.com/platform/build/blueprint/+/refs/heads/main/parser/parser.go

### DIFF
--- a/blueprint/README.md
+++ b/blueprint/README.md
@@ -1,0 +1,13 @@
+# Blueprint Grammar
+
+## Description
+
+This is an Antlr4 grammar for Blueprint (also known as Android.bp), the build language used by Android build system.
+
+## License
+
+[MIT](https://opensource.org/licenses/MIT)
+
+## Links
+* [Android Blueprint Documentation](https://source.android.com/docs/setup/reference/androidbp)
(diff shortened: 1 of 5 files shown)


kaby76:
Please remove the PHP target. It does not work for this grammar.  https://github.com/antlr/grammars-v4/blob/a2a54d12e660299a74d6bf4b3515b0772b4044bc/blueprint/desc.xml#L3

xndcn:
> Please remove the PHP target. It does not work for this grammar.  https://github.com/antlr/grammars-v4/blob/a2a54d12e660299a74d6bf4b3515b0772b4044bc/blueprint/desc.xml#L3

Thanks, have removed

kaby76:
Looks good.
* The one test file covers the grammar.
* There's no ambiguity detected in the example.

Could you add a readme.md that generally follows the format used elsewhere (e.g., [here](https://github.com/antlr/grammars-v4/blob/master/antlr/antlr4/readme.md)), and includes your description above to the doc and parser source?

PR #4443, opened 2025-03-15
Mark the PT grammar as being deprecated.

mike-lischke:
Added notes to the Positive Technologies MySQL grammar about the deprecation process that started for it.


Naros (approved):
Thanks @mike-lischke

PR #4455, opened 2025-03-25
Fix: identifier name cannot include '<' or '>'

arbinada-com:
Expressions like "Abc<>Cde" (without spaces) is considered as a single identifier


teverett:
@arbinada-com  thanks!

PR #4470, opened 2025-04-13
[c] Fix #4463 -- unused rule removed.

### LINKED ISSUE #4463: Rule "nestedParenthesesBlock" never used or referenced in C.g4 grammar
What is the purpose of rule [nestedParenthesesBlock](https://github.com/antlr/grammars-v4/blob/master/c/C.g4#L349C1-L349C6) in `C.g4`? It's never referenced or used by any other rules in the grammar which already handle nested parentheses anyway.

```antlr
nestedParenthesesBlock
    : (~('(' | ')') | '(' nestedParenthesesBlock ')')*
    ;
```

kaby76:
This PR removes an unused rule, fixing #4463.

There should be a check in Trash to find rules that take part of a recursive cycle but are unused in any rules outside of the cycle (i.e., disconnected cycles from the start rule).

### DIFF
--- a/c/C.g4
+++ b/c/C.g4
@@ -346,10 +346,6 @@ gccAttribute
     ('(' argumentExpressionList? ')')?
     ;
 
-nestedParenthesesBlock
-    : (~('(' | ')') | '(' nestedParenthesesBlock ')')*
-    ;
-
 pointer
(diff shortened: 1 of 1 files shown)


teverett:
@kaby76  thanks!

Issue #4469, opened 2025-04-13
[golang] Ambiguity in `arguments`.


kaby76:
Continuing to disambiguate the [golang grammar](https://github.com/antlr/grammars-v4/tree/master/golang).

In [anonymousMethods.go](https://github.com/antlr/grammars-v4/blob/master/golang/examples/anonymousMethods.go), line 7 has the string `fmt.Println(s)`. This string is ambiguous because the `s` can be parsed two ways:
```
[...]

PR #4488, opened 2025-05-16
[golang] Fix for #4469--Add symbol table.

kaby76:
First stab at a real symbol table for Go to avoid problems in ambiguity.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,189 +3,189 @@
   "isRoot": true,
   "tools": {
     "trcaret": {
-      "version": "0.23.17",
+      "version": "0.23.18",
       "commands": [
         "trcaret"
       ],
       "rollForward": false
     },
     "trcover": {
-      "version": "0.23.17",
+      "version": "0.23.18",
       "commands": [
         "trcover"
       ],
       "rollForward": false
     },
     "trgen": {
-      "version": "0.23.17",
+      "version": "0.23.18",
       "commands": [
         "trgen"
       ],
       "rollForward": false
     },
     "trglob": {
-      "version": "0.23.17",
+      "version": "0.23.18",
       "commands": [
         "trglob"
       ],
       "rollForward": false
     },
     "triconv": {
-      "version": "0.23.17",
+      "version": "0.23.18",
       "commands": [
         "triconv"
       ],
       "rollForward": false
     },
     "trparse": {
-      "version": "0.23.17",
+      "version": "0.23.18",
       "commands": [
         "trparse"
       ],
       "rollForward": false
     },
     "trquery": {
-      "version": "0.23.17",
+      "version": "0.23.18",
       "commands": [
         "trquery"
       ],
       "rollForward": false
     },
     "trtext": {
-      "version": "0.23.17",
+      "version": "0.23.18",
       "commands": [
         "trtext"
       ],
       "rollForward": false
     },
     "trwdog": {
-      "version": "0.23.17",
+      "version": "0.23.18",
       "commands": [
         "trwdog"
       ],
       "rollForward": false
     },
     "trxgrep": {
-      "version": "0.23.17",
+      "version": "0.23.18",
       "commands": [
         "trxgrep"
       ],
       "rollForward": false
     },
     "trxml": {
-      "version": "0.23.17",
+      "version": "0.23.18",
       "commands": [
         "trxml"
       ],
       "rollForward": false
     },
     "trxml2": {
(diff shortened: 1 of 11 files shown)


kaby76:
Annoyingly, Antlr does not parse correctly when predicates are placed after a decision point. This occurs when there are two deriviations that lead to the same thing, as with input `make([]fs.DirEntry, 1)` in reader.go in the Go runtime. For this input, there are two trees:
```
expressionStmt
└── expression
    └── primaryExpr
        ├── primaryExpr
        │   └── operand
        │       └── operandName
        │           ├── Attribute TERMINATOR Value '\n' chnl:HIDDEN
        │           ├── Attribute WS Value '\t' chnl:HIDDEN
        │           └── IDENTIFIER
        │               └── "make"
        └── arguments
            ├── Attribute OTHER Value '' chnl:HIDDEN
            ├── L_PAREN
            │   └── "("
            ├── expressionList
            │   ├── expression
            │   │   └── primaryExpr
            │   │       └── methodExpr
            │   │           ├── type_
            │   │           │   └── typeLit
            │   │           │       └── sliceType
            │   │           │           ├── L_BRACKET
            │   │           │           │   └── "["
            │   │           │           ├── R_BRACKET
            │   │           │           │   └── "]"
            │   │           │           └── elementType
            │   │           │               └── type_
            │   │           │                   └── typeName
            │   │           │                       ├── Attribute OTHER Value '' chnl:HIDDEN
            │   │           │                       └── IDENTIFIER
            │   │           │                           └── "fs"
            │   │           ├── Attribute OTHER Value '' chnl:HIDDEN
            │   │           ├── DOT
            │   │           │   └── "."
            │   │           └── IDENTIFIER
            │   │               └── "DirEntry"
            │   ├── Attribute OTHER Value '' chnl:HIDDEN
            │   ├── COMMA
            │   │   └── ","
            │   └── expression
            │       └── primaryExpr
            │           └── operand
            │               └── literal
            │                   └── basicLit
            │                       └── integer
            │                           ├── Attribute WS Value ' ' chnl:HIDDEN
            │                           └── DECIMAL_LIT
            │                               └── "1"
            ├── Attribute OTHER Value '' chnl:HIDDEN
            └── R_PAREN
                └── ")"
```
and
```
expressionStmt
└── expression
    └── primaryExpr
        ├── primaryExpr
        │   └── operand
        │       └── operandName
        │           ├── Attribute TERMINATOR Value '\n' chnl:HIDDEN
        │           ├── Attribute WS Value '\t' chnl:HIDDEN
        │           └── IDENTIFIER
        │               └── "make"
        └── arguments
            ├── Attribute OTHER Value '' chnl:HIDDEN
            ├── L_PAREN
            │   └── "("
            ├── type_
            │   └── typeLit
            │       └── sliceType
            │           ├── L_BRACKET
            │           │   └── "["
            │           ├── R_BRACKET
            │           │   └── "]"
            │           └── elementType
            │               └── type_
            │                   └── typeName
            │                       └── qualifiedIdent
            │                           ├── Attribute OTHER Value '' chnl:HIDDEN
            │                           ├── IDENTIFIER
            │                           │   └── "fs"
            │                           ├── Attribute OTHER Value '' chnl:HIDDEN
            │                           ├── DOT
            │                           │   └── "."
            │                           └── IDENTIFIER
            │                               └── "DirEntry"
            ├── Attribute OTHER Value '' chnl:HIDDEN
            ├── COMMA
            │   └── ","
            ├── expressionList
            │   └── expression
            │       └── primaryExpr
            │           └── operand
            │               └── literal
            │                   └── basicLit
            │                       └── integer
            │                           ├── Attribute WS Value ' ' chnl:HIDDEN
            │                           └── DECIMAL_LIT
            │                               └── "1"
            ├── Attribute OTHER Value '' chnl:HIDDEN
            └── R_PAREN
                └── ")"
```

With this input, we see two derivations for `arguments`: one that uses only `expressionList`, and the other that expands with `type_ ',' expressionList`. If we place the predicate at the end of `type_`, the first parse would be eliminated because `fs` is not a type, but a package name. So, it cannot be a receiver. The second choice must be made. However, this is too late for a one-pass parser.

One hack is to perform the parse twice at the beginning of `argments`, then force one or the other choice depending on the result.

kaby76:
It turns out that we should have both the ambiguous grammar from spec without symbol table and one with symbol table. Using the strategy I outline above to implement disambiguation predicates, the original grammar without predicates is exceedingly important. This is because as one implements the actions and semantic predicates, we need to understand where the ambiguity is actually coming from so we can add in the "syntactic predicate" hack.

Issue #4492, opened 2025-05-31
[lua] Add a significant, real regression test suite.


kaby76:
As mentioned in PR https://github.com/antlr/grammars-v4/pull/4491, the test suite for the grammar is small. I expanded the test suite from a whopping three files to 42 files. While still inadequate, I did what I could to beef up the completely inadequate test suite with something that tests most of the grammar. The time has come to add a large test suite for regression testing.

I mentioned in https://github.com/antlr/grammars-v4/pull/4491, I created a larger test suite of .lua files from the links in the comments in the .g4. That test suite is here: [lua-test-suite.tar.gz](https://github.com/user-attachments/files/20521147/lua-test-suite.tar.gz). I recommend that the files that don't parse be kept with the test suite for this grammar, but renamed with a `.lua.fail` extension to denote that the grammar does not work with the input.

In addition, we need to employ Trash XPath expressions that check the correctness of the parse trees. We cannot keep using .tree files to test parse tree output. This is because of two reasons:
* Small changes in the grammar create large deltas for the .tree files because the [Antlr4 ToStringTree()](https://github.com/antlr/antlr4/blob/09d527fddd8dcd3733d7275cf3436a1a9e1e12db/runtime/CSharp/src/Tree/Trees.cs#L37) displays the entire parse tree on one line. Naive programmers do not check the parse tree.
* Antlr4 ToStringTree() does not output the token types for leaf nodes, but instead only outputs the token text. This is terrible because a grammar could be changed in which the token type is wrong, but appear correct. Antlr4-style parse tree display for testing parse tree correctness is completely inadequate and must be abandoned.

So, to test that https://github.com/antlr/grammars-v4/blob/a5cc838f34ccde47ba9fba1f08e50774b08cf467/lua/examples/3885.lua actually parses with the correct tree structure, the tester should check something like this, which tests that the parse results in two statements, not four.
```
$ trparse in.txt | trxgrep 'count(//stat)'
CSharp 0 in.txt success 0.0408087
4
05/31-07:46:06 ~/issues/g4-current/lua/Generated-CSharp
$ trparse in.txt | trxgrep 'count(//stat) = 2'
CSharp 0 in.txt success 0.042838
False
05/31-07:46:26 ~/issues/g4-current/lua/Generated-CSharp
$ cat in.txt
local _addr = tostring(_obj).sub(8)
local a = GetClass(data).OtherFunc(self, ...)
[...]

PR #4491, opened 2025-05-30
fix 3885 and 3923

matthargett:
fixes #3885 and #3923, and repairs a few dozen other regressions since my last contribution. (there are still several dozen other regressions that remain, but I wanted to keep this PR small and focused on user-reported issues.)

 when all of my previous contributions were wiped out, introduced were incorrect handling of left-recursive patterns common in Lua expressions, especially in how suffixes (like function arguments (), member access .name, or indexing []) are applied to a base expression. 

I had previously done extensive regression testing on numerous public real-world Lua codebases, but it looks like the person who removed all my changes did significantly less testing. the changes I contributed were used in pmd's copy-paste detector tool, which I deployed to ~20 Lua repositories at Roblox that ran on ~400KLOC of Lua code a minimum of dozens of times per workday on code written by over 100 global engineers. it's a super bummer to see my rigor wiped out and my name completely removed.

this fix should help the two issues reported by users, and fix some of the new regressions introduced in parsing the luvit codebase, and stylua test suite. if the ANTLR admins/maintainers don't want the change, and prefer to leave Lua users without a working grammar (and great ANTLR-based tools like pmd's copy-paste detector), no problem. I think it's probable that the ANTLR admins like @teverett  didn't quite realize the extent of the regression, and sadly I wasn't available to do an in-depth review.

Issue #3923, opened 2024-01-11
Problem when there are consecutive function calls on the right side of the assignment statement


Dongyang0810:
`local a = GetClass(data).OtherFunc(self, ohterData)`
GetClass is a function that returns an object with a member function OtherFunc.
GetClass(data).OtherFunc(self, otherData) should be parsed as a tree structure where the root node is a function call, and the child node is also a function call.

The parsing result of GetClass(data).OtherFunc(self, ohterData) when parsing in a single line is inconsistent with the parsing result when placed on the right side of the assignment statement.

```
local a = GetClass(data).OtherFunc(self, ohterData)
GetClass(data).OtherFunc(self, ohterData)
```
![image](https://github.com/antlr/grammars-v4/assets/37564912/793918cc-8fd6-4a2e-9aae-92ffeb77abdf)

How can I resolve this issue?

KvanTTT:
Please clarify the grammar you are talking about.

KvanTTT:
Got it, I've seen `Lua` on the image.

bkiers:
Instead of:

```
prefixexp
    : NAME ('[' exp ']' | '.' NAME)*
    | functioncall ('[' exp ']' | '.' NAME)*
    | '(' exp ')' ('[' exp ']' | '.' NAME)*
    ;
```

try this:


```
prefixexp
    : functioncall ('[' exp ']' | '.' NAME)*
    | NAME ('[' exp ']' | '.' NAME)*
    | '(' exp ')' ('[' exp ']' | '.' NAME)*
    ;
```

bkiers:
This seems to be an exact duplicate of https://github.com/antlr/grammars-v4/issues/3885

Dongyang0810:
> This seems to be an exact duplicate of https://github.com/antlr/grammars-v4/issues/3885
yes,I saw the question after I asked it.

Dongyang0810:
> Instead of:
> 
> ```
> prefixexp
>     : NAME ('[' exp ']' | '.' NAME)*
>     | functioncall ('[' exp ']' | '.' NAME)*
>     | '(' exp ')' ('[' exp ']' | '.' NAME)*
>     ;
> ```
> 
> try this:
> 
> ```
> prefixexp
>     : functioncall ('[' exp ']' | '.' NAME)*
>     | NAME ('[' exp ']' | '.' NAME)*
>     | '(' exp ')' ('[' exp ']' | '.' NAME)*
>     ;
> ```

[...]

Issue #3885, opened 2023-12-12
Lua grammar handles method calls incorrectly


Martmists-GH:
Sample snippet: `local _addr = tostring(_obj):sub(8)`
![image](https://github.com/antlr/grammars-v4/assets/16361449/4b1b97f2-1c5a-4801-a3a7-40d327b755cb)

kaby76:
This input is ambiguous.

Lua allows two statements without a semi-colon separator. `print(1) print(2)`. So, for `local _addr = tostring(_obj).sub(8)`, we have two statements, `local _addr = tostring` and `(_obj).sub(8)`.

But, clearly the intent is for it to be one statement.

I haven't looked at the source code for the interpreter yet, but I suspect there are missing disambiguation rules for the parse.

***Reordering rule alts is insufficient to fix this problem!***

PR #4493, opened 2025-05-31
[lua] Fix for #4492 -- add large test suite.

kaby76:
This is a fix for #4492.

In this PR, I add 1557 .lua files/817920 lines of Lua code from the links of open-source software towards implementing a real regression test suite for the [Lua grammar](https://github.com/antlr/grammars-v4/tree/master/lua). This addresses a legitimate complaint, that we don't have a large regression test suite. https://github.com/antlr/grammars-v4/pull/4491#issue-3102100717 (Note, I've been saying for years that we don't have very good test suites for many grammars. It takes time to add good testing.) With this PR, we now cover almost 100% of the grammar: 
[cover.html.txt](https://github.com/user-attachments/files/20540139/cover.html.txt)

This regression test suite is to validate that the parser accepts Lua source code. For the moment, the regression test suite does not test two other important requirements.
* Tests to reject invalid Lua code. Of the 1557 files in the test suite, 1411 .lua files are valid Lua code that work with the current grammar (https://github.com/antlr/grammars-v4/tree/3e371d221c783c8e4b143409695da137d538521f/lua). The rest of the Lua code that do not parse (179 files, which I assume are valid Lua), these were renamed to end with a `.lua.fail` extension. These files can be renamed back to the `.lua` extension as the grammar is refined in future PRs.
* Validate the tree output. We know that there are problems with ambiguity in the EBNF grammar (#3885 and #3923). These can be fixed with semantic predicates, but I haven't had time to fix them yet. The alternative fix is to refactor the grammar as with #4491.
 
The desc.xml script was updated to test only files with the `.lua` extension, and only for the CSharp target because full checking is unnecessary across all ports. In fact, we can't do full testing across all ports because some targets, like PHP and Python3, are slow.

This PR updates the test scripts due to performance issues for testing a large test suite. On my machine, prior to the changes contained in this PR to the testing templates, the test suite took 10 minutes. This was too long because we specifically implemented a 5-minute timeout threshold for testing in order to keep costs for using Github Actions low, and because people don't like waiting a long time for the results of a change they are testing.

Part of the problem was the call to iconv to test whether the input is valid utf-8. The filter was removed because it didn't add much value, and really slowed the testing down. Further, some of the .lua files in the test suite were flagged as not being utf-8, and some crashed the triconv tool (a platform-independent implementaiton of Linux iconv). I have yet to understand why iconv insists some of the .lua are not in utf-8. But, the .lua files were parsing anyways.

With the change, the longest _Github Action Job_ was for the CSharp/Windows job, and it completed in 7 minutes, which means that the entire testing took only slightly longer than that. In my opinion, this is an acceptable time to wait.

## This PR does not correct outstanding Issues on the lua grammar
Fixes for #3885 and #3923 should include checks to validate the parse tree output from the parser. Trash XPath predicates are perfect for this task! The old method of testing .tree file output is baroque. It should be abandoned because it is too sensitive to small changes in the grammar. Additionally, time and again, I've had to correct problems people introduced because they didn't actually go through the tree diffs and validate that the changes were correct.

Prior to this PR, no parse trees were actually tested for the [lua grammar](https://github.com/antlr/grammars-v4/tree/master/lua). Before the last major overhaul (PR https://github.com/antlr/grammars-v4/pull/3752 https://github.com/antlr/grammars-v4/commit/4d3b08fa77c5820a1cdeaf6a016aee7ce1c7fdeb), we didn't do any significant testing (roughly 60% of the grammar), and there was ***zero*** testing of tree output. When I made the change, I didn't know of any further requirements. Issues #3885 and #3923 are legitimate complaints and will be addressed, and with them, the parse tree structure requirements will be made explicit and validated.

PR https://github.com/antlr/grammars-v4/pull/4491 includes a nice small two-line test for the correct parse tree, which should have two `stat`'s, not four. But the PR does not include a .tree to validate the parse tree. The whole reason for PR https://github.com/antlr/grammars-v4/pull/4491 is to correct the parse tree structure in Issues #3885 and #3923, but it doesn't even test that!

Note, in my opinion, semantic predicates are the preferred approach to fixing these issues because the grammar is scraped from the EBNF in the manual/spec, and refactoring would cause the grammar to diverge from the well-established, officially published EBNF. The other approach is to scrape the grammar from the source code of the implementation, but this would be harder, likely resulting in the same EBNF as published in the manual. https://www.lua.org/source/5.4/lparser.c.html The manual specifically mentions the known ambiguity of the EBNF, and mentions that it should be resolved in a semantic approach. I do know there are other ambiguities in the grammar. I don't know whether a symbol table is going to be required to resolve those ambiguities, but the Antlr4 grammar was split long ago (https://github.com/antlr/grammars-v4/commit/4d3b08fa77c5820a1cdeaf6a016aee7ce1c7fdeb) in anticipation of resolving these ambiguities. All the changes I made then were deliberate, well-thought out, and corrected problems introduced in PR https://github.com/antlr/grammars-v4/pull/2728.

Refactoring the grammar to correct the ambiguity has not worked (https://github.com/antlr/grammars-v4/pull/4491 https://github.com/antlr/grammars-v4/pull/2728 https://github.com/antlr/grammars-v4/pull/4491#issuecomment-2924251763). Refactoring is a multi-step algorithm, and it is prone to errors, even for developers who work regularly with Antlr and have years of experience. This is exactly why I wrote Trash: refactoring cannot be done by hand, off the top of one's head.

------

***Let me be crystal clear about using this test suite: the old Maven tester has been update to parse the valid Lua code in the 1411 .lua files here. But, it will not work beyond that when I expand on the test suite shortly. This is because of many limitations in the old tester. The old Maven tester, which uses the pom.xml, uses a version of Antlr4 that is several years old. If you need to test this locally, I suggest that you use Trash to generate a driver app to test parsing for your OS and target. Alternatively, you will need to write your own driver app for parsing.***


kaby76:
This PR is all set.

kaby76:
I checked whether the Lua compiler is able to compile the new test suite. It appears that most, if not all, of the .lua.fail files that the lua grammar fails on do not parse, and all except for a few of the .lua files compile except for additional static semantics!! This is excellent news because it says that our new test suite is correct, and that the grammar is actually quite correct. This is indeed fabulous news!!!

I tested it with this code:
```
for i in `find . -name '*.lua.fail'`; do echo $i; /c/Users/Kenne/Downloads/lua-5.4.2_Win64_bin/luac54.exe $i; if [ $? -ne 0 ]; then echo non-zero return; fi; done
```

I should rename these files to something else to note that my assumptions are correct, and include the script in the readme.md file. But, for now, this PR is fine to go.

teverett:
@kaby76  thanks!

PR #4497, opened 2025-06-02
[lua] Fix for #4496 -- add the Antlr4ng port, and fix several broken ports.

### LINKED ISSUE #4496: [lua] Target Antlr4ng needs to be added.
While going through the lua grammar, I've noticed that there is no Antlr4ng port. This is a mistake because the new Antlr4ng parser generator is much, much better for TypeScript than the "official" Antlr4 JavaScript/TypeScript target. It needs to be added.

kaby76:
This PR fixes #4496.

For this PR, I added the Antlr4ng port. A comment in the grammar, which transformGrammar.py modifies to a `@header`, was changed to remove the string "C++" because it has to be modified for Antlr4ng as well.

In addition, while perusing the code, I noticed that several ports were broken, and the test scripts were also broken. This PR fixes the Cpp, JavaScript, and TypeScript ports for the grammar, but the test scripts are fixed in another PR. https://github.com/antlr/grammars-v4/pull/4501


teverett:
@kaby76 thanks!

PR #4500, opened 2025-06-03
pragma udf support

michaelhansen079:
Added support for _pragma udf_ introduced in Oracle 12.

https://docs.oracle.com/database/121/LNPLS/udf_pragma.htm#LNPLS1851


teverett:
@michaelhansen079  thanks!

Issue #4499, opened 2025-06-02
[build] Add parse tree validation via XPath predicates.


kaby76:
The .tree files are ok for validation of whole parse trees but it can be too sensitive to tiny, insignificant charges in the parse tree. For example, if a node is renamed,the entire .tree changes. People then have to remaster the file.

As an alternative, XPathe expressions can be used to make assertions on significant parts of the tree and disregard others. We can implent this easily with Trash.

PR #4501, opened 2025-06-04
[build] Fix for #4499 -- Add parse tree validation via XPath predicates.

kaby76:
I'm really excited about this PR! It introduces a new way to validate the parse tree outputted from a parse.

This PR introduces a new approach to validating parse trees with the latest Trash toolkit. It can be used in conjunction with the old Maven-tester style `.tree` validation, or it can be used as an alternative. The only caveat is that the grammar has to work for the `CSharp` target, because trparse only works with a `CSharp` parser app.

The old Maven-tester style of validating parse trees is declarative by the existence of a companion `.tree` file to an input test file. When the tester is run, the parser driver outputs a .tree file of the parse. The .tree file is compared to the previous version of the file, and if it has changed, the test fails.

The tree displayed in the .tree file is the output from the Antlr4 runtime `ToStringTree()`. This output is a Lisp-like expression containing internal and leaf nodes of the parse tree. When someone changes the grammar, we'd expect the parse tree to change. The developer is then required to go through the .tree file and verify that the change is correct.

The problem with this validation approach is that small changes in the grammar will cause the whole file to change. Some of these changes may not be important, such as renaming a parser rule. But the developer must meticulously compare the trees and make sure it is valid.

With the latest version of Trash, [trquery](https://github.com/kaby76/Trash/tree/main/src/trquery) has the `assert` command, which validates the parse tree using a predicate as an XPath expression. This PR includes an example in the Lua grammar to demonstrate the test, as described below.

Consider input stats.lua.
```
a = b + c
(print or io.write)('done')
```

The parse tree for this file is:
```
start_
├── chunk
│   └── block
│       ├── stat
│       │   ├── varlist
│       │   │   └── var
│       │   │       └── NAME
│       │   │           └── "a"
│       │   ├── Attribute WS Value ' ' chnl:HIDDEN
│       │   ├── EQ
│       │   │   └── "="
│       │   └── explist
│       │       └── exp
│       │           ├── exp
│       │           │   └── prefixexp
│       │           │       ├── Attribute WS Value ' ' chnl:HIDDEN
│       │           │       └── NAME
│       │           │           └── "b"
│       │           ├── Attribute WS Value ' ' chnl:HIDDEN
│       │           ├── PLUS
│       │           │   └── "+"
│       │           └── exp
│       │               └── prefixexp
│       │                   ├── Attribute WS Value ' ' chnl:HIDDEN
│       │                   └── NAME
│       │                       └── "c"
│       └── stat
│           └── functioncall
│               ├── Attribute WS Value '\r' chnl:HIDDEN
│               ├── Attribute NL Value '\n'
│               ├── OP
│               │   └── "("
│               ├── exp
│               │   ├── exp
│               │   │   └── prefixexp
│               │   │       └── NAME
│               │   │           └── "print"
│               │   ├── Attribute WS Value ' ' chnl:HIDDEN
│               │   ├── OR
│               │   │   └── "or"
│               │   └── exp
│               │       └── prefixexp
│               │           ├── Attribute WS Value ' ' chnl:HIDDEN
│               │           ├── NAME
│               │           │   └── "io"
│               │           ├── DOT
│               │           │   └── "."
│               │           └── NAME
│               │               └── "write"
│               ├── CP
│               │   └── ")"
│               └── args
│                   ├── OP
│                   │   └── "("
│                   ├── explist
│                   │   └── exp
│                   │       └── string
│                   │           └── CHARSTRING
│                   │               └── "'done'"
│                   └── CP
│                       └── ")"
├── Attribute WS Value '\r' chnl:HIDDEN
├── Attribute NL Value '\n'
└── EOF
    └── ""
```

_NB: This parse tree contains two `stat` nodes. As explained in the Lua manual, the parse tree should contain one `stat` node. This is caused by the ambiguity in the grammar. Unfortunately, the Antlr parser makes the wrong choice. This problem will be addressed in another PR. https://github.com/antlr/grammars-v4/pull/4502_

We know that the parser should output two `stat` nodes (until we fix the grammar). We can test whether this is expected by adding a companion `.trq` file for the input.

stat.lua.trq:
```
(: Check whether there is a correct number of "stat".

    NB!!! This is a placeholder to implement a requirement on
    the number of stat-nodes. This is known to be incorrect for
    the current version of the lua grammar. When tree structure
    is corrected, this file must be fixed.

    For now, the count is 2.
:)

assert count(//stat) = 2, "Invalid number of 'stat' nodes.";
```
The test is performed in the test scripts thus: `dotnet trparse stats.lua | dotnet trquery -c stats.lua.trq`.

Note that we can write multiple tests in one file, and the .trq can include comments that explain the purpose of the predicate. The predicate is followed by a string to print when the predicate fails.

Trash requires a `CSharp` port, so we only validate using the .trq for the `CSharp` target.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,189 +3,189 @@
   "isRoot": true,
   "tools": {
     "trcaret": {
-      "version": "0.23.18",
+      "version": "0.23.19",
       "commands": [
         "trcaret"
       ],
       "rollForward": false
     },
     "trcover": {
-      "version": "0.23.18",
+      "version": "0.23.19",
       "commands": [
         "trcover"
       ],
       "rollForward": false
     },
     "trgen": {
-      "version": "0.23.18",
+      "version": "0.23.19",
       "commands": [
         "trgen"
       ],
       "rollForward": false
     },
     "trglob": {
-      "version": "0.23.18",
+      "version": "0.23.19",
       "commands": [
         "trglob"
       ],
       "rollForward": false
     },
     "triconv": {
-      "version": "0.23.18",
+      "version": "0.23.19",
       "commands": [
         "triconv"
       ],
       "rollForward": false
     },
     "trparse": {
-      "version": "0.23.18",
+      "version": "0.23.19",
       "commands": [
         "trparse"
       ],
       "rollForward": false
     },
     "trquery": {
-      "version": "0.23.18",
+      "version": "0.23.19",
       "commands": [
         "trquery"
       ],
       "rollForward": false
     },
     "trtext": {
-      "version": "0.23.18",
+      "version": "0.23.19",
       "commands": [
         "trtext"
       ],
       "rollForward": false
     },
     "trwdog": {
-      "version": "0.23.18",
+      "version": "0.23.19",
       "commands": [
         "trwdog"
       ],
       "rollForward": false
     },
     "trxgrep": {
-      "version": "0.23.18",
+      "version": "0.23.19",
       "commands": [
         "trxgrep"
       ],
       "rollForward": false
     },
     "trxml": {
-      "version": "0.23.18",
+      "version": "0.23.19",
       "commands": [
         "trxml"
       ],
       "rollForward": false
     },
     "trxml2": {
-      "version": "0.23.18",
+      "version": "0.23.19",
       "commands": [
         "trxml2"
       ],
       "rollForward": false
     },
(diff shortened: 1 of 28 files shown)


kaby76:
@teverett Please merge https://github.com/antlr/grammars-v4/pull/4497 first before this PR. Thanks. --Ken

teverett:
@kaby76  thanks!

PR #4508, opened 2025-06-16
[Golang] fix operandName rule

GeorgeAlexandria:
Apply fix to operandName rule by go spec https://go.dev/ref/spec#OperandName

### DIFF
--- a/golang/GoParser.g4
+++ b/golang/GoParser.g4
@@ -451,6 +451,7 @@ integer
 
 operandName
     : IDENTIFIER
+    | qualifiedIdent
     ;
 
 qualifiedIdent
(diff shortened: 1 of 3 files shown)


kaby76:
The build failed for the Antlr TypeScript port because it runs in an older environment, as compared to Antlr4ng. For now, delete the string `;TypeScript` from the desc.xml:
```
$ git diff .
diff --git a/golang/desc.xml b/golang/desc.xml
index a183b954..274523ad 100644
--- a/golang/desc.xml
+++ b/golang/desc.xml
@@ -1,5 +1,5 @@
    <antlr-version>^4.13.0</antlr-version>
-   <targets>Antlr4ng;Cpp;CSharp;Dart;Go;Java;JavaScript;Python3;TypeScript</targets>
+   <targets>Antlr4ng;Cpp;CSharp;Dart;Go;Java;JavaScript;Python3</targets>
 </desc>
06/19-06:12:38 ~/issues/g4-4508/golang
$
```
This will remove the TypeScript target from testing until I can figure out the best way forward. We're using `ES6` for TypeScript, `ESNext` for Antlr4ng. I can change it for TypeScript, but I will need to do a complete regression test of all of grammars-v4.

Or, we may be able to just change `at()` to `charAt()`.

GeorgeAlexandria:
@kaby76, the TypeScript target was removed.

teverett:
@GeorgeAlexandria  thanks!

PR #4509, opened 2025-06-19
[Cobol85] fix preprocessor compilation errors

hanohrs:
- Fixed compile errors of the Cobol85 preprocessor
- Added Cobol85 as a module
- Configured the test suite to run on the Cobol85 example1.txt


kaby76:
Testing is broken. We get `System.Exception: Can't figure out the grammar name`, which means that we have multiple top-level .g4's. The desc.xml needs to be fixed. trgen is used to generate a standardized application across all targets and OSes.

Basically, trgen looks at the grammars-v4/cobol85 directory, and sees that there are two .g4's. trgen reads every .g4 and constructs a graph describing grammar usage: what .g4 `import`'s what .g4; what .g4 has a `options { tokenVocab = ...; }` usage. trgen then performs a topological sort of the graph to determine which .g4 is the "top-level" grammar for the application.

trgen does not have a notion of changed grammars from a preprocessor into some other grammar down a pipeline. Haven't had time to add it.

So, for now, just add back the `grammar-name` element in desc.xml. That should test the grammar minimally. It won't test the preprocessor, but we don't have tests for that anyways.

hanohrs:
I'm sorry for breaking the tests. I had run mvn test, but not trgen.

I've reverted the line `<grammar-name>Cobol85</grammar-name>`. I had removed it earlier because desc.xml is inconsistent with _scripts/desc.xsd. It seems that desc.xsd should include the line `<xs:element name="grammar-name" type="xs:string"/>`.

kaby76:
@hanohrs Thank you for bringing up this important issue regarding the desc.xsd. The file is has not been updated with more recent changes to trgen. I will update the file. https://github.com/antlr/grammars-v4/issues/4516

teverett:
@kaby76 good to merge this?

kaby76:
@teverett Sure!

teverett:
@hanohrs  thanks!

PR #4512, opened 2025-06-19
[build] Fixes for 4510 and 4511.

kaby76:
Changed test script call to trgen, and test proper exit code from it. Migrate to ESNext for TypeScript to fix build issue https://github.com/antlr/grammars-v4/pull/4508#issuecomment-2987546260.


teverett:
@kaby76  thanks!

Issue #4428, opened 2025-02-27
[sql/hive/v3] ResourcePlanParser.g4 does not define "identifier".


kaby76:
For the latest version of the Trash Toolkit, I modified the code to do a better, bug-free analysis of the grammars. The top-level grammar ResourcePlanParser.g4 is not included by any other grammar (so it is top-level), and it does not import other .g4's. The grammar is maldefined.

PR #4513, opened 2025-06-19
[hive/v3] Fix for #4428: rename ResourcePlanParser.g4.

kaby76:
This is a fix for #4428.

The purpose of the grammar ResourcePlanParser.g4 is unknown. It requires parser rule `identifier` to be defined, but that's defined in IdentifierParser.g4. This grammar *does not* include IdentifiersParser.g4, or vice versa, and no other .g4 includes ResourcePlanParser.g4.

I'm renaming this to have it not interfere with trgen. Otherwise, it's a "top-level" grammar, and all top-level grammars are processed by the Antlr4 tool, which would flag correctly as an illegal grammar. Perhaps an older version of the Antlr4 tool allowed such a thing??


teverett:
@kaby76  thanks

Issue #4514, opened 2025-06-20
Add in Rust target.


kaby76:
See https://github.com/antlr/antlr4/issues/1839#issuecomment-2992098905

PR #4517, opened 2025-06-21
[build] Fix for #4514: Add Rust target.

kaby76:
* Check to see what if Rust is installed in default VM.


kaby76:
Closing. The Antlr4 Rust target is way too unstable. There is not one example where the tool generates compilable code.

PR #4519, opened 2025-06-22
[build] Fix "No test cases provided" error.

kaby76:
Some of the latest changes I made to the build scripts broke parser testing. You can tell if it's broken by looking at the GitHub Actions output `Test` step and seeing whether there is a "No test cases provided" error, meaning that nothing was parsed. That in itself is not an error because some grammars actually aren't tested beyond running the grammar through the Antlr4 tool. However, most grammars have parsing tests. So, a few of the recent PRs haven't tested properly. 

While investigating the problem, I noticed that there may be a bug with the [`System.IO.StreamWriter()` call in the code I added to write out file names to test](https://github.com/antlr/grammars-v4/blob/891f7290ec7ab436a5479bff88be04b0cd0d4a04/_scripts/templates/CSharp/st.test.ps1#L30-L35). `SystemIO.StreamWriter()` doesn't seem to handle the relative path file name "tests.txt". I tried replacing it with "./tests.txt", or ".\\tests.txt", neither worked. The solution was to change the relative file path name "tests.txt" to a full path file name. This PR fixes the problem by using the full path name to "tests.txt".

### DIFF
--- a/_scripts/templates/Antlr4cs/st.test.ps1
+++ b/_scripts/templates/Antlr4cs/st.test.ps1
@@ -1,7 +1,7 @@
 # Generated from trgen <version>
 
 $workingDirectory = Get-Location
-$filePath = "$workingDirectory\tests.txt"
+$filePath = "$workingDirectory/tests.txt"
 
 $Tests = "<if(os_win)>../<example_files_win><else>../<example_files_unix><endif>"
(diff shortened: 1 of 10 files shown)


kaby76:
@teverett This PR is ready. It fixes a testing regression.

teverett:
@kaby76  thanks!

PR #4523, opened 2025-06-27
[build; lua, tcpheader, trapc, plsql] Fixes for #4521

### LINKED ISSUE #4521: [build] tcpheader and lua broken.
Looks like the CSharp targets for lua and tcpheader are broken.

For the lua grammar, the .trq file needs to be updated with the correct number of `stat` nodes after the recent fixes. That's an easy fix.

For tcpheader, prior to recent changes, binary files were not parsed, at least by the trgen parse tester. Now,`tcpheader\examples\example1.bin` fails to parse; we don't filter out non-Unicode input files because this took too long for large test suites. The Maven tester works fine because it uses file encoding "ISO_8859_1". The trgen tester does not implement encodings in a desc.xml. For tcpheader, we'll need a new version of trgen/desc.xml that has the file encoding and binary file string output for tree display.

kaby76:
This PR fixes #4521.

### Fixed grammars
* [lua](https://github.com/antlr/grammars-v4/tree/e1491245d045474fa5393fb7be03b3b8ed22080f/lua): [stat.lua.trq](https://github.com/antlr/grammars-v4/blob/e1491245d045474fa5393fb7be03b3b8ed22080f/lua/examples/stats.lua.trq) in the `examples/` directory needed to be updated with the expected count of `stat` nodes in the parse tree of [stats.lua](https://github.com/antlr/grammars-v4/blob/e1491245d045474fa5393fb7be03b3b8ed22080f/lua/examples/stats.lua).
* [tcpheader](https://github.com/antlr/grammars-v4/tree/e1491245d045474fa5393fb7be03b3b8ed22080f/tcpheader): This grammar requires the "latin1" file encoding. To facilitate file encodings, trgen was updated to extract file encoding from the desc.xml file. I will need to update the desc.xsd file with this information, but will do so in another PR. 
* [plsql](https://github.com/antlr/grammars-v4/tree/e1491245d045474fa5393fb7be03b3b8ed22080f/sql/plsql): This grammar contains a custom [CMakeLists.txt](https://github.com/antlr/grammars-v4/blob/e1491245d045474fa5393fb7be03b3b8ed22080f/sql/plsql/Cpp/st.CMakeLists.txt) in template form for the Cpp target. It needed to be updated to compile and link with the new file stream encoding API.
* [trapc](https://github.com/antlr/grammars-v4/tree/e1491245d045474fa5393fb7be03b3b8ed22080f/trapc): I touched one of the files to force a test of this grammar. trgen wasn't computing the correct file copy "to" locations, which has been corrected with version 0.23.23.

### Changed targets
* Antlr4ng, CSharp, Dart, Go, Java, JavaScript, Python3, TypeScript: File encoding for parsing, and the class `BinaryCharStream` for debugging parse trees, were added.
* Cpp: `ANTLRFileStream` was replaced with `EncodingFileStream`, a rewrite of the ANTLRFileStream class for file encodings. Currently, only "latin1" and "utf8" are recognized.
* Various: Templates were reformatted.

### Updated trgen
* This PR requires a fix to trgen for file encoding. This PR requires version 0.23.23.


kaby76:
@teverett All set with this PR.

teverett:
@kaby76  thanks

PR #4525, opened 2025-07-01
update corundum grammar

Puzakov1:
add array_definition to initial_array_assignment, as well as a corresponding example to test.rb


teverett:
@Puzakov1  thanks!

PR #4536, opened 2025-07-09
Add missing REQUIRED keyword to Protobuf3.g4

dsyer:
Fields in proto3 can be "optional" or "required" or "repeated". One of those appears to be missing in the sample.

### DIFF
--- a/protobuf3/Protobuf3.g4
+++ b/protobuf3/Protobuf3.g4
@@ -53,6 +53,7 @@ optionName
 // Normal Field
 fieldLabel
     : OPTIONAL
+    | REQUIRED
     | REPEATED
     ;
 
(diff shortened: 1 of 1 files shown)


kaby76:
I was going through some examples and `required` is not a valid field label in version 3. I took this example [5.proto](https://github.com/kaby76/grammars-v4/blob/8455170b55fe1578949e570f8561e39ff39fd83f/protobuf/protobuf3/examples/x/5.proto.v2), added `syntax="proto3";` at the top, and ran it through the most recent compiler. Without the `syntax` it compiles fine. But with `syntax`, it outputs an error. I'm not sure what is the proper thing to do here.

dsyer:
My bad. It's not allowed in proto3.

PR #4538, opened 2025-07-11
[protobuf2, 3] Fix for #4535: disambiguate grammar via symbol table

### LINKED ISSUE #4535: Protobuf3.g4 cannot distinguish between a field of type message and enum
Example:

```proto
syntax = "proto3";
enum TestEnum {
        UNKNOWN = 0;
        FOO = 1;
        BAR = 2;
}
message TestMessage {
        TestEnum value = 1;
}
```

`TestMessage.value` is of type `TestEnum` (which is an enum by construction). The grammar can't tell the difference between a message type field and an enum type, so it matches as a message type (I assume because it is listed first in the grammar). The grammar looks like this:

```g4
...
messageType
    : (DOT)? (ident DOT)* messageName
    ;

enumType
    : (DOT)? (ident DOT)* enumName
    ;
...
messageName
    : ident
    ;

enumName
    : ident
    ;
...
```

Maybe there is no way for Antlr to express this difference? it's definitely a problem for anyone who wanted to use the generated parser.

Example command line to test:

```bash
$ grun Protobuf3 proto -tree
syntax = "proto3";
enum TestEnum {
        UNKNOWN = 0;
        FOO = 1;
        BAR = 2;
}
message TestMessage {
        TestEnum value = 1;
}
(proto (syntax syntax = "proto3" ;) (topLevelDef (enumDef enum (enumName (ident TestEnum)) (enumBody { (enumElement (enumField (ident UNKNOWN) = (intLit 0) ;)) (enumElement (enumField (ident FOO) = (intLit 1) ;)) (enumElement (enumField (ident BAR) = (intLit 2) ;)) }))) (topLevelDef (messageDef message (messageName (ident TestMessage)) (messageBody { (messageElement (field (type (messageType (messageName (ident TestEnum)))) (fieldName (ident value)) = (fieldNumber (intLit 1)) ;)) }))) <EOF>)
```

kaby76:
This PR fixes #4535.

* I reorganized the protobuf grammars. `protobuf2` and `protobuf3` are now nested under a directory called `protobuf`. There is a newer version of the language, [Protocol Buffers 2023](https://protobuf.dev/reference/protobuf/edition-2023-spec/), but I didn't add that.
* `protobuf2` is unchanged. So, `protobuf2` still has ambiguity in the grammar.
* The ambiguities in `protobuf3` has been removed. You can see where the ambiguity is by looking for the semantic predicates added in the grammar.
* Much of the ambiguity was caused in discerning enum and message types. I added code for a symbol table of type declarations.
* The parser performs two passes for input.
   * In the first pass, actions in the grammar enter type declarations into the symbol table. The first pass is performed in `DoRewind()`.
   * The second pass happens with a second call to `proto()`. In the second pass, predicates are evaluated to turn on or off rules that should apply for the symbol with the given type.
* I added the example from #4535 into the tests directory along with some others. These added examples compile fine with the protobuf compiler.
* Added `package` symbol table implementation. 
* Added `import` parsing. Checks are performed to ensure not files are imported multiple times, or the parse gets stuck in a cycle of `import` references.


teverett:
@kaby76  thanks

Issue #4539, opened 2025-07-11
[_scripts/mkindex.py] Script generates the wrong entry for BibTeX.


kaby76:
https://github.com/antlr/grammars-v4/pull/4537

kaby76:
The pom.xml is wrong (mkindex.py uses it). *At some point, mkindex.py should not be using the pom.xml. trgen outputs a graph on the dependency analysis.*

PR #4540, opened 2025-07-11
[bibtex] Fix for #4539: pom.xml wrong for bibtex.

kaby76:
The pom.xml was wrong. The parser grammar file name was fixed.


teverett:
@kaby76  thanks

PR #4546, opened 2025-07-16
Fix for #4545

### LINKED ISSUE #4545: python3_12 module reference should be python3_13 in python/pom.xml
The [python/pom.xml](https://github.com/antlr/grammars-v4/blob/master/python/pom.xml) file currently references a module named `python3_12`, which does not exist in the repository. The correct module name is `python3_13`.

**Current (incorrect):**

```xml
<module>python3_12</module>
```

**Expected (correct):**

```xml
<module>python3_13</module>
```

This causes a build error:

```
Module 'python3_12' not found
```

hanohrs:
This PR fixes #4545

### DIFF
--- a/python/pom.xml
+++ b/python/pom.xml
@@ -16,6 +16,6 @@
         <module>python2</module>
         <module>python3</module>
         <module>python2_7_18</module>
-        <module>python3_12</module>
+        <module>python3_13</module>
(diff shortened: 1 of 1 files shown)


teverett:
@hanohrs  thanks

PR #4548, opened 2025-07-16
Fix for #4547

### LINKED ISSUE #4547: ruby module reference should be corundum in pom.xml
The [pom.xml](https://github.com/antlr/grammars-v4/blob/master/pom.xml) file currently references a module named `ruby`, which does not exist in the repository. The correct module name is `corundum` due to the change in #4533.

hanohrs:
This PR fixes #4547.


teverett:
@hanohrs  thanks

Issue #4552, opened 2025-07-19
[build] mvn clean broken.


kaby76:
Looks like there are a couple of some problems that didn't get cleaned up recently in the "mvn clean test" build. The javacc/pom.xml uses the wrong `artifactId`.

hanohrs:
It looks like `mvn` runs without issues, but the following `artifactId`s are also duplicated:

* `abb`
* `antlr4`
* `bnf`
* `java`
* `mysql`

In addition, `python2` and `python3` differ only in the case of their initial letters. I’ve created a branch that fixes these duplicates (including the case-only ones), for your reference:
[https://github.com/hanohrs/grammars-v4/tree/g4-4552](https://github.com/hanohrs/grammars-v4/tree/g4-4552)

There’s still room for further cleanup. According to [this page](https://maven.apache.org/guides/mini/guide-naming-conventions.html#Artifact_identifier) `artifactId`s should only consist of lowercase letters, digits, and hyphens.

All current `artifactId`s that don’t follow that convention are listed below:

* `ASL`
* `C`
* `CPP14`
* `CSS3`
* `EVMB`
* `IDL`
* `LPC`
* `PMMN`

kaby76:
@hanohrs, please submit a PR for your changes. In the meantime, I'll update the build to add these checks. I'll add some Maven testing back into the build. I'll also add integrity checking for the problems you mention.

hanohrs:
@kaby76
Thank you for your response. I've submitted a PR for the changes: #4558.

PR #4553, opened 2025-07-19
[build] Fixes for #4552: mvn clean test broken.

kaby76:
Fixed regression in build using Maven tester. `javacc/pom.xml` had a typo. `protobuf/protobuf3` has proto2 code, but can't be parsed using the protobuf3 grammar. Apparently, the protobuf compiler can dynamically switch to older syntaxes, but that wasn't the problem because the trgen tester can tell to test files in `examples/*.proto`, not recursively, so the proto2 code is not parsed. The Maven tester cannot be directed to test only `examples/*.proto`, i.e., only the examples in the directory, not sub-directories. So, parse testing with Maven was removed.

__Output from `mvn clean test`. Note, runtime is not realistic because I put my machine in sleep mode during the test. But, it does illustrate that the build works.__ 
[out.txt](https://github.com/user-attachments/files/21330765/out.txt)

### DIFF
--- a/javacc/pom.xml
+++ b/javacc/pom.xml
@@ -1,6 +1,6 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 	<modelVersion>4.0.0</modelVersion>
-	<artifactId>antlr4</artifactId>
+	<artifactId>javacc</artifactId>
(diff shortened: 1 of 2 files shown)


teverett:
@kaby76  thanks!

kaby76:
Ugh. The PR didn't quite fix things because I didn't resolve the "predefined.tokens" issues with the antlr/antlr4/ grammar. When Mike changed this grammar, he introduced a file called "predefined.tokens" to load in predefined token type values. Apparently, he makes use of these tokens and their order. The Maven tester works find in antlr/antlr4, but doesn't in the root directory for the repo because it cannot find "predefined.tokens".

The easy hack is to add a symbolic link to the file in the root directory, but it isn't nice.

PR #4556, opened 2025-07-20
[java] Fix for #4530: update recordComponent.

kaby76:
This PR updates the [grammar-v4/java/java/ grammar](https://github.com/antlr/grammars-v4/tree/b590556410b619930a1a3d4574a43b19e673e4c9/java/java) to be more in line with the [JLS 24](https://docs.oracle.com/javase/specs/jls/se24/html/jls-19.html) with respect to `RecordComponent`:

    RecordComponent:
       { RecordComponentModifier } UnannType Identifier
       VariableArityRecordComponent

    VariableArityRecordComponent:
        { RecordComponentModifier} UnannType { Annotation } ... Identifier

    RecordComponentModifier:
        Annotation

For the final form in the grammar,  I refactored the EBNF to result in a slightly more efficient parse. But, it is not required for Antlr4: the rules are not refactored in the java20 grammar.  https://github.com/antlr/grammars-v4/blob/b590556410b619930a1a3d4574a43b19e673e4c9/java/java20/Java20Parser.g4#L670-L681

Changes:
1)  Updated rules according to Spec: `recordComponent:
 recordComponentModifier* typeType identifier |
variableArityRecordComponent; variableArityRecordComponent: recordComponentModifier* typeType annotation* '...' identifier; recordComponentModifier: annotation;`.
1) Unfolded `recordComponentModifier` and deleted (two applied occurrences). `recordComponent:
 annotation* typeType identifier |
variableArityRecordComponent; variableArityRecordComponent: annotation* typeType annotation* '...' identifier;`
1) Unfolded `variableArityRecordComponent` and deleted (one applied occurrence). `recordComponent:
 annotation* typeType identifier | annotation* typeType annotation* '...' identifier;`.
1) Grouped common terms.  `recordComponent: annotation* typeType (annotation* '...')? identifier;`

I added a semantic predicate to implement the static semantics constraint that the ellipsis be mentioned only in the last record component of a list. However, this might be better implemented in a tree walker after parsing. Predicates that disambiguate the grammar need to be performed during the parse if the parser needs to make a choice in the parse; this predicate is not one of those.

I am adding to the PR scripts and test suites for performance analysis. It is a challenge to repeat tests without knowing the exact test suite and script.

_NB: This PR only implements the CSharp and Java targets. Once I have the grammar updated to parse the JDK version 24, I'll add the other ports._

### DIFF
--- a/java/java/CSharp/JavaParserBase.cs
+++ b/java/java/CSharp/JavaParserBase.cs
@@ -1,0 +1,28 @@
+using Antlr4.Runtime;
+using System.IO;
+using System.Linq;
+
+public abstract class JavaParserBase : Parser {
+	private readonly ITokenStream _input;
+
+	protected JavaParserBase(ITokenStream input, TextWriter output, TextWriter errorOutput)
+			: base(input, output, errorOutput) {
+		_input = input;
+	}
+
+	public bool DoLastRecordComponent()
+	{
+		var ctx = this.Context;
+		var tctx = ctx as JavaParser.RecordComponentListContext;
+		var rcs = tctx.recordComponent();
+		if (rcs.Any()) return true;
+		var count = rcs.Count();
+		for (int c = 0; c < count; ++c)
+		{
+			var rc = rcs[c];
+			if (rc.ELLIPSIS() != null && c+1 < count)
+				return false;
+		}
+		return true;
+	}
+}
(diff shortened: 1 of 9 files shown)


teverett:
@kaby76  thanks!

PR #4558, opened 2025-07-21
Fix duplicate artifactId (case-insensitive) (#4552)

hanohrs:
This PR fixes issues with duplicate `artifactId`s, including case-insensitive duplicates (e.g., `python3` vs `Python3`).
It addresses and includes the fix for #4552.

Note: According to [this page](https://maven.apache.org/guides/mini/guide-naming-conventions.html#Artifact_identifier), `artifactId`s should only consist of lowercase letters, digits, and hyphens. However, this PR does not modify `artifactId`s that contain uppercase letters and violate this convention.


teverett:
@hanohrs  thanks!

Issue #4560, opened 2025-07-22
[java] grammar doesn't parse openjdk version 9 and newer.


kaby76:
PR https://github.com/antlr/grammars-v4/pull/4556 contains a script for validation and performance testing of the java grammar. It can download a number of versions of the JDK but only one is programmed because the parser does not work on version 9 or newer.

The grammar needs to be updated so it can parse all available versions, which is currently version 24.

PR #4561, opened 2025-07-23
[java] Fix for #4560; Update grammar to parse JDK 21.

kaby76:
This is a fix for #4560. I updated the grammar enough in order to parse the OpenJDK to version 21. Additional work will need to be done to parse version 22 and newer for the OpenJDK (expressions).
* The supported targets are restricted to just Java and CSharp for now. When the grammar is extended to the current version of the JLS, the targets can be fixed.
* I've added a script to test the parser grammar against the JDK, version 21.
* I've included a test for #4560 the `...`-operator in records.


teverett:
@kaby76  thanks

Issue #4562, opened 2025-07-23
[clickhouse] Add other targets to grammar--especially CSharp.


kaby76:
The [Clickhouse grammar](https://github.com/antlr/grammars-v4/tree/master/sql/clickhouse) only has the [Java port](https://github.com/antlr/grammars-v4/blob/b590556410b619930a1a3d4574a43b19e673e4c9/sql/clickhouse/desc.xml#L7-L8) listed as viable for this grammar. This is a drawback because the C # port, which works, is required to obtain important information about the grammar, such as ambiguity.

We need to add as many of the other targets as possible for the grammar.

PR #4563, opened 2025-07-23
[clickhouse] Fix for #4562: Add CSharp, and other. targets.

kaby76:
The CSharp target is a very important target to have. It is used to obtain critical information on the grammar that people use to review changes to the grammar.


teverett:
@kaby76  thanks!

PR #4564, opened 2025-07-23
Adding gherkin parser and examples

incident-recipient:
Providing a parser, examples and documentation for a gherkin parser originally created in a separate project, but suitable for general use.


kaby76:
#### 1. Please add a desc.xml and a pom.xml, and update root pom.xml.
These two files describe how to test the grammar. Here's what I suggest:

For desc.xml
```
   <targets>Antlr4ng;CSharp;Cpp;Dart;Go;Java;Python3</targets>
   <!-- PHP, JavaScript, TypeScript don't work, likely binary input.
   -->
</desc>
```
For pom.xml
```
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
	<modelVersion>4.0.0</modelVersion>
	<artifactId>gherkin</artifactId>
	<packaging>jar</packaging>
	<name>Gherkin grammar</name>
	<parent>
		<groupId>org.antlr.grammars</groupId>
		<artifactId>grammarsv4</artifactId>
		<version>1.0-SNAPSHOT</version>
	</parent>
        <build>
		<plugins>
			<plugin>
				<groupId>org.antlr</groupId>
				<artifactId>antlr4-maven-plugin</artifactId>
				<version>${antlr.version}</version>
				<configuration>
					<sourceDirectory>${basedir}</sourceDirectory>
					<includes>
					   <include>GherkinLexer.g4</include>
					   <include>GherkinParser.g4</include>
					</includes>
					<visitor>true</visitor>
					<listener>true</listener>
				</configuration>
				<executions>
					<execution>
						<goals>
							<goal>antlr4</goal>
						</goals>
					</execution>
				</executions>
			</plugin>
			<plugin>
				<groupId>com.khubla.antlr</groupId>
				<artifactId>antlr4test-maven-plugin</artifactId>
				<version>${antlr4test-maven-plugin.version}</version>
				<configuration>
					<verbose>false</verbose>
					<showTree>false</showTree>
					<entryPoint>gherkinDocument</entryPoint>
					<grammarName>Gherkin</grammarName>
					<packageName></packageName>
					<exampleFiles>examples/</exampleFiles>
					<testFileExtension>.feature</testFileExtension>
				</configuration>
				<executions>
					<execution>
						<goals>
							<goal>test</goal>
						</goals>
					</execution>
				</executions>
			</plugin>
		</plugins>
	</build>
</project>
```
You will need to change the root pom.xml: You will need to add an entry `<module>cucumber-gherkin</module>` to the pom.xml at the top-level directory for the repo, before this line: https://github.com/antlr/grammars-v4/blob/cd8affd1e37b2a8c43dbf4f2d4da36db251c594d/pom.xml#L182

#### 2. Please rename `README` to `README.md`.
No system understands that this is a Markdown file unless you add the `.md` extension.

#### 3. Rename directory cucumber-gherkin to gherkin?
I'm not sure why you gave the directory containing your `Gherkin*.g4` grammar as `cucumber-gherkin`. Suggest you just rename it `gherkin`.

incident-recipient:
I added the pom.xml and desc.xml file. I also renamed the directory to "gherkin" and added the `.md` extension to the README

I was able to run `mvn compile` on the gherkin project, but got an error when I tried to do it on the root directory. It isn't clear to me if this is a problem or not. If so, the stack track is this:

```
Scanning for projects...
[ERROR] [ERROR] Project 'org.antlr.grammars:antlr4:1.0-SNAPSHOT' is duplicated in the reactor @ 
```

kaby76:
The "duplicated" error is fixed in https://github.com/antlr/grammars-v4/pull/4558. It needs to be merged. @teverett 

Thanks for the changes.

incident-recipient:
There are test failures for the C++ target. The errors appear to be due to characters with diacritics or emojis.

I'm not sure how the tests are being run, but can I get a confirmation on whether or not the C++ tests support UTF-8? Given the tests pass in all other targets I suspect that the issue lies with the C++ tests. 

If we can't fix it from the C++ side, the "solution" would be to remove the feature files that are in non-English locales. I can do this but it wouldn't be my first choice.

kaby76:
I'm not sure why the `Cpp` target isn't working in Github Actions; the target works fine on my machine, which is Windows. You can punt on trying to solve it by just removing `Cpp;` from the desc.xml file. But if you want to solve the problem, the code that is run is in template form here: https://github.com/antlr/grammars-v4/blob/1dc19f1279cb6a74d2b2c7c87a052e731a81b37f/_scripts/templates/Cpp/st.Test.cpp

Issue #4555, opened 2025-07-20
[build] Add Maven testing, add integrity checking.


kaby76:
The Maven tester was removed years ago because it "didn't do incremental testing." Well, that was wrong because you could `cd` to the grammar that was changed and then type `mvn clean test`.

Since people still want the Maven tester, the plan is to add the tester back into the build. But, since we don't want to do a full build, we'll need to have a script to selectively test only the grammars that changed. In addition, there are a few integrity checks added to the build, mentioned in here: https://github.com/antlr/grammars-v4/issues/4552#issuecomment-3092695464.

PR #4565, opened 2025-07-23
[build] Fix for #4555: reintroduce Maven testing.

kaby76:
_Please merge https://github.com/antlr/grammars-v4/pull/4570 first._

This is a fix for #4555. With this PR, the Maven tester is once again checked. While it misses many features that the trgen tester does, the Maven tester still is a good, quick check, and people seem to prefer that.

This PR changes:
* Github workflows file [main.yml](https://github.com/antlr/grammars-v4/blob/master/.github/workflows/main.yml). The job "maven" has been added. It computes the before and after Git SHA's for a PR, then calls a new script, maven.sh.
* _scripts/maven.sh: This Bash script computes the directories changed in a PR, then processes them with the Maven tester. We don't do a full repo test.
* At the end of the grammar tests, `mvn clean` is done at the root for a quick sanity check of the root pom.xml.

Some of the grammars had problems in their pom.xml: bison, clif, ocl, pike, ruleworks, spl, teal, tinyos_nesc. These were corrected in various ways.


teverett:
@kaby76  thanks

PR #4568, opened 2025-07-29
snowflake - is distinct from expr

mlorek:
support fr `is distinct from` expr

### DIFF
--- a/sql/snowflake/SnowflakeParser.g4
+++ b/sql/snowflake/SnowflakeParser.g4
@@ -4052,8 +4052,7 @@ expr
     | trim_expression
     | function_call
     | subquery
-    | expr IS null_not_null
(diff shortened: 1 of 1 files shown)


sql/snowflake/SnowflakeParser.g4
```diff
@@ -4053,6 +4053,7 @@ expr
     | function_call
     | subquery
     | expr IS null_not_null
+    | expr IS NOT? DISTINCT FROM expr
```
KvanTTT: It probably makes sense to simplifying a little bit.

```suggestion
    | expr IS (null_not_null | NOT? DISTINCT FROM expr)
```

teverett:
@mlorek thanks

Issue #4569, opened 2025-07-30
[antlr4] predefined.tokens interferes with "mvn clean test" from root.


kaby76:
The Maven tester is broken when run in the root directory for the repo. This is because the [antlr4](https://github.com/antlr/grammars-v4/tree/master/antlr/antlr4) grammar references the file [predefined.tokens](https://github.com/antlr/grammars-v4/blob/1dc19f1279cb6a74d2b2c7c87a052e731a81b37f/antlr/antlr4/predefined.tokens). If the Maven tester is run in the directory `grammars-v4/antlr/antlr4`, it finds the file with no problem. However, when run from any other directory, the tester cannot find the file.

`predefined.tokens` was inttroduced by Mike with PR https://github.com/antlr/grammars-v4/pull/4424 for Antlr-ng. It forces a specific, total order of the token type values for the lexer grammar vocabulary. This ordering is assumed in the Antlr tool code. For the Antlr4 Tool, the order was fixed because the Antlr4 Tool depended on Antlr3, which supported a way to predeclare the token types and values. However, Antlr-ng removed the Antlr3 dependency, and needed another way to declare the total ordering, which is why he added the [tokenVocab option to the lexer grammar](https://github.com/antlr/grammars-v4/blob/1dc19f1279cb6a74d2b2c7c87a052e731a81b37f/antlr/antlr4/ANTLRv4Lexer.g4#L54).

While the total ordering is a requirement for Antlr-ng, it is not for any other tool that uses this grammar. The `tokenVocab` option should be removed or another way to support the total ordering found.

PR #4570, opened 2025-07-30
[antlr4, maven testing] Fix for #4569: predefined.tokens interfering with testing.

kaby76:
This is a fix for #4569.

The file predefined.tokens, introduced in https://github.com/antlr/grammars-v4/pull/4424, breaks Maven testing. People still use the tester, so we are trying to reintroduce it but in a way that is more "incremental".

The predefined.tokens file is replaced with an ordered list of tokens in the grammar. I would recommend the file be removed, but kept it for reference. In addition, I noticed an error in the file: `TOKEN_REF` is listed twice. https://github.com/antlr/grammars-v4/blob/1dc19f1279cb6a74d2b2c7c87a052e731a81b37f/antlr/antlr4/predefined.tokens#L9-L10


teverett:
@kaby76  thanks

Issue #4574, opened 2025-08-06
[plantuml] Grammar not integrated into build.


kaby76:
plantuml was recently added. I noted in my review that it should be integrated into the build, and that the grammar was ambiguous. Neither of these problems was fixed.

At a minimum, the grammar needs to be added to the build.

PR #4575, opened 2025-08-06
[plantuml] Fixes for #4574: add to testing, fix incorrect check-in line endings for PlantUML.g4.

kaby76:
This PR fixes testing of the plantUML grammar.

In addition, the initial check-in of the PlantUML.g4 file was incorrect: apparently, the line endings were wrong for the host environment for the check-in. The file should be considered a text file, which means that you need to have the correct line endings for the initial `git commit`. This is why `git clone` of the repo under the MSYS2 and Linux environments both produce the wrong line endings (with `\r\n` instead of `\n`). According to .gitattributes, [all .g4's are text](https://github.com/antlr/grammars-v4/blob/260cc3a272bb633bc69833ab527f4b9d34a3502c/.gitattributes#L36), which is correct. This is why the file is "changed" with this PR.

I changed the readme.md file to follow the standardized format that we use for this repo. I added comment and links to the readme so we can find the "spec" and "doc" for this language without searching.

* There are still:
    * unused symbols (INTEGRAL, SIGMA)
    * ambiguities (classElement, internalAttribute)

cc: @kevinlano


teverett:
@kaby76  thanks

PR #4577, opened 2025-08-07
[antlr/antlr4] Fix for #4576 (duplicate TOKEN_REF).

### LINKED ISSUE #4576: `ANTLRv4Lexer.g4` contains a duplicate token...
... following [this recent commit](https://github.com/antlr/grammars-v4/commit/98465e177d4ff882092bd530e4a3fddbf2c601c5)

https://github.com/antlr/grammars-v4/blob/260cc3a272bb633bc69833ab527f4b9d34a3502c/antlr/antlr4/ANTLRv4Lexer.g4#L75

I'm guessing this is not intentional.

kaby76:
This PR fixes #4576. There is a duplicate for `TOKEN_REF` in the Antlr4 lexer grammar. It should not be there. As noted in the comments in the grammar (see `options { tokenVocab ... }`), the order of the tokens is important, but the values are not. So, the gap in token values is not necessary.

### DIFF
--- a/antlr/antlr4/ANTLRv4Lexer.g4
+++ b/antlr/antlr4/ANTLRv4Lexer.g4
@@ -72,7 +72,6 @@ tokens {
     SEMPRED,
     STRING_LITERAL,
     TOKEN_REF,
-    TOKEN_REF,
     UNICODE_ESC,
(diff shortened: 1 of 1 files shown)


teverett:
@kaby76  thanks

Issue #4549, opened 2025-07-16
OSSRH Sunset: Update pom.xml


hanohrs:
[As of June 30, 2025, the OSSRH service has officially reached end-of-life and has been shut down](https://central.sonatype.org/pages/ossrh-eol/). The current [pom.xml](https://github.com/antlr/grammars-v4/blob/master/pom.xml) still references OSSRH endpoints such as:

```xml
<distributionManagement>
    <snapshotRepository>
        <id>ossrh</id>
        <url>https://oss.sonatype.org/content/repositories/snapshots</url>
    </snapshotRepository>
    <repository>
        <id>ossrh</id>
        <url>https://oss.sonatype.org/service/local/staging/deploy/maven2/</url>
    </repository>
</distributionManagement>
```

How about removing the related configuration, since as far as I know this repository is not actually distributed?

Thanks!

PR #4584, opened 2025-08-12
Fix for #4549: Remove references to ossrh.

kaby76:


teverett:
@kaby76  should that be changed to a more modern repo url?

PR #4586, opened 2025-08-13
[java20 and java] Fix for #4391

### LINKED ISSUE #4391: Java 20 grammar bug(?) in localVariableDeclaration
In [Java20Parser.g4](https://github.com/antlr/grammars-v4/blob/master/java/java20/Java20Parser.g4)'s localVariableDeclaration rule (line 887), variableDeclaratorList is optional. The rule is based on Paragraph 14.4 of [the Java 20 language specification](https://docs.oracle.com/javase/specs/jls/se20/jls20.pdf), which does not specify variableDeclaratorList as optional.
Due to this,
`class Foo { void bar () { boolean; } }`
successfully parses, although it violates the specification as a variableDeclaratorList is expected after the typeIdentifier (boolean).

kaby76:
This PR fixes #4391, a problem specifically in the Java 20 grammar. Note, I checked the java/java grammar and it works correctly, reporting the parse error.

I corrected Java 20 grammar and added a test for the problem. I copied the test to the [java/java grammar](https://github.com/antlr/grammars-v4/tree/master/java/java) as well.

### DIFF
--- a/java/java/examples/Foo4391.java
+++ b/java/java/examples/Foo4391.java
@@ -1,0 +1,1 @@
+class Foo4391 { void bar () { boolean; } }
--- a/java/java/examples/Foo4391.java.errors
+++ b/java/java/examples/Foo4391.java.errors
@@ -1,0 +1,1 @@
+line 1:37 no viable alternative at input 'boolean;'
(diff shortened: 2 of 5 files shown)


teverett:
@kaby76  thanks

PR #4588, opened 2025-08-14
[java] Fix for #4583: update readme's.

### LINKED ISSUE #4583: [java] The readme needs to be updated.
There are numerous errors in the readme file for the [java/java/ grammar](https://github.com/antlr/grammars-v4/tree/d0f72ce9467cc2b78c0eb2acd95e0a86377914ac/java/java). These need to be corrected.

### "Tokens using"

In a combined Antl4 grammar, the use of a string literal in a parser rule can define a token for the string literal if there is no explicit rule in the grammar. This *can* cause problems when trying to refer to the string literal token in an Antlr visitor or listener. But, this cannot happen in a split Antlr4 grammar because the Antlr Tool outputs warnings for "implicitly" defined symbols. The java/java grammar is a split grammar.

We define a grammar as ***string-literal normalized*** if the parser rules have no string literals on the right-hand side of a rule. A parser rule such as `packageDeclaration` is ***string-literal folded*** for `PACKAGE`, but not for [SEMI/';'](https://github.com/antlr/grammars-v4/blob/d0f72ce9467cc2b78c0eb2acd95e0a86377914ac/java/java/JavaLexer.g4#L149).
```
packageDeclaration
    : annotation* PACKAGE qualifiedName ';'
    ;
```

String-literal normalization is not required for this grammar, or any other grammar in this repo: it's easy to convert the applied occurrences of string-literals using Trash. And, it doesn't make the grammar any more readable.

### "Grammar style"

This section needs to be deleted. We use an explicit, mechanically enforced, standard coding style across all grammars (https://github.com/antlr/grammars-v4/wiki#is-there-a-coding-standard-for-antlr4-grammars).

### "Benchmarks"

This section is obsolete, and it does not mention anything about ambiguity or max-k.

kaby76:
This PR fixes #4583, and other documentation issues for the Java grammars, e.g., mentioned in https://github.com/antlr/grammars-v4/discussions/4530. In addition, it updates to the latest of the Trash toolkit because the documentation for the java/ grammar now mentions trunfoldlit.


teverett:
@kaby76  thanks

PR #4600, opened 2025-08-22
Fix PRO/PROMPT conflict in PlSql grammar

thiagopenatti:
This PR fixes an issue where identifiers named PRO or PROMPT caused conflicts
with the SQL*Plus PROMPT command.

Changes included:
- Removed PROMPT_MESSAGE token from lexer to avoid hardcoding PROMPT as keyword.
- Added prompt_message parser rule to handle PRO and PROMPT correctly.
- Added a sample (plsql_prompt_conflict.sql) to reproduce the issue.

This ensures that users can declare identifiers named PRO without breaking parsing,
while still supporting SQL*Plus PROMPT command.

### DIFF
--- a/sql/plsql/PlSqlParser.g4
+++ b/sql/plsql/PlSqlParser.g4
@@ -6807,13 +6807,19 @@ sql_plus_command_no_semicolon
 
 sql_plus_command
     : EXIT
-    | PROMPT_MESSAGE
+    | prompt_message
     | SHOW (ERR | ERRORS)
     | whenever_command
     | timing_command
     | start_command
-    | set_command
-    | clear_command
+    ;
+
+prompt_message
+    : id=ID {
+          String t = $id.text.toUpperCase();
+          if (!t.equals("PRO") && !t.equals("PROMPT")) {
+          }
+    }
     ;
 
 start_command


kaby76:
There are still testing errors. There seems to be more missing updates (`Added prompt_message ***parser rule*** to handle PRO and PROMPT correctly.).

thiagopenatti:
@kaby76 @KvanTTT

Sorry, when I created the Pull Request I forgot to commit the .g4 files. I've just updated the PR with the missing changes and adjustments. Thank you!

thiagopenatti:
@kaby76 @KvanTTT
I am seeing test failures related to the way the solution was implemented. The solution was tested for Java compilation. I’m not sure if there is a better way to fix the problem.

kaby76:
> @kaby76 @KvanTTT I am seeing test failures related to the way the solution was implemented. The solution was tested for Java compilation. I’m not sure if there is a better way to fix the problem.

@thiagopenatti Sorry, I didn't really read your code thoroughly. For what purpose is this action?

https://github.com/thiagopenatti/grammars-v4/blob/0bd987a64ae61fb57e568b5fa11546e1062adcb4/sql/plsql/PlSqlParser.g4#L6818C13-L6822C20

The code does absolutely nothing of value. It's a no-op because there are no side effects, and it's not a predicate.

Unless you want to fork out a solution for 9 targets, I suggest you just remove the Java-specific action `{ String t = $id.text.toUpperCase(); if (!t.equals("PRO") && !t.equals("PROMPT")) {  }  }` completely. In addition, I don't know why you want a label for `ID` in `prompt_message : id=ID ;`. It is useless, so remove it. `prompt_message : ID ;`.

If you remove the grammar no-op action, the grammar will probably work for all 9 targets.

thiagopenatti:
@kaby76 

The proposed solution was a way I found to fix a false positive related to syntax validation in PL/SQL files (e.g., procedure), where there is a "PRO" declaration in the first column of the line, and the validation interprets it as the PROMPT command, when in fact it is just a variable name. From what I researched, the approach was to remove the PROMPT_MESSAGE token from the lexer and move a check to the parser, where it verifies whether the identifier PRO or PROMPT is coming only from an SQL*Plus script. This worked very well in Java, as shown in the example file added in the PR.

kaby76:
I deleted the no-op action. It does not work, even for Java.


> @kaby76
> 
> The proposed solution was a way I found to fix a false positive related to syntax validation in PL/SQL files (e.g., procedure), where there is a "PRO" declaration in the first column of the line, and the validation interprets it as the PROMPT command, when in fact it is just a variable name. From what I researched, the approach was to remove the PROMPT_MESSAGE token from the lexer and move a check to the parser, where it verifies whether the identifier PRO or PROMPT is coming only from an SQL*Plus script. This worked very well in Java, as shown in the example file added in the PR.

Does it work? See https://github.com/antlr/grammars-v4/actions/runs/17217170402/job/48843593968?pr=4600#step:23:545

The Java target fails to parse sql/plsql/more/comments.sql.
```
$ bash run.sh ../more/comments.sql
line 8:9 no viable alternative at input 'prompt . #'
Java 0 ../more/comments.sql fail 0.288
Total Time: 4.687
08/25-16:48:57 ~/issues/g4-4600/sql/plsql/Generated-Java-more
```

kaby76:
Looks like the pom.xml for testing hasn't been updated for a while, and contains some errors. `mvn clean test` works, but it only tests `examples/**/*.sql`, not `more/**/*.sql`. #4602 There are two directories of tests because, for some targets, the parsing is terribly slow.

thiagopenatti:
@kaby76

Indeed, the proposed fix fails with the comments.sql example. I tested it locally using only Java and with the false positive example I added. I can’t see a way to adjust it in a way that is agnostic to the targets while at the same time fixing the false positive in the example I added. Do you see a solution? Thanks.

kaby76:
> @kaby76
> 
> Indeed, the proposed fix fails with the comments.sql example. I tested it locally using only Java and with the false positive example I added. I can’t see a way to adjust it in a way that is agnostic to the targets while at the same time fixing the false positive in the example I added. Do you see a solution? Thanks.

If you fix it for Java, I can help you with a target agnostic solution.

thiagopenatti:
@kaby76
After reassessing the issue and the possible solution path, I have decided to abandon the Pull Request, since the workaround in the code is simple and easy to apply, whereas a full fix across all targets could be complex and problematic.

Issue #4604, opened 2025-08-26
[build] Getting npm 403 errors.


kaby76:
Something has changed in the last day with Github Actions. We are getting an npm install issue.

PR #4605, opened 2025-08-27
[build] npm install -g causes E403 on build.

kaby76:
* Add subtle change to force build for #4604.


kaby76:
The "npm i -g" E403 build error is not deterministic. For now, ignore.

PR #4611, opened 2025-09-02
[angelscript] Clean up grammar start rule.

kaby76:
This PR fixes the EOF-terminated start rule for angelscript. The [script rule](https://github.com/antlr/grammars-v4/blob/37146747969be81255787b80d476873ec24d2626/angelscript/angelscript.g4#L38) was used in [another rule](https://github.com/antlr/grammars-v4/blob/37146747969be81255787b80d476873ec24d2626/angelscript/angelscript.g4#L67), so it should not have been labeled as the start rule. I added `start_` as the corrected start rule.


teverett:
@kaby76  thanks!

PR #4612, opened 2025-09-02
[asn_3gpp] Clean up grammar start rule.

kaby76:
This PR fixes the EOF-terminated start rule for asn_3gpp. There was an EOF on `moduleDefinition`, which should not be. (I erroneously added it.) I changed the grammar to be "target agnostic," and added CSharp and Antlr4ng ports as well.


teverett:
@kaby76  thanks!

PR #4613, opened 2025-09-03
[hypertalk] Clean up grammar start rule.

kaby76:
This PR fixes the EOF-terminated start rules for hypertalk. I changed the desc.xml and pom.xml to test the two entry points.


teverett:
@kaby76  thanks!

PR #4622, opened 2025-09-06
PLSQL slightly more accepting for XMLATTRIBUTES and LISTAGG functions

ahjulstad:
The AS in XMLATTRIBUTES is optional. LISTAGG may take arbitraty expressions as delimiter.


teverett:
@ahjulstad  thanks

PR #4625, opened 2025-09-09
[javascript]: Allow only single variable bindings in `for in` and `for of` loops

seclerp:
Current grammar is too permissive - it allows passing multiple `var` declarations in the `for in` and `for of` loop variants:

```js
for (var a, b, c in someObject) // ...
```

```js
for (var a, b, c of someIterable) // ...
```

It is not supported by the runtime and spec in both cases:
```
Uncaught SyntaxError: Invalid left-hand side in for-of loop: Must have a single binding.
```
Moreover, such syntax doesn't make any sense.

So only single binding is meaningful and should be supported:
```js
for (var a of someIterable) // ...
```

Which is covered by this change.

Let me know if you want me to cover it with an example in `examples/` folder or something.


teverett:
@seclerp  thanks!

PR #4626, opened 2025-09-18
[golang] Fix go_parser_base table update

JSYoo5B:
When the package name aliased, table should register aliased name. 
But previous code registers aliased name on debug condition.

Other language base parser implementation registers aliased names. 
Only golang had this bug. Fix debug condition.

### DIFF
--- a/golang/desc.xml
+++ b/golang/desc.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <desc xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../_scripts/desc.xsd">
    <antlr-version>^4.13.0</antlr-version>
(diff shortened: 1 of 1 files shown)


kaby76:
@teverett Please approve a workflow for https://github.com/antlr/grammars-v4/pull/4626. The change fixes a bug.

JSYoo5B:
The CI failure seems unrelated to this PR.  
It looks like some core methods (such as `myreset()`) are missing in the JS base parser.

Since the JS base parser seems incomplete, maybe those tests should be skipped until it’s properly implemented.  
Otherwise, you could just merge this PR and keep the failing tests for now.  

Up to you, but my suggestion would be to exclude the incomplete parser from CI.

kaby76:
The problem is that the changes for the target were never completed. So, JavaScript should not be in the desc.xml file. Please remove it and the extra semicolon. In other words, the target does not work for the grammar.

JSYoo5B:
@teverett could you re-approve workflow for this PR? It changed test targets to skip failing jobs.

teverett:
@JSYoo5B  thanks!

PR #4627, opened 2025-09-24
fix: fix clickhouse

yywing:
- copy some change from  https://github.com/ClickHouse/ClickHouse/tree/master/utils/antlr
- add `ColumnExprMapAccess`


KvanTTT (commented):
Is it possible to rewrite keywords using [caseInsensitive](https://github.com/antlr/antlr4/blob/master/doc/options.md#caseinsensitive) option?

yywing:
> Is it possible to rewrite keywords using [caseInsensitive](https://github.com/antlr/antlr4/blob/master/doc/options.md#caseinsensitive) option?

I think this file should align as closely as possible with the official documentation. So we can diff and update.

teverett:
@yywing  thanks

PR #4633, opened 2025-09-26
[protobuf] Add reserved keyword to enumElement

dsyer:
I found examples of this in both proto2 and proto3 files at https://github.com/protocolbuffers/protobuf


kaby76:
Please add a GitHub Issue to describe what the problem is. Give a concrete example and/or show in the grammar the fault.

That said, I do see that the grammars don't agree with either version 2 (https://protobuf.dev/reference/protobuf/proto2-spec/#enum_definition) or version 3 (https://protobuf.dev/reference/protobuf/proto3-spec/#enum_definition) of the spec.

https://github.com/antlr/grammars-v4/blob/6b517735620223475eefaa85c92f8d6bce15f360/protobuf/protobuf2/Protobuf2.g4#L175-L179

https://github.com/antlr/grammars-v4/blob/6b517735620223475eefaa85c92f8d6bce15f360/protobuf/protobuf3/Protobuf3.g4#L175-L179

Also, all your PRs have grammar changes without a single example added. We must have examples to compare before and after changes to the grammars.

dsyer:
I don't understand how to set up the proto3 tests. For proto2 I find I can just add a file to the examples directory. If you want proto3 samples you might have to explain, or add some explicit configuration for the Maven build so I can understand.

kaby76:
I think all you have to do is put the corresponding .proto files in `protobuf2/examples/` and `protobuf3/examples/`. If no `<inputs>` element is specified in the desc.xml, as with protobuf2 testing, the default is to just look for all files recursively in `protobuf2/examples/`. For protobuf3, the desc.xml has an "inputs" element that looks just for the "top-level" .proto files in `prtobuf3/examples/` (i.e., it's not recursive because I didn't want to address the cross parsing issue yet of proto3 importing proto2 files.

The Maven tester will generally do the same thing with its specification, the pom.xml. For protobuf2, it's set to `<exampleFiles>examples/</exampleFiles>`, which is identical to what the new trgen tester does. For protobuf3, the problem is that `<exampleFiles>examples/</exampleFiles>` would start parsing every file in the directory recursively. I had to punt on this when setting up testing for protobuf3 because of this issue, so made it explicitly "don't test the parser".

For both grammars, the EBNF is ambiguous. To fix the problem, as a first attempt, I added a symbol table to protobuf3 and made it into a two-pass parser. This fixes gathering types from `import`. If the import file is not found, the parse may end up with the wrong type.

Sorry, testing is not documented at all. There's a Gihub Issue to track fixing this problem, but I haven't gotten around to fixing that either.

dsyer:
That last change added samples, and enabled the Maven build for all files ending in .proto in proto3 (ignoring errors, which seems to be the only way to filter out the proto2 code using that Maven plugin).

kaby76 (approved):

teverett:
@dsyer  thanks

Issue #4635, opened 2025-09-26
[protobuf2] The applied occurrence of `syntax` should not be mandatory.


kaby76:
In [protobuf2](https://github.com/antlr/grammars-v4/tree/6b517735620223475eefaa85c92f8d6bce15f360/protobuf/protobuf2), the syntax for `proto` has `syntax` as mandatory.

https://github.com/antlr/grammars-v4/blob/6b517735620223475eefaa85c92f8d6bce15f360/protobuf/protobuf2/Protobuf2.g4#L16-L18

This is wrong. The Protobuf2 spec says: `proto = [syntax] { import | package | option | topLevelDef | emptyStatement }` (https://protobuf.dev/reference/protobuf/proto3-spec/#proto_file).

Consequently, it fails to parse the first example in the Extensions section in the programming guide (https://protobuf.dev/programming-guides/proto2/#ext-example).
```
// file kittens/video_ext.proto

import "kittens/video.proto";
import "media/user_content.proto";

package kittens;

// This extension allows kitten videos in a media.UserContent message.
extend media.UserContent {
  // Video is a message imported from kittens/video.proto
  repeated Video kitten_videos = 126;
}
```
I mentioned this problem in another comment, https://github.com/antlr/grammars-v4/issues/4629#issuecomment-3337458963, chasing down another problem.

PR #4636, opened 2025-09-26
[protobuf2] Fix for #4635: `start` symbol is optional.

kaby76:
As mentioned in the comment in #4635, the grammar deviates from the spec. This change resolves the issue, and the tests now include an additional test to verify its effectiveness.


teverett:
@kaby76  thanks!

PR #4638, opened 2025-09-27
[PlSql] Introduce support for annotations_clause

Naros:


Naros:
@teverett rebased, should be g2g.

teverett:
@Naros  thanks!

PR #4640, opened 2025-09-29
[PlSql] Multiple grammar improvements

Naros:
This PR includes a myriad of grammar improvements, separated by commit. Given that many of these may touch or adjust various similar files, it made more sense to put these into 1 PR where they can be reviewed by commit rather than multiple PRs that would require continuous rebases.

### DIFF
--- a/sql/plsql/Python3/PlSqlParserBase.py
+++ b/sql/plsql/Python3/PlSqlParserBase.py
@@ -3,7 +3,7 @@ from antlr4 import *
 class PlSqlParserBase(Parser):
 
     _isVersion10 = False
-    _isVersion11 = False
+    _isVersion11 = True
     _isVersion12 = True
 
     def isVersion10(self):


Naros:
Hi @teverett I'm trying to understand the various failures, but it's not clear what is broken. When running the tests via Maven, everything passes. Any pointers on what I may have missed?

kaby76:
The error is caused by time-outs on the network file transfers from Maven. There are a number of ways I could fix this:
* Fork https://github.com/antlr/antlr4-tools and add in a robust network file transfer that doesn't stop until N tries, something I've been pointing out years ago. https://github.com/antlr/antlr4-tools/issues/13
* Write my own tool that is robust. Of course, people won't use it, just like Trash I spent years writing.
* Resubmit the job (via a NOP-change, like add another newline to the the last line of the file) to force the rebuild during hours that don't pulverize Github Action's puny servers or the lethargic Maven Central.
* Have @teverett try to rerun the workflow for the PR at something like before 10 AM UTC.

The old Antlr Maven tester is old and hasn't been updated since 4.11.1 or 2. Plus, it does not test any targets except Java.

My recommendation is to try making a change and pushing it around 6 am EDT.

Naros:
@kaby76 are you sure that's the case for all of them, e.g. ubuntu-latest Go
```
New errors in output.
/home/runner/work/grammars-v4/grammars-v4/sql/plsql/examples/examples-sql-script/with_clause_in_exists_block_in_procedure.sql.tree
line 226:9 no viable alternative at input 'RELATIONAL'
line 226:19 missing ';' at '('
```

kaby76:
You're right. Not all of the problems were network timeouts.

For CSharp, TypeScript, Python, and maybe others, you made the bool `_isVersion11 = false`. It should consistently be "true" across targets for the code to work.

For Go, There's no "base class" initializers! To get around this mess, I had to implement a global, static struct. You'll have to initialize _isVersion11 to "true" in `func init()`, where it initializes other bools.

Naros:
@teverett @kaby76 where do we stand on this PR?

kaby76:
The Python3 port is broken still. (The parse tests are only run on Linux and Windows because the third OS--MacOS--is slower than mollassas; the tests have timeouts, so the OS is not reliable for parsing test. But we still test the grammar there and do compilations/links.)

Here's what the Python3 port does.
```
$ bash run.sh ../examples/create_table.sql
line 226:9 no viable alternative at input 'RELATIONAL'
line 226:19 missing ';' at '('
Python3 0 ../examples/create_table.sql fail 25.162585
Total Time: 25.165641
```

Here's what the CSharp port does.
```
$ bash run.sh ../examples/create_table.sql
CSharp 0 ../examples/create_table.sql success 2.1247596
Total Time: 2.5166988
10/13-11:08:30 ~/issues/g4-4640/sql/plsql/Generated-CSharp
```

The Python3 port still has ` _isVersion11 = False`. It should be ` _isVersion11 = True`. If you change that, the parse works.

Naros:
Thanks @kaby76 things are all green now, appreciate the pointer.

kaby76:
> Thanks @kaby76 things are all green now, appreciate the pointer.

Looks good.

For your other PR, https://github.com/antlr/grammars-v4/pull/4638, I published a new version of antlr4-tools (more robust) and will update the builds in this repo to use that. That should take care of the download issues from Maven Central.

teverett:
@Naros  thanks!

PR #4653, opened 2025-10-25
[build] Fix for #4647: Maven Central unreliable, use Antlr download page at GitHub.

### LINKED ISSUE #4647: [Build] Maven Central unreliable in builds.
As described in another thread (https://github.com/antlr/grammars-v4/pull/4631#issuecomment-3352514715), Maven Central is used in the builds to fetch a version of the Antlr Tool .jar. The problem is that on a heavily loaded Maven Central, the server will deny download requests when heavily loaded, even though the GitHub Actions VM servers are not.

Additionally, the 'antlr4' tool used doesn't perform retries; if it fails once, the request is never attempted again.

https://github.com/antlr/antlr4-tools/blob/dd4ea2883ddcfe023c8ffc3b7f96289bef9ef0c9/antlr4_tool_runner.py#L54-L69

I have a modification of antlr4-tools that fixes these problems. See https://github.com/kaby76/antlr4-tools/tree/tools-fix-fetches. However, the change hardwires the .py to run the command-line tool rather than the GUI. To call, run`python3 antlr4_tool_runner.py`; there won't be a published version in PyPl because I don't understand Python at all.

kaby76:
This PR fixes #4647.

Briefly, Maven Central blocks requests when bandwidth usage exceeds a threshold (https://central.sonatype.org/faq/403-error-central/#question). This can occur with builds, particularly with PRs that modify more than one grammar.

The solution implemented here uses a version of the antlr4-tools that I updated to download from https://github.com/antlr/website-antlr4/tree/gh-pages/download instead of https://repo1.maven.org/maven2/org/antlr/antlr4/. In addition, the Python tool was not programmed to be very robust: with slow network communications, requests would time out. Code was added to attempt the download multiple times, with a short pause between attempts, until it either succeeds or gives up trying.

For more information on the updated antlr4-tools, see https://github.com/kaby76/antlr4-tools/tree/tools-fix-fetches and  https://pypi.org/project/antlr4-tools-ken.domino/.


kaby76:
@teverett Please consider merging this PR. Maven Central is not a reliable service for downloading the Antlr Tool jar during a build because it treats the downloads as a DOS attack. It might be best to keep a cache of the most used .jars in this repo.

kaby76:
I'm getting these errors almost daily, for months and months. I'm closing this PR and re-opening it with conflicts resolved. We need this fixed.

PR #4655, opened 2025-10-27
[Snowflake]Add new supported syntax in IN clause of PIVOT

MasterKuat:
Doc: https://docs.snowflake.com/fr/sql-reference/constructs/pivot#syntax
Pivot now support dynamic pivot.


teverett:
@MasterKuat  thanks!

PR #4661, opened 2025-10-29
[java20] Fix for #4660: location of parsing tests incorrect.

### LINKED ISSUE #4660: [java20] Wrong directory specified using `<inputs>`.
Looks like there's a mistake in the location of the `examples/` directory for the java20 grammar. It should say `examples/**/*.java`, not `**/*.java`.

https://github.com/antlr/grammars-v4/blob/ec9ccb72950766262bcba769e2a943aa5765822a/java/java20/desc.xml#L4

I noticed the problem when I tried to run performance tests on the grammar, and found that the generated parser and lexer .java files were added to the test.

kaby76:
This PR fixes #4660, where the location of the parser test input files was incorrect. Trash trgen assumes the initial directory is where the desc.xml file resides. Any example inputs should be relative to that. So, "examples/" must be part of the string because "&#42;&#42;/*.java" includes not only those .java files with "examples/&#42;&#42;/&#42;.java", but also with "target/&#42;&#42;/&#42;.java" and "Generated-Java/&#42;&#42;/&#42;.java"!


teverett:
@kaby76 thanks!

Issue #4667, opened 2025-10-31
[acme] "values unreachable" warnings in acme.g4.


kaby76:
While checking out a full build of grammars-v4 with the Rust target, I noticed warnings outputted from the Antlr Tool for the acme grammar.
```
warning(184): acme.g4:886:0: One of the token PATHSEPARATOR values unreachable. : is always overlapped by token COLON
warning(184): acme.g4:886:0: One of the token PATHSEPARATOR values unreachable. . is always overlapped by token DOT
warning(184): acme.g4:886:0: One of the token PATHSEPARATOR values unreachable. - is always overlapped by token MINUS
warning(184): acme.g4:910:0: One of the token PLUS values unreachable. + is always overlapped by token PATHSEPARATOR
warning(184): acme.g4:946:0: One of the token REM values unreachable. % is always overlapped by token PATHSEPARATOR
warning(184): acme.g4:987:0: One of the token SLASH values unreachable. / is always overlapped by token PATHSEPARATOR
warning(184): acme.g4:886:0: One of the token PATHSEPARATOR values unreachable. : is always overlapped by token COLON
warning(184): acme.g4:886:0: One of the token PATHSEPARATOR values unreachable. . is always overlapped by token DOT
warning(184): acme.g4:886:0: One of the token PATHSEPARATOR values unreachable. - is always overlapped by token MINUS
warning(184): acme.g4:910:0: One of the token PLUS values unreachable. + is always overlapped by token PATHSEPARATOR
warning(184): acme.g4:946:0: One of the token REM values unreachable. % is always overlapped by token PATHSEPARATOR
warning(184): acme.g4:987:0: One of the token SLASH values unreachable. / is always overlapped by token PATHSEPARATOR
```
`PATHSEPARATOR` was mentioned before in https://github.com/antlr/grammars-v4/issues/2308. A fix was suggested in https://github.com/ispras/grammars-v4/commit/3c3b17d985aad14de56129a05b9782b12d09a43e. But, it doesn't look correct, and was never merged.

PR #4670, opened 2025-11-11
[acme] Fix for #4667: warnings from Antlr Tool for lexer symbol unreachable.

kaby76:
Fix for #4667. ANTLR grammars define lexer rules after first noting all string literals. If the lexer rules don't define a single lexer name for each string literal, then the folding is never performed by the Antlr Tool. The tool now provides a check and warning for lexer rules that overlap with other rules. I changed the rule to a parser rule, which avoids the problem.

### DIFF
--- a/acme/acme.g4
+++ b/acme/acme.g4
@@ -883,7 +883,7 @@ OR
     : O R
     ;
 
-PATHSEPARATOR
+pathseparator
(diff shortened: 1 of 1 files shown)


teverett:
@kaby76  thanks!

PR #4673, opened 2025-11-12
[acme] Fix for #4672: use case insensitive option; fix unary floats.

### LINKED ISSUE #4672: [acme] Various questions on correctness.
* Please compare https://github.com/antlr/grammars-v4/blob/c164eb4228e925754c8b4ebab303d4be1cf3cf04/acme/acme.g4#L513 versus https://github.com/antlr/grammars-v4/blob/c164eb4228e925754c8b4ebab303d4be1cf3cf04/acme/acme.g4#L1175. Generally speaking, you don't define a lexer value with a `'-'`-sign and turn around and have unary expressions with a `MINUS` lexer symbol in a parser rule. It doesn't make a lot of sense.
* Examine https://github.com/antlr/grammars-v4/blob/c164eb4228e925754c8b4ebab303d4be1cf3cf04/acme/acme.g4#L1187. Is it correct to have hyphens in identifiers given expressions can have `'-'`-sign operators?
* Consider using the case-insensitive option for lexer rules.

kaby76:
This PR fixes #4672. The grammar was simplified using the lexer caseInsensitive option, and the auxiliary rules that implement case (e.g., `A : 'a' | 'A';`) were removed, simplifying the grammar. Note, this allows direct use of the string literal for keywords, which is sometimes preferable. Also, the arithmetic sign was removed from the float lexer rule and moved to the parser.

### DIFF
--- a/acme/desc.xml
+++ b/acme/desc.xml
@@ -1,4 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <desc xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../_scripts/desc.xsd">
+   <antlr-version>^4.10</antlr-version>
    <targets>CSharp;Cpp;Dart;Go;Java;JavaScript;Python3;TypeScript;Antlr4ng</targets>
 </desc>


teverett:
@kaby76  thanks!

PR #4680, opened 2025-11-17
[vhdl] Add grammar to support the following features of vhdl2008:

LucasMoss:
* block comments

 * process(all)

 * if-else generate, elsif generate, for <range> generate

 * 'subtype attribute in signal declarations

 * AND, OR, XOR, NAND, XNOR unary reduction functions


teverett:
@LucasMoss thanks!

vhdl/vhdl2008/examples/add_pkg.vhd
```diff
@@ -0,0 +1 @@
+../../vhdl/examples/add_pkg.vhd
```
kaby76, line 1: Antlr is not a preprocessor. There is no code in the grammar that opens another input file and starts processing that file. While the input at `../../vhdl/examples/add_pkg.vhd` may parse (haven't checked), the input string `"../../vhdl/examples/add_pkg.vhd"` does not parse. Neither Maven tester or the Trgen tester passed, at least not on Windows. https://github.com/antlr/grammars-v4/issues/4690. These could be symbolic links, but git doesn't seem to support them. But, maybe hard links are.

There are other issues with testing. I'll be cleaning this up today. Thanks.

Issue #4676, opened 2025-11-13
[C] declarationSpecifiers is too greedy, take initDeclaratorList as typedefName


kevemueller:
I am parsing C with the current C grammar in Java language and try to make sense of the parse tree.
I believe to have found a bug similar to #1979, in which a typedef appears instead of a directDeclarator.
As illustration, first a good parse:

```c
typedef unsigned const int* x;
```
parses as:

<img width="881" height="375" alt="Image" src="https://github.com/user-attachments/assets/51538784-d55b-4608-b362-ae4f3fe3d6a3" />

Note the clear distinction between ```declarationSpecifiers``` and ```initDeclaratorList```.

Whereas the following
```c
typedef unsigned const int x;
```
parses as 

<img width="861" height="375" alt="Image" src="https://github.com/user-attachments/assets/9c984540-48c8-4d4f-8fc2-f8e3431961e9" />

The ```initDeclaratorList``` is missing, instead of being a ```directDeclarator``` the ```x``` is treated as yet another ```typeSpecifier```.

The same erratic behaviour can be observed with 

```c
extern int* x;
extern int x;
```
The former being parsed properly, the latter not.

I am currently trying to write some post-processing logic to detect this, but it is not easy to distinguish the cases and it definitely messes with the situation at the wrong place (in the visitor) instead of in the grammar.

Any help is appreciated, this is currently a major blocker for proceeding with ANTLR (and this grammar).

kaby76:
In the input string `extern int x;`, the `x` is parsed as a [declarationSpecifier](https://github.com/antlr/grammars-v4/blob/c164eb4228e925754c8b4ebab303d4be1cf3cf04/c/C.g4#L182). It should be parsed as an [initDeclaratorList](https://github.com/antlr/grammars-v4/blob/c164eb4228e925754c8b4ebab303d4be1cf3cf04/c/C.g4#L190). You can see this because the grammar is ambiguous:
```
11/14-06:37:56 ~/issues/g4-current/c/Generated-CSharp
$ trparse --ambig x | trtree
CSharp 0 x success 0.0449465
compilationUnit
├── translationUnit
│   ├── externalDeclaration
│   │   └── declaration
│   │       ├── declarationSpecifiers
│   │       │   ├── declarationSpecifier
│   │       │   │   └── storageClassSpecifier
│   │       │   │       └── Extern
│   │       │   │           └── "extern"
│   │       │   └── declarationSpecifier
[...]

Issue #4380, opened 2025-01-11
[c] C grammar really should have a basic symbol table.


kaby76:
https://groups.google.com/g/antlr-discussion/c/nGvUxmnOxsI

Issue #1979, opened 2021-01-05
[C] struct members are parsed as typedefs


jirislaby:
https://github.com/antlr/grammars-v4/blob/4c783b6b0f10927485c4e63007088ddbb3f777c0/c/C.g4#L245-L248
improperly parses:
```
struct X {
        int val1;
};
```
into:
```
(structOrUnionSpecifier 
    (structOrUnion struct) X { 
    (structDeclarationList 
        (structDeclaration 
            (specifierQualifierList 
                (typeSpecifier int) 
                (specifierQualifierList 
                    (typeSpecifier 
                        (typedefName val1)))) ;)) }))
```

Note the `typedefName val1`. val1 is not type specifier, but identifier.

We used GnuC antlr3 grammar in https://github.com/jirislaby/stanse/blob/master/src/cz/muni/stanse/cparser/GNUCa.g with `typedef` handling taken over from https://github.com/antlr/grammars-v3/blob/master/ANSI-C/C.g. I.e. conditional parsing of identifiers. This was eliminated in this v4 grammar, so that the grammar is language agnostic. But it obviously doesn't work here.

jirislaby:
Another example:
```C
(void *)(addr) + 1;
```
results in:
```
(castExpression ( 
    (typeName 
        (specifierQualifierList 
            (typeSpecifier 
                (typeSpecifier void) 
                (pointer *)))) ) 
    (castExpression ( 
        (typeName 
            (specifierQualifierList 
                (typeSpecifier 
                    (typedefName addr)))) ) 
        (castExpression 
            (unaryExpression 
                (unaryOperator +) 
                (castExpression 
                    (unaryExpression 
                        (postfixExpression 
                            (primaryExpression 1))))))))
```
Opposing to:
```C
(void *)addr + 1
```
which results in correct:
```
(additiveExpression 
    (additiveExpression 
        (multiplicativeExpression 
            (castExpression ( 
                (typeName 
                    (specifierQualifierList 
                        (typeSpecifier 
                            (typeSpecifier void) 
                            (pointer *)))) ) 
                (castExpression 
                    (unaryExpression 
                        (postfixExpression 
                            (primaryExpression addr))))))) + 
    (multiplicativeExpression 
        (castExpression 
            (unaryExpression 
                (postfixExpression 
                    (primaryExpression 1))))))
```

kaby76:
[...]

Issue #1115, opened 2018-04-15
c: unable to parse FunctionPointer.c


peterremote1980:
Unable to parse the example https://github.com/antlr/grammars-v4/blob/master/c/examples/FunctionPointer.c

```
CLexer lexer = new CLexer(new ANTLRInputStream(IOUtils.toString(new FileInputStream("/Users/peter/Desktop/FunctionPointer.c"), "utf-8")));
CommonTokenStream tokenStream = new CommonTokenStream(lexer);
CParser parser = new CParser(tokenStream);
ParserRuleContext t = parser.primaryExpression();

ParseTreeWalker walker = new ParseTreeWalker();
MyCListener listener = new MyCListener(parser);
walker.walk(listener, t);
```

line 2:0 no viable alternative at input 'typedef'

Please help

PR #4682, opened 2025-11-21
[c] Update the C grammar to the latest ISO Spec.

kaby76:
This PR is a fix for issue #4676, where some declarations were parsed incorrectly. But this PR fixes much more than that. This fix addresses long-standing problems in the grammar for C.

* Fixes most ambiguity in the grammar by incorporating a symbol table. This is necessary because, plain and simple, the C language EBNF is not context-free. Further, ANTLR 4 does not perform well on grammars with ambiguity and large max-k values; it often makes the wrong choice among all possible parses.
* Calls a native preprocessor on the C input to support the symbol table disambiguation rather than requiring people to run the preprocessor manually.
* Updates the grammar to the latest official ISO Spec (purchased, rather than draft), and includes extensions for GCC, Clang, and Visual Studio C 2026.
* Expands testing with several thousand tests.
* Outputs correct line/column information for error messages.
* Upgrades Trash, necessary to handle the added preprocessor step, for globbing patterns for selecting various tests from the test suites (see `<inputs>` elements in the desc.xml file), and for command line args for the lexer and parser.

The point of this PR is to bring the grammar to a high-quality state.

For further information, please see the new readme.md.

# Issues to close
* #4676
* #1979
* #4683
* #4380
* #1065
* #1066
* #1082
* #1115


kaby76:
BS! Some of the examples in the test suite are not ***strict*** ISO C. For example, this should not parse.
```
int p1, p2;
int a2(int param1, param2);
```
But, it does!
```
$ gcc -c -std=c24 p.c
gcc.exe: error: unrecognized command-line option '-std=c24'; did you mean '-std=c2x'?
11/22-08:50:09 ~/issues/g4-current/c/Generated-CSharp
$ gcc -c -std=c2x p.c
p.c:2:20: error: unknown type name 'param2'
    2 | int a2(int param1, param2);
      |                    ^~~~~~
11/22-08:50:17 ~/issues/g4-current/c/Generated-CSharp
$ cat p.c
int p1, p2;
int a2(int param1, param2);
11/22-08:51:49 ~/issues/g4-current/c/Generated-CSharp
$ ./bin/Debug/net8.0/Test.exe p.c
CSharp 0 p.c success 0.0453763
Total Time: 0.1107057
11/22-08:52:29 ~/issues/g4-current/c/Generated-CSharp
$
```
It can't parse because a `declarator` *must* include a type, as with `int param2`. `int param1, param2` currently parses because `param2` is parsed as a type id, as though you were give it `int param1, int`. We cannot at the same time ignore this, and add a symbol table to solve #4676.

So, several of the examples provided should not parse, but they do.

kaby76:
@teverett After many months of working on the C grammar, the work is complete.

kaby76:
Squashed PR. The replacement is https://github.com/antlr/grammars-v4/pull/4735

Issue #2827, opened 2022-09-16
SQLite grammar: top level `parse` rule accepts statements not separated by semicolons


juretta:
The SQLite language syntax defines a `sql-stmt-list` that consists of zero, one or more `sql-stmt`s. 
Each `sql-stmt` should be separated by a semicolon. See [https://www.sqlite.org/lang.html](https://www.sqlite.org/lang.html).

Here is a simple example with SQLite (SQLite 3.37.0):

```
sqlite> SELECT Title FROM Album LIMIT 2; SELECT ArtistId FROM Album LIMIT 3;
For Those About To Rock We Salute You
Balls to the Wall
1
1
2
```
Omitting the first semicolon (and hence not properly separating the `SELECT` statement) leads to an error:

```
sqlite> SELECT Title FROM Album LIMIT 2 SELECT ArtistId FROM Album LIMIT 3;
Error: in prepare, near "SELECT": syntax error (1)
```

The ANTLR grammar currently accepts the second form but IMHO shouldn't. 

## Root cause

The grammar-v4 grammar for the SQLite SQL dialect is defined as follows:

https://github.com/antlr/grammars-v4/blob/07314e4615982ba77864d7b8cd804c7b5d803bb0/sql/sqlite/SQLiteParser.g4#L37-L42

So `parse` is also valid for a list of `sql_stmt_list`s that each consist of a single `sql_stmt` only, hence accepting the SQL statements not separated by semicolons.

Instead the `parse` rule should probably be:

```
parse: (sql_stmt_list)? EOF
;
```

which only accepts a single `sql_stmt_list` at most and hence requires `sql_stmt`s to be separated by `SCOL` (semicolon).

juretta:
Here is a change with accompanying tests: https://github.com/antlr/grammars-v4/compare/master...juretta:grammars-v4:master#diff-8bf47bc01d73b92fd763a6f79c0a9827e7006434ce463139e2f40cecc6a2ee1a

Changing `parse` causes the generated parser to omit a single (optional) `sql_stmt_list` instead of that being a list. Not entirely sure about the ramifications here, this seems to break all code that relies on the previous behaviour. So updating to the new grammar would be a breaking change.

michaelmcleodnz:
This was fixed by #4688.

Issue #2640, opened 2022-06-03
[SQLite] Unexpected errors with legit SELECT statement using `parse` rule


ryangawei:
Hi there,

I'm using the SQLite grammar, 

```
def parse(query: str):
    input_stream = InputStream(query)
    lexer = SQLiteLexer(input_stream)
    stream = CommonTokenStream(lexer)
    parser = SQLiteParser(stream)

    # Weird errors happen when it starts with `parse()`
    # tree = parser.parse()
    tree = parser.select_stmt()

    return tree
```

if I start with the rule `parse`, I get errors using the following queries,

```
>>> SELECT count(*) FROM postseason WHERE YEAR  =  1885 AND ties  =  1;
line 1: 52 extraneous input 'AND' expecting {<EOF>, ';', 'ALTER', 'ANALYZE', 'ATTACH', 'BEGIN', 'COMMIT', 'CREATE', 'DEFAULT', 'DELETE', 'DETACH', 'DROP', 'END', 'EXPLAIN', 'INSERT', 'PRAGMA', 'REINDEX', 'RELEASE', 'REPLACE', 'ROLLBACK', 'SAVEPOINT', 'SELECT', 'UPDATE', 'VACUUM', 'VALUES', 'WITH'}

>>> SELECT document_name FROM documents GROUP BY document_type_code ORDER BY count(*) DESC LIMIT 3 INTERSECT SELECT document_name FROM documents GROUP BY document_structure_code ORDER BY count(*) DESC LIMIT 3
line 1: 95 extraneous input 'INTERSECT' expecting {<EOF>, ';', 'ALTER', 'ANALYZE', 'ATTACH', 'BEGIN', 'COMMIT', 'CREATE', 'DEFAULT', 'DELETE', 'DETACH', 'DROP', 'END', 'EXPLAIN', 'INSERT', 'PRAGMA', 'REINDEX', 'RELEASE', 'REPLACE', 'ROLLBACK', 'SAVEPOINT', 'SELECT', 'UPDATE', 'VACUUM', 'VALUES', 'WITH'}
```

And if I switch to `select_stmt`, the errors are gone. And it seems to me that the two SQL queries have correct syntax. What's happening here?

kaby76:
[...]

Issue #2266, opened 2021-07-23
Infinite recursion in SQLite grammar


Changochen:
As shown in https://github.com/antlr/grammars-v4/blob/cda43d1e0e58baccdf62da56d01b86ab76e88d93/sql/sqlite/SQLiteParser.g4#L605,  the rule seems wrong as it will never stop expanding.

```antlr
order_by_expr_asc_desc:
    ORDER_ BY_ order_by_expr_asc_desc
;
```

Changochen:
I think it should be
```antlr
order_by_expr_asc_desc:
    ORDER_ BY_ expr_asc_desc
;
```

kaby76:
Dup of https://github.com/antlr/grammars-v4/issues/2039

The main problem is that it's a useless symbol. It is used only by `window_function`, which is used only in `window_function_invocation`---and that symbol is not used anywhere. Trash's tranalyze has graph analysis of symbol uses and finds all useless symbols. I think Bison has a similar feature.

In fact, there's an ominous "// unknown" comment right before this production.

dpmm99:
The original issue was fixed [here](https://github.com/antlr/grammars-v4/commit/76b1d66d768902f821b7a7440507bfdb9b6a32f9); do you want to rename this issue because of the unused rule or open a new issue?

kaby76:
I think close this issue out. All grammars should be analyzed for useless rules, so the issue is not limited to this grammar alone.

dpmm99:
Makes sense. Also, I just noticed #1493 covers the window functions not being supported.

Issue #2151, opened 2021-04-20
sqlite grammar: BETWEEN expr AND expr


ysdg:
I found a problem of sqlite grammar when BETWEEN AND + AND expr. For example:
# Result Now
```sql
SELECT a FROM tb WHERE a BETWEEN 100 AND 200 AND b > 0
```
Now, just analyze the WHERE condition expr.
```sql
a BETWEEN 100 AND 200 AND b > 0
```
The grammar now is :
```sql
expr1 BETWEEN expr2 AND expr3
```
Then, parser to:
|  expr   | content  |
|  ----  | ----  |
| expr1  | a |
| expr2  | **100 AND 200** |
| expr3  | **b > 0** |

# Result Expected
But, by the priorty of sqlite(AND > BETWEEN), it should parser AND first to:
AND left expr:

```sql
a BETWEEN 100 AND 200
```
AND right expr:
```sql
b > 0
```
Then, parser BETWEEN to:
|  expr   | content  |
|  ----  | ----  |
| expr1  | a |
| expr2  | **100** |
| expr3  | **200** |

# Solution Maybe

I think expr of BETWEEN should not contain AND.
So, may split the expr(AND, OR) to loginExpr

PR #4688, opened 2025-11-26
More fixes for SQLite

michaelmcleodnz:
Here are a few small things I didn't catch in my last update to the SQLite grammar (#4679), typically because they are more to do with parse tree shape or only specified on subpages of the SQLite docs:

- There was no reason for `column_alias` to have different semantics than the rest of the identifiers.
- The page on [vaccum statements](https://sqlite.org/lang_vacuum.html) says `filename` "can be an arbitrary SQL expression that evaluates to a string".
- The rule for identifiers (`any_name`) was allowing any identifier to be surrounded by parentheses, but I can't find anywhere in the docs suggesting this should be the case, it seems to be incorrect in practice, and the tests continue to pass when I remove the option, so I'm fairly confident it should be removed.
- `over_clause` refactored to use the same `order_clause` helper rule that everything else with an ORDER BY was using.
- Whole `FROM` clause was missing from `update_stmt_limited`.
- Operator precedence in `expr` has been creating incorrect parse trees [since the original repo](https://github.com/bkiers/sqlite-parser/pull/16), so I've reordered the branches of `expr` to create correct parse trees as per [the order in the docs](https://sqlite.org/lang_expr.html#operators_and_parse_affecting_attributes). See the pull request in the original repo for example screenshots of the parse trees before and after the fix.
- I fixed another bug [discovered in the original repo](https://github.com/bkiers/sqlite-parser/pull/14) that was causing `type_name` to match `column_constraint`s, which I fixed by making `type_name` [non-greedy](https://github.com/antlr/antlr4/blob/master/doc/wildcard.md).
- The top level rules weren't allowing multiple semicolons without a statement (e.g. `;;;`), and were allowing multiple statements in a row that weren't separated by semicolons (e.g. `SELECT 1 SELECT 2`). These have been fixed so that statements must be separated but can be empty.

### DIFF
--- a/sql/sqlite/SQLiteParser.g4
+++ b/sql/sqlite/SQLiteParser.g4
@@ -511,7 +511,7 @@ filename
 ;
 
 module_argument
-    : (~(OPEN_PAR | CLOSE_PAR | COMMA) | OPEN_PAR module_argument* CLOSE_PAR)+
+    : (~(OPEN_PAR | CLOSE_PAR) | OPEN_PAR module_argument* CLOSE_PAR)+?
 ;
 
 keyword
--- a/sql/sqlite/examples/sql2.sql
+++ b/sql/sqlite/examples/sql2.sql
@@ -25,7 +25,9 @@ CREATE TABLE f
     m  REAL
 ) STRICT;
 
+-- Module arguments 
 CREATE VIRTUAL TABLE f USING module_name(abc()d(te(s)t)f,gh!j%$m@p);
+CREATE VIRTUAL TABLE f USING module_name(column VARCHAR(1,2), secondarg);
 
 -- All identifiers are keywords (https://sqlite.org/c3ref/keyword_check.html)
 CREATE TABLE BEGIN(REPLACE,PRAGMA,END);


michaelmcleodnz:
(I'm especially happy with the `type_name` fix, because the suggested fix in the original pull request was to disallow keywords as identifiers, which would have broken a lot of valid SQLite queries, and implementing it the way SQLite does with a custom parser that supports a `%fallback` rule seems to be impossible in Antlr because the parser can't ask the lexer to reinterpret tokens. But so far as I can tell, the current change fixes the only issue that not having `%fallback` was causing.)

kaby76:
I'm noticing new ambiguities in the grammar.

Prior to your last PR (in https://github.com/antlr/grammars-v4/tree/b019c956c3405ebcb7d78f38ee9a902c598519ef), ambiguities were in these 12 rules:
```
expr
frame_left
frame_right
frame_single
indexed_column
insert_stmt
select_core
sql_stmt
sql_stmt_list
table_or_subquery
type_name
update_stmt
```
With your last PR (in https://github.com/antlr/grammars-v4/commit/b1f9d3d92f002cb2e2d1e61faa98d98f04cb369c), the ambiguities were in 13 rules:
```
expr
frame_left
frame_right
frame_single
indexed_column
insert_stmt
module_argument
select_core
sql_stmt
sql_stmt_list
table_or_subquery
type_name
update_stmt
```
`module_argument` is ambiguous.

With this PR (https://github.com/michaelmcleodnz/grammars-v4/commit/b6be1f98b2905780e79f865142b1ea8cf4340497), there are more ambiguities, now 14 rules:
```
column_def
expr
frame_left
frame_right
frame_single
indexed_column
insert_stmt
module_argument
select_core
sql_stmt
table_or_subquery
type_name
update_stmt
update_stmt_limited
```
Could you address these ambiguities?

In addition, I'm somewhat concerned that you are using the online railroad diagrams. Typically, online documentation trails the current implementation and is often notoriously wrong because people scrape the grammar from the source code by hand. SQLite is one of the few grammars where the Bison grammar is open source (https://github.com/sqlite/sqlite/blob/master/src/parse.y).

michaelmcleodnz:
I'd love to address these, but I'm at a loss for how to identify ambiguities; what tool are you using for this?

With regards to using the railroad diagrams, I'm unsure if you're saying that scraping the grammar from the source code would be better or worse, but the current version is already based on the diagrams, which you can tell because it includes orphaned rules like `simple_select_stmt` and `window_function_invocation` that are only specified there because they're used as examples on other pages of the documentation. It seems the diagrams are [generated](https://pikchr.org/home/doc/trunk/doc/sqlitesyntax.md) by the SQLite maintainers, and not a community wiki.

I would be surpised if my changes so far have fixed all the issues with the current grammar, and would be more than happy for someone to read over parse.y to find points of departure. I've just been trying to fix the errors and ommissions I could easily identify, and have been testing that I can execute any changes against an actual SQLite database as well.

I don't fully understand what makes Antlr see a rule as ambiguous, but what I can say about the new ambiguities:
- `update_stmt_limited` was missing a subsection that depends on the ambiguous `table_or_subquery` rule, and is now almost identical to the already ambiguous `update_stmt` which I haven't touched;
- `column_def` was returning incorrect parse trees before (in situations where `type_name` was missing), and the change here seems to have resolved that issue, but I don't know if 'ambiguous' means it can't be actually trusted to parse correctly in all situations;
- `module_argument` is the only change I've made where I resorted to trying to interpret [parse.y](https://www.sqlite.org/src/file?name=src/parse.y) because the [documentation](https://sqlite.org/lang_createvtab.html) is a bit thin, and I would trust my implementation of this far less than the rules that there are syntax diagrams for. The previous solution had a comment `TODO check what exactly is permitted here`, which I did what I could to resolve but could potentially reinstate.

As to resolving all the older ambiguities, the more I can do the better because I want to be able to use this grammar to parse SQLite statments, but I'm also aware that it would be outside my area of expertise and might be something best done by someone else in a separate pull request. At least keeping them would be no worse than doing nothing, which would also result in keeping them.

The grammar should be far more aligned with SQLite's behaviour and spec after these changes than it was before.

kaby76:
> I'd love to address these, but I'm at a loss for how to identify ambiguities; what tool are you using for this?

I'm using my own [Trash toolkit](https://github.com/kaby76/Trash) to find ambiguity. You may not want to use it because it requires .NET, generate a C# driver, and build. But, if you have it installed, then you can install the toolkit (see https://github.com/kaby76/Trash?tab=readme-ov-file#install-globally), generate a C# driver (`dotnet trgen -t CSharp`), build (`bash build.sh` or `pwsh build.ps`), then run `trperf ../examples/*.sql`
The easiest is probably best to use the Antlr plugin/Intellij environment. The toolkit is used in this repo for testing because it supports all the Antlr targets, and all OSes. But, you can see the ambiguity test in one of the build outputs (https://github.com/antlr/grammars-v4/actions/runs/19701451910/job/56438509540#step:15:115).

It's probably ok for now to ignore the ambiguity as long as it doesn't cause the build to go over the 5-minute per grammar time limit.

Thank you for updating this grammar. Really appreciate all the help in maintaining this repo.

michaelmcleodnz:
Thanks for that; I'd been running `trgen` and `build.sh` and `test.sh` as per the [wiki](https://github.com/antlr/grammars-v4/wiki#using-trgen-all-targets), but didn't know about `trperf`.
I will check it out later today my time and probably add a note about it to the wiki.

michaelmcleodnz:
I've tried a few attempts to resolve the ambiguities and run some tests with `test-ambiguity.sh` and `trperf`, but I think I'm going to have to admit that I don't understand the concept or problem space well enough to resolve these myself.

Should I raise a separate issue noting that the SQLite grammar has ambiguites?

In good news, while looking at the issues, I discovered that:
- [#2640](https://github.com/antlr/grammars-v4/issues/2640) was resolved by commit cc5420e in the last pull request (at least the first example that tries to use a keyword as a column name; the second example is actually invalid SQLite and will produce the error "ORDER BY clause should come after INTERSECT not before" when run in SQLite, so doesn't need changing);
- [#1493](https://github.com/antlr/grammars-v4/issues/1493) is resolved after a first attempt was made in [#1790](https://github.com/antlr/grammars-v4/issues/1790) and a second in [#1808](https://github.com/antlr/grammars-v4/issues/1808) that would have been correct if not for some conflicts with the first attempt that led to errors like [#2266](https://github.com/antlr/grammars-v4/issues/2266), and the broken code from the first attempt was finally removed in 2ecd6c1 of the last pull request;
- [#2827](https://github.com/antlr/grammars-v4/issues/2827) will be resolved after merging b6be1f9 from this current pull request.

Despite the fixes to `expr` precedence made in this pull request, they don't fix the problem in [#2151](https://github.com/antlr/grammars-v4/issues/2151), which would be the only remaining open issue with a `sqlite` tag, and funnily enough removing the ambiguities would probably fix that issue.

teverett:
@michaelmcleodnz thanks

PR #4691, opened 2025-11-30
Adding multiple line srt text capture

JoaoMirandaXP:
May today be the day that I join the legendary league of heroes which dive into the depths of grammars ?


teverett:
@JoaoMirandaXP thanks!

Issue #4690, opened 2025-11-30
[vhdl2008] Grammar fails tests.


kaby76:
```
vhdl/vhdl2008,Java:
 Build 00:00:07
 Test 00:00:07

========
Parsing tests of vhdl/vhdl2008 for Java failed.
dos2unix: converting file ../examples/add_pkg.vhd.errors to Unix format...
dos2unix: converting file ../examples/add_pkg.vhd.tree to Unix format...
dos2unix: converting file ../examples/arith.vhd.errors to Unix format...
dos2unix: converting file ../examples/arith.vhd.tree to Unix format...
dos2unix: converting file ../examples/misc.vhd.errors to Unix format...
dos2unix: converting file ../examples/misc.vhd.tree to Unix format...
dos2unix: converting file ../examples/numeric_bit-body.vhd.errors to Unix format...
dos2unix: converting file ../examples/numeric_bit-body.vhd.tree to Unix format...
dos2unix: converting file ../examples/numeric_bit.vhd.errors to Unix format...
dos2unix: converting file ../examples/numeric_bit.vhd.tree to Unix format...
dos2unix: converting file ../examples/numeric_std-body.vhd.errors to Unix format...
dos2unix: converting file ../examples/numeric_std-body.vhd.tree to Unix format...
dos2unix: converting file ../examples/numeric_std.vhd.errors to Unix format...
dos2unix: converting file ../examples/numeric_std.vhd.tree to Unix format...
dos2unix: converting file ../examples/signed.vhd.errors to Unix format...
dos2unix: converting file ../examples/signed.vhd.tree to Unix format...
dos2unix: converting file ../examples/standard.vhd.errors to Unix format...
dos2unix: converting file ../examples/standard.vhd.tree to Unix format...
dos2unix: converting file ../examples/std_logic_1164.vhd.errors to Unix format...
dos2unix: converting file ../examples/std_logic_1164.vhd.tree to Unix format...
dos2unix: converting file ../examples/std_logic_1164_body.vhd.errors to Unix format...
dos2unix: converting file ../examples/std_logic_1164_body.vhd.tree to Unix format...
dos2unix: converting file ../examples/std_logic_textio.vhd.errors to Unix format...
dos2unix: converting file ../examples/std_logic_textio.vhd.tree to Unix format...
dos2unix: converting file ../examples/test_generate.vhd.errors to Unix format...
dos2unix: converting file ../examples/test_generate.vhd.tree to Unix format...
[...]

PR #4692, opened 2025-12-01
[vhdl2008] Fixes for #4690

kaby76:
* Replace symbolic links with duplication of file contents.


kaby76:
Doesn't work because git still thinks we have symbolic links. Symbolic links don't work on Windows. Do not use!

kaby76:
Replaced with multi-part PR. https://github.com/antlr/grammars-v4/pull/4693

PR #4698, opened 2025-12-05
V next

ravituringworks:


kaby76:
Your PR brings up an interesting point on how to address parsing for a particular version of a language. For most languages, we don't have a different grammar for each version of a language. The closest we do have for this is for parsing Java (i.e., we have `java/java/`, `java/java8`, `java/java9`, `java/java20`). The "java/java/" grammar is the most "up to date" version, but you can never put your finger on what version it's supposed to parse.

For protobuf, we're going to have to have a way to dynamically switch the parser because protobuf3 files can import protobuf2 files.

But, we're also dealing with the issue that Antlr doesn't have a good way to do target-independent grammars. `@members` in .g4's is not the preferred way to write target-independent grammars because we'd need a .g4 for every target, setting up the grammar for version skews between targets. The preferred method instead is to use ["target agnostic format"](https://github.com/antlr/antlr4/blob/dev/doc/target-agnostic-grammars.md), where the action code is wrapped in a "method call", placed in some support code in a directory named after the target, and `superClass` used in the .g4.

I'm going to start a discussion for this issue to spec out the requirements of how to address dynamically switching versions of the parser.

Issue #3358, opened 2023-04-30
[all] Unreferenced parser rule symbols


kaby76:
I wrote a little Trash script to see what I would find in grammars-v4 for unreferenced parser rule symbols, where there is a parser rule defined, but the LHS symbol of the rule is never reference anywhere in the grammar files. I was surprised to find a lot of parser rules that aren't used.

Perhaps the references to these rules were refactored out of the grammar purposefully. For example, to eliminate indirectly left recursion, unfolding is usually performed. But, afterward, unused rules should have been then been deleted.

Unused rules likely indicate an error in the grammar.

Here is the script and output:
### Find unused parser symbols
```
#
for i in `find . -name desc.xml | grep -v Generated\*`
do
  echo $i
  d=`dirname $i`
  pushd $d > /dev/null 2>&1
  # Parse all grammar files so that any imports can also be checked.
  trparse *.g4 2> /dev/null | trxgrep ' //parserRuleSpec[not(doc("*")//ruleBlock//RULE_REF/text() = ./RULE_REF/text()) and not(./ruleBlock//TOKEN_REF/text() = "EOF")]/RULE_REF' | trtext
  popd > /dev/null 2>&1
done
```
### Unused symbols in grammars-v4
```
$ bash ~/unused.sh
./abb/desc.xml
./abnf/desc.xml
./acme/desc.xml
acmePropertyElement
acmeViewDeclaration
acmeViewTypeDeclaration
terminatedDesignRuleExpression
./ada/ada2005/desc.xml
Ada2005Parser.g4:logical_operator
Ada2005Parser.g4:highest_precedence_operator
./ada/ada2012/desc.xml
AdaParser.g4:logical_operator
AdaParser.g4:highest_precedence_operator
AdaParser.g4:quantified_expression
./ada/ada83/desc.xml
Ada83Parser.g4:logical_operator
Ada83Parser.g4:highest_precedence_operator
./ada/ada95/desc.xml
Ada95Parser.g4:logical_operator
Ada95Parser.g4:highest_precedence_operator
Ada95Parser.g4:restriction
./agc/desc.xml
./alef/desc.xml
./algol60/desc.xml
integer
number
destination
./alloy/desc.xml
./alpaca/desc.xml
./angelscript/desc.xml
./antlr/antlr2/desc.xml
ANTLRv2Parser.g4:exceptionSpecNoLabel
ANTLRv2Parser.g4:action
./antlr/antlr3/desc.xml
[...]

PR #4700, opened 2025-12-06
Remove ambiguities from SQLite grammar; fixes #2151

### LINKED ISSUE #2151: sqlite grammar: BETWEEN expr AND expr

michaelmcleodnz:
After spending a bit of time wrapping my head around ambiguities and some moderate restructuring to `expr` and `keyword` especially, I've managed to remove all ambiguities from the SQLite grammar ~~except (ironically) the one I introduced to `module_argument` myself in [#4679](https://github.com/antlr/grammars-v4/pull/4679).~~

I'd be fairly confident this hasn't introduced any regressions in what can be parsed, although it has created some divergence from the official [syntax diagrams](https://sqlite.org/syntaxdiagrams.html), which created inherent ambiguities by giving certain subsets their own path. I've added comments against these rules explaining why they diverge so that someone doesn't come along and undo the changes to bring the rules in line with the diagrams. All tests are passing.

The rest of this is really long, but I thought it best to explain what changes I made in detail.

## Disambiguations

The ambiguity fixes fall into a few main categories:

### Explicit `expr` precedence

I split up the `expr` rule into a bunch of subrules that have to happen in a certain order. This is already the approach of other grammars like [Postgres](https://github.com/antlr/grammars-v4/blob/f4b7e36bcf8345206e74f1020d28523f3dcfc0ab/sql/postgresql/PostgreSQLParser.g4#L3485), and it fixes the precedence issues in [#2151](https://github.com/antlr/grammars-v4/issues/2151). It also made it easier to see which `expr` subrules had ambiguities.

### Limiting the keyword fallback list

[parse.y](https://sqlite.org/src/file/src/parse.y) has a `%fallback` rule that only lists a certain subset of keywords that can be used as identifiers. By removing any tokens from the `keyword` rule (which I've renamed `fallback` to clarify its function) that weren't in this list, some rules were instantly disambiguated:

- `column_def` and `type_name` because there was no longer any ambiguity bewteen `type_name` and `column_constraint`,
- the subset of `expr` that allowed parentheses around other `expr`s or `select_stmt`s, because it could tell them apart.

Other rules allowed slightly different subsets of keywords that required creating separate fallback lists (I just left the specific conlicting words off the main list and created extra rules that allowed the extra keywords as well):

- the `function_name` subset of `expr` clashed with `raise_function` unless `RAISE` was excluded,
- the `column_name` subset of `expr` clashed with `literal_value` unless `STRING_LITERAL` was excluded,
- `table_or_subquery` allowed a `table_alias` unpreceeded by `AS` to clash with join operator keywords unless those were excluded ([parse.y](https://sqlite.org/src/file/src/parse.y) handles this the same way, and I was able to confirm that the join keywords can be used as aliases only if they are preceeded by `AS` in practice),
- `frame_left`, `frame_right` and `frame_single` all allow both `UNBOUNDED` or `expr` in the same places, despite `UNBOUNDED` actually being in [parse.y](https://sqlite.org/src/file/src/parse.y)'s `%fallback` list. Because of this, I opted to treat the `UNBOUNDED` options as subsets of the `expr` options, so that `UNBOUNDED` can continue to remain valid expressions everywhere else.

Some keywords are always treated as identifiers by [parse.y](https://sqlite.org/src/file/src/parse.y), which then checks that they have specific values in the areas where they are used. Namely, `TRUE`, `FALSE`, `STRICT`, `ROWID`, `STORED` and `VIRTUAL` (when used on generated columns but not when used in `CREATE VIRTUAL TABLE`). These need to be part of the fallback list despite not being included in [parse.y](https://sqlite.org/src/file/src/parse.y). I put comments against these keywords to explain why they are in the fallback list.

### Merging subsets that didn't need special-casing

There are a number of pathways in the [syntax diagrams](https://sqlite.org/syntaxdiagrams.html) that didn't need to be added because one option was already covered as a subset of the other option, but they must have been added separately because they reflected the most common features of the rule.

- `insert_stmt` allowed `select_stmt` or `values_clause` even though `select_stmt` can be a `values_clause`,
- `indexed_column` allowed `expr` or `column_name` even though `expr` can be a `column_name`,
- `select_core`, `update_stmt`, `update_stmt_limited` and `table_or_subquery` all allowed a comma-separated list of `table_or_subquery`s or a `join_clause`, even though `join_clause` can be a comma-separated list of `table_or_subquery`s.
- `sql_stmt` allowed both `delete_stmt` and `delete_stmt_limited`, and `update_stmt` and `update_stmt_limited`, despite the only difference in the `_limited` versions being the extra optional clauses on the end. The `_limited` versions of the rule are only valid when the [SQLITE_ENABLE_UPDATE_DELETE_LIMIT](https://sqlite.org/lang_delete.html#optional_limit_and_order_by_clauses) compile-time option is set, but when it is, then *all* delete and update statements should have these features. This means that merging the `_limited` rules with the main ones would have also fixed a previously unidentified bug disallowing `limit_clause`s in `create_trigger_stmt`s.

I removed all these subsets and left comments so that future devs would know not to add them back.

## Other fixes

While I was changing things, I also made some other small  additions:

- Added support for the special syntax of the [percentile extension](https://sqlite.org/percentile.html#the_percentile_cont_y_p_aggregate_function), which is the only extension I've come across that adds syntax rules, and should probably be supported for full compatibility with databases that use it,
- I decided to comment out the orphaned rules from the the syntax diagrams, seeing that [this issue](https://github.com/antlr/grammars-v4/issues/3358) implies there shouldn't be any unreferenced rules in grammars. I commented them out rather than removed them, so that someone comparing the grammar to the syntax diagrams in the future understands that they aren't missing.
- Removed `RANK` from the list of tokens because it was unused and should've been removed when the rest of the clashing `window_function` code was removed.

---

~~I still have no idea how to disambiguate `module_argument`, which was what I set out to do in the first place, but~~ I am glad to have squashed everything else. I did make some updates to `module_argument` so that it conforms to my best understanding of the SQLite parser and the behaviour I see in practice, which required getting the [wasm project](https://sqlite.org/wasm/doc/trunk/index.md) running among other things (or compiling from source, but I wanted to avoid C), so that I could actually create a module function and see how the arguments are passed to it. Weirdly, while SQLite will accept 'just about' anything with balanced parentheses as module arguments, the contents of the arguments [have to be composed of valid tokens](https://sqlite.org/src/doc/trunk/doc/lemon.html#:~:text=4.4.25), so there can be `@`s and `$`s in the middle of the argument because they will parse to `BIND_PARAMETER`s, but there can't be `!`s unless they are quoted.

With this pull request, all currently opened issues with the sqlite tag can now be closed.

### DIFF
--- a/sql/sqlite/SQLiteParser.g4
+++ b/sql/sqlite/SQLiteParser.g4
@@ -578,10 +578,17 @@ filename
 ;
 
 module_argument
-    : (
-        ~(OPEN_PAR | CLOSE_PAR | UNEXPECTED_CHAR)
-        | OPEN_PAR module_argument CLOSE_PAR
-    )*?
+    : module_argument_outer*
+;
+
+module_argument_outer
+    : ~(OPEN_PAR | CLOSE_PAR | UNEXPECTED_CHAR | COMMA)
+    | OPEN_PAR module_argument_inner* CLOSE_PAR
+;
+
+module_argument_inner
+    : ~(OPEN_PAR | CLOSE_PAR | UNEXPECTED_CHAR)
+    | OPEN_PAR module_argument_inner* CLOSE_PAR
 ;
 
 // Only some keywords are allowed as identifiers.


michaelmcleodnz:
I've just noticed that [one of the GitHub action checks](https://github.com/antlr/grammars-v4/actions/runs/19987124455/job/57322697656?pr=4700#step:15:1260) suggests this grammar be reformatted.

I should point out that I've been adding the additional `antlr-format` rules:
```json
{ "main": { "alignColons": "hanging" }, "lexer": { "alignColons": "trailing" } }
```
to keep the formatting consistent with the state I found it in, without altering the inline formatting options that the [wiki](https://github.com/antlr/grammars-v4/wiki) says shouldn't be changed in a pull request.

I'd be happy to either reformat the file or update the formatting options.

kaby76:
Looks fabulous. The formatting change is fine. The format check in the build is just a "heads up", not an error, and the wiki could be loosened up instead of saying you can't change the formatting, especially if you are using the formatter. So, if there's formatting options set in the .g4, then just use that.

kaby76:
For `module_arguments`, the problem is that it's in the context of enclosing parentheses with comma separations. So, input `CREATE VIRTUAL TABLE f USING module_name(abc()d(te(s)t)f,gh`!`j%$m@p);` can be parsed two ways.
```
$ trparse --ambig xsql2.sql | trtree -a
CSharp 0 sql2.sql success 0.0380996
sql2.sql.d=254.a=1: (parse (sql_stmt_list (sql_stmt (create_virtual_table_stmt (CREATE_ "CREATE") (VIRTUAL_ "VIRTUAL") (TABLE_ "TABLE") (table_name (any_name (IDENTIFIER "f"))) (USING_ "USING") (module_name (any_name (IDENTIFIER "module_name"))) (OPEN_PAR "(") (module_argument (IDENTIFIER "abc") (OPEN_PAR "(") (module_argument) (CLOSE_PAR ")") (IDENTIFIER "d") (OPEN_PAR "(") (module_argument (IDENTIFIER "te") (OPEN_PAR "(") (module_argument (IDENTIFIER "s")) (CLOSE_PAR ")") (IDENTIFIER "t")) (CLOSE_PAR ")") (IDENTIFIER "f")) (COMMA ",") (module_argument (IDENTIFIER "gh") (IDENTIFIER "`!`") (IDENTIFIER "j") (MOD "%") (BIND_PARAMETER "$m") (BIND_PARAMETER "@p")) (CLOSE_PAR ")"))) (SCOL ";")) (EOF ""))
sql2.sql.d=254.a=2: (parse (sql_stmt_list (sql_stmt (create_virtual_table_stmt (CREATE_ "CREATE") (VIRTUAL_ "VIRTUAL") (TABLE_ "TABLE") (table_name (any_name (IDENTIFIER "f"))) (USING_ "USING") (module_name (any_name (IDENTIFIER "module_name"))) (OPEN_PAR "(") (module_argument (IDENTIFIER "abc") (OPEN_PAR "(") (module_argument) (CLOSE_PAR ")") (IDENTIFIER "d") (OPEN_PAR "(") (module_argument (IDENTIFIER "te") (OPEN_PAR "(") (module_argument (IDENTIFIER "s")) (CLOSE_PAR ")") (IDENTIFIER "t")) (CLOSE_PAR ")") (IDENTIFIER "f") (COMMA ",") (IDENTIFIER "gh") (IDENTIFIER "`!`") (IDENTIFIER "j") (MOD "%") (BIND_PARAMETER "$m") (BIND_PARAMETER "@p")) (CLOSE_PAR ")"))) (SCOL ";")) (EOF ""))

12/06-09:35:50 ~/issues/g4-2151/sql/sqlite/Generated-CSharp
```
It might be that what they say in the text isn't quite right. Perhaps you can't have commas in the name either. Or, it may be that the parser has some type checking to count the number of arguments, and disallow certain parses based on the count it is expecting.

michaelmcleodnz:
That's got it! Thanks for your help here.

I was about to comment that the issue is `module_argument` *does* allow commas, provided that they're inside balanced parentheses, because column definitions are meant to be valid arguments among other things (module functions can be created by the user, so SQLite just passes the arguments through raw).

But then I realised that I can specfiy explicitly that the "outer" level of the module_argument can't contain commas, while any "inner" part inside parentheses can, and that fixed the ambiguity while still parsing everything as intended.

(SQLite's internal implementation of `module_argument` is called `vtabarg` and can be found [in this section of parse.y](https://github.com/sqlite/sqlite/blob/6cc0da6e7d5427b18e48e906ddc1d48f7f5c26c3/src/parse.y#L1925), where `ANY` is defined [here](https://github.com/sqlite/sqlite/blob/6cc0da6e7d5427b18e48e906ddc1d48f7f5c26c3/src/parse.y#L296) as the only time the `%wildcard` rule is used. `%wildcard` is explained [here](https://sqlite.org/src/doc/trunk/doc/lemon.html#:~:text=4.4.25), and there are [two](https://sqlite.org/lang_createvtab.html) [pages](https://sqlite.org/vtab.html) where the intent of `module_argument` is explained in the docs. In practice, I was able to confirm that the examples I added in the test cases are parsed as one would expect.)

teverett:
@michaelmcleodnz  thanks!

PR #4702, opened 2025-12-08
[vhdl2008] - Reintroduce example files and grammar improvements

LucasMoss:
* Added literals package to test string, real and integer literals.

 * Added support for generic types, as used in the generic_mux example file.

 * Tidied generate statement rules.

 * Tidied name rule.

 * Reintroduced example files for VHDL2008 as plain files rather than symlinks.


teverett:
@LucasMoss  thanks!

Issue #4717, opened 2025-12-29
[lua] The grammar for Lua still contains keyword "continue".


kaby76:
See https://github.com/antlr/grammars-v4/commit/ff6ef1e876060e9e66535e88879e1ccd045647e4#r173636608. The "continue" keyword is a remnant of the (incorrect) merge of grammar rules for Luau into Lua. `Luau != Lua!!` In fact, the ["official" grammar for Luau](https://luau.org/grammar) contains many errors. At some point, we could have a grammar for Luau, but it should be a separate grammar placed under a directory of various flavors of Lua.

PR #4720, opened 2025-12-29
Fixes #4719 - remove non-compliant `continue` keyword

### LINKED ISSUE #4719: [Lua] Invalid reserved keyword continue prevents parsing valid Lua 5.1+ code
The current Lua grammar includes `CONTINUE` as a reserved keyword. However, `continue` is not a reserved word in standard Lua (including 5.1, 5.2, 5.3, and 5.4). While some custom Lua dialects (like Luau or GMod Lua) support it, its inclusion in the base grammar breaks compatibility with standard Lua code where `continue` is a valid identifier (e.g., as a table key or variable name).

### Steps to reproduce

Attempting to parse the following valid Lua code fails:

```lua
storage = { }
storage.continue = true
return storage['continue']
```

**Actual Lexer/Parser behavior**: The lexer matches continue as a `CONTINUE` token instead of a `NAME`. Because the parser expects `NAME` after a dot (.), it throws an error: `Lua: no viable alternative at input 'storage.continue'`

### Root Cause

In `LuaLexer.g4`:

``` antlr
CONTINUE : 'continue';
```

### Suggested Fix

Remove the `CONTINUE` token from `LuaLexer.g4`.

pzurowski:
fixes #4719

### DIFF
--- a/lua/LuaLexer.g4
+++ b/lua/LuaLexer.g4
@@ -32,7 +32,6 @@ LOCAL    : 'local';
 LT       : '<';
 GT       : '>';
 RETURN   : 'return';
-CONTINUE : 'continue';
 CC       : '::';
 NIL      : 'nil';
 FALSE    : 'false';


pzurowski:
duplicate of #4717

PR #4732, opened 2026-01-28
[PlSql] Support relaxed and extended partition list syntax

Naros:
Upstream port of changes from https://github.com/debezium/debezium/pull/7036


teverett:
@Naros thanks!

Issue #4733, opened 2026-01-28
golang parser crashes with java.lang.ArrayIndexOutOfBoundsException


andrewstein:
I am using the `golang/GoLexer.g4`, `golang/GoParser.g4`  and `golang/Java/GoParserBase.java` with the following code snippet:
```
        import ("..")
```

The `GoParser::importDecl` method crashes with the following stack trace:
```
java.lang.ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 0

	at org.antlr.parser.golang.GoParserBase.addImportSpec(GoParserBase.java:66)
	at org.antlr.parser.golang.GoParser.importSpec(GoParser.java:549)
	at org.antlr.parser.golang.GoParser.importDecl(GoParser.java:471)
```

andrewstein:
The line that is problematic in `GoParserBase.java` is
```
            String fileName = fileArr[fileArr.length - 1];
```
where the `fileArr` is empty.

kaby76:
I fixed the base classes for the grammar to handle all possible types of "import" statements. The example is still illegal because there's no "package" statement. https://github.com/antlr/grammars-v4/pull/4736

PR #4736, opened 2026-01-30
[go] Fix for #4733: crash in symbol table set up.

kaby76:
This is a fix for #4733 where there was a crash with input "import (..)". While the input is invalid (a package statement is required), and "import (..)" is no longer valid code because of Go modules, the code did not handle the various forms of "import". I added an example to test the code and updated all the ports. Also, `--debug` on the command line is now recognized. The Antlr4ng template was updated to fix a crash when reporting errors.


teverett:
@kaby76 thanks!

andrewstein:
Yes, I did simplify it. The original did have a package statement, several imports and the a lot of code. I was trying to report the minimal reproduction of the issue.

Issue #4737, opened 2026-01-31
[c] Minor problem in using `trparse -i 'void (*(*f[])())();'`


kaby76:
I was reading a [post in Reddit](https://www.reddit.com/r/Compilers/comments/1qqy9wm/understanding_c_declarators_by_writing_a_minimal/) for parsing and understanding declarations like `void (*(*f[])())();`. I tried to do that using trparse, but I noticed that I'm getting an empty tree. This is due to a bug in the code that reads command-line input strings. The fix is very simple, e.g., for CSharp:
```
diff --git a/c/CSharp/CLexerBase.cs b/c/CSharp/CLexerBase.cs
index b35100c57..3f20bbe97 100644
--- a/c/CSharp/CLexerBase.cs
+++ b/c/CSharp/CLexerBase.cs
@@ -50,7 +50,7 @@ public abstract class CLexerBase : Lexer
             File.WriteAllText(output_name, x1);
             return CharStreams.fromString(x1);
         }
-        File.WriteAllText(output_name, x1);
+        if (source_name == "stdin.c") File.WriteAllText(source_name, x1);

         if (gcc)
         {
```
Essentially, when a string input is used, the string should be output to "stdin.c", rather than "stdin.c.p", because the preprocessor needs to be applied to the string first. If stdin.c is empty, the parse is for an empty file. The fix should be applied across targets.

PR #4738, opened 2026-01-31
[c] Fix stdin.c handling for string input to driver program

kaby76:
Fix for #4737. Only write input to stdin.c when the source comes from a string (not a file). Previously, temp files with timestamps were created unnecessarily.


teverett:
@kaby76  thanks!

Issue #4734, opened 2026-01-29
Broken link in documentation


YamadaHideki:
In the documentation `https://github.com/antlr/grammars-v4/tree/master/vba/vba_cc` there is a broken link
`https://msopenspecs.azureedge.net/files/MS-VBAL/%5bMS-VBAL%5d.pdf`

<img width="746" height="211" alt="Image" src="https://github.com/user-attachments/assets/ccdf6877-be34-42bf-b308-d2bca7c594db" />

mlorek:
is this it? https://learn.microsoft.com/en-us/dotnet/visual-basic/programming-guide/program-structure/conditional-compilation

YamadaHideki:
I think so. I don't know what was in that link.

kaby76:
The way you fix these kinds of things is to use the [waybackmachine](web.archive.org), then analyze what is going on. The .pdf is supposed to be the specification of the language. I would replace it with current web doc (https://learn.microsoft.com/en-us/openspecs/microsoft_general_purpose_programming_languages/ms-vbal/d5418146-0bd2-45eb-9c7a-fd9502722c74) and mention what version the grammar is trying to implement. You may add a link to the web.archive.org as an alternative or replacement. 

Grammars should say exactly what version it is implementing because it's always unclear what it should be supporting. 

We don't have a script that verifies web links. I'll have Claude write something and add it to the builds once a week to do the checking.

PR #4740, opened 2026-02-01
[build] Fix for detecting stale http links.

kaby76:
This is a fix for #4734. I had Claude write a Bash script that detects stale links. *(It took a few iterations to find and fix a few problems, but the code is pretty robust now. I am sold: AI agents are a major efficiency multiplier.)*

The check is done per PR, per grammar, and is also done weekly. The script is in file `_scripts/find-stale-links.sh` and includes the `--replace` option, which modifies stale links and replaces them with web.archive.org links.

A couple of bad links are fixed here, which I did for testing. I'll have separate PRs to fix the remaining bad links later.

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -8,7 +8,9 @@
       "Bash(echo:*)",
       "Bash(curl:*)",
       "Bash(git add:*)",
-      "Bash(git commit:*)"
+      "Bash(git commit:*)",
+      "WebFetch(domain:productionresultssa18.blob.core.windows.net)",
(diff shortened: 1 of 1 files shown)


kaby76:
99 stale links detected: 
[script.output.txt](https://github.com/user-attachments/files/25008137/script.output.txt)

kaby76:
Still getting timeouts with 30s wait on http links. May need to run the downloads in parallel so that it doesn't take too long with many links, and bump up to 1 minute wait.

teverett:
@kaby76  thanks this is cool!

Issue #4739, opened 2026-02-01
[golang] https://github.com/mna/pigeon/blob/096f8b447491463228b465bc82166c1a4067bbfa/bootstrap/cmd/bootstrap-pigeon/bootstrap_pigeon.go#L1939 does not parse.


kaby76:
See https://github.com/mna/pigeon/blob/096f8b447491463228b465bc82166c1a4067bbfa/bootstrap/cmd/bootstrap-pigeon/bootstrap_pigeon.go#L1939

This fails to parse because the parser makes the wrong parse choice for an ambiguity, but then runs into a failed predicate afterwards. The grammar needs additional predicates to make the right choice in an ambiguity with a `make()` call.

PR #4741, opened 2026-02-03
[golang] Fix for #4739: can't parse certain `make()` calls.

kaby76:
This PR is a fix for #4739, where certain calls to `make()` were not parsing. Parsing the test without disambiguating predicates revealed we were missing a disambiguating predicate. Predicates are needed because the official EBNF for the language is not context-free.

This change adds semantic predicates to the `arguments` rule, specifically at the choice between `Type` or `ExpressionList` (in the official EBNF).

The code was written using Claude, taking after how I implemented context-sensitivity in the grammar for C. Claude is able to use Trash tools to automatically find and fix grammar ambiguity.


teverett:
@kaby76  thanks!

PR #4748, opened 2026-02-10
[PlSql] Support multi-columned partition lists

Naros:
Based on https://github.com/debezium/debezium/pull/7069


kaby76:
@Naros Don't worry about the state links check failure. I'm going through the entire repo cleaning up these bad links.

@teverett All set.

Issue #4752, opened 2026-02-14
[ada2012] The Ada 2012 grammar does not parse many files of the Ada Conformity Assessment Test Suite (ACATS)


kaby76:
Ada is an old programming language, but it is seeing a resurgence in popularity. https://thenewstack.io/2025-the-year-of-the-return-of-the-ada-programming-language/. The grammars-v4 repo has grammars for Ada, but they are not very good.

Ada compilers have a conformance test suite: http://www.ada-auth.org/acats.html. The ada2012 grammar does not parse most of the *.ada files.

It looks like there were transcription errors due to scraping the grammar via eyeball. E.g., missing ';' in `subtype_declaration`.

kaby76:
Yeah, there are a lot of errors in the grammar. We can't transcribe grammars by eyeball. It leads to situations like this. Using Claude is invaluable in finding these problems.

PR #4753, opened 2026-02-15
[Ada 2012] Major improvements in grammar quality.

kaby76:
@teverett This PR is good to go.

This is a fix for #4752. It improves the grammar by adding the Ada Conformity Assessment Test Suite (ACATS). Many errors were identified and corrected, resulting in a much better-quality grammar.

***This PR does not address the ambiguity in the EBNF, which is inherent in the grammar in the Spec. Most .ada files now parse, but very slowly. I will be implementing a second PR to address major ambiguity and add testing of the ACATS.***

## What was fixed

 - Add missing SEMI terminators to 11 parser rules: subtype_declaration, generic_instantiation (3 alternatives), record_representation_clause, number_declaration, abstract_subprogram_declaration, goto_statement, procedure_call_statement (1st alternative), delay_until_statement, delay_relative_statement, abort_statement, formal_object_declaration (2nd alternative)
  - Fix record_representation_clause: remove incorrect optional marker on END RECORD
  - Fix subtype_mark to accept dotted names (identifier (DOT identifier)*) instead of just identifier, allowing qualified names like PACK2.ACC1
  - Fix qualified_expression: remove extra parentheses around aggregate alternative, matching Ada RM definition
  - Fix formal_signed_integer_type_definition: use RANGE_ token instead of range parser rule
  - Merge '!' into VL token to support Ada 83 obsolescent replacement character (RM J.2)
  - Add lexer predicate IsCharLiteralAllowed() to disambiguate character literals from attribute ticks, using AdaLexerBase superclass to track previous token
  - Add AdaLexerBase implementations for all targets: CSharp, Java, Python3, Dart, JavaScript, TypeScript, Antlr4ng, Cpp, Go
  - Add transformGrammar.py for Python3, Antlr4ng, Cpp, and Go targets
  - Add the ACATS.

## AdaLexer.g4 changes
  - Two-pass pragma parsing: Added channels { PRAGMA_CHANNEL } and a PRAGMA_MODE lexer mode. When the lexer sees pragma, it emits the token on PRAGMA_CHANNEL and enters PRAGMA_MODE, where all subsequent tokens through ; are also emitted on PRAGMA_CHANNEL with type() mappings to existing token types. This makes pragmas invisible to the main parser while preserving them for second-pass parsing.

  - Ada 83 obsolescent forms (RM Annex J.2):
    - BASED_LITERAL: Accept : as alternative delimiter to # (e.g., 2:10: equivalent to 2#10#)
    - STRING_LITERAL_: Accept % as alternative string delimiter with %% for embedded percent
    - CHARACTER_LITERAL_: Allow apostrophe and backslash as the enclosed character (~['\\\r\n] → ~[\r\n]), matching RM 2.6 graphic_character definition

## AdaParser.g4 changes
  - Two-pass pragma parsing: Added superClass = AdaParserBase to options. Added {this.ParsePragmas();} action in the compilation rule.
  - Added pragmaRule and pragma_argument_association rules per RM 2.8. The ParsePragmas() method (in the target-specific AdaParserBase files) collects pragma tokens from PRAGMA_CHANNEL and parses each one individually using pragmaRule.
  - Attribute designator fix: Added keyword tokens ACCESS, DELTA, DIGITS, MOD as alternatives in attribute_designator. Previously only the title-case-only tokens (ACCESS__, DELTA__, etc.) were accepted, which failed for all-uppercase attribute references like T'DELTA.

Almost all the work was done via Claude.


kaby76:
The performance of the grammar is terrible. Many .ada files in ACATS are really quite slow.

## Ambiguity

Input:  `( false and true )` via `primary`. https://github.com/antlr/grammars-v4/blob/6dd67a28ffe4d3a3e7529d96976337ab42cc52d6/ada/ada2012/examples/pkg1.adb#L13C23-L13C39

### The Problem

Ada RM 4.4 defines `primary` with (among others) these two alternatives:

  1. `aggregate` — which includes record_aggregate → '(' record_component_association_list ')'
  2. `'(' expression ')'` — a parenthesized expression

  For input like `(false and true)`, both paths match:

  - Path 1: primary → aggregate → record_aggregate → ( record_component_association_list ) — where the list is a single
  record_component_association containing just an expression (false and true)
  - Path 2: primary → ( expression ) — directly parenthesizing the expression false and true

  These consume exactly the same tokens, producing a genuine syntactic ambiguity.

### How Ada Resolves It

The Ada RM resolves this semantically, not syntactically. The compiler uses type context to determine whether (expr) is a parenthesized expression or a one-element aggregate. For example:

  - `X : Integer := (1 + 2);` — parenthesized expression (Integer is not an aggregate type)
  - `X : Arr := (1 + 2);` — aggregate with one positional component (if Arr is an array type)

teverett:
@kaby76  thanks!

PR #4755, opened 2026-02-16
[ada2012] Implement disambiguating semantics

### LINKED ISSUE #4754: [ada2012] Ambiguity in the grammar and poor performance.
The grammar's performance is terrible. Many .ada files in ACATS are really quite slow.

## Ambiguity

Input:  `( false and true )` via `primary`. https://github.com/antlr/grammars-v4/blob/6dd67a28ffe4d3a3e7529d96976337ab42cc52d6/ada/ada2012/examples/pkg1.adb#L13C23-L13C39

### The Problem

Ada RM 4.4 defines `primary` with (among others) these two alternatives:

  1. `aggregate` — which includes record_aggregate → '(' record_component_association_list ')'
  2. `'(' expression ')'` — a parenthesized expression

  For input like `(false and true)`, both paths match:

  - Path 1: primary → aggregate → record_aggregate → ( record_component_association_list ) — where the list is a single
  record_component_association containing just an expression (false and true)
  - Path 2: primary → ( expression ) — directly parenthesizing the expression false and true

  These consume exactly the same tokens, producing a genuine syntactic ambiguity.

### How Ada Resolves It

The Ada RM resolves this semantically, not syntactically. The compiler uses type context to determine whether (expr) is a parenthesized expression or a one-element aggregate. For example:

  - `X : Integer := (1 + 2);` — parenthesized expression (Integer is not an aggregate type)
  - `X : Arr := (1 + 2);` — aggregate with one positional component (if Arr is an array type)

kaby76:
This PR fixes #4754, addressing (partially) the inherent syntactic ambiguity in the grammar. As a first step to address ambiguity, this PR adds disambiguating predicates. To handle `with` statements in Ada, the parser now reads package .ads files. Options are provided to specify the search path for Ada packages, and debug the symbol table constructed for disambiguation.

Almost all ports are supported. I didn't bother with PHP because it has serious bugs.

A new readme.md file has been written for a full description of the grammar. It's modeled after the one for the [C grammar](https://github.com/antlr/grammars-v4/tree/master/c).

### DIFF
--- a/ada/ada2012/readme.md
+++ b/ada/ada2012/readme.md
@@ -55,7 +55,7 @@ The parser base class implements the following command-line options:
 
 <img src="./times.svg">
 
-Runtime of `examples/*.c` on AMD Ryzen 7 2700 Eight-Core Processor; 16GB DDR4;
+Runtime of `examples/acats/a/*.ada` on AMD Ryzen 7 2700 Eight-Core Processor; 16GB DDR4;
 Samsung SSD 990 EVO Plus 2TB;
 Windows: Version 10.0.26200.7623 (this is a Windows 11 Insider Preview build); 
 .NET SDK: 10.0.102. Sample size 20.


teverett:
@kaby76  thanks!

PR #4763, opened 2026-02-21
[c] Fix #4762: various ambiguities.

### LINKED ISSUE #4762: [c] Ambiguities in the C grammar.
Attached are three single-point ambiguities in the C grammar for three examples.

[ambiguity-1065.c.txt](https://github.com/user-attachments/files/25457717/ambiguity-1065.c.txt)
[ambiguity-FunctionPointer.c.txt](https://github.com/user-attachments/files/25457716/ambiguity-FunctionPointer.c.txt)
[ambiguity-ll.c.txt](https://github.com/user-attachments/files/25457718/ambiguity-ll.c.txt)

kaby76:
Fixes for #4762, where there are ambiguities in examples/{ll,FunctionPointer,1065}.c.

* Added semantic predicate support method IsGnuAttributeBeforeDeclarator() to all CParserBase targets and use it in IsDeclarationSpecifier() to prevent __attribute__((...)) from being treated as a declaration specifier when it precedes a declarator (identifier, *, or parenthesized declarator). This eliminates the AdaptivePredict ambiguity without grammar changes and without introducing parse errors.
* Add in GCC keyword extensions.
* Clean up badly refactored `unaryExpression`.
* Added semantic predicates and method `IsSomethingOfTypeName()` to support correct alts in `unaryExpression`.
* Added semantic predicate and method `IsInitDeclaratorList()` to `declaration`.
* Added semantic predicate and method `IsTypedefName()` to `typedefName`. Note: it is not at the beginning of a decision state, but double-checks the semantics here to make sure nothing bad happens.
* Add in base "int k" for lookahead in semantic predicates.

_NB: We need to add checks on ambiguity in cleaned-up examples. trparse takes `--ambig`, trquery does not do the parse. Perhaps, we need a combined trparse/trquery to test ambiguities._

### DIFF
--- a/c/CParser.g4
+++ b/c/CParser.g4
@@ -482,7 +482,7 @@ alignmentSpecifier
 // This rule is basically what was implemented in the GCC compiler.
 // https://github.com/gcc-mirror/gcc/blob/f5cda36f16d447198c1e00b191d720b6f4a02876/gcc/c/c-parser.cc#L4975-L4995
 declarator
-    : (gnuAttribute? pointer declarationSpecifiers?)* (gnuAttribute* directDeclarator gccDeclaratorExtension*) {this.EnterDeclaration();}
+    : (gnuAttribute? pointer /* declarationSpecifiers? */)* (gnuAttribute* directDeclarator gccDeclaratorExtension*) {this.EnterDeclaration();}
     ;
 
 // ISO C: direct-declarator (6.7.7.1)
--- a/c/CSharp/CParserBase.cs
+++ b/c/CSharp/CParserBase.cs
@@ -597,7 +597,13 @@ public abstract class CParserBase : Parser
     public bool IsInitDeclaratorList()
     {
         // Cannot be initDeclaratorList if the first thing is a type.
-        // Types need to go to preceeding declarationSpecifiers.
+	// Types need to go to preceeding declarationSpecifiers.
+	// A declarator must start with:
+	//  identifier
+	//  *
+	//  (
+	// It cannot start with __attribute__.
+	
         if (no_semantics.Contains("IsInitDeclaratorList")) return true;
         var ts = this.InputStream as CommonTokenStream;
         var lt1 = (this.InputStream as CommonTokenStream).LT(1);
@@ -609,7 +615,9 @@ public abstract class CParserBase : Parser
         {
             result = true;
         }
-        else if (resolved.Classification.Contains(TypeClassification.TypeQualifier_) || resolved.Classification.Contains(TypeClassification.TypeSpecifier_))
+        else if (resolved.Classification.Contains(TypeClassification.TypeQualifier_)
+            || resolved.Classification.Contains(TypeClassification.TypeSpecifier_)
+            || text == "__attribute__")
             result = false;
         else
             result = true;
--- a/c/examples/1065.c
+++ b/c/examples/1065.c
@@ -1,3 +1,7 @@
+// NB: GCC preprocessor changes __cdecl to __attribute__((__cdecl__))
+// under the covers, even though there are no #include files anywhere.
+// VSC and CLANG do not alter.
+
 typedef int UINTN;
 void * __cdecl  LibPcdGetPtr (UINTN TokenNumber);
 void __cdecl LibPcdGetPtr (UINTN TokenNumber);


kaby76:
```
❯ Claude, please read C:\msys64\home\Kenne\issues\g4-4762-kaby76\c\Generated-CSharp\xx.c and
  C:\msys64\home\Kenne\issues\g4-4762-kaby76\c\examples\FunctionPointer.c.

● Read 2 files (ctrl+o to expand)

● Here are both files:

  Generated-CSharp/xx.c (1 line):
  struct s1 { char x[1]; } s1 = { "0" };
  This declares a struct type s1 with a char x[1] field, then declares a variable also named s1 and initializes it with { "0" }. It's
   the test case for the IsInitDeclaratorList fix — s1 after the struct definition must be parsed as an initDeclaratorList (variable
  declaration), not as a continuation of declarationSpecifiers.

  examples/FunctionPointer.c (32 lines):
  Three typedef declarations of function-pointer types (f1, f2, f3). The tricky case is f3, which uses MyStruct (a typedef'd name) as
   the return type, so the parser must correctly distinguish MyStruct in declarationSpecifiers from a potential initDeclaratorList.
  This is the core ambiguity that IsInitDeclaratorList resolves.

  Is there something specific you'd like me to do with these files?

❯ FunctionPointer.c parses but is ambiguous if "--no-semantics=IsInitDeclaratorList" is not specified. This is because I
  inserted a semantic predicate IsInitDeclaratorList() to take care that MyStruct on line 26 is parsed under
  declarationSpecifier, not under initDeclaratorList. However, the semantic predicate brakes xx.c because the second "s2" is a
  variable being declared, not a type. What do you recommend I do the fix the problem?

● The root cause is in EnterDeclaration(). When the outer declaration struct s1 { char x[1]; } s1 = ... is processed, this block
  runs:

  var sous = ds.typeSpecifier()?.structOrUnionSpecifier();
  if (sous != null) {
      var idToken = sous.Identifier()?.Symbol;  // ← registers "s1" as TypeSpecifier_
      _st.Define(new Symbol() { ..., Classification = TypeSpecifier_ });
  }

  That registers the struct tag s1 as TypeSpecifier_. So when IsInitDeclaratorList() later looks up the variable s1 it finds it in
  the symbol table as TypeSpecifier_ and returns false, blocking the initDeclaratorList branch.

  The core issue: struct/union/enum tags (the s1 in struct s1 { }) occupy a separate namespace in C from ordinary identifiers. A tag
  can only be referenced with the struct/union/enum keyword prefix — s1 alone is never a valid standalone type name. Registering it
  as TypeSpecifier_ conflates the two namespaces.

  My recommendation — Option A (simplest): Simply do not register struct/union/enum tags from the structOrUnionSpecifier inside
  EnterDeclaration(). The struct keyword itself handles disambiguation via IsStructOrUnionSpecifier(), so the tag never needs to be
  in the ordinary-identifier symbol table.

  This leaves FunctionPointer.c correct because MyStruct is registered as TypeSpecifier_ through the typedef declarator path in
  EnterDeclaration() (the initDeclaratorList branch, not the struct-tag branch), which is the correct mechanism.

  Option B (more precise but more work): Add a StructTag_ (and EnumTag_) value to TypeClassification, register struct tags with that
  new value instead of TypeSpecifier_, and leave IsInitDeclaratorList() as-is since it only blocks on TypeQualifier_ and
  TypeSpecifier_.

  Option A requires removing one block from EnterDeclaration() in all targets. Option B requires adding a new enum value to
  TypeClassification in all targets plus updating the registration code.

  I'd go with Option A — it correctly reflects C's namespace rules and is the minimal change. Shall I apply it?

✻ Sautéed for 4m 18s

❯ Is there any code in C where "s1" the struct tag can be used anywhere? Examples please?

● In standard C, no — a struct/union/enum tag name can never be used as a standalone identifier. Tags occupy a completely separate
  namespace from ordinary identifiers. You always need the keyword prefix:

  struct Point { int x, y; };
  struct Point p1;   // OK — must use "struct" keyword
  // Point p2;       // ERROR in C (valid in C++, not C)

  enum Color { RED, GREEN, BLUE };
  enum Color c1;     // OK — must use "enum" keyword
  // Color c2;       // ERROR in C
  int n = RED;       // OK — enum *constants* are in the ordinary namespace, but not the tag

  The only case where the same name as a struct tag works without a prefix is when it is also a typedef name — but that's a separate
  registration:

  typedef struct s1 { int x; } s1;   // "s1" is both a struct tag AND a typedef name
  s1 var;            // OK — but "s1" here is the typedef, not the struct tag

  In that case, s1 as a typedef gets registered through the initDeclaratorList path in EnterDeclaration() (the typedef declarator),
  independently of the struct tag. So removing the struct tag registration has no effect on this case.

  Conclusion: Option A is completely safe. Struct/union/enum tags need not appear in the ordinary-identifier symbol table at all. The
   struct/union/enum keyword itself is always present when a tag is referenced, and the grammar handles it through
  IsStructOrUnionSpecifier() / IsEnumSpecifier() without consulting the symbol table.
```

kaby76:
1065.c now fixed.

teverett:
@kaby76  thanks!

PR #4767, opened 2026-02-23
Fix: Make PlantUML more lenient

david-xander:
## Summary

Relax the grammar so it matches the level of permissiveness accepted by PlantUML.

The current grammar is stricter than the actual language and rejects inputs that PlantUML parses successfully. It should not prematurely restrict inputs that are valid according to PlantUML.

The parser should focus on recognizing the high-level structure of the diagram while remaining tolerant of loosely structured content. 

Detailed interpretation (attributes, types, modifiers, etc.) should be performed in a later semantic analysis phase based on an AST rather than directly from the parse tree.

## Example

The following diagram is valid PlantUML and can be parsed using the online server:
https://www.plantuml.com/

```plantuml
@startuml

enum Position <<enum>> {
    Goalkeeper
    LeftBack
    ...
}

enum Recommendation <<enum example of allowed commentary>> {
    KeyPlayer
    FirstTeamPlayer
    ReserveTeamPlayer
    ProspectivePlayer
    NotRecommended
}

enum ScoutingStatus <<enum>> {
    LongListed
    ShortListed {another} example of allowed input here
    RecommendedForSigning
    OfferMade
}

enum ScoutKind <<enum>> {
    RegularScout
    HeadScout
}

abstract class Person {
    String FirstName
    String LastName
}

class Player {
    ScoutingStatus Status
}

class HeadCoach {}

class Director {}

class Scout {
    ScoutKind ScoutKind
}

class Club {}

class Offer {
    Int Value
}

class ScoutingAssignment {}

class ScoutReport {
    {String} Pro
    [String] I can write whatever I want {String}
    Recommendation Recommendation
}

class PlayerProfile {
    Position Position
}

class PlayerAttribute {
    Name: String
    Int Value
}

Person <|-- Player
Person <|-- HeadCoach
Person <|-- Director
Person <|-- Scout

Director *--> "0..*" Offer
Club *--> "0..1" Director
Club *--> "0..1" HeadCoach
Club *--> "0..*" Player
Club *--> "0..*" Scout
Scout *--> "0..*" ScoutingAssignment
ScoutingAssignment *--> "0..1" ScoutReport 
HeadCoach *--> "0..*" PlayerProfile
PlayerProfile *--> "0..*" PlayerAttribute

Player "0..1" <--> "0..* another comment here" Offer
Player "0..1" <--> "0..1" PlayerProfile
Player "1..1" <--> "0..*" ScoutingAssignment
ScoutReport "0..1" <--> "0..1" ScoutReport : nextReport example comment
Player "1..1" <--> "0..*" ScoutingAssignment

@enduml
```

It produces the following diagram:

example

The example shows that the PlantUML domain-specific language allows additional text in several places where the existing ANTLR PlantUML grammar is too strict and fails to parse the input.

## Changes

* Make enum names more permissive (allow spaces and inline text).
* Allow arbitrary content inside class bodies.
* Relax parsing of association cardinalities and labels to allow free text.
* Introduce a lexer using ANTLR modes to tolerate free-form text in these contexts.

## Notes

The parser should focus on structural recognition. Detailed interpretation (attributes, types, etc.) should happen later using an AST.

This level of tolerance is also necessary for constructs embedding other syntaxes such as HTML or CSS.

### DIFF
--- a/plantUML/PlantUML.g4
+++ b/plantUML/PlantUML.g4
@@ -10,218 +10,87 @@
 * SPDX-License-Identifier: EPL-2.0
 * *****************************/
 
-grammar PlantUML;	
-	
-classDiagram
-  : '@startuml' cdelement* '@enduml' EOF
-  ;
 
-cdelement
-    : 'class' identifier stereotype? classDefinition?
-    | 'interface' identifier stereotype? classDefinition?
-    | 'abstract' 'class' identifier stereotype? classDefinition?
-    | 'abstract' identifier stereotype? classDefinition?
-    | 'entity' identifier stereotype? classDefinition?
-    | 'enum' identifier stereotype? enumDefinition?
-    | 'annotation' identifier
-    | inheritance
-    | association
-    | aggregation
-    | attribute
-    | composition
-    | dependency
-    | method
+parser grammar PlantUML;
+
+options { tokenVocab=PlantUMLLexer; }  // tokens from lexer file
+
+uml:
+    (NEWLINE | COMMENT)* STARTUML (IDENT)? (NEWLINE | class_dclr | enum_dclr | association_dclr | associative_class_dclr | COMMENT)* ENDUML (NEWLINE | COMMENT)*
     ;
 
-classDefinition
-    : '{' classElement* '}'
-    ; 
-
-classElement
-    : internalMethod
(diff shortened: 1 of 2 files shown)


teverett:
@david-xander  thanks!

PR #4769, opened 2026-02-25
[Snowflake]Add Hybrid Table and Index

MasterKuat:
Add support for hybrid table (https://docs.snowflake.com/en/sql-reference/sql/create-hybrid-table) and related INDEX statements (https://docs.snowflake.com/en/sql-reference/sql/create-index and so on)


teverett:
@MasterKuat  thanks

PR #4770, opened 2026-02-25
Implement support for Kotlin context parameters and explicit backing fields

KitsuneAlex:
This adds support for [context parameters](https://kotlinlang.org/docs/context-parameters.html) and [explicit backing fields](https://github.com/Kotlin/KEEP/blob/explicit-backing-fields/proposals/explicit-backing-fields.md) into the informal Kotlin grammar.

These are experimental features which may be subject to change, but having them in the informal grammar would greatly broaden the codebases which can be parsed using this grammar, since the aforementioned features are already being used quite extensively from what i've seen, especially context parameters.

### DIFF
--- a/kotlin/kotlin/KotlinParser.g4
+++ b/kotlin/kotlin/KotlinParser.g4
@@ -790,6 +790,7 @@ simpleIdentifier
     | ANNOTATION
     | BY
     | CATCH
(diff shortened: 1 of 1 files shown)


teverett:
@KitsuneAlex thanks!

Issue #4758, opened 2026-02-20
[PL/SQL] Exponential parse time on sum(column) + sum(column) + sum(column) + ...


mpeschel10:
Good afternoon, all,

I ran into a select statement in the wild with a bajillion sum statements added together. The parser was getting stuck at 100% CPU without terminating or giving an error. I'm using C# and redownloaded PlSqlParser.g4, PlSqlLexer.g4, PlSqlLexerBase.cs, and PlSqlParserBase.cs a few hours ago.

Here's an example statement that should reproduce the issue (It takes about a minute for me to parse after 17 SUM functions):
```SQL
SELECT
    SUM(COLUMN_NAME) + SUM(COLUMN_NAME) + SUM(COLUMN_NAME) + SUM(COLUMN_NAME) +
    SUM(COLUMN_NAME) + SUM(COLUMN_NAME) + SUM(COLUMN_NAME) + SUM(COLUMN_NAME) +
    SUM(COLUMN_NAME) + SUM(COLUMN_NAME) + SUM(COLUMN_NAME) + SUM(COLUMN_NAME) +
    SUM(COLUMN_NAME) + SUM(COLUMN_NAME) + SUM(COLUMN_NAME) + SUM(COLUMN_NAME) +
    SUM(COLUMN_NAME) + SUM(COLUMN_NAME) + SUM(COLUMN_NAME) + SUM(COLUMN_NAME) +
    SUM(COLUMN_NAME) + SUM(COLUMN_NAME) + SUM(COLUMN_NAME) + SUM(COLUMN_NAME) +
    SUM(COLUMN_NAME) + SUM(COLUMN_NAME) + SUM(COLUMN_NAME) + SUM(COLUMN_NAME)
FROM TABLE_NAME;
```

I asked Gemini for advice, and it said the issue is some overlap in the definition of concatenation and interval_expression:
```ANTLR4
concatenation
    : model_expression (AT (LOCAL | TIME ZONE concatenation) | interval_expression)? (
        ON OVERFLOW_ (TRUNCATE | ERROR)
    )?
    | concatenation op = DOUBLE_ASTERISK concatenation
    | concatenation op = (ASTERISK | SOLIDUS | MOD) concatenation
    | concatenation op = (PLUS_SIGN | MINUS_SIGN) concatenation
    | concatenation BAR BAR concatenation
    | concatenation COLLATE column_collation_name
    ;

interval_expression
[...]

PR #4774, opened 2026-02-27
[PlSql] Fix for #4758, slow parse.

kaby76:
Fix for #4758, extremely slow parse, which involves aggregate function ambiguity with IsNotNumericFunction predicate.

Add semantic predicate IsNotNumericFunction() to all target base classes to prevent SUM/COUNT/AVG/MIN/MAX/ROUND/LEAST/GREATEST followed by '(' from being parsed as atom or other_function instead of numeric_function_wrapper.

In addition to fixing ambiguity, the parse times are slightly faster.


teverett:
@kaby76  thanks!

Issue #4757, opened 2026-02-19
Missing Ada 2022


kaby76:
Officially Ada 2022 is the only Ada language right now ISO/IEC 8652:2023.

mlorek:
will add it at some point

kaby76:
Hi @mlorek , If you like, otherwise I was going to work on this this week. I just updated the ada2012 grammar with the TC1 grammar correction. We should add the corresponding conformance test suite for ada2022.

mlorek:
go for it

kaby76:
Fixed with https://github.com/antlr/grammars-v4/pull/4775. There's no official conformance test suite for Ada 2022 yet. Instead, examples written to test Ada 2022 were added and tested with `gnat -gnat2022 -gnatwa" for version 15, the latest free version. adacore.com implements a commercial compiler, version 26 for Ada 2022, but no free version.

Irvise:
Thank you for the work @kaby76, I just found it! Also the "paid" version of AdaCore's compiler is just the GCC/GNAT version with some bugs fixed and generally one release behind, so what you tested is what can be tested :)

Again, thanks! Best regards,
Fer

P.S: the ACATS testing suite for Ada 2022 should be coming soonish

PR #4775, opened 2026-03-01
[new grammar] ada2022 (#4757)

kaby76:
This PR adds an ANTLR4 grammar for Ada 2022, derived from the existing Ada 2012 grammar and extended to cover all language additions introduced in the http://www.ada-auth.org/standards/ada22.html.

### New language features supported

  - parallel loops and blocks — parallel for loops and parallel do ... and ... end do block statements
  - Target name @ — shorthand for the LHS in assignments (X := @ + 1;)
  - Bracket aggregates [...] — array and container aggregates using square brackets, including null aggregate []
  - Delta aggregates — (X with delta F => V) for record and array delta aggregates
  - Container aggregates — positional and named container aggregates with key => value associations
  - declare expressions — (declare ... begin expr) inline declarations in expressions
  - Iterator filters — when condition on for loop iterators
  - Reduction attributes — Arr'Reduce("+", 0) and related 'Reduce_Right, etc.
  - Global aspects — with Global => ... subprogram contracts
  - or use formal type defaults — optional default in generic formal type declarations
  - aspect_specification on additional constructs — parameters, discriminants, entry bodies, extended return objects

### Symbol table for disambiguation

Ada has inherent syntactic ambiguities (e.g., Foo'(...) as a qualified expression vs. an attribute reference; Foo(X) as a type conversion vs. a function call). The grammar includes a symbol table implemented across all target runtimes that tracks type definitions as they are parsed and uses semantic predicates (IsTypeName, IsAggregate) to resolve these ambiguities at parse time.

The symbol table supports:

  - Proper lexical scoping (push/pop on block and declarative region entry/exit)
  - Cross-file symbol resolution via with-clauses, following GNAT file naming conventions
  - Cycle detection and per-file caching of parsed packages

### Runtimes

Runtime support files (LexerBase, ParserBase, Symbol, SymbolTable, TypeClassification) are provided for: C#, C++, Dart, Go, Java, JavaScript, Python3, TypeScript, Antlr4ng

### Test examples

10 example .adb files in examples/ cover each major new feature: target names, delta aggregates, bracket aggregates, iterated components, parallel loops, declare expressions, iterator filters, null array aggregates, formal defaults, and global aspects.

  ---

### Notes on Implementation
This PR was written mostly by Claude. However, I did need to guide Claude to a good implementation. Claude was going to copy the entire test suite to ada2022, but there is no conformance test suite for Ada 2022--just reuse ada2012/examples, and add specific tests for Ada 2022 features. Claude was going to add an alternative to get around the `"| {this.IsAggregate()}? aggregate` problem with input `"[]"`. The proper solution was to fix `IsAggregate()`. The EBNF from the Spec for Ada 2022 introduces a new mutual left recursion. Claude was going to implement a wacky solution until I stepped in to teach it how to unfold a rule, remove parentheses, and reformat.


teverett:
@kaby76 thanks!

PR #4776, opened 2026-03-01
Fix #4725 function precedence in SQLite

### LINKED ISSUE #4725: [sqlite] fails to parse SELECT with a function
The query

```sql
select sql from sqlite_schema where lower(name) = lower('MyTable')
```

got error

unexpected error: 1:48: mismatched input '=' expecting <EOF>:

michaelmcleodnz:
Well, that was embarrassing. I was convinced I'd fixed everything in the last PR, but `expr_recursive` should have come *after* eveything else, not *before*. Just spotted this issue today.

By moving `expr_recursive`'s precedence, the issue is resolved, the automated tests still pass, and a manual inspection of the parse trees for the tests with `expr`s in them suggests that they are all parsing as intended, so I'm pretty sure everything should work as intended now.

Sorry for the regression!

### DIFF
--- a/sql/sqlite/SQLiteParser.g4
+++ b/sql/sqlite/SQLiteParser.g4
@@ -238,7 +238,8 @@ drop_stmt
 expr
     : literal_value
     | BIND_PARAMETER
-    | ((schema_name DOT)? table_name DOT)? column_name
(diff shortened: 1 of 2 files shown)


teverett:
@michaelmcleodnz thanks!

PR #4799, opened 2026-04-04
[csharp/v8-spec] Continuation of improvements.

kaby76:
This is a continuation on the improvements for the csharp/v8-spec/ grammar. The changes here refactor a few more rules to reduce max-k's. The parser now completes the test suite twice as fast as compared to the grammar from the specification. 

There are still quite a few performance issues, but I'm committing this group of changes as a checkpoint.

@teverett All set for merge. Ty.


teverett:
@kaby76  thanks!

Issue #4807, opened 2026-04-12
[PL/SQL] KEEP Clause


la1ty:
I'm using antlr4 to parse some analytical stored procedures, in which they use a special grammar called `KEEP clause`:

```sql
SELECT dept, MIN(salary) KEEP (DENSE_RANK FIRST ORDER BY hire_date)
FROM emp
GROUP BY dept;
```

Currently in grammar-v4 it returns an error:

```
line 2:13 no viable alternative at input 'MIN'
```

Does anyone know how to handle it? Thanks in advance.

la1ty:
After carefully examining the stored procedures, I found two more unsupported PL/SQL syntaxes:

- `MERGE INTO` a subquery: `MERGE INTO (SELECT 1 FROM DUAL) T1 USING (SELECT 2 FROM DUAL) T2 ON (1=1) WHEN MATCHED THEN UPDATE SET COL=1`
- CASE clause in a function: `UPDATE TABLE T SET COLUMN = ROUND(COLUMN / 10000 , CASE WHEN 1 = '2' THEN 6 ELSE 4 END)`

I found that sqlglot can handle these syntaxes without any problems. Is it possible to migrate sqlglot's parsing logic to antlr4?

la1ty:
Just create a small PR to mitigate this issue. 

I don't understand antlr's grammar, so the fix may not be correct.

PR #4808, opened 2026-04-12
Fix some plsql parsing errors in merge_statement and numeric_function grammar

la1ty:
A quick fix of #4807 .


teverett:
@la1ty  thanks!

PR #4814, opened 2026-04-14
Antlr upgrade

teverett:


kaby76:
@teverett 

This upgrade is incomplete. We get `"ANTLR Tool version 4.13.2 used for code generation does not match the current runtime version 4.11.1; ANTLR Runtime version 4.13.2 used for parser compilation does not match the current runtime version 4.11.1"`. This is because there hasn't been an update to the Maven Antlr4 tester for years.

[<antlr.version>4.13.2</antlr.version>](https://github.com/antlr/antlr4test-maven-plugin/blob/fbba99e1ecebe03681e1269a0a1fba95f75550c3/pom.xml#L41)

The tester has some open PRs, including https://github.com/antlr/antlr4test-maven-plugin/pull/81.

I just tried `mvn clean test` from the root, and there are some other problems, e.g., ada/ada2012/. I'll fix those since I probably broke all that.

Issue #4817, opened 2026-04-15
[build] "mvn clean test" grammar testing from root broken.


kaby76:
Looks like the Maven tester is broken with some recent changes (ada/ada2012/). And, for that PR, looks like the tester didn't work correctly, but reported "success" anyways. https://github.com/antlr/grammars-v4/actions/runs/22505376133/job/65202585977

We see the failure, but it's ignored. I'm not sure why. 

[x.txt](https://github.com/user-attachments/files/26745723/x.txt)

PR #4818, opened 2026-04-15
[build, selected grammars] Maven tester from root broken.

kaby76:
This is a fix for #4817, where a number of grammars were failing the Maven tester. There were actually several problems.

* The script _scripts/maven.sh wasn't failing for grammars that had parse failures.
* For several grammars, the pom.xml didn't parse the right input files in the examples/ directory, or the pom.xml file was missing. Significant changes were made to the plantUML grammar because the relation operator didn't work. I didn't do a detailed investigation, but just changedwhat Icould to get the input to parse. /ada2012, csharp/v7, csharp/v8-spec, eiffel, plantUML.

I tested this via "mvn clean test" in the root.


teverett:
@kaby76  thanks!

PR #4820, opened 2026-04-17
sql/snowflake: allow 2-argument form of LAG/LEAD

ivanovvmilen:
The Snowflake documentation states LAG/LEAD accept 1, 2, or 3 arguments
(expression, optional offset, optional default). The grammar previously
allowed only 1 or 3 by making the offset and default a single optional
group: `(COMMA expr COMMA expr)?`.

With 2 arguments (e.g. `LAG(x, 3)`), the `ranking_windowed_function`
rule does not match. The input still parses — LAG falls through to the
generic `function_call: object_name '(' expr_list? ')'` alternative and
the trailing `OVER (...)` is picked up separately by `expr over_clause`
— but the resulting parse tree is structurally wrong: the window
function is no longer represented as a `ranking_windowed_function`, and
the OVER clause is detached from the function call that semantically
owns it. Downstream tools that rely on recognizing window functions via
`ranking_windowed_function` miss these forms, and the detached OVER
clause can associate incorrectly inside binary expressions.

Change the trailing arguments from `(COMMA expr COMMA expr)?` to
`(COMMA expr (COMMA expr)?)?` so the default value is independently
optional. All three arities (1, 2, 3) now match `ranking_windowed_function`.

Reference: https://docs.snowflake.com/en/sql-reference/functions/lag

Added two example queries (one LAG, one LEAD, both 2-arg) to
sql/snowflake/examples/select.sql as documentation that these forms
are intentionally supported. Note that the existing CI only checks
that inputs parse, not which grammar rule matched, so these examples
pass with or without the fix — but they help future readers
understand the intended shape of the rule.


kaby76 (approved):

teverett:
@ivanovvmilen  thanks!

PR #4821, opened 2026-04-19
[build] Fix for #4647: get Antlr4 Tool .jar from github.

### LINKED ISSUE #4647: [Build] Maven Central unreliable in builds.

kaby76:
This PR fixes #4647, where builds would fail due to Maven Central rejecting download requests, typically during a weekday afternoon. To fix this, a copy of the antlr4-tools Python package was added to the repo, and the tool was modified to work against github.com instead of Maven Central. This is the best way to avoid download problems because Maven Central rejects download requests when it is heavily loaded, whereas GitHub does not. Besides, the Antlr.org download page itself points to GitHub, not Maven Central (https://www.antlr.org/download.html, last sentence on webpage). This change also eliminates the extra dependency on the external antlr4 Python wrapper tool. Builds will still fail when other servers are heavily loaded, but it's less likely to occur.


teverett:
@kaby76 thanks

Issue #4823, opened 2026-04-23
Release all grammars on maven central


mickroll:
Hi there,

great work!

Is there any reason for the `org.antlr.grammars` modules to not to be released on maven central?
The testing plugin is already released, so infrastructure for releases should already exist :) see #466

greets,
Michael

kaby76:
The [antlr4test-maven-plugin](https://github.com/antlr/antlr4test-maven-plugin) hasn't been updated in 4 years and is built against an old version of the Antlr4 Java library. That has to be fixed first. I can create a build artifact in the meantime, if you are interested.

PR #4824, opened 2026-04-23
[build] Add weekly workflow to produce combined all-grammars JAR

kaby76:
This PR adds a new GitHub Actions workflow (weekly-jar.yml) that runs every Monday and on workflow_dispatch, builds all grammars via "mvn clean test package -Pgrammarv4", then combines every module JAR into a single all-grammars-1.0-SNAPSHOT.jar using the new _scripts/assemble-all-grammars.sh script. https://github.com/antlr/grammars-v4/issues/4823

### DIFF
--- a/.github/workflows/weekly-jar.yml
+++ b/.github/workflows/weekly-jar.yml
@@ -1,0 +1,30 @@
+name: Weekly All-Grammars JAR
+
+on:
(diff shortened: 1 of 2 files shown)


teverett:
@kaby76  thanks.  This is cool!

PR #4825, opened 2026-04-27
Fix lexer rule for Turtle IRIREF

nglgzz:
Currently, in the Turtle grammar, the IRIREF definition doesn't match [the spec](https://www.w3.org/TR/turtle/#sec-grammar-grammar). One case I noticed is that IRIs containing `=` are considered invalid by the current antlr grammar, but they are considered valid in the spec.

This PR changes the rule to use an exclusion list rather than an allow list for the characters forming the IRIREF, matching what is in the spec:

```ebnf
IRIREF	::=	'<' ([^#x00-#x20<>"{}|^`\] | UCHAR)* '>' /* #x00=NULL #01-#x1F=control codes #x20=space */
```

Disclaimer: I have used copilot to help fix the issue, though I have manually double checked that the change actually fixes the original issue I had.

### DIFF
--- a/turtle/TURTLE.g4
+++ b/turtle/TURTLE.g4
@@ -148,10 +148,9 @@ PN_PREFIX
     : PN_CHARS_BASE ((PN_CHARS | '.')* PN_CHARS)?
     ;
 
-//IRIREF	        :	'<' (~(['\u0000'..'\u0020']|'<'|'>'|'"'|'{'|'}'|'|'|'^'|'`'|'\\') | UCHAR)* '>'; /* \u00=NULL #01-\u1F=control codes \u20=space */
-
+// \u0000=NULL \u0001-\u001F=control codes \u0020=space
 IRIREF
-    : '<' (PN_CHARS | '.' | ':' | '/' | '\\' | '#' | '@' | '%' | '&' | UCHAR)* '>'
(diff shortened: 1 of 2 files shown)


kaby76 (approved):
Please add a test input file .ttf to the `examples/` directory.

Copilot-written code is fine as long as it is backed up with a good test suite and checked. I've been using Claude Code to craft PRs for many months. I pair the CC-development against a GPT-5.2 review, which tends to catch problems before the final merge.

I've added a CLAUDE.md file to the repo to aide in automating general development of Antlr grammars. It also contains general instructions on how to fix ambiguity and max-k problems using the Trash Toolkit.

nglgzz:
Thank you @kaby76 for the quick review and for the pointer. 

I've added a file in the examples folder to test valid IRIs and ran the test.sh script at the root of the repo to verify that tests pass.

kaby76:
Looks fine. Thanks.

I'm adding to CLAUDE.md information on how to run the Maven Antlr4 Tester. "test.sh" is fine, but it tests every grammar; you're only interested in running it for turtle/. In that case, you just "cd turtle/; mvn clean test". https://github.com/antlr/grammars-v4/pull/4831

teverett:
@nglgzz  thanks!

PR #4827, opened 2026-04-28
antlr/grammars-v4#4826: Fix issues in the Oracle MySQL grammar

### LINKED ISSUE #4826: Potential issues in the Oracle MySQL grammar
The Oracle MySQL grammar in `sql/mysql/Oracle/` contains **5 possible bugs**
   that cause it to reject valid MySQL 8.0/9.0 syntax. I verified all issues  against official MySQL 8.0.46 and 9.0.1 instances and cross-referenced with MySQL documentation.

 Issue #1: Missing GEOMCOLLECTION Keyword

  MySQL supports `GEOMCOLLECTION` as a synonym for `GEOMETRYCOLLECTION` in
  multiple contexts (data types, functions, CAST operations), but the grammar only recognizes the latter.

  **Test**:
  ```sql
  -- GEOMCOLLECTION as column type
  mysql> CREATE TABLE t1 (geom GEOMCOLLECTION);
  Query OK, 0 rows affected (0.02 sec)

  mysql> CREATE TABLE t2 (geom GEOMETRYCOLLECTION);
  Query OK, 0 rows affected (0.01 sec)

  -- GEOMCOLLECTION in CAST
  mysql> SELECT CAST(ST_GeomFromText('POINT(1 1)') AS GEOMCOLLECTION);
  +--------------------------------------------------------+
  | CAST(ST_GeomFromText('POINT(1 1)') AS GEOMCOLLECTION)  |
  +--------------------------------------------------------+
  | 0x000000000700000001000000010100000000000000000000F0... |
  +--------------------------------------------------------+
```
  Proposed Fix:
  - Add GEOMCOLLECTION_SYMBOL token to lexer
  - Add GEOMCOLLECTION_SYMBOL to geometryFunction rule 
  - Add GEOMCOLLECTION_SYMBOL to castType rule 
  - Add GEOMCOLLECTION_SYMBOL to dataType rule 
  - Add GEOMCOLLECTION_SYMBOL to labelKeyword rule 
  - Add GEOMCOLLECTION_SYMBOL to functionNameBase rule 
  ---
  Issue #2: Uppercase Charset Names Rejected

  The grammar's UNDERSCORE_CHARSET pattern only accepts lowercase [a-z0-9]+,
  but MySQL accepts uppercase charset names like __UTF8MB4, _LATIN1._

  Test:
  ```sql
  mysql> SELECT _UTF8MB4'test', _utf8mb4'test';
  +----------------+----------------+
  | _UTF8MB4'test' | _utf8mb4'test' |
  +----------------+----------------+
  | test           | test           |
  +----------------+----------------+

  mysql> SELECT _LATIN1'hello', _latin1'hello';
  +----------------+----------------+
  | _LATIN1'hello' | _latin1'hello' |
  +----------------+----------------+
  | hello          | hello          |
  +----------------+----------------+
  ```

Proposed fix: Change _UNDERSCORE_CHARSET_ pattern from [a-z0-9]+ to [a-zA-Z0-9]+ 

  ---
 Issue #3: Incorrect Backtick Escaping

  The grammar's BACK_TICK_QUOTED_ID pattern doesn't correctly handle doubled
  backticks (MySQL's escape mechanism for backticks inside identifiers).

  Test:
  ```sql
  mysql> CREATE TABLE `test``table` (`col``name` INT);
  Query OK, 0 rows affected (0.02 sec)

  mysql> SHOW TABLES;
  +----------------+
  | Tables_in_test |
  +----------------+
  | test`table     |
  +----------------+

  mysql> INSERT INTO `test``table` (`col``name`) VALUES (42);
  Query OK, 1 row affected (0.01 sec)

  mysql> SELECT `col``name` FROM `test``table`;
  +-------------+
  | col`name    |
  +-------------+
  |          42 |
  +-------------+
  ```

  Proposed fix: Updated BACK_TICK_QUOTED_ID pattern to recognize BACK_TICK BACK_TICK as
  escaped backtick sequence.

  ---
  Issue #4: ALTER EVENT Qualified Identifier Missing

  The grammar uses identifier for RENAME TO clause in ALTER EVENT, but MySQL
  accepts qualified identifiers (e.g., db1.event → db2.event), allowing events
  to be moved between databases.

  Test:
  ```sql 
 mysql> CREATE DATABASE db1;
  Query OK, 1 row affected (0.01 sec)

  mysql> CREATE DATABASE db2;
  Query OK, 1 row affected (0.00 sec)

  mysql> CREATE EVENT db1.evt ON SCHEDULE EVERY 1 DAY DO SELECT 1;
  Query OK, 0 rows affected (0.01 sec)

  mysql> ALTER EVENT db1.evt RENAME TO db2.evt;
  Query OK, 0 rows affected (0.01 sec)

  mysql> SELECT EVENT_SCHEMA, EVENT_NAME FROM information_schema.EVENTS WHERE
  EVENT_NAME='evt';
  +--------------+------------+
  | EVENT_SCHEMA | EVENT_NAME |
  +--------------+------------+
  | db2          | evt        |
  +--------------+------------+
  ```

  Proposed Fix: Changed identifier to qualifiedIdentifier in alterEvent rule's RENAME TO
   clause.

  Note: qualifiedIdentifier allows optional database qualification:
  - identifier: Only accepts evt (single name)
  - qualifiedIdentifier: Accepts both evt and db2.evt (optional database
  prefix)

  ---
  Issue #5: DEFAULT Keyword Version Restriction

  The grammar blocks DEFAULT in charsetName and collationName rules for MySQL
  8.0.11+ with predicate {this.isServerVersionLt80011()}?, but MySQL accepts
  SET CHARSET DEFAULT in all versions including 8.0.11+.

  Test:
  ```sql  
mysql> SELECT VERSION();
  +-----------+
  | VERSION() |
  +-----------+
  | 8.0.46    |
  +-----------+

  mysql> SET CHARSET DEFAULT;
  Query OK, 0 rows affected (0.00 sec)

  mysql> SET CHARACTER SET DEFAULT;
  Query OK, 0 rows affected (0.00 sec)

  mysql> SET NAMES DEFAULT;
  Query OK, 0 rows affected (0.00 sec)

  mysql> SHOW VARIABLES LIKE 'character_set_client';
  +----------------------+---------+
  | Variable_name        | Value   |
  +----------------------+---------+
  | character_set_client | utf8mb4 |
  +----------------------+---------+
  ```

  Documentation: MySQL 8.0 SET CHARACTER SET states "The default character set
  mapping can be restored by using the value DEFAULT" with no version
  restrictions.

  Proposed Fix: Remove {this.isServerVersionLt80011()}? predicate from both charsetName
   and collationName rules

 
  ---
  References

  - MySQL 8.0 Spatial Data Types:
  https://dev.mysql.com/doc/refman/8.0/en/spatial-type-overview.html
  - MySQL 8.0 Character Sets:
  https://dev.mysql.com/doc/refman/8.0/en/charset.html
  - MySQL 8.0 SET Statements:
  https://dev.mysql.com/doc/refman/8.0/en/set-variable.html
  - MySQL 8.0 Events: https://dev.mysql.com/doc/refman/8.0/en/events.html
  - Debezium MySQL Grammar (reference):
  https://github.com/debezium/debezium/tree/main/debezium-ddl-parser/src/main/a
  ntlr4/io/debezium/ddl/parser/mysql/generated

### PR DESCRIPTION
**Pull request by vsantonastaso**
closes https://github.com/antlr/grammars-v4/issues/4826

  This PR fixes **5 possible bugs** in the Oracle MySQL grammar that caused it
   to reject valid MySQL 8.0/9.0 syntax. All fixes were verified against
  official MySQL 8.0.46 and 9.0.1 Docker instances and cross-referenced with
  MySQL documentation.

  ##### 1. Missing GEOMCOLLECTION Keyword
  - **Issue**: Grammar only recognized `GEOMETRYCOLLECTION`, not its synonym
  `GEOMCOLLECTION`
  - **Impact**: Rejected valid column types, functions, and CAST operations
  - **Fix**: Added `GEOMCOLLECTION_SYMBOL` token and updated all relevant
  parser rules

  ##### 2. Uppercase Charset Names Rejected
  - **Issue**: `UNDERSCORE_CHARSET` pattern only accepted lowercase `[a-z0-9]+`
  - **Impact**: Rejected valid charset literals like `_UTF8MB4'test'`,
  `_LATIN1'hello'`
  - **Fix**: Changed pattern to `[a-zA-Z0-9]+` to accept uppercase

  ##### 3. Incorrect Backtick Escaping
  - **Issue**: `BACK_TICK_QUOTED_ID` didn't handle doubled backticks correctly
  - **Impact**: Rejected identifiers with embedded backticks like ``
  `test``table` ``
  - **Fix**: Updated pattern to recognize `BACK_TICK BACK_TICK` as escape
  sequence

  ##### 4. ALTER EVENT Qualified Identifier Missing
  - **Issue**: `RENAME TO` clause used `identifier` instead of
  `qualifiedIdentifier`
  - **Impact**: Rejected moving events between databases: `ALTER EVENT db1.evt
  RENAME TO db2.evt`
  - **Fix**: Changed to `qualifiedIdentifier` to allow optional database prefix

  ##### 5. DEFAULT Keyword Version Restriction
  - **Issue**: Incorrect predicate `{this.isServerVersionLt80011()}?` blocked
  DEFAULT in charset/collation
  - **Impact**: Rejected `SET CHARSET DEFAULT` in MySQL 8.0.11+
  - **Fix**: Removed version restriction (DEFAULT works in all MySQL versions)

  All fixes verified with MySQL 8.0.46 (Docker: `mysql:8.0.46`)
and MySQL 9.0.1 (Docker: `mysql:9.0.1`)

### DIFF
--- a/sql/mysql/Oracle/MySQLLexer.g4
+++ b/sql/mysql/Oracle/MySQLLexer.g4
@@ -1171,6 +1171,10 @@ GEOMETRYCOLLECTION_SYMBOL
     : G E O M E T R Y C O L L E C T I O N
     ;                                                                                     // MYSQL
 
+GEOMCOLLECTION_SYMBOL
+    : G E O M C O L L E C T I O N -> type(GEOMETRYCOLLECTION_SYMBOL)
+    ;                                                                                     // Synonym
+
 GEOMETRY_SYMBOL
     : G E O M E T R Y
     ;
@@ -3318,6 +3322,11 @@ SOURCE_COMPRESSION_ALGORITHM_SYMBOL
     : S O U R C E '_' C O M P R E S S I O N '_' A L G O R I T H M                         {this.isServerVersionGe80023()}?
     ;                                                                                     // MYSQL
 
+SOURCE_COMPRESSION_ALGORITHMS_SYMBOL
+    : S O U R C E '_' C O M P R E S S I O N '_' A L G O R I T H M S                       {this.isServerVersionGe80023()}?
+      -> type(SOURCE_COMPRESSION_ALGORITHM_SYMBOL)
+    ;                                                                                     // MYSQL
+
 SOURCE_CONNECT_RETRY_SYMBOL
     : S O U R C E '_' C O N N E C T '_' R E T R Y                                         {this.isServerVersionGe80023()}?
     ;                                                                                     // MYSQL
@@ -3587,7 +3596,7 @@ INVALID_INPUT
 // The underscore charset token is used to defined the repertoire of a string, though it conflicts
 // with normal identifiers, which also can start with an underscore.
 UNDERSCORE_CHARSET
-    : UNDERLINE_SYMBOL [a-z0-9]+ { this.doUnderscoreCharset(); }
+    : UNDERLINE_SYMBOL [a-zA-Z0-9]+ { this.doUnderscoreCharset(); }
     ;
 
 // TODO: check in the semantic phase that starting and ending tags are the same.
@@ -3624,7 +3633,7 @@ fragment DOUBLE_QUOTE
     ;
 
 BACK_TICK_QUOTED_ID
-    : BACK_TICK (({this.isBackTickQuotedId()}? '\\')? .)*? BACK_TICK
+    : BACK_TICK ( BACK_TICK BACK_TICK | ~[`] )* BACK_TICK
     ;
 
 DOUBLE_QUOTED_TEXT
--- a/sql/mysql/Oracle/MySQLParser.g4
+++ b/sql/mysql/Oracle/MySQLParser.g4
@@ -163,7 +163,7 @@ alterDatabaseOption
 alterEvent
     : definerClause? EVENT_SYMBOL eventRef (ON_SYMBOL SCHEDULE_SYMBOL schedule)? (
         ON_SYMBOL COMPLETION_SYMBOL NOT_SYMBOL? PRESERVE_SYMBOL
-    )? (RENAME_SYMBOL TO_SYMBOL identifier)? (
+    )? (RENAME_SYMBOL TO_SYMBOL qualifiedIdentifier)? (
         ENABLE_SYMBOL
         | DISABLE_SYMBOL (ON_SYMBOL replica)?
     )? (COMMENT_SYMBOL textLiteral)? (DO_SYMBOL compoundStatement)?
@@ -2355,7 +2355,7 @@ optionValueListContinued
 
 optionValueNoOptionType
     : lvalueVariable equal setExprOrDefault
-    | charsetClause
+    | charset (charsetName | DEFAULT_SYMBOL)
     | userVariable equal expr
     | AT_AT_SIGN_SYMBOL setVarIdentType? lvalueVariable equal setExprOrDefault
     | NAMES_SYMBOL (
@@ -3839,12 +3839,10 @@ functionDatetimePrecision
 charsetName
     : textOrIdentifier
     | BINARY_SYMBOL
-    | {this.isServerVersionLt80011()}? DEFAULT_SYMBOL
     ;
 
 collationName
     : textOrIdentifier
-    | {this.isServerVersionLt80011()}? DEFAULT_SYMBOL
     | {this.isServerVersionGe80018()}? BINARY_SYMBOL
     ;
 

### CODE REVIEW

File: sql/mysql/Oracle/MySQLLexer.g4 [hunk 1]
```diff
@@ -1171,6 +1171,10 @@ GEOMETRYCOLLECTION_SYMBOL
     : G E O M E T R Y C O L L E C T I O N
     ;                                                                                     // MYSQL
 
+GEOMCOLLECTION_SYMBOL
```
kaby76, line 1174: Change:

```
GEOMCOLLECTION_SYMBOL
      : G E O M C O L L E C T I O N -> type(GEOMETRYCOLLECTION_SYMBOL)
      ;  // Synonym
```

sql/mysql/Oracle/MySQLLexer.g4
```diff
@@ -1171,6 +1171,10 @@ GEOMETRYCOLLECTION_SYMBOL
...
+    : G E O M C O L L E C T I O N
+    ;                                                                                     // MYSQL
+
```
kaby76, line 1177: Looks like the grammar is also missing SOURCE_COMPRESSION_ALGORITHM(S). Suggest also adding this in its alphabetic-sorted location in the grammar:
```
SOURCE_COMPRESSION_ALGORITHM_SYMBOL: S O U R C E '_' C O M P R E S S I O N '_' A L G O R I T H M;
  SOURCE_COMPRESSION_ALGORITHMS_SYMBOL: S O U R C E '_' C O M P R E S S I O N '_' A L G O R I T H M S
      -> type(SOURCE_COMPRESSION_ALGORITHM_SYMBOL);
```

sql/mysql/Oracle/MySQLLexer.g4
```diff
@@ -3587,7 +3591,7 @@ INVALID_INPUT
 // The underscore charset token is used to defined the repertoire of a string, though it conflicts
 // with normal identifiers, which also can start with an underscore.
 UNDERSCORE_CHARSET
-    : UNDERLINE_SYMBOL [a-z0-9]+ { this.doUnderscoreCharset(); }
+    : UNDERLINE_SYMBOL [a-zA-Z0-9]+ { this.doUnderscoreCharset(); }
```
kaby76, line 3599: Perfect!

sql/mysql/Oracle/MySQLParser.g4
```diff
@@ -163,7 +163,7 @@ alterDatabaseOption
 alterEvent
     : definerClause? EVENT_SYMBOL eventRef (ON_SYMBOL SCHEDULE_SYMBOL schedule)? (
         ON_SYMBOL COMPLETION_SYMBOL NOT_SYMBOL? PRESERVE_SYMBOL
-    )? (RENAME_SYMBOL TO_SYMBOL identifier)? (
+    )? (RENAME_SYMBOL TO_SYMBOL qualifiedIdentifier)? (
```
kaby76, line 166: Perfect!

sql/mysql/Oracle/MySQLParser.g4
```diff
@@ -3128,7 +3128,7 @@ returningType
     ;
 
 geometryFunction
-    : GEOMETRYCOLLECTION_SYMBOL OPEN_PAR_SYMBOL exprList? CLOSE_PAR_SYMBOL
+    : (GEOMETRYCOLLECTION_SYMBOL | GEOMCOLLECTION_SYMBOL) OPEN_PAR_SYMBOL exprList? CLOSE_PAR_SYMBOL
```
kaby76: Collapse back to the original rule.

sql/mysql/Oracle/MySQLParser.g4
```diff
@@ -3252,6 +3252,7 @@ castType
         | MULTIPOINT_SYMBOL
         | MULTILINESTRING_SYMBOL
         | MULTIPOLYGON_SYMBOL
+        | GEOMCOLLECTION_SYMBOL
```
kaby76: Collapse back to the original rule. Do for all other rules with GEOMCOLLECTION_SYMBOL occurrences. You won't need to change any of these rules.

sql/mysql/Oracle/MySQLLexer.g4
```diff
@@ -3624,7 +3628,7 @@ fragment DOUBLE_QUOTE
     ;
 
 BACK_TICK_QUOTED_ID
-    : BACK_TICK (({this.isBackTickQuotedId()}? '\\')? .)*? BACK_TICK
+    : BACK_TICK ( BACK_TICK BACK_TICK | ({this.isBackTickQuotedId()}? '\\')? ~[`] )* BACK_TICK
```
kaby76: I think this is still wrong. For example, does it work for  ``CREATE TABLE `logs\` (id INT); --table name is logs\`` or ``SELECT `col\` FROM t;  -- column name is col\``. The rule should match the server exactly — doubles only, no backslash special-casing:
```
BACK_TICK_QUOTED_ID
    : BACK_TICK ( BACK_TICK BACK_TICK | ~[`] )* BACK_TICK
    ;
```

kaby76 (changes requested):

vsantonastaso:
@kaby76  thanks for the feedback. I pushed a new version with all comments addressed.

kaby76 (approved):

sql/mysql/Oracle/MySQLParser.g4
```diff
@@ -3839,12 +3839,12 @@ functionDatetimePrecision
 charsetName
     : textOrIdentifier
     | BINARY_SYMBOL
-    | {this.isServerVersionLt80011()}? DEFAULT_SYMBOL
+    | DEFAULT_SYMBOL
```
mike-lischke: This is wrong! We discussed this already elsewhere. You cannot allow DEFAULT for any charset name.

sql/mysql/Oracle/MySQLParser.g4
```diff
@@ -3839,12 +3839,12 @@ functionDatetimePrecision
...
     ;
 
 collationName
     : textOrIdentifier
-    | {this.isServerVersionLt80011()}? DEFAULT_SYMBOL
+    | DEFAULT_SYMBOL
```
mike-lischke: Same here. This change is not correct.

mike-lischke (changes requested):
I recommend to consult the server grammar file (https://github.com/mysql/mysql-server/blob/trunk/sql/sql_yacc.yy) before you change this ANTLR grammar or you may introduce bugs, if you just try to follow the documentation (which simplifies things for better understanding). For example the charset_name rule is:

```
charset_name:
          ident_or_text
          {
            if (!($$=get_charset_by_csname($1.str,MY_CS_PRIMARY,MYF(0))))
            {
              my_error(ER_UNKNOWN_CHARACTER_SET, MYF(0), $1.str);
              MYSQL_YYABORT;
            }
            YYLIP->warn_on_deprecated_charset($$, $1.str);
          }
        | BINARY_SYM { $$= &my_charset_bin; }
        ;

```

Take away the actions and you'll end up with `text | BINARY`, nothing else. If you miss a keyword in a statement, follow that statement through entire grammar to see where you have to add that keyword.

Take for example this ANTLR rule:

```
optionValueNoOptionType:
    lvalueVariable equal setExprOrDefault
    | charsetClause
    | userVariable equal expr
    | AT_AT_SIGN_SYMBOL setVarIdentType? lvalueVariable equal setExprOrDefault
    | NAMES_SYMBOL (
        equal expr
        | charsetName collate?
        | {this.serverVersion >= 80011}? DEFAULT_SYMBOL
    )
;
```

It uses the `charsetName` rule and explicitly adds the `DEFAULT` keyword as alternative (starting 8.0.11, which is the version where `DEFAULT` was removed from `charsetName`).

sql/mysql/Oracle/MySQLLexer.g4
```diff
@@ -3624,7 +3633,7 @@ fragment DOUBLE_QUOTE
     ;
 
 BACK_TICK_QUOTED_ID
-    : BACK_TICK (({this.isBackTickQuotedId()}? '\\')? .)*? BACK_TICK
+    : BACK_TICK ( BACK_TICK BACK_TICK | ~[`] )* BACK_TICK
```
mike-lischke, line 3636: This is not correct. Back tick identifiers do not support doubled quotation characters like single quote and double quote strings do. Check out https://dev.mysql.com/doc/refman/8.4/en/string-literals.html.

Back tick strings are not strings in the normal sense, but identifiers! If you want backticks in such an identifier you have to escape it using the backslash (which the lexer grammar handles already).

mike-lischke (changes requested):

sql/mysql/Oracle/MySQLLexer.g4
vsantonastaso: Thanks for the feedback, @mike-lischke. I tested the backtick escaping behavior against MySQL 8.0.46 to verify. What I Found

  Doubled backticks work:
```sql
  mysql> CREATE TABLE `test``table` (id INT);
  Query OK, 0 rows affected

  mysql> SHOW TABLES;
  test`table         -- actual table name contains a backtick

  mysql> SHOW CREATE TABLE `test``table`;
  CREATE TABLE `test``table` (...)   -- MySQL's own output uses doubled backticks
```

  Backslash escaping fails:
```sql
  mysql> CREATE TABLE `test\`table2` (id INT);
  ERROR 1064 (42000): You have an error in your SQL syntax
```

  This happens in both default mode and with SET sql_mode = 'NO_BACKSLASH_ESCAPES'.

The documentation states:

  ▎ _"If the character to be included in the identifier is the same as that used to quote the identifier itself, then  you need to double the character."_
And provides this example:
  CREATE TABLE `a``b` (`c"d` INT);
 The String Literals page you mentioned describes backslash escaping, but only for string literals ('...' and "..."), not identifiers.
Could you clarify what I'm missing?

kaby76: Exactly. 

`` CREATE TABLE `test\`table` (`col\`name` INT); `` does not parse in the mysql cli. But it does for Workbench.

`` CREATE TABLE test`table (col`name INT); `` parses in the mysql cli. But it does for Workbench.

Both are version 8, Windows.

Strictly speaking, the mysql cli (.yy) and the Workbench/Plugin (.g4) are two different SQL languages because of the above. Further, in the .g4: ["The coverage of the MySQL language should be 100%, but there might still be bugs or omissions."](https://github.com/antlr/grammars-v4/blob/55d2bd37ca9b4271ff1f5cb3868e7d58e68f5a0f/sql/mysql/Oracle/MySQLParser.g4#L14)

This seems to be the intent. Are we to have two grammars, one to read one or the other? Besides, Workbench is behind a version from the server.

mike-lischke: I stand corrected (https://dev.mysql.com/doc/refman/5.7/en/sys-quote-identifier.html). However, I would implement that like in the single and double quote rules, for consistency.

@kaby76 Don't use Workbench anymore for any parser related comparison :-) It's far behind the MySQL development and wasn't updated for years.

vsantonastaso:
@mike-lischke @kaby76 I updated changes based on your feedback. Please take a look when you have a chance. Thanks

mike-lischke (approved):

vsantonastaso:
@mike-lischke How Can I proceed to merge this PR? 
I see this in the PR actions detail 
```
1 workflow awaiting approval
This workflow requires approval from a maintainer.
```

mike-lischke:
One of the maintainers has to approve the workflow - likely @teverett.

teverett:
@vsantonastaso  thanks!

PR #4831, opened 2026-04-30
[ai] Update Claude settings and info.

kaby76:
* Update CLAUDE.md with info on Maven testing.
* Remove .claude/ directories.
* Update the _scripts/maven.sh file to not create a .jar. Just to clean/test.


teverett:
@kaby76  thanks

PR #4848, opened 2026-05-10
feat(wkt): CLOTHOID support in COMPOUNDCURVE (per proposal #4847)

grootstebozewolf:
Adds support for the `CLOTHOID` keyword as a member of `COMPOUNDCURVE`, per the
proposal at [grammars-v4 discussion #4847](https://github.com/antlr/grammars-v4/discussions/4847).

A clothoid (Euler / Cornu spiral) is the curve whose curvature varies linearly
with arc length — `κ(s) = κ₀ + (κ₁ − κ₀)·s/L`. It's the standard transition
between a straight and a circular arc in road and rail design (LandXML, IFC,
ProRail, Deutsche Bahn, etc.); without it, real-world track / highway alignments
have to be approximated with chord polylines or stored out-of-band in custom
attributes.

## Grammar diff

| Addition | Form |
|---|---|
| Lexer token | `CLOTHOID : C L O T H O I D ;` |
| Parser rule | `clothoidGeometry : CLOTHOID LPAR ordinate COMMA ordinate COMMA ordinate RPAR ;` |
| Wired into | `compoundCurveMember` |

Three scalars: `(startKappa, endKappa, length)`. Coordinates of the clothoid's
endpoints are derived analytically from the preceding member's end-state
(start point + tangent), per §3.3 of the proposal — so the keyword is only
valid in non-leading positions inside a `COMPOUNDCURVE`. The grammar accepts
it; the WKT reader rejects it as the first member with a clear message.

The `ordinate` production gives `NaN` / `Inf` syntactic acceptance for free;
the runtime construct rejects them (length must be positive finite, kappas
must be finite) per §3.4.

## Examples (in `examples/example2.txt`)

```wkt
COMPOUNDCURVE ((0 0, 100 0), CLOTHOID (0, 0.005, 80))
COMPOUNDCURVE ((0 0, 100 0), CLOTHOID (0, 0.005, 48), CIRCULARSTRING (147.93 1.92, 200 50, 150 100), CLOTHOID (0.005, 0, 42))
```

`mvn clean test` passes locally with both new examples parsing.

## Reference implementation

The runtime semantics (`ClothoidSegment` geometry type, Simpson integrator,
`CurvedWKTReader/Writer`, `CompoundCurve` with structural members,
`Linearizable.toLinear(tolerance)`, analytical bounding box, `reverse()` with
κ-sign flip, `equalsExact` with start-state, junction-drift warning,
TestBuilder editing/inserting UI) is implemented on a JTS playground branch:

https://github.com/grootstebozewolf/jts/tree/feature/sfa-curve-clothoid-playground

Reviewers who want to see the parsed geometry render and operate end-to-end
can run the TestBuilder on that branch — `bin/testbuilder.sh` — and load
`CurveExample → clothoidRailBend` (a real ProRail track 823_12V_4.3 in
EPSG:28992) or `arcChainNoSpirals` (G1-only chain to test the "Insert spiral
before next arc" action).

## Implementation notes (for reviewers asking what's in scope here vs the JTS side)

**1. Buffering / spatial ops.** Currently goes through full linearization:
`bufferCurveWithParams` does `((Linearizable) g).toLinear(tol)` and feeds the
result into the standard JTS `BufferOp`. This is intentional for v1 — the
parallel of a clothoid is *not* itself a clothoid (no closed-form arc-length
parameterization), and the parallel of a circular arc remains a circular arc
only if you're willing to accept a different radius (and the construction
degenerates for inward offsets when distance ≥ R).

For v1 we intentionally retain the densify-then-buffer approach. It is
predictable, the geometric error is explicitly bounded by the caller-supplied
tolerance, and it avoids the substantially larger problem of a true
curve-aware offset algorithm (which belongs in a separate proposal).

**2. Buffer artifacts to watch.**
   - Tangent kinks at user-authored junctions render as the configured join
     style (Round / Mitre / Bevel) instead of smoothing — a feature, not a
     bug, but worth knowing if a buffer comes back with corners.
   - Very tight clothoids: the densifier's chord-to-true-curve drift is
     bounded by `tol`, but the buffer polygon's vertices inherit that drift.
     Visible as minor faceting / short straight segments on the offset
     polygon in regions of very high curvature.
   - Cap orientation at clothoid endpoints uses the analytical end tangent,
     not chord direction. A naive last-two-coords implementation would be
     subtly wrong here.

**3. Cost vs plain `LineString` of equivalent densified-point count.**
Equivalent. `CompoundCurve` adds the structural member array (a few hundred
bytes for typical chains) but the flat coord sequence concatenates member
coords, so spatial ops see the same point count. Repeated `toLinear(tol)`
calls aren't memoized in v1 — that's a possible future optimisation.

## What this PR does NOT include

- No code changes to other grammars in this repo.
- No JTS / NetTopologySuite library code (that's the linked playground
  branch — separate repo, separate PR when this lands).
- No changes to `MULTICURVE` / `CURVEPOLYGON`. We intentionally scoped
  `CLOTHOID` to `compoundCurveMember` only for v1: a leading clothoid has no
  preceding tangent state (proposal §3.3), so allowing it in `curveMember`
  would require the WKT reader to special-case "non-leading except when the
  parent is COMPOUNDCURVE" — weirder than just keeping the grammar tight.
  The proposal foreshadows a 6-arg standalone form for the future "clothoid
  with absolute start state" use case.

## AI Disclosure

Per Eclipse Foundation Generative AI Usage Guidelines (applied as best
practice): proposal text and grammar diff drafted with `Claude (Opus-4.7)`
and `xAI Grok (grok-4.3)`, human-reviewed.


teverett:
@grootstebozewolf  thanks!

PR #4861, opened 2026-05-16
[build] Update Trash Toolkit.

kaby76:
* Update remaining Trash Toolkit app versions.
* Sync templates with current version of trgen.


teverett:
@kaby76  thanks!

PR #3788, opened 2023-10-19
snowflake - warehouse new features

mlorek:

PR #3538, opened 2023-06-16
snowflake - event tables

mlorek:

PR #4863, opened 2026-05-22
snowlfake: add location to non reserved words

svetoslav-petkov-atscale:
## Summary

The Snowflake grammar over-reserves `LOCATION`. Per Snowflake's own [reserved-keywords reference](https://docs.snowflake.com/en/sql-reference/reserved-keywords), `LOCATION` is **not** a reserved keyword — it should be usable as a column, alias, or table name. Today the grammar lexes `LOCATION` as a dedicated token (`SnowflakeLexer.g4:516`) but does not list it in either `keyword` or `non_reserved_words`, so the `id_` rule cannot match it and any identifier named `location` fails to parse.

This is reproducible with a one-line query that any user with a `location` column would write:

```sql
SELECT LOCATION FROM my_table;
-- line 1:7 mismatched input 'LOCATION' expecting { ACCOUNT, ACCOUNTADMIN, ... }
```

It also fails as a qualified reference (`location.location`) and as an alias (`x AS location`), which is how I hit it — parsing a Snowflake Semantic View definition where a dimension was declared `LOCATION.LOCATION AS LOCATION`.

## Fix

Add `LOCATION` to `non_reserved_words` in `SnowflakeParser.g4`, in alphabetic order between `LOCAL` and `LOW` (per the rule's own `// please add in alphabetic order for easy reading` comment):

```antlr
    | LOCAL
    | LOCATION
    | LOW
```

`LOCATION` continues to function in its structural positions (`CREATE STAGE … LOCATION = …`, `ALTER EXTERNAL TABLE … DROP PARTITION LOCATION '…'`, `STORAGE_ALLOWED_LOCATIONS = (…)`) because those rules reference the `LOCATION` token directly, not through `id_`. The change only widens what `id_` accepts.

## Why this is the right shape

1. **Snowflake itself agrees.** `LOCATION` is absent from the reserved-keywords list. The grammar's behavior is stricter than the database it models.
2. **Cross-grammar precedent in this repo.** The sibling PostgreSQL grammar lists `LOCATION` in its `unreserved_keyword` rule.
3. **Mechanically identical merged PRs.** #3788 (`MAX_CONCURRENCY_LEVEL`, `WAREHOUSE_TYPE`), #3962, #3538, #3359, #3132 — all single-line additions to `non_reserved_words` for the same class of issue. None of the merged ones introduced ambiguity.

## Risk

Minimal. Adding a token to `non_reserved_words` only relaxes `id_`; it does not change how the parser handles `LOCATION` in any rule that references the token directly. I ran the existing Snowflake parser test suite locally with the change applied and saw no regressions.


teverett:
@svetoslav-petkov-atscale thanks!

Issue #1894, opened 2020-10-16
Duplicate GreaterEqual token in CPP14Parser theOperator rule


mycroft8:
This specifically refers to the September 8th version of CPP14Parser.g4
For the rule "theOperator", that starts at line 768, the token "GreaterEqual" appears twice.  Once at line 783, once at line 799.
I assume that the token at line 783 should be "Greater".
Thanks,
Dave

mycroft8:
The "theOperator" rule is also missing the 'DivAssign" operator ( /= ).  A consequence is that operator overloading of DivAssign isn't recognized.  As the other assignment operators, like "*=" are listed under "theOperator", I added it to my copy.
I've now reached the point where my version of the CPP parser, with modifications for handling the GCC dialect, can now take my main test program; run it through my nextToken based preprocessor, run it through my modified parser, and parse successfully.  The main program includes many GLIBC complex header files including C++ STL headers, plus the CPP antlr4 runtime headers.  That comes to over 400K tokens; and it takes a while.  
Eventually I may stick this on github somewhere.

PR #4874, opened 2026-05-29
Added DivAssign to theOperator in C++ parser

TwoOfTwelve:
I tried to update the C++ grammar in our project [JPlag](https://github.com/jplag/JPlag) and noticed the current C++ grammar is not able to pass one of our test files. Specifically the following code breaks the grammar:

```C++
    HDRColorA& operator/=(float f) noexcept
    {
        const float fInv = 1.0f / f;
        r *= fInv;
        g *= fInv;
        b *= fInv;
        a *= fInv;
        return *this;
    }
```

It seems to me that as mentioned in #1894 the DivAssign operator is simply missing in the "theOperator" rule. As far as I can tell there are no issues with just adding it. Please let me know if this is an issue.

### DIFF
--- a/cpp/CPP14Parser.g4
+++ b/cpp/CPP14Parser.g4
@@ -1042,6 +1042,7 @@ theOperator
     | PlusAssign
     | MinusAssign
     | StarAssign
+    | DivAssign
     | ModAssign
     | XorAssign
     | AndAssign


teverett:
@TwoOfTwelve  thanks!

Issue #4876, opened 2026-05-31
[Postgresql] Fix uescape_ rule placement in identifier


capjck:
Description:
Currently, the `identifier ` rule applies `uescape_?` to standard `Identifier ` instead of `UnicodeQuotedIdentifier`. According to the PostgreSQL specification, the UESCAPE clause is only valid after a Unicode escape string/identifier literal (e.g., U&"..." UESCAPE '!').

Here is the proposed fix:

Current: [PostgreSQLParser.g4](https://github.com/antlr/grammars-v4/blob/master/sql/postgresql/PostgreSQLParser.g4), line 5463
```
identifier
    : Identifier uescape_?
    | QuotedIdentifier
    | UnicodeQuotedIdentifier
    | PLSQLVARIABLENAME
    ;
```

Should be:
```
identifier
    : Identifier
    | QuotedIdentifier
    | UnicodeQuotedIdentifier uescape_?
    | PLSQLVARIABLENAME
    ;
```

PR #4877, opened 2026-05-31
Fix uescape_ rule placement in identifier (#4876)

capjck:


teverett:
@capjck thanks!

kaby76:
(This was merged before testing. The Cpp build broke because there were new rules added, which broke COFF restrictions on VS2026, and probably VS2022.)

PR #4880, opened 2026-06-02
[vhdl] Added correct signature rule to VHDL2008 Grammar

LucasMoss:
Added signature rule to VHDL2008 Grammar, in accordance with IEEE 1076-2008:

4.5.3 Signatures
A signature distinguishes between overloaded subprograms and overloaded enumeration literals based on
their parameter and result type profiles. A signature can be used in a subprogram instantiation declaration,
attribute name, entity designator, or alias declaration.
signature ::= [ [ type_mark { , type_mark } ] [ return type_mark ] ]
(Note that the initial and terminal brackets are part of the syntax of signatures and do not indicate that the
entire right-hand side of the production is optional.)

The old signature rule never worked because of the confusion regarding the enclosing LBRACKET and RBRACKET.

### DIFF
--- a/vhdl/vhdl2008/examples/test_pkg.vhd
+++ b/vhdl/vhdl2008/examples/test_pkg.vhd
@@ -38,10 +38,12 @@ package test_pkg is
   end component component_with_trailing_label;
 
   function signature_test(data : unsigned) return integer;
-
+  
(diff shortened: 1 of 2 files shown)


teverett:
@LucasMoss thanks!

PR #4889, opened 2026-06-09
R - Fixed compiler error for RFilter.g4

TwoOfTwelve:
When updating the R grammar I noticed that the RFilter.g4 does not compile due to a missing semicolon. This PR adds it.


teverett:
@TwoOfTwelve thanks!

PR #4892, opened 2026-06-09
Fixed errors in TypeScriptParser.g4

TwoOfTwelve:
Using the current version of the [typescript grammar](https://github.com/antlr/grammars-v4/tree/master/javascript/typescript) the AST generated for the test files found [here](https://github.com/jplag/JPlag/tree/main/languages/typescript/src/test/resources/de/jplag/typescript) seem to be wrong in some cases:

[forLoops.ts](https://github.com/jplag/JPlag/blob/main/languages/typescript/src/test/resources/de/jplag/typescript/forLoops.ts): The second and third for loops get turned into an expression rule instead of the iteration rule
[methods.ts](https://github.com/jplag/JPlag/blob/main/languages/typescript/src/test/resources/de/jplag/typescript/methods.ts): The second and third methods get turned into expression rules instead of function declarations
[allTokens.ts](https://github.com/jplag/JPlag/blob/main/languages/typescript/src/test/resources/de/jplag/typescript/allTokens.ts): The namespace gets turned into a variable declaration instead of a namespace declaration

I was able to fix those errors by changing the order of the rules. I'm not sure if the current order is perfect, but it seems to work for the test files. If someone knows more about this grammar I'd appreciate some feedback.

### DIFF
--- a/javascript/typescript/TypeScriptParser.g4
+++ b/javascript/typescript/TypeScriptParser.g4
@@ -345,6 +345,7 @@ sourceElement
 
 statement
     : block
+    | namespaceDeclaration //ADDED
     | variableStatement
     | importStatement
     | exportStatement
@@ -352,9 +353,7 @@ statement
     | abstractDeclaration //ADDED
     | classDeclaration
     | functionDeclaration
-    | expressionStatement
     | interfaceDeclaration //ADDED
-    | namespaceDeclaration //ADDED
     | ifStatement
     | iterationStatement
     | continueStatement
@@ -372,6 +371,7 @@ statement
     | typeAliasDeclaration //ADDED
     | enumDeclaration      //ADDED
     | Export statement
+    | expressionStatement
     ;
 
 block
--- a/javascript/typescript/examples/allTokensJPlag.ts
+++ b/javascript/typescript/examples/allTokensJPlag.ts
@@ -1,0 +1,70 @@
+import {test} from 'test';
+
+namespace Name {
+    enum Values {
+        val1 = 3,
+        val2
+    }
+}
+
+interface Inter {
+    value: number;
+}
+
+class Class {
+    private x;
+
+    constructor(x) {
+        this.x = x;
+    }
+
+    public test(y: number) {
+        return this.x + y;
+    }
+}
+
+let a = 3;
+const d = 4;
+
+function c() {
+    let b;
+    b = 8;
+    return a - b;
+}
+
+switch (c()) {
+    case 1:
+        break;
+    default:
+        console.log(c())
+}
+
+try {
+    throw "Error"
+} catch (e) {
+    console.log('Error', e);
+} finally {
+    console.log()
+}
+
+for (let i = 0; i < a; i++) {
+
+}
+
+while(a > 3) {
+    a--;
+    if (d > 5) {
+        continue;
+    }
+}
+
+export default Name;
+export { c };
+
+if (d) {
+
+} else if (d < 10) {
+
+} else {
+    
+}
--- a/javascript/typescript/examples/forLoops.ts
+++ b/javascript/typescript/examples/forLoops.ts
@@ -1,0 +1,13 @@
+const are = ["Test", "Test2"];
+
+for (let i = 0; i < are.length; i++) {
+    console.log(are[i])
+}
+
+for (const a in are) {
+    console.log(a)
+}
+
+for (const a of are) {
+    console.log(a)
+}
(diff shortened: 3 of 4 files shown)


kaby76:
Please add the three .ts files to .../javascript/typescript/examples/.

Changing the order of rules for `statement` will change the derivations, but it ultimately means that the grammar is incomplete: it requires disambiguation predicates.

For now, it's ok to change the order.

#### method.ts
```
...
const add2 = function (a: number, b: number) {
    return a + b;
}
...
```

For the "add2" function in method.ts, there are (at least) two parse trees. and those involving `statement` are:
```
truncated-method.ts.d=85.a=2: (program (sourceElements (sourceElement (statement (variableStatement (varModifier (Const "const")) (variableDeclarationList (variableDeclaration (identifierOrKeyWord (identifier (Identifier "add2"))) (Assign "=") (singleExpression (anonymousFunction (Function_ "function") (OpenParen "(") (formalParameterList (formalParameterArg (assignable (identifier (Identifier "a"))) (typeAnnotation (Colon ":") (type_ (unionOrIntersectionOrPrimaryType (primaryType (predefinedType (Number "number"))))))) (Comma ",") (formalParameterArg (assignable (identifier (Identifier "b"))) (typeAnnotation (Colon ":") (type_ (unionOrIntersectionOrPrimaryType (primaryType (predefinedType (Number "number")))))))) (CloseParen ")") (OpenBrace "{") (functionBody (sourceElements (sourceElement (statement (expressionStatement (expressionSequence (singleExpression (identifierName (reservedWord (keyword (Return "return")))) (singleExpression (identifierName (identifier (Identifier "a"))) (singleExpression (Plus "+") (singleExpression (identifierName (identifier (Identifier "b")))))))) (SemiColon ";")))))) (CloseBrace "}"))))))))) (EOF ""))
truncated-method.ts.d=85.a=9: (program (sourceElements (sourceElement (statement (expressionStatement (expressionSequence (singleExpression (identifierName (reservedWord (keyword (Const "const")))) (singleExpression (singleExpression (identifierName (identifier (Identifier "add2")))) (Assign "=") (singleExpression (anonymousFunction (Function_ "function") (OpenParen "(") (formalParameterList (formalParameterArg (assignable (identifier (Identifier "a"))) (typeAnnotation (Colon ":") (type_ (unionOrIntersectionOrPrimaryType (primaryType (predefinedType (Number "number"))))))) (Comma ",") (formalParameterArg (assignable (identifier (Identifier "b"))) (typeAnnotation (Colon ":") (type_ (unionOrIntersectionOrPrimaryType (primaryType (predefinedType (Number "number")))))))) (CloseParen ")") (OpenBrace "{") (functionBody (sourceElements (sourceElement (statement (expressionStatement (expressionSequence (singleExpression (identifierName (reservedWord (keyword (Return "return")))) (singleExpression (identifierName (identifier (Identifier "a"))) (singleExpression (Plus "+") (singleExpression (identifierName (identifier (Identifier "b")))))))) (SemiColon ";")))))) (CloseBrace "}")))))))))) (EOF ""))
```

The choice between alt 2 and 9 has to be disambiguated via mutually exclusive predicates for the two alts (the expressionStatement alt would need to exclude var-modifier tokens), something like this:
```
In TypeScriptParserBase:
 protected notVarModifier(): boolean {
      const t = this._input.LT(1).type;
      return t !== TypeScriptParser.Var
          && t !== TypeScriptParser.Let
          && t !== TypeScriptParser.Const;
  }
```
And, it would have to be inserted before `expressionStatement` in the grammar:

https://github.com/antlr/grammars-v4/blob/923a1a90f3b9e28c8ac08e170e582ceb15d6f0a9/javascript/typescript/TypeScriptParser.g4#L355

The `variableStatement` would need to be `this.isVarModifier()`, the negated value of the above function.

TwoOfTwelve:
Thank you for the quick reply.

I added the test files.

I found an alternative solution for the first two issues:
The rule "identifierName" allows for either an identifier or a reserved word. While I'm not an expert on typescript, as far as I know only a handful of keywords (those mentioned in identifier explicitly) can be used as an identifier. Removing the option for reservedWord solves both issues for me. I wasn't really comfortable changing the grammar like this, but it might be the more elegant solution, as it removes the ambiguity all together.

The issue with the namespace seems to be a bit more complicated. The solution using the parser base for the variable could be replaced (I think) by just making the varModifier in variableStatement mandatory. However, the file is still not parsed correctly because of this rule in singleExpression: `| identifierName singleExpression? # IdentifierExpression`. This rule seems to be a little too open in general in my opinion and the ambiguity cannot be solved easily, because without strict mode namespace is a valid identifier.

teverett:
@TwoOfTwelve  thanks!

PR #4895, opened 2026-06-12
[c] Fix ports for the C grammar.

kaby76:
* Implemented the Go port.
* Renamed the Octave script to not end in `.m` because the Go compiler sucks and ends up with this: `package example.com/Test: Objective-C source files not allowed when not using cgo or SWIG: zzz.m`.
* Updated the symbol table entries with new data because the C compiler environments have changed everywhere.
* Added missing empty .h files in the sqlite example, which should be generated by Bison, so that the C preprocessor runs without error on the examples--not that it's respected in the code anyway!
* Circular imports in Node changed at some point for ESM so that it broke referencing CParser.RULE_functionDefinition in CParserBase altogether. An explanation of why this broke is below.

##### Explanation of ESM brokenness and the new maintenance nightmare introduced.
```
❯ Claude, what version of the compiler broke this circular importing for TypeScript that required you to change
TypeScript/CParserBase.ts?

● It's not a TypeScript compiler version issue — the root cause is the ES module specification itself, specifically how Node.js ESM
  handles circular dependencies with default exports.

  The short version: in ESM, when module A imports module B which imports module A back, one of them will see the other's exports as
  uninitialized (in the Temporal Dead Zone). Which one "loses" depends on the evaluation order and export style.

  The reason the Antlr4ng target works despite the same circular structure is that antlr4ng uses named exports:
  // antlr4ng generated CParser
  export class CParser extends CParserBase { ... }  // named export

  While the TypeScript (antlr4) target uses default exports:
  // antlr4 generated CParser
  export default class CParser extends CParserBase { ... }  // default export

  In Node.js ESM, export default class Foo and export class Foo have subtly different initialization semantics in the circular
  dependency case. Named class exports appear to be treated more leniently by the module linker, while export default class is more
  prone to the TDZ error in a cycle.

  So this isn't something that "broke" at a particular version — the TypeScript target's use of default exports combined with the
  circular dep between CParser and CParserBase was always going to be fragile under Node.js ESM. It would have worked under CommonJS
  (where require() returns a partially-constructed object rather than a TDZ binding), which is probably how the TypeScript target was
   originally tested.

❯ I thought that ESM is the "new" way to do things. Why is it so "fragile"?

● ESM is indeed the modern standard, but the "fragility" with circular imports is actually intentional design, not a flaw. The
  reasons:

  CommonJS was lenient (and arguably wrong)
  In CommonJS, require() returns whatever has been exported so far — a partially-constructed object. This meant circular deps often
  "worked" by accident, but could produce subtle bugs depending on initialization order. The behavior was unpredictable and hard to
  reason about.

  ESM is strict (and arguably correct)
  ESM uses live bindings with the TDZ rule: a binding exists from the moment the module is linked, but accessing it before the
  exporting module has executed that line is a hard error. This makes the problem visible rather than silently returning undefined or
   a half-built object. The spec designers considered this the right tradeoff — fail loudly rather than silently misbehave.

  The real lesson: circular imports are a design smell
  Both CommonJS and ESM are just revealing a pre-existing problem — CParser and CParserBase are genuinely tightly coupled in a way
  that creates a dependency cycle. The "correct" fix isn't to make the runtime more lenient; it's to break the cycle, which is what
  we did by removing the CParser import from CParserBase.ts.

  Why does Antlr4ng "get away with it"?
  Likely a coincidence of evaluation order. Named exports (export class Foo) happen to get initialized before the class body runs in
  the cycle, whereas export default class Foo does not. It's not that named exports are inherently safer for cycles — the Antlr4ng
  version is just as fragile in principle, it happens to not trigger the error with the current Node.js evaluation order.

  So ESM isn't fragile — it's honest. The circular dep was always a latent problem; ESM just makes it impossible to ignore.

❯ How come other OO languages handle initialization order correctly if it is "arguably wrong" to have circular references?

● That's a fair pushback. The key distinction is what kind of circular reference we're talking about:

  What OO languages handle fine
  class A {
      B b = new B();  // A references B
  }
  class B {
      A a = new A();  // B references A
  }
  In Java, C#, C++, etc., the class definitions (types) are resolved at compile time, and instances are created at runtime. The
  compiler sees the full picture before any code runs, so circular type references are fine.

  What ESM is dealing with — a different problem
  ESM modules are not just type definitions — they are executable code that runs top to bottom. The circular dependency is between
  module initialization scripts, not type declarations. The question becomes: which script runs first, and what does the other one
  see while waiting?

  CParser.ts executes → needs CParserBase to be a class already →
  CParserBase.ts executes → needs CParser.ts to... wait, it's already running

  This is closer to this Java scenario, which Java also can't handle:
  class A {
      static int x = B.y + 1;  // needs B.y at class load time
  }
  class B {
      static int y = A.x + 1;  // needs A.x at class load time
  }
  Java silently returns 0 for the uninitialized static — exactly the kind of "wrong" behavior ESM refuses to do.

  The real difference
  OO languages separate compilation (resolving types) from instantiation (running code). ESM module evaluation blurs that line —
  importing a module both declares types and runs initialization code in the same step. That's what makes circular module
  initialization genuinely tricky, regardless of language.

  So ESM isn't being stricter than Java about OO design — it's being stricter about static initializer order, which Java itself gets
  wrong in the circular case.

❯ But, if you look at the use of CParserBase, the use of the field CParser.RULE_functionDefinition is in a method in CParserBase
that does not execute ever in initialization. Why is that so "wrong"?

● You're absolutely right — and this is where ESM's design has a legitimate criticism.

  The TDZ error we hit is too conservative. The actual runtime usage is:

  // In CParserBase — only called at parse time, never during initialization
  public EnterDeclaration(): void {
      if (isFunctionDefinitionContext(context)) { // uses CParser.RULE_functionDefinition

  By the time EnterDeclaration() is ever called, both modules are fully initialized. The circular reference is safe because the
  values from CParser are only needed at call time, not at import time. This is the classic distinction between:

  - Eager use — needed when the module initializes → genuinely problematic
  - Lazy use — needed only when a method is called later → perfectly safe

  CommonJS actually handles this correctly in this specific case. Because require() is a runtime function call (not a declarative
  link), by the time any method is called, both modules are fully loaded. The laziness is implicit.

  ESM's weakness is that it cannot (or does not) distinguish between these two cases. It sees a circular binding and applies TDZ to
  the entire binding, even if every actual use of that binding is safely deferred inside a method body.

  Languages like Java handle this correctly because method bodies are never executed during class loading — the JVM explicitly
  separates the <clinit> (static initializer) phase from method dispatch. ESM has no equivalent separation.

  So your observation is accurate: the strictness ESM applies here is broader than necessary. The fix we made (hardcoding the rule
  indices) works around a real limitation in the ESM design, not a real design problem in the original code.

❯ That's right, and that's why we have to hard code "92" for "CParser.RULE_functionDefinition". It's a terrible "feature".

● Agreed — it's a real maintenance hazard. If the grammar ever changes and RULE_functionDefinition gets a different index assigned by
   ANTLR (which can happen any time rules are added, removed, or reordered), the hardcoded 92 silently becomes wrong. No compiler
  error, no warning — just incorrect parse behavior that could be very hard to track down.

  The "correct" fix that ESM's design forces on us is worse than the original code. The original intent was clear and
  self-documenting:

  return x?.ruleIndex === CParser.RULE_functionDefinition;  // obvious

  The workaround is fragile and opaque:
  return x?.ruleIndex === 92;  // what is 92? why?

  A slightly less terrible alternative would have been a dynamic import:

  // In the method body, lazily load CParser only when first called
  const { default: CParser } = await import('./CParser.js');

  But that makes every semantic predicate method async, which would cascade through the entire ANTLR runtime and the grammar's action
   code — arguably even worse.

  The least-bad real solution would be for the ANTLR4 TypeScript target to switch from export default class CParser to export class
  CParser (named export), matching the Antlr4ng convention. That would likely resolve the ESM circular dependency issue without
  requiring any workaround at all — and it's a one-line change in the code generator template.
```

Further:
```
  The only things that would actually solve it (short of hardcoding) are:
  - Remove the import from CParserBase entirely (what we did)
  - Use a dynamic import() inside the method body (defers the import past initialization)
  - Change antlr4's code generator to use named exports instead of default exports

```
I changed the generated code to stop using "default" on everything, and it all works again. But, this would require antlr4 to stop adding "default" to the generated classes. Here's the app: 
[Generated-TypeScript.zip](https://github.com/user-attachments/files/28952208/Generated-TypeScript.zip)


teverett:
@kaby76  thanks!

PR #4902, opened 2026-06-16
Fix weekly dev build, along with other things that broke with #4875

kaby76:
When #4875 was made, I forgot to account for other scripts that use the Trash. This change fixes those files, which includes the weekly-dev workflow.


teverett:
@kaby76 thanks!

PR #4904, opened 2026-06-17
Fix README badges to link to workflow run pages

kaby76:
Wrap each badge in a Markdown link so clicking it navigates to the corresponding GitHub Actions workflow page instead of just the SVG image.


teverett:
@kaby76  thanks!

PR #4906, opened 2026-06-18
Add grammar for vba 7.1

Beakerboy:
The vba 7.1 grammar has an official specification from Microsoft. This lexer and parser follows that specification as close as possible. The existing vba6 grammar is widely used, but differs so much that changing it to match the specification would introduce breaking changes.


teverett:
@Beakerboy thanks!

Issue #4908, opened 2026-06-18
Python/TypeScript ports cannot build.


kaby76:
```
Build of python/python2_7_18 for TypeScript failed.

changed 20 packages in 2s

added 12 packages, and audited 13 packages in 1s

3 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
PythonLexerBase.ts:32:30 - error TS2307: Cannot find module 'typescript-collections' or its corresponding type declarations.

32 import * as Collections from "typescript-collections";
                                ~~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error in PythonLexerBase.ts:32

```

kaby76:
Fixed in https://github.com/antlr/grammars-v4/pull/4909.

PR #4909, opened 2026-06-19
[python2_7_18, python3_13] Add dependency to typescript-collections (#4908)

kaby76:
Updated package.json in template form to TypeScript/ for the two Python grammars so that they build and run.


teverett:
@kaby76  thanks!

PR #4916, opened 2026-06-25
Add MAML

antonmedv:
Hi! This adds a grammar for [MAML](https://maml.dev), which is basically JSON
with comments, unquoted keys, optional commas, and multiline strings.

One thing worth knowing: newlines are significant, so a comma or a newline
separates items. I kept newlines as real tokens for that reason.

Tested with ANTLR 4.13.2 and the examples all parse cleanly. Thanks!

### DIFF
--- a/maml/MAML.g4
+++ b/maml/MAML.g4
@@ -1,0 +1,141 @@
+/*
+ MAML - Minimal Abstract Markup Language.
+
+ "JSON with comments, unquoted keys, optional commas, and multiline strings."
+
+ Derived from the MAML v0.1 specification: https://maml.dev/spec/v0.1
+ See also https://maml.dev/ and ../json5/JSON5.g4 for prior art.
+
+ Note on separators: per the spec, members and items are separated by a comma
+ OR a newline (separator = "," / newline). A comma is "optional" only in the
+ sense that a newline may stand in for it - two values on the SAME line with
+ only spaces between them are NOT valid. Newlines are therefore significant
+ tokens here, not skipped whitespace.
+ */
+
+// $antlr-format alignTrailingComments true, columnLimit 150, minEmptyLines 1, maxEmptyLinesToKeep 1, reflowComments false, useTab false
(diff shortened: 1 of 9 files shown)


maml/MAML.g4
```diff
@@ -0,0 +1,141 @@
...
+    ;
+
+NULL
+    : 'null'
+    ;
+
+// A hash marks the rest of the line as a comment, except inside a string. The
+// terminating newline is left for NL so it can still act as a separator.
+COMMENT
+    : '#' ~[\r\n]* -> skip
```
kaby76: Thanks for the addition.

For text reconstruction, a better choice would be  "-> channel(HIDDEN)" rather than "-> skip". "skip" eliminates the token from the token stream completely; "channel(HIDDEN)" places the token on the stream but hides it from the parse. Similar change for WS below.

maml/desc.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<desc xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../_scripts/desc.xsd">
+   <antlr-version>^4.7</antlr-version>
+   <targets>CSharp;Cpp;Dart;Go;Java;JavaScript;PHP;Python3;TypeScript;Antlr4ng</targets>
```
kaby76: Change `<targets>...</targets>` to `<targets>Antlr4ng;CSharp;Cpp;Dart;Go;Java;JavaScript;PHP;Python3;Rust;TypeScript</targets>`. It works for Rust, too.

kaby76 (changes requested):

maml/MAML.g4
antonmedv: Done

maml/desc.xml
antonmedv: Done

kaby76 (approved):

teverett:
@antonmedv thanks!

Issue #4915, opened 2026-06-24
[c] Regression in sqlite/src/test1.c?


kaby76:
Looks like there is a regression in test1.
```
$ ./bin/Debug/net10.0/Test.exe ../examples/sqlite/src/test1.c
"../examples/sqlite/src/test1.c" line 1078, .p 144481:2 no viable alternative at input 'i64'
CSharp 0 ../examples/sqlite/src/test1.c fail 20.6561346
Total Time: 21.8167668
06/24-18:32:47 ~/issues/g4-current/c/Generated-CSharp
$
```

This occurs for at least the Java port as well, and likely all others.

kaby76:
The behavior is expected (but incorrect). I'm not sure why it wasn't caught before.

Consider this example:
```
typedef long long int sqlite_int64;
typedef sqlite_int64 sqlite3_int64;
typedef sqlite_int64 i64; /* 8-byte signed integer */

static void inttoptrFunc(
){
  sqlite3_int64 i64;
  i64 = 1;
}

struct Point { int x; int y; };   // declares struct tag, no variable
union Value { int i; float f; };  // declares union tag, no variable
enum Color { Red, Green, Blue };  // declares enum, no variable
```
The problem is because "i64" on line 7 was previously declared as a type on line 3, so it choses to parse i64 as a declarationSpecifier, not an initDeclaractorList. But, once we have a type that was defined as a typedef, like sqlite3_int64, we cannot have another typedef type in the declarationSpecifiers. The constraint needs to be added to the grammar.

This is why a grammar in which the parse depends on the semantics is a bad grammar. It's what I call "semantics-dependent parsing," and it occurs in the EBNF of many popular languages because language designers backport the EBNF from the AST, conflating the AST with the CST.

PR #4917, opened 2026-06-28
[c] Fix bad parses involving typedefs (#4915).

kaby76:
Fixed regression in parsing examples/sqlite/src/test1.c, where a symbol was first declared a typedef, then in a new block scope, declared as a variable. This caused an error due to a flaw in the symbol table code.


teverett:
@kaby76 thanks!

PR #4919, opened 2026-07-02
[CI] Add performance information for all grammars in repo.

kaby76:
This PR adds a table on the performance for all grammars in the repo, navigatable from a new section in the top-level readme.md. With the parser rate measurement (aka "tokens per second", TPS), this table gives a clear picture of the overall performance of each grammar.

* Add performance.md, a table of performance for every grammar in this repo.
* Update Trash and testing app templates.
* Removed Nuget updating because Trash should only be updated manually.

The script used to output the performance.md table performs tests on the CSharp target, and runs the parser app once to get past the Windows Virus Scan delay.

I implemented the performance table performance.html, and render it using htmlpreview.github.io for a dynamic table presentation. One can sort rows via the rate of parsing (TPS) by clicking on the column header. See https://htmlpreview.github.io/?https://github.com/kaby76/grammars-v4/blob/8b2b00c952310eed72b86bb464dcb905040144a0/performance.html. Surprisingly, the best performing grammar is antlr/antlr4!


teverett:
@kaby76 thanks!

Issue #4924, opened 2026-07-04
[java/java] Java grammar only supports Java and CSharp targets.


kaby76:
The grammar needs to be updated with the other targets, at least Cpp, Dart, Go, TypeScript, and Rust.

PR #4925, opened 2026-07-04
[java/java] Add ports for the java/java/ grammar (#4924).

kaby76:
This PR adds the Cpp, Dart, Go, Rust, and TypeScript ports to the java/java grammar. I also collected parser performance data for the targets against the JDK SDK source, and added a table with the information to the readme.md.


teverett:
@kaby76 thanks!

PR #4926, opened 2026-07-05
[c] Add Rust target support

kaby76:
Add Rust port for the grammar for C. 

- c_parser_base.rs: thread-local symbol table tracking C typedef names, with stateless predicate helpers (is_declaration_specifier, is_init_declarator_list, is_cast, is_statement, etc.) that use a backward token scan for context instead of mutable state. The main issue is that the Rust API does not offer access to the partially completed parse tree.
- transformGrammar.py: two-phase script that transforms grammar predicates/actions for antlr4rust and fixes codegen issues. NB: The transformGrammar.py script should not modify main.rs, but this is acceptable for now.
- Update desc.xml to include Rust in targets and test targets.
- Updated perf data with table.
- _NB: This is a preliminary implementation. Some of the important command-line options are not available, and the `--nopp` option is in a modified main.rs. It should not._


teverett:
@kaby76  thanks!

PR #4927, opened 2026-07-07
Fix link to perf data.

kaby76:
This PR fixes the link to the performance table. I had to test using my cloned repo, but didn't update the link correctly when submitting https://github.com/antlr/grammars-v4/pull/4919.


teverett:
@kaby76  thanks!

PR #4929, opened 2026-07-08
[abnf] Add abnf in abnf as an example.

kaby76:
This PR adds abnf.abnf, the definition of ABNF in ABNF, for testing. I was going through the performance data, and there were only [three examples](https://github.com/antlr/grammars-v4/tree/e3f7c992259b5208b166a931100b735c246779a6/abnf/examples) for the abnf grammar. That's when I remembered that ABNF must be in ABNF itself, and copied the grammar to the new example. I also updated the examples/ directory with a few more from apg-java but renamed several because they are not ABNF. I updated the grammar with one extension for the underscore in rule names.


teverett:
@kaby76  thanks!

PR #4943, opened 2026-07-20
[perf] Update perf table with grammar action count.

kaby76:
This PR adjusts the perf table and adds the count of the number of action blocks in the grammar. The perf table now looks like this: https://htmlpreview.github.io/?https://raw.githubusercontent.com/kaby76/grammars-v4/refs/heads/g4-update-perf-table-with-action-info/performance.html


teverett:
@kaby76  thanks!

PR #4941, opened 2026-07-18
[muddb] Fixed link (#4939).

kaby76:
The link to the Wikipedia web page changed (#4939). Adjusted here.

PR #4934, opened 2026-07-16
[perf] Several fixes for obtaining performance data for more grammars (#4933)

kaby76:
This PR fixes several of the problems described in #4933. I only focused on those grammars for "important" languages: Ada (currently in the top 20 of TIOBE); 8080 assembly (not really used nowadays, but since it's an assembly language, I deemed it of value); Swift (currently in the top 20 of TIOBE); XPath (not on TIOBE, but I use it quite often with Trash). Most problems remain because many grammars have only one parser test file. I didn't feel the need to address the issues at this point.

* [ada/ada2012](https://github.com/antlr/grammars-v4/tree/master/ada/ada2012) and [ada/ada2022](https://github.com/antlr/grammars-v4/tree/master/ada/ada2022): The preferred test suite of the parser should be the ACATS, which is implemented for testing by reordering the tests in the desc.xml file.
* [asm/asm8080](https://github.com/antlr/grammars-v4/tree/master/asm/asm8080): More tests were added. This required updates to the grammar.
* [asm/asmZ80](https://github.com/antlr/grammars-v4/tree/master/asm/asmZ80): Tests added, requiring updates to grammar.
* [swift/swift5](https://github.com/antlr/grammars-v4/tree/master/swift/swift5): The _scripts/test-tps.sh required updates to fix a bug that was exposed with this grammar (spaces in file names). Other grammars with similar issues now fixed:  arden, asm/ptx, pgn, yini.
* [xpath/xpath31](https://github.com/antlr/grammars-v4/tree/master/xpath/xpath31): Tests were split into multiple files. Statistics computed by test-tps.sh require tests in multiple files, not one.
* I updated the `performance.*` files with the latest data. This is with the latest changes to `_scripts/test-tps.sh` script in this PR to output the number of ambiguities detected on input. It now looks like this: https://htmlpreview.github.io/?https://raw.githubusercontent.com/kaby76/grammars-v4/refs/heads/g4-4933/performance.html

PR #4931, opened 2026-07-13
[CI] TypeScript broken (#4930).

kaby76:
This PR fixes #4930.

PR #4947, opened 2026-07-23
[PlSql] Allow `index_properties` to be empty

Naros:
This allows for using simply `USING INDEX` without any further clauses.


kaby76:
The TypeScript testing is failing because features were removed in the TypeScript compiler. https://github.com/antlr/grammars-v4/pull/4931 fixes testing. Merging that into you PR will fix the tests.

@teverett Can we please merge in the completed and functioning PRs, especially https://github.com/antlr/grammars-v4/pull/4931 ? Also ready for review/merge are: https://github.com/antlr/grammars-v4/pull/4932, https://github.com/antlr/grammars-v4/pull/4934, https://github.com/antlr/grammars-v4/pull/4935, https://github.com/antlr/grammars-v4/pull/4936, https://github.com/antlr/grammars-v4/pull/4941.

Briseglace:
Hi @Naros,

I know the rule is presented this way in the documentation, but I am wondering if having a rule that matches an empty string could be problematic. I have not found any other rule in the grammar that is implemented this way.
Also, the proposed changes allow for more than one `INDEXTYPE IS ...` per usage of the `index_properties` rule. I might be wrong, but this seems incorrect.

I suggest adding the `?` operator to the usages of the rule instead. The rule has only two usages and one of them is already marked with `?`. It would only need to be added to the usage in `using_index_clause` which seems to be your initial goal.

I would appreciate your thoughts on the subject.

kaby76:
Please sync with latest on master to resolve build failures.

teverett:
@Naros thanks!

PR #4949, opened 2026-07-24
[mariadb] Fix DECLARE .. HANDLER syntax to allow for bare return, #4948

### LINKED ISSUE #4948: [mariadb] Unable to parse bare return in `DECLARE ... HANDLER` syntax
The MariaDB grammar fails to parse this line from the Sakila sample database: 

https://github.com/jOOQ/sakila/blob/e089a5b1ec9af0df7a9c6a5d47d49fa1736a4e84/mysql-sakila-db/mysql-sakila-schema.sql#L595

```sql
DECLARE EXIT HANDLER FOR NOT FOUND RETURN NULL;
```

This is because the `routineBody` rule does not allow for a bare `returnStatement`. The Oracle MySQL grammar uses `compoundStatement` for a declare handler rule instead, which could probably be a viable fix by itself, but a safer change that doesn't remove the transitive `sqlStatement` would be to change the declare handler rule to use `(compoundStatement | sqlStatement)` instead of `routineBody`, since it is a superset of the `routineBody` rules already (`routineBody` is `blockStatement | sqlStatement`, and `compoundStatement` includes `blockStatement`):

```antlr
declareHandler
    : DECLARE handlerAction = (CONTINUE | EXIT | UNDO) HANDLER FOR handlerConditionValue (
        ',' handlerConditionValue
    )* (compoundStatement | sqlStatement)
    ;
```

A narrower possible fix would be to simply add `returnStatement` as an alternative:

```antlr
declareHandler
    : DECLARE handlerAction = (CONTINUE | EXIT | UNDO) HANDLER FOR handlerConditionValue (
        ',' handlerConditionValue
    )* (routineBody | returnStatement)
    ;
```

... but I think using compoundStatement gets this closer to the Oracle MySQL grammar, and this won't cover other valid, bare control-flow statements.

I have tested that this works for the Sakila sample database and will submit a PR after further testing.

paulirwin:
This fixes the MariaDB grammar to allow for bare return (and other control-flow statements) as the body for a handler declaration.

Note that `routineBody` is defined as `blockStatement | sqlStatement`, and `compoundStatement` includes `blockStatement`, so this is a superset of the previous rules. I am open to alternative approaches, however.

Fixes #4948

### DIFF
--- a/sql/mariadb/MariaDBParser.g4
+++ b/sql/mariadb/MariaDBParser.g4
@@ -1528,7 +1528,7 @@ declareCursor
 declareHandler
     : DECLARE handlerAction = (CONTINUE | EXIT | UNDO) HANDLER FOR handlerConditionValue (
         ',' handlerConditionValue
-    )* routineBody
+    )* (compoundStatement | sqlStatement)
     ;
 
 handlerConditionValue
--- a/sql/mariadb/examples/fast/ddl_create.sql
+++ b/sql/mariadb/examples/fast/ddl_create.sql
@@ -791,3 +791,13 @@ SELECT
      ,@N_latin;
 END
 #end
+
+#begin
+-- Create function with exit handler containing a bare RETURN statement
+-- src: https://mariadb.com/docs/server/reference/sql-statements/programmatic-compound-statements/declare-handler#syntax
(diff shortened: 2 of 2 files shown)


teverett:
@paulirwin thanks!

PR #4963, opened 2026-08-05
[mysql/Oracle] Fix URL identifier parsing (#4962)

### LINKED ISSUE #4962: [MySQL] URL cannot be used as an unquoted identifier on 8.0.32+
### Grammar

`sql/mysql/Oracle`

### Reproducer

Configure both the lexer and parser with `serverVersion = 80032` or later, then parse:

```sql
ALTER TABLE URL ADD account_id BIGINT;
ALTER TABLE URL ADD INDEX idx_account_id (account_id);
ALTER TABLE auth.page_view CHANGE url url VARCHAR(700);
```

### Actual behavior

The grammar rejects `URL` in table and column identifier positions.

### Expected behavior

MySQL documents `URL` as a nonreserved keyword added in 8.0.32, so it can be used as an unquoted identifier.

https://dev.mysql.com/doc/refman/8.0/en/keywords.html

### Cause

`MySQLLexer.g4` emits `URL_SYMBOL` when `serverVersion >= 80032`, but `MySQLParser.g4` does not include `URL_SYMBOL` in `identifierKeywordsUnambiguous`. Before 8.0.32 the same text remains an `IDENTIFIER`, which hides the missing parser alternative.

### Proposed fix

Add a version-gated `URL_SYMBOL` alternative to `identifierKeywordsUnambiguous` and add regression examples for table and column identifiers.

minleejae:
## Summary

- allow `URL_SYMBOL` as an unquoted identifier on MySQL 8.0.32 and later
- add regression examples covering table and column identifiers

## Why

MySQL documents `URL` as a nonreserved keyword added in 8.0.32. The lexer emits `URL_SYMBOL` for those versions, but the parser did not include it in `identifierKeywordsUnambiguous`, so valid statements such as the following were rejected:

```sql
ALTER TABLE URL ADD account_id BIGINT;
```

Closes #4962.

## Testing

- `mvn clean test` in `sql/mysql/Oracle`: passed
- verified the regression examples with `serverVersion` set to both `80031` and `80032`


kaby76:
Prerequisite for testing: https://github.com/antlr/grammars-v4/pull/4965

PR #4967, opened 2026-08-06
Fix Cypher grammar lexer/parser bugs

knoguchi:
# Summary
Found several bugs and a missing literal by cross-checking the grammar against the openCypher [Technology Compatibility Kit (TCK)](https://github.com/opencypher/openCypher/tree/master/tck).

**Bug fixes**
`ID` , `CHAR_LITERAL` , `ERRCHAR` , `DIGIT`, `HexDigit` -> `HexDigits`, `Digits`, `multiPartQ`, `notExpression`.

**New addition:**
`OctalDigits` for octal literals (0o1, 0o77, etc).

# Result
 The TCK contains 3,880 scenarios.  3,185 are positive scenarios. The original grammar correctly accepted 2,837 but incorrectly rejected 348. With this patch, all 3,185 positive scenarios are accepted correctly. Both the original grammar and this patch correctly reject 22 negative syntax scenarios. The remaining 673 negative scenarios are semantic (type/scope/arity checks, etc) and are accepted by both versions, as expected, since semantic validation is outside the grammar's scope.

# Details

CypherLexer.g4:
- ID must begin with a letter, followed by letters or digits. Original allowed IDs to begin with, or be entirely composed of, digits.
- CHAR_LITERAL (single-quoted strings) must allow any number of characters. Original capped it at 0-1 characters (`?` instead of `*`): `RETURN 'ab'` silently parsed as the bare identifier `ab` (quotes vanished, no error raised), and `RETURN 'a:b'` as a symbol with a label predicate (`a:b`) -- both wrongly accepted instead of erroring.
- An unrecognized character must raise a syntax error. Original's ERRCHAR sent it to the HIDDEN channel instead, silently discarding it -- combined with the CHAR_LITERAL bug, `RETURN 'hello world'` did raise an error, but ANTLR's error recovery still produced a mangled tree (`hello`/`world` as two bare tokens, quotes and space gone) rather than cleanly failing.
- DIGIT must not include a sign -- unary minus is the parser's job (unaryAddSubExpression). Original's leading SUB? let maximal-munch swallow a binary minus into the next operand's token, so `5-1` tokenized as two adjacent DIGIT tokens with no operator between them.
- Hex literals use the '0x'-prefixed HexDigits fragment. Original referenced the single-char HexDigit fragment instead, so hex literals like `0x1` never tokenized. Octal literals need an explicit '0o' prefix, which didn't exist (only a non-standard bare-leading-zero form). Decimal integers may begin with 0 (`007` is valid); original required a leading 1-9.

CypherParser.g4:
- Both reading statements (MATCH/UNWIND/CALL) and updating statements (CREATE/MERGE/DELETE/SET/REMOVE) must be allowed between a multi-part query's WITH boundaries. Original only allowed updating statements, so chains like `WITH ... UNWIND ... WITH ...` couldn't parse.
- NOT must allow repetition for chained negation. Original allowed at most one NOT, so `NOT NOT true` couldn't parse.

New examples exercise each fix.

### DIFF
--- a/cypher/CypherLexer.g4
+++ b/cypher/CypherLexer.g4
@@ -128,19 +128,18 @@ OF         : 'OF';
 ADD        : 'ADD';
 DROP       : 'DROP';
 
-ID: LetterOrDigit+;
+ID: Letter LetterOrDigit*;
 
 ESC_LITERAL    : '`' .*? '`';
-CHAR_LITERAL   : '\'' (~['\\\r\n] | EscapeSequence)? '\'';
+CHAR_LITERAL   : '\'' (~['\\\r\n] | EscapeSequence)* '\'';
 STRING_LITERAL : '"' (~["\\\r\n] | EscapeSequence)* '"';
 
-DIGIT : SUB? (HexDigit | OctalDigit | Digits | FLOAT);
+DIGIT : HexDigits | OctalDigits | Digits | FLOAT;
 FLOAT : (Digits '.' Digits | '.' Digits) ExponentPart? [fd]? | Digits (ExponentPart [fd]? | [fd]);
 
 WS           : [ \t\r\n\u000C]+ -> channel(HIDDEN);
 COMMENT      : '/*' .*? '*/'    -> channel(COMMENTS);
 LINE_COMMENT : '//' ~[\r\n]*    -> channel(COMMENTS);
-ERRCHAR      : .                -> channel(HIDDEN);
 
 fragment EscapeSequence:
     '\\' [btnfr"'\\]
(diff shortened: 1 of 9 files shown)


knoguchi:
Pushed one more fix in the same TCK/spec-cross-checking series: `END` moved from `reservedWord` to `symbol` (identifier position), with an example.

Verified against openCypher's official grammar artifact (`openCypher.bnf`): `<identifier> ::= ... | <non-reserved word>`, and `<non-reserved word>` explicitly includes `END`. It's safe under this grammar's LL(*) design because `END` only ever terminates a `CASE` expression — no clause-start ambiguity. Practical motivation: `MATCH (start)-[r]->(end)` is the standard endpoint naming in exported/generated Cypher and previously failed to parse.

Broader note from the same cross-check, for context rather than action: the official `<non-reserved word>` list makes nearly every keyword (~55 words: `MATCH`, `CREATE`, `ORDER`, ...) a valid identifier — the spec assumes context-sensitive resolution. This grammar's narrower `symbol` list (`COUNT | FILTER | EXTRACT | ANY | NONE | SINGLE`, now `END`) is a reasonable LL(*) simplification; `END` is just the one remaining member of that class that's unambiguous for free.

teverett:
@knoguchi thanks

PR #4970, opened 2026-08-07
[dot] Add examples.

kaby76:
The test suite for the dot grammar is too small. Since it is one of the faster target-language-free grammars, I've added more test input. Most of the .dot files are Antlr4 Tool generated ATN graphs. I also included a couple of large examples from the graphviz source.


teverett:
@kaby76 thanks

PR #4985, opened 2026-08-14
[awk] Updated awk grammar and added more example input.

kaby76:
## Summary

- Fix grammar to handle real-world AWK scripts: leading blank lines, multi-line function parameter lists, whole-array \`delete\`, \`.5\`/\`5.\` float literals, ERE character classes with \`/\` inside \`[...]\`, and \`COMMENT\` consuming trailing newlines
- Add \`_afterExpr\` semantic predicate on the ERE lexer rule to correctly distinguish \`/\` (division) from \`/…/\` (ERE) based on the preceding token
- Add 33 real-world AWK example scripts covering arithmetic, string processing, CSV, chemistry, sorting, and system utilities; all parse without errors
- Add \`awkLexerBase\` base class for Antlr4ng, Cpp, CSharp, Dart, Go, Java, Python3, TypeScript targets implementing the \`_afterExpr\` logic via \`NextToken\`/\`nextToken\` override
- Add OphiRust (Rust/antlr4-runtime) target port with a \`LexerBase\` implementing \`SemanticHooks\` for the ERE predicate and token-emitted/reset lifecycle; uses \`AwkLexer::with_hooks()\` so the predicate is actually evaluated
- Add Trash/trgen templates for OphiRust (\`_scripts/templates/OphiRust/\`): \`st.build.sh\`, \`st.build.ps1\`, \`st.main.rs\`, \`st.lexer_base.rs\`, \`src/gen/st.mod.rs\`, \`Cargo.toml\`
- Pin CI install of \`antlr4-rust-gen\` to git tag \`v0.33.1\` (via \`cargo install --locked\`) and cache the binary to avoid recompiling \`rustpython-vm\` on every run
- Fix \`.gitignore\`: add negation rules for \`_scripts/templates/**/gen\` so template files in \`gen/\` subdirectories (e.g. \`st.mod.rs\`) are tracked normally without requiring \`git add -f\`
- Fix Maven tester configuration


teverett:
@kaby76 thanks!

PR #4986, opened 2026-08-14
[PlSql] fix for alter type * ATTRIBUTE-keyword support

frolovgo:
Hello ANTLR-team.
Looks like the "alter_attribute_definition" rule is missing in the "alter_type" rule:
https://docs.oracle.com/en/database/oracle/oracle-database/19/lnpls/ALTER-TYPE-statement.html#GUID-A8B449E7-E3A8-48F4-A4C6-5BB87B1841CD__I2112566

image

So false posittive diagnostics generated:
image


teverett:
@frolovgo  thanks!

PR #4987, opened 2026-08-16
[iXML] Add ANTLR4 grammar for Invisible XML 1.0

kaby76:
## Summary

Adds a new grammar for [Invisible XML 1.0](https://invisiblexml.org/1.0/).

- Split lexer/parser grammar (`ixmlLexer.g4` + `ixmlParser.g4`) to support lexer modes; `HEX_MODE` isolates hex-digit tokenization after `#` so that name-follower characters (hyphen, letters a–f) do not interfere with the `NAME` rule
- Parser rule names mirror the iXML 1.0 spec grammar exactly
- `desc.xml` uses multiple `<inputs>` patterns with negation to include `examples/*.ixml` and `examples/ixml/samples/**/*.ixml` while excluding `*.decorated.ixml` files (which use non-standard annotations outside the iXML 1.0 spec)
- Includes `ixml.ixml` (the self-describing grammar from the spec) plus sample grammars from the iXML community test suite

## Test plan

- [x] Grammar parses `ixml.ixml` (the self-describing iXML grammar)
- [x] Grammar parses 15/15 standard sample grammars from the iXML community test suite
- [x] `*.decorated.ixml` files are excluded from test runs (they use non-standard `[jl:opt]` annotations)
- [x] `pom.xml` Maven test passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)


kaby76:
(https://github.com/antlr/grammars-v4/pull/4985 needs to be merged first.)

PR #4989, opened 2026-08-18
promql: add mad_over_time function to the lexer

alexpir76:
Prometheus supports `mad_over_time` (median absolute deviation over time; experimental, gated by `--enable-feature=promql-experimental-functions`, implemented in [promql/functions.go as `funcMadOverTime`](https://github.com/prometheus/prometheus/blob/main/promql/functions.go)). The lexer's function list is missing it, so such queries fail to parse.

Adds the token alongside the other `*_over_time` functions, plus an example query.


teverett:
@alexpir76 thanks

PR #4997, opened 2026-08-28
[wkt] ISO/IEC 13249-3 5.1.67: add 8 geometry types, complete 3 more

grootstebozewolf:
Completes the WKT grammar's coverage of subclause 5.1.67, `<well-known text representation>`, of ISO/IEC 13249-3 (SQL/MM Part 3: Spatial).

## New types

`CIRCLE`, `GEODESICSTRING`, `ELLIPTICALCURVE`, `NURBSCURVE`, `SPIRALCURVE`, `COMPOUNDSURFACE`, `BREPSOLID` — together with the component productions they are built from: reference locations and affine placements, vectors, angles in degrees/gradians/radians, NURBS control points and knots, and the scale-factor, distance and curvature scalars.

```wkt
CIRCLE Z (0 0 0, 1 1 0, 2 0 0)
NURBSCURVE (DEGREE 3, CONTROLPOINTS (NURBSPOINT (WEIGHTEDPOINT (0 0), WEIGHT 1), NURBSPOINT (WEIGHTEDPOINT (1 2), WEIGHT 0.5)), KNOTS (KNOT (VALUE 0, MULTIPLICITY 4)))
BREPSOLID Z (POLYHEDRALSURFACE Z (PATCHES (POLYGON Z ((0 0 0, 0 1 0, 1 1 0, 0 0 0)))))
```

## Completed types

Three types previously accepted only the OGC Simple Features form:

| Type | ISO form now accepted |
| --- | --- |
| `TRIANGLE` | `TRIANGLE (0 0, 1 0, 0 1)` — three bare points |
| `POLYHEDRALSURFACE` | `POLYHEDRALSURFACE Z (PATCHES (POLYGON Z ((...))))` |
| `TIN` | `TIN Z (PATCHES ((...)) ELEMENTS (...) MAXSIDELENGTH 5)`, with all eleven TIN element types and their `ID` / `TAG` labels |

**The OGC forms still parse.** Dropping them would reject data that existing implementations emit, including two of this repo's own committed examples.

## CLOTHOID

`CLOTHOID` gains the ISO form next to the `(k0, k1, L)` JTS extension added in #4848. They share the token and the `clothoidGeometry` rule and are told apart by the token after the opening parenthesis — a number for the JTS form, `REFERENCELOCATION` for the ISO one. Both are accepted; `CLOTHOID_PROPOSAL.md` §5.1 records the coexistence.

`compoundCurveMember` is kept as a named rule (rather than folded into `curveMember`, which the standard defines identically) so #4848's wiring is undisturbed; it now delegates to `curveMember`, so a compound curve accepts every ISO curve type.

## On the source document

The grammar was written against the **DIS ballot text** — ISO/IEC JTC 1/SC 32 N 2593, 5th edition, stage 40 — not the published ISO/IEC 13249-3:2016, which is paywalled. `README.MD` now says so plainly rather than implying the published standard was consulted, and records the three places the draft contradicts itself along with how the grammar resolves each:

- `<compoundsurface text body>` is referenced by `<surface text>` and `<shell text>` but never defined.
- `<angle text>` is cross-referenced to subclause 15.1.21; the definition is in 16.1.21.
- `<vector text representation>` is cross-referenced to 16.2.22; the definition is in 17.2.22.

## Known limitation

Free-form `SPIRALTYPE` values are left out. The BNF defines `<spiraltype text>` as `<letters>`, a production admitting parentheses and spaces, so it cannot be lexed without swallowing its own terminator. The grammar accepts the five names 4.2.12 lists (`CLOTHOID`, `BLOSS`, `BIQUADRATIC`, `SINE`, `COSINE`) instead of inventing a quoting syntax the standard does not define.

## Testing

- `examples/example3.txt` — the new types, including `EMPTY` at every level the standard permits it
- `examples/example4.txt` — the completed `TRIANGLE` / `POLYHEDRALSURFACE` / `TIN` forms and all eleven TIN element types
- `examples/example5.txt` — both `CLOTHOID` forms, including one of each in a single `COMPOUNDCURVE`

`mvn clean test` passes. ANTLR generates the grammar with no warnings. The examples introduce no ambiguity beyond the pre-existing one in the `point` rule, verified by generating the parser from `master` and reproducing that same report.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


teverett:
@grootstebozewolf thanks!

PR #5005, opened 2026-09-12
[clif] Add start rule and targets for CLIF grammar.

### LINKED ISSUE #5004: Clif grammar should have an EOF-terminated start rule.

kaby76:
Fixes #5004 

While evaluating Trash's complete replacement of Antlr4, I saw that CLIF.g4 did not declare an EOF-terminated start rule. ChatGPT suggested `cltext` from the original, hand-written parser.

### DIFF
--- a/clif/CLIF.g4
+++ b/clif/CLIF.g4
@@ -42,6 +42,8 @@
 
 grammar CLIF;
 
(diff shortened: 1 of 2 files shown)


teverett:
@kaby76 thanks!

PR #5010, opened 2026-09-18
[java] Add OphiRust port.

kaby76:
Adds the OphiRust port for the `java/java` grammar. OphiRust renames functions to follow a specific coding style, so the start symbol was renamed to avoid this behavior. Trash version and templates upgraded to support changes with new trxquery.


teverett:
@kaby76 thanks!

PR #5012, opened 2026-09-20
Fix Cpp build problem.

### LINKED ISSUE #5011: CI fails to build Cpp ports for some grammars.
Something changed in the environment so that some grammars that used to work for the Cpp target now don't on Windows.

kaby76:
Fixes #5011 

As it turns out, there was only one grammar that was not fixed: snowflake. That is the only grammar change in this PR. 
[out.txt](https://github.com/user-attachments/files/32445396/out.txt)


teverett:
@kaby76 thanks!

